PR #1, opened 2015-08-13
copy over unreferenced contrib to new contrib repo

mikedanese:
When this goes in, it should be the opposite of kubernetes/kubernetes#12693


mikedanese:
@googlebot are you here?

PR #6, opened 2015-08-14
remove extra godeps that snuck in somehow when we first moved the repo over

mikedanese:
I am godep inept :frowning:


eparis:
Should I do something like https://github.com/kubernetes/kubernetes/pull/12188 over here?

mikedanese:
@eparis I'm attempting to bring over the verify checks and add them to travis (see https://github.com/kubernetes/contrib/blob/master/.travis.yml#L13-L14). I'm not sure how best to do this without duplicating a bunch of code. Feel free to knock some of the checks out if you have a good idea on how to do it.

PR #11, opened 2015-08-18
Add a pod YAML file for the submit-queue.

brendandburns:


brendandburns:
@ixdy @zmerlynn can I get a quick review?

submit-queue/pod.yaml
```diff
@@ -0,0 +1,22 @@
...
+metadata:
+  name: submit-queue
+  namespace: default
+spec:
+  containers:
+  - image: gcr.io/google_containers/submit-queue:0.1
+    command:
+      - bash
+      - -c
+      - "/submit-queue --user-whitelist=/whitelist.txt --jenkins-host=http://jenkins-master:8079 --token=$(cat /etc/secret-volume/token) --alsologtostderr"
```
ixdy: where does this jenkins-host address come from?

brendandburns: That's the name of the VM that the jenkins instance runs on.  The cluster runs in the same GCE project, so it just uses the built in project DNS.

ixdy: OK, but I think Jenkins uses port 8080, not 8079?

ixdy: also, won't passing the token in as a cmdline argument expose it via ps?

brendandburns: Switched to 8080 (I swear it was 8080 before...)

wrt to the command line, that's true, and we should prob. fix eventually, but for now, if you can log onto the machine, you can talk to the kubernetes API server and read the secret anyway.

brendandburns:
@ixdy comments addressed, please take another look.

Thanks!
--brendan

PR #13, opened 2015-08-19
ansible: factor out kube_cert_ip

jlebon:
This will help users more easily specify the IP addresses for which the
certificate should be valid. The make-ca-cert.sh script allows for many
"magic strings" such as '_use_aws_external_ip_', which can automatically
query the metadata service to retrieve the public IP. This modification
makes changing that setting much easier (and less likely to change).


roberthbailey:
LGTM.

PR #15, opened 2015-08-19
Small iptables fixes

jlebon:
The first patch ensures that we wait for the lock even when only listing iptables entries. Otherwise we may sometimes get a busy lock error.

The second patch removes an unnecessary iptables restart, which coincidentally resolves some issues it caused.


brendandburns:
Assign to @eparis because Ansible.

PR #17, opened 2015-08-21
Ansible: fix etcd connection refuse

Arano-kai:
Connection fail if fqdn not resolved

### DIFF
--- a/ansible/roles/etcd/defaults/main.yaml
+++ b/ansible/roles/etcd/defaults/main.yaml
@@ -16,10 +16,10 @@ etcd_peer_key_file: "{{ etcd_conf_dir }}/peer.key"
 etcd_initial_cluster_state: new
 etcd_initial_cluster_token: etcd-k8-cluster
 
-etcd_initial_advertise_peer_urls: "{{ etcd_peer_url_scheme }}://{{ inventory_hostname }}:{{ etcd_peer_port }},{{ etcd_peer_url_scheme }}://127.0.0.1:{{ etcd_peer_port }}"
-etcd_listen_peer_urls: "{{ etcd_peer_url_scheme }}://{{ inventory_hostname }}:{{ etcd_peer_port }},{{ etcd_peer_url_scheme }}://127.0.0.1:{{ etcd_peer_port }}"
-etcd_advertise_client_urls: "{{ etcd_url_scheme }}://{{ inventory_hostname }}:{{ etcd_client_port }},{{ etcd_peer_url_scheme }}://127.0.0.1:{{ etcd_client_port }}"
-etcd_listen_client_urls: "{{ etcd_url_scheme }}://{{ inventory_hostname }}:{{ etcd_client_port }},{{ etcd_peer_url_scheme }}://127.0.0.1:{{ etcd_client_port }}"
+etcd_initial_advertise_peer_urls: "{{ etcd_peer_url_scheme }}://{{ inventory_hostname }}:{{ etcd_peer_port }}"
+etcd_listen_peer_urls: "{{ etcd_peer_url_scheme }}://{{ inventory_hostname }}:{{ etcd_peer_port }}"
+etcd_advertise_client_urls: "{{ etcd_url_scheme }}://{{ inventory_hostname }}:{{ etcd_client_port }}"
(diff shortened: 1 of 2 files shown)


jeffbean:
I am curious if you left off the 127.0.0.1 address from the list did it also have issues conneting? using only the inventory_hostname?

eparis:
@detiber @timothysc always localhost seems wrong to me but you guys know it better than me.

timothysc:
I believe the original was just to use inventory_hostname.  Are you trying to setup in a non-DNS env?

detiber:
@eparis all three are sort of wrong in different ways.

localhost will prevent any clustering from working, or even connecting to a single etcd from a remote system.

inventory_hostname does not necessarily contain the value you want either, and it might not even be an IP or fqdn.

ansible_fqdn might also be wrong since it tries to do a reverse DNS lookup on the IP address with the default route.

This was one of the main reasons we created openshift_facts, so that we could attempt to compute some defaults based on the environment (for example using the cloud provided metadata when available).

That said, it depends on how smart you want the role to be vs. when a user should have to explicitly override the default values.

jasonbrooks:
I believe this needs to be 0.0.0.0, rather than 127.0.0.1. Will additional nodes be able to contact etcd if the listen url is 127.0.0.1?

It works for me with this change:

```
-etcd_advertise_client_urls: "{{ etcd_url_scheme }}://{{ ansible_fqdn }}:{{ etcd_client_port }}"
-etcd_listen_client_urls: "{{ etcd_url_scheme }}://{{ ansible_fqdn }}:{{ etcd_client_port }}"
+etcd_advertise_client_urls: "{{ etcd_url_scheme }}://0.0.0.0:{{ etcd_client_port }}"
+etcd_listen_client_urls: "{{ etcd_url_scheme }}://0.0.0.0:{{ etcd_client_port }}"
```

See also https://access.redhat.com/solutions/1498143.

detiber:
@jasonbrooks Does ADVERTISE_CLIENT_URLS handle setting 0.0.0.0 properly for clustered hosts?  I've only tested with specifying valid IPs and hostnames.

jasonbrooks:
@detiber I tested with a master and two nodes -- is that what you mean by clustered? If you're talking about multi-master, I haven't tried that.

detiber:
@jasonbrooks no, with multiple etcd hosts defined.

timothysc:
@Arano-kai do you have your inventory file you could post?  Also, is there a reason for dns-fail?

Arano-kai:
@timothysc, reason for dns-fail is lack of dns, only static IP.
I experiment with kube as main structure, who contain base services (dns, freeipa, etc). So, no dns at this point.

Arano-kai:
@jasonbrooks, 0.0.0.0 will listen on all interfaces what may not be safe.
127.0.0.1 left for debug reasons only (you know, local query etcd for see what he did there) and may be safely removed. Not so familiar with the etcd yet...

Arano-kai:
Woops, my bad, didn't know about etcdctl -C key...
127.0.0.1 will removed in next commit (now testing multiple etcd hosts)

timothysc:
Can anyone verify that this is still ok in a fqdn setup?

jasonbrooks:
In my tests w/ these commits and CentOS 7, I had to remove kube-master from the localhost line in my `/etc/hosts`. Without that change, etcd seemed to treat kube-master as 127.0.0.1 rather than as the public IP.

As for multi-etcd, these coreos docs provide guidance: https://coreos.com/etcd/docs/latest/clustering.html. They're citing values like:

``` 
-listen-client-urls http://10.0.1.10:2379,http://127.0.0.1:2379 \
  -advertise-client-urls http://10.0.1.10:2379 \
```

And providing for the cluster w/ other values like:

```
-initial-cluster infra0=http://10.0.1.10:2380,infra1=http://10.0.1.11:2380,infra2=http://10.0.1.12:2380
```

eparis:
I'm thinking we want to use `ansible_default_ipv4.address` and `hostvars[host]['ansible_default_ipv4']['address']`   @pmorie did that work for you?  @rrati for you?

rrati:
Switching to ansible_default_ipv4.address and hostvars[host]['ansible_default_ipv4'] solved my etcd problem.

eparis:
closing this PR, as I believe this should be working now.

PR #20, opened 2015-08-24
Add rules for flannel and kube-proxy traffic and re-enable support for firewalld

jlebon:
Small patches to get flannel to work on Atomic.

### DIFF
--- a/ansible/roles/flannel/handlers/main.yml
+++ b/ansible/roles/flannel/handlers/main.yml
@@ -5,16 +5,17 @@
     - stop docker
     - delete docker0
     - start docker
-  when: inventory_hostname in groups['nodes']
 
 - name: stop docker
   service: name=docker state=stopped
+  when: '"docker0" in ansible_interfaces'
 
 - name: delete docker0
   command: ip link delete docker0
-  ignore_errors: yes
+  when: '"docker0" in ansible_interfaces'
 
 - name: start docker
   service: name=docker state=started
-  # This might fail if docker isn't installed yet
-  ignore_errors: yes
+  # NB: even if docker0 is deleted above, ansible_interfaces is not affected
+  # since it is set once at the beginning of the role
+  when: '"docker0" in ansible_interfaces'
--- a/ansible/roles/flannel/tasks/firewalld.yml
+++ b/ansible/roles/flannel/tasks/firewalld.yml
@@ -1,0 +1,26 @@
+---
+- name: Open firewalld vxlan port for flanneld
+  firewalld: port=8472/udp permanent=false state=enabled
+  # in case this is also a node where firewalld turned off
+  ignore_errors: yes
+
+- name: Save firewalld vxlan port for flanneld
+  firewalld: port=8472/udp permanent=true state=enabled
+  # in case this is also a node where firewalld turned off
+  ignore_errors: yes
+
+  # unfortunately, the ansible firewalld module doesn't support direct rules so
+  # we have to manually invoke firewall-cmd
+- name: Open flanneld subnet traffic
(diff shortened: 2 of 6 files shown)


ansible/roles/flannel/handlers/main.yml
```diff
@@ -5,7 +5,6 @@
     - stop docker
     - delete docker0
     - start docker
-  when: inventory_hostname in groups['nodes']
```
eparis: docker is no necessarily install on non-nodes, so this is not acceptable as I recall...

ansible/roles/flannel/tasks/firewalld.yml
```diff
@@ -0,0 +1,5 @@
+---
+# https://bugzilla.redhat.com/show_bug.cgi?id=1033606 and I think others say firewalld+docker == bad
+# It is also disabled by the node role.
+- name: disable firewalld
+  service: name=firewalld enabled=no state=stopped
```
eparis: Is there an actual incompatibility here? Docker has been fixed. If there is a flannl incompatibility here fix it, don't paper over it. Let kube turn off firewalld (as it does today) until https://github.com/kubernetes/kubernetes/pull/12396 lands.

jlebon: So should we reverse the similar disabling done in the nodes role as well?

ansible/roles/flannel/handlers/main.yml
jlebon: Understood. As it stands, the docker bridge on the master is on the wrong subnet, but I suppose as long as we don't actually use docker there, it can last until the next reboot.

ansible/roles/flannel/tasks/firewalld.yml
eparis: or dream up a way to know if docker is running....   @detiber

detiber: Okay, so within openshift-ansible we disabled firewalld across the board, mainly as a way to ensure that it is disabled before we install, start and configure iptables. The tricky part being that iptables doesn't provide a native or supported way to atomically update and persist firewall rules. It looks like the --wait flag was added as a workaround, but it seems like we still have plenty of room for race conditions with updating the firewall rules.

We sort of worked around this by only updating a single chain, to limit the potential for rule change collisions to happen only when adding the jump rule to our new chain.

With that background out of the way...  openshift-sdn currently manages adding it's own port to the firewall (https://github.com/openshift/openshift-sdn/blob/1694e9adbaae43a42944de1c5243a12dca46a535/ovssubnet/controller/kube/bin/openshift-sdn-kube-subnet-setup.sh for example), so we aren't currently managing those rules through openshift-ansible.

As for detecting docker running, I believe there is also code for that in openshift-sdn when setting up the bridge, but @rajatchopra would be able to better speak to that.

I'd say the more you can avoid messing with iptables rules in ansible the better.

jlebon:
I rebased and updated the PR as follow:

1. Detect whether docker was running before restarting it in the flannel role.
2. Re-enable firewalld support in the node role.
3. Implement firewalld rules for the flannel traffic.
4. Add rules to allow kube-proxy redirected traffic coming from the docker containers.

I successfully tested this on the latest AH tree, as well as with a custom tree with firewalld.

ansible/roles/flannel/handlers/main.yml
```diff
@@ -5,16 +5,17 @@
...
 
 - name: stop docker
   service: name=docker state=stopped
+  when: '"docker0" in ansible_interfaces'
```
eparis: awesome/WAY better way to solve the problem.

PR #22, opened 2015-08-26
Ensure that kube-apiserver starts after etcd

sgallagher:
If etcd is running on the same system as kube-apiserver and is
being started as part of the same series of services (such as with a
custom systemd target), we need to ensure that it orders later in
startup than etcd or it will fail to find the local service.

Note: this does not attempt to start etcd on the local system if it
is not already part of the transaction, so it will have no effect
if someone calls `systemctl start kube-apiserver.service` directly.

This is the master-branch version of https://github.com/kubernetes/kubernetes/pull/12939/

### DIFF
--- a/init/systemd/kube-apiserver.service
+++ b/init/systemd/kube-apiserver.service
@@ -2,6 +2,7 @@
 Description=Kubernetes API Server
 Documentation=https://github.com/GoogleCloudPlatform/kubernetes
 After=network.target
+After=etcd.service
 
 [Service]
 EnvironmentFile=-/etc/kubernetes/config


justinsb:
This seems sensible.  A question though: don't we also need this with other configurations like Salt?

e.g. https://github.com/kubernetes/kubernetes/blob/master/cluster/saltbase/salt/kube-apiserver/kube-apiserver.manifest

I don't see how we introduce the dependency that etcd must run first when we're running everything under the kubelet (i.e. the 'kube-up' configuration).  If so, should kube-apiserver in fact tolerate etcd not being started yet?

sgallagher:
I'm not an expert here; I was trying to start up kubernetes locally on my Fedora system and discovered that the order the local services started mattered (but only in this one case). This local patch solved the problem for me, so I offer it to upstream as a simple solution.

jeffbean:
This would then not work for the configuration of not running etcd on the master nodes (valid use case for production level configurations).

jeffbean:
Oh, and as a note I think these init files are used to create the RPMs that are build for distribution. So RedHat pulls these files in to use for the installable. If we add this it might break the yum installable packages.. 

@eparis do you know if this is the case? or who would know?

sgallagher:
@jeffbean The way systemd works is that it has separate directives for requirement and ordering. The "After=" line only applies if (and only if) etcd.service is also being started as part of the same transaction (such as during system boot-up). If etcd.service isn't also in the transaction, this change has no effect. (Meaning it will not negatively impact cases where etcd isn't running on this system).

I included this in the commit message for the patch.

jeffbean:
@sgallagher Ahh sorry I was just playing with this and that make way more sense.

In that case as least for Ansible, as long as these are the systemd files are brought into the distributed packages, we would not need to update anything.

eparis:
so this is fine, but it shouldn't be a problem/be needed...  the api-server will block for quite a while as I recall waiting on etcd...  So even if etcd starts on the system later the apiserver shouldn't fail and should just wait for it.  (For all the same reasons we cannot promise ordering between machines we shouldn't try to force ordering on a single machine).  Did you actually have trouble (I guess so since you opened this PR)

sgallagher:
@eparis Yes, it hung for a considerable amount of time and then failed. When I added the After=, it completed successfully. I suspect (but cannot support) that there may have been a deadlock occurring. The etcd.service *had* started successfully once the transaction concluded.

eparis:
I'm going to merge, it doesn't hurt, but it seems wrong...

eparis:
s/wrong/should be unneeded/

PR #24, opened 2015-08-27
Add a PR-size munger for github

thockin:
This analyzes the additions/deletions stats and adds a label about the size of
the PR.  The heuristic is totally arbitratry for now.

This makes PR munging much slower because it has to get every PR to get the
stats.  C'est la vie.

@brendandburns @lavalamp

### DIFF
--- a/mungegithub/Makefile
+++ b/mungegithub/Makefile
@@ -1,7 +1,7 @@
 all: push
 
-mungegithub: mungegithub.go
-	CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo -ldflags '-w' ./mungegithub.go
+mungegithub:
+	CGO_ENABLED=0 GOOS=linux godep go build -a -installsuffix cgo -ldflags '-w' ./mungegithub.go
 
 container: mungegithub
 	docker build -t gcr.io/google_containers/mungegithub:0.3 .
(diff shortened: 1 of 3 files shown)


lavalamp:
LGTM

Sad that we're writing our own tools like this :/

derekwaynecarr:
In future, can we label changes that make updates to godeps differently so they are all not XXL?

thockin:
The heuristic is open for changes once this gets merged.  I'll go make the
GH labels

thockin:
GH labels now exist, so this can be merged.  @k8s-oncall to actually run this :)

brendandburns:
We're actually bumping up against quota limits on github...  I'm going to add throttles to both the munger and the submit-queue...

brendandburns:
LGTM.

brendandburns:
This is now running in the jenkins utility cluster.

lavalamp:
Is there a reason why it also comments when it applies a label?

thockin:
Honestly: So I can add gmail labels and can see which PRs are easiest to
review.

eparis:
what are you thoughts on how to make it ignore Godeps?  /me wants to write one to make it ignore automatically generated files...

eparis:
@derekwaynecarr #32 should ignore Godeps

PR #31, opened 2015-08-28
Updating kubelet reviewers.

vishh:


yujuhong:
I didn't know we had this file. LGTM.

PR #32, opened 2015-08-29
Do not counts Godeps changes in the sizer

eparis:
Github hides these in review, we typically don't review them closely, and hopefully the godep checker will tell us if someone isn't pushing what they say they are...

### DIFF
--- a/mungegithub/.gitignore
+++ b/mungegithub/.gitignore
@@ -1,0 +1,1 @@
+mungegithub
--- a/mungegithub/issues/issues.go
+++ b/mungegithub/issues/issues.go
@@ -20,6 +20,8 @@ import (
 	"fmt"
 	"strings"
 
+	"k8s.io/contrib/mungegithub/opts"
+
 	"github.com/golang/glog"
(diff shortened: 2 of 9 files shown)


eparis:
@thockin I can break this into 2 parts, one with all my unrelated changes and one with exactly those directed at making the size labeler ignore Godeps, if you like...   Should be easy to review per commit...

derekwaynecarr:
@eparis - I need to think of other useful things I can randomly ask for that you just go ahead and do!

eparis:
@derekwaynecarr all my PRs got labeled XXL (I also wrote a follow on PR to ignore autogenerated files like docs, deep-copy, conversions, and swagger). But waiting for this to merge first.

eparis:
needs rebase after #37

thockin:
I like it

thockin:
LGTM

@brendandburns does this get auto-deployed or do we have to do something?

thockin:
@k8s-oncall

I don't think this will get merge-botted, so please merge when you're comfortable with the answer from @brendandburns

PR #34, opened 2015-08-29
mungegithub: Fix PR List options

eparis:
Sort: desc

Is not valid according to
https://godoc.org/github.com/google/go-github/github#PullRequestListOptions

So we have been sorting ascending. Continue to do that, but at least
make the options correct/obvious. If we want descending that's fine, but
this PR has NO actual change to the operation of the code


eparis:
@brendandburns since you put that line in there.

Issue #19, opened 2015-08-21
mergebot shouldn't self-assign PRs


ixdy:
I created kubernetes/kubernetes#12950 and didn't assign someone, and the mergebot later assigned me (probably because of the files I was touching).

It probably shouldn't self-assign PRs.

PR #35, opened 2015-08-30
mungegithub: blunderbuss: do not self-assign issues

eparis:
If I submit a PR, probably best if it is not assigned to me.


eparis:
@brendandburns 
Should fix #19

PR #36, opened 2015-08-30
mungegithub: Only do github operations once per PR

eparis:
not once per munger


eparis:
@brendandburns since you wrote it...

PR #37, opened 2015-08-30
mungegithub: Create an options structure instead of individual variables

eparis:
Makes adding new command line args which get passed to mungers much more sane.


eparis:
should be rebased after #36

PR #39, opened 2015-08-30
mungegithub: sizer: Skip generated files

eparis:
We automatically generated deep copies, conversions, docs, bash completions, swagger stuff. Humans don't look at those things closely, so stop counting them in the 'size' of a PR.

To be honest, this doesn't currently ignore the 'docs' but there is an open PR in kubernetes which redoes how we generate docs so this will hopefully start working once that lands. (currently we scatter the '.generated_docs' files all over the repo instead of just one file at the root of the repo, this expects it to be in the root)

### DIFF
--- a/mungegithub/.gitignore
+++ b/mungegithub/.gitignore
@@ -1,0 +1,1 @@
+mungegithub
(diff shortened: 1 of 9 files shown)


eparis:
should be rebased after #32

brendandburns:
@eparis same here, LGTM, I don't think it actually needs a rebase...

eparis:
ahhh, the magic of git

PR #40, opened 2015-08-31
Make the http cache optional and default to false

brendandburns:
Caching PRs seems to mess up the mergeability checks.


eparis:
failed on gofmt

brendandburns:
Fixed...

PR #42, opened 2015-08-31
Add two new mungers for managing the submit-queue

brendandburns:
@eparis @zmerlynn


eparis:
not running gofmt on save huh?  otherwise LGTM.  I wish we called to shippable (don't know if you can with travis) and restarted the job instead of screwing with open/close the PR, but this is certainly easier.

brendandburns:
Done... (argh, need to set up that repo with the right hooks...)

eparis:
oh, I do it in VIM on every `:w`  :)

PR #43, opened 2015-08-31
Update the munger to 0.4

brendandburns:
@thockin @eparis

### DIFF
--- a/mungegithub/run.sh
+++ b/mungegithub/run.sh
@@ -16,7 +16,7 @@
 
 while true; do
     /mungegithub \
-      --pr-mungers=blunderbuss,lgtm-after-commit,remove-needs-rebase \
+      --pr-mungers=blunderbuss,remove-needs-rebase,ok-to-test,ping-ci,size \
       --blunderbuss-config=/blunderbuss.yml \
       --token-file=/etc/secret-volume/token
     sleep 600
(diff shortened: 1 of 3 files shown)


mungegithub/run.sh
```diff
@@ -16,7 +16,7 @@
 
 while true; do
     /mungegithub \
-      --pr-mungers=blunderbuss,lgtm-after-commit,remove-needs-rebase \
```
thockin: did you mean to remove lgtm-after-commit ?

brendandburns: yes, see: https://github.com/kubernetes/contrib/pull/45

thockin: What's the correlation?  I'm missing somethign

brendandburns: We weren't searching enough events, so we removed for some edge cases what
we should not have...

thockin: I can't quite parse that sentence, but if you say it is intentional, I buy it.

thockin:
LGTM

mungegithub/run.sh
brendandburns: silly phone auto-complete...

We weren't searching all of the events, so we didn't find an LGTM when it existed and we would spuriously remove LGTMs...

PR #47, opened 2015-09-01
Move mergeability detection into the main pull loop.

brendandburns:
Add a munger that adds the "needs-rebase" label.

@jlowdermilk


mungegithub/pulls/needs_rebase.go
```diff
@@ -26,31 +24,47 @@ import (
...
+		glog.Infof("Skipping %d since mergeable is nil", *pr.Number)
 	}
 	if pr.Mergeable != nil && *pr.Mergeable {
 		if opts.Dryrun {
 			glog.Infof("Would have removed needs-rebase for %d", *pr.Number)
 		} else {
-			client.Issues.RemoveLabelForIssue(opts.Org, opts.Project, *pr.Number, "needs-rebase")
+			client.Issues.RemoveLabelForIssue(opts.Org, opts.Project, *pr.Number, needsRebase)
+		}
+	}
```
eparis: Why a whole new munger instead of just doing it here? We will always trigger one of the other of these 2 mungers, so the HasLabel() early exit stuff isn't actually helping anything.  Doesn't really matter, but other mungers, like the sizer, will remove and add new labels rather than having two mungers messing with the same labels...

eparis:
The command "./hack/verify-gofmt.sh" exited with 1.

https://github.com/fatih/vim-go

mungegithub/pulls/needs_rebase.go
brendandburns: Done.  (I had thought it would be an ugly mess of if/else conditions, but it's not that bad...)

brendandburns:
Comments addressed (and gofmt'd https://github.com/GoClipse/goclipse/issues/142)

jlowdermilk:
lgtm

PR #51, opened 2015-09-02
Add a few more top level catch-alls so that we don't miss on some PRs

brendandburns:
@thockin @smarterclayton @bgrant0607 @derekwaynecarr 

Moar owners for moar issuez

### DIFF
--- a/mungegithub/blunderbuss.yml
+++ b/mungegithub/blunderbuss.yml
@@ -118,12 +118,30 @@ prefixMap:
     - jdef
     - karlkfi
 
-  hack/:
+  docs:
+    - thockin
+    - smarterclayton
+    - brendandburns
+    - bgrant0607
+
+  cluster:
+    - robertbailey
+    - davidopp
+    - zmerlynn
+    - brendandburns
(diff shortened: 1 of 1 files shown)


mungegithub/blunderbuss.yml
```diff
@@ -118,6 +118,24 @@ prefixMap:
...
+    - zmerlynn
+    - brendandburns
+    - eparis
+
+  examples:
+    - bgrant0607
+    - thockin
+    - brendandburns
+
   hack/:
```
bgrant0607: Does it matter that hack ends in `/`, but other directories don't?

bgrant0607:
LGTM

mungegithub/blunderbuss.yml
brendandburns: only in that it confuses people.  fixed.

eparis:
you can add me to hack if you want.

eparis:
but LGTM

brendandburns:
done, and merged on green.

PR #52, opened 2015-09-02
Add support for a 'safe-to-merge' label

brendandburns:
@ixdy @lavalamp @quinton-hoole

### DIFF
--- a/submit-queue/github/github.go
+++ b/submit-queue/github/github.go
@@ -77,7 +77,7 @@ func MakeClient(token string) *github.Client {
 	return github.NewClient(client)
 }
 
-func hasLabel(labels []github.Label, name string) bool {
+func HasLabel(labels []github.Label, name string) bool {
 	for i := range labels {
 		label := &labels[i]
 		if label.Name != nil && *label.Name == name {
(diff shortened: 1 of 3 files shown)


davidopp:
Will it be confusing if we have ok-to-merge and safe-to-merge?

eparis:
yes  :)

brendandburns:
I'm open to other names.  But I'd like a marker for PRs that can't possibly effect e2e (e.g. all Markdown PRs)

brendandburns:
This is so that we can purge the queue and focus efforts on the PRs that really need it.

I can also add in heuristics (e.g. if all files end in `.md`) if that is preferable to an explicit label.

thockin:
docs-only ?

no-e2e-required?

quinton-hoole:
What thockin said.

brendandburns:
switched to `e2e-not-required`

thockin:
should we maybe flip the polarity and annotate any PR which does NOT match
this as "e2e-required: yes"?

brendandburns:
There's no automated matching right now...  It requires manual reviewer intervention.

(but yes, if it was automatic, flipping the parity might make sense)

quinton-hoole:
LGTM!  Thanks @brendandburns 
For what it's worth, I like the parity as it is ("don-require-e2e", rather than "require-e2e").  And I like the explicit label rather than implicit opaque heuristics.

thockin:
I meant to imply that every PR might require exactly one of "e2e-required:
yes" or "e2e-required: no" eventually

brendandburns:
Sure, that's fine, we can write a munger to ask the reviewer for a label...  (or attempt to heuristically add it's own)

--brendan

PR #53, opened 2015-09-02
fix issue kubernetes/kubernetes#12663

3cky:
podex: support Kubernetes v1 API / Docker Registry 2.0
podex: support insecure HTTP/self-signed registries


brendandburns:
LGTM, thanks for the PR!

PR #54, opened 2015-09-02
Ansible: Fix docker startup on btrfs

Arano-kai:
Also, move startup of docker dependent services to handlers

### DIFF
--- a/ansible/roles/docker/handlers/main.yml
+++ b/ansible/roles/docker/handlers/main.yml
@@ -2,7 +2,6 @@
 - name: restart docker
   command: /bin/true
   notify:
-    - check docker existence
     - stop docker
     - delete docker0
     - start docker
@@ -10,17 +9,9 @@
 - name: start docker
   command: /bin/true
   notify:
-    - check docker existence
     - docker graph driver
     - docker fix start on btrfs
     - docker start service
-
-- name: check docker existence
-  command: "docker version"
-  # if rc!=2 (No such file or directory), assume that docker has installed
-  register: has_docker
-  changed_when: false
-  failed_when: false
 
 - name: stop docker
   service: name=docker state=stopped
@@ -34,7 +25,6 @@
 - name: docker graph driver 
   command: findmnt -no FSTYPE -T /var/lib/docker
   register: docker_driver
-  when: has_docker.rc|int != 2
   changed_when: false
 
   # TODO: fail if no --selinux-enabled in 'OPTIONS'
@@ -43,8 +33,7 @@
     dest="{{ docker_config_dir }}/docker"
     regexp="^(OPTIONS=.*)(--selinux-enabled(?!=false)([^\s\"\']*))"
     replace="\1--selinux-enabled=false"
-  when: has_docker.rc|int != 2 and docker_driver.stdout == 'btrfs'
+  when: docker_driver.stdout == 'btrfs'
 
 - name: docker start service
   service: name=docker state=started
-  when: has_docker.rc|int != 2
--- a/ansible/roles/docker/handlers/main.yml
+++ b/ansible/roles/docker/handlers/main.yml
@@ -1,3 +1,50 @@
 ---
 - name: restart docker
-  service: name=docker state=restarted
+  command: /bin/true
+  notify:
+    - check docker existence
+    - stop docker
+    - delete docker0
+    - start docker
+
+- name: start docker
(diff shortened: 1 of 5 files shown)


ansible/roles/docker/handlers/main.yml
```diff
@@ -1,3 +1,50 @@
...
+
+- name: start docker
+  command: /bin/true
+  notify:
+    - check docker existence
+    - docker graph driver
+    - docker fix start on btrfs
+    - docker start service
+
+- name: check docker existence
```
eparis: If we are running the docker role, doesn't that mean docker must be installed (or we would have installed it?)

ansible/roles/docker/handlers/main.yml
```diff
@@ -1,3 +1,50 @@
...
+  service: name=docker state=stopped
+  when: '"docker0" in ansible_interfaces'
+
+- name: delete docker0
+  command: ip link delete docker0
+  when: '"docker0" in ansible_interfaces'
+
+  # TODO: work only on default installation...
+- name: docker graph driver 
+  command: findmnt -no FSTYPE -T /var/lib/docker
```
eparis: How about looking at the output of `docker info` and seeing what the graph driver is?

ansible/roles/flannel/handlers/main.yml
```diff
@@ -14,8 +35,21 @@
   command: ip link delete docker0
   when: '"docker0" in ansible_interfaces'
 
-- name: start docker
+  # TODO: work only on default installation...
+- name: docker graph driver 
+  command: findmnt -no FSTYPE -T /var/lib/docker
```
eparis: I don't want to duplicate extra stuff from the docker role in flannel. Can we just leave the network stuff here and not do the BTRFS stuff?  Do we need to do the docker0 stuff in the docker role?

ansible/roles/docker/handlers/main.yml
Arano-kai: if docker daemon not started (as on btrfs without `--selinux-enabled=false`) `docker info` will fail

ansible/roles/flannel/handlers/main.yml
Arano-kai: Since flannel set-up invokes docker restart, I assume fix for docker goes here respectively.
There is possibility that docker installed before docker role (manually/by other play etc).
Solution for code de-duplication You can see in my fork on ansible/rewrite

ansible/roles/docker/handlers/main.yml
Arano-kai: Going at night with red eyes =P
Will remove that check here.

ansible/roles/docker/handlers/main.yml
```diff
@@ -1,3 +1,39 @@
...
+  register: docker_driver
+  changed_when: false
+
+  # TODO: fail if no --selinux-enabled in 'OPTIONS'
+- name: docker fix start on btrfs
+  replace:
+    dest="{{ docker_config_dir }}/docker"
+    regexp="^(OPTIONS=.*)(--selinux-enabled(?!=false)([^\s\"\']*))"
+    replace="\1--selinux-enabled=false"
```
eparis: do you need a space before the --selinux in case there was no trailing space previously?

ansible/roles/docker/handlers/main.yml
eparis: obviously.

ansible/roles/docker/handlers/main.yml
Arano-kai: Nope, since regex operate exact on `--selinux-enabled` substring.
No trailing space before is out-of-scope for this fix.

ansible/roles/docker/handlers/main.yml
Arano-kai: And obviously docker may fail by any other reason, not only because btrfs. (KO)
We touch part of config here only in case btrfs.
Also, I think we need act in same way in `Turn down docker logging` task (change only `--log-level` or add it if missing).

ansible/roles/flannel/handlers/main.yml
Arano-kai: `docker0` rise only after docker successfully start-up, so we **can** use it in stop/remove tasks and **can't** in fix/start (first run)

PR #55, opened 2015-09-03
Push a new submit queue.

brendandburns:
@thockin @quinton-hoole @ixdy 

Update the submit queue, and add more jenkins-jobs to monitor.


quinton-hoole:
Two of those Jenkins jobs are not currently green.  I assume that nothing will auto-merge until they're fixed?  Is this by design?

quinton-hoole:
@brendandburns  ^^

brendandburns:
Yep, that's by design...

eparis:
closing just merged something else called 0.4.   crap.

ixdy:
we lost the changes to `--jenkins-jobs` then...

PR #59, opened 2015-09-03
Include all users with commit access in the submit queue whitelist

eparis:
Continue to include those in the whitelist.txt


eparis:
@mikedanese @lavalamp @brendandburns You three have played in the submit queue

submit-queue/whitelist.txt
```diff
@@ -1,41 +1,6 @@
-brendandburns
-thockin
-mikedanese
-a-robinson
-saad-ali
-lavalamp
```
lavalamp: I think your script is trigger happy...

eparis: you have commit access, so you don't need to be in the whitelist.

eparis: better explanation, the point of the PR was to pull people with commit access from github and consider them 'whitelisted'  I'm not trying to maintain that list in the whitelist.txt.  We just pull it when the submit-queue launches. So now we only have to maintain in this file those people without commit access that we want to be able to use the submit queue without the extra label.

lavalamp: Oh, I see, sorry!

submit-queue/whitelist.txt
```diff
@@ -1,41 +1,6 @@
...
-markturansky
-pmorie
-yujuhong
-roberthbailey
-vishh
-deads2k
-bprashanth
-cjcullen
-liggitt
 bgrant0607
```
lavalamp: bgrant0607 doesn't have committ access?

lavalamp:
LGTM

lavalamp:
I think we shouldn't have enough committers that this would make us run out of quota.

submit-queue/whitelist.txt
eparis: Surprised me so I had to look it up. https://github.com/orgs/kubernetes/teams/kubernetes-maintainers  only bgrant0607-nocc has commit access.

eparis:
Should be about 3ish github API calls each time you start the submit-queue daemon. Not even 3 per poll period.

submit-queue/whitelist.txt
bgrant0607: I could fix that.

eparis: doesn't matter from my PoV.

lavalamp:
Planning on merging this and rebuilding/starting the merge queue in the coming hour or two.

lavalamp:
Merge queue has been restarted.

lavalamp:
When DNS propagates, the submit queue status should be at http://submit-queue.k8s.io/

Until then, it's at http://104.154.50.45/.

Clearly it could stand to be improved.

lavalamp:
Printing this error now. @brendandburns perhaps the token that you gave it is overly limited? Or maybe that's by design.

```
E0905 00:17:00.926084       1 github.go:263] GET https://api.github.com/orgs/kubernetes/teams?page=1&per_page=100: 403 Must have admin rights to Repository. []
```

PR #61, opened 2015-09-04
Skip CI check when a PR is not mergeable

### LINKED ISSUE #60: mungegithub e2e re-starter in endless loop
For example
https://github.com/kubernetes/kubernetes/pull/13393

My guess is the user in question is not in the e2e whitelist and thus e2e will only run when told to run. Open/close will do nothing.

Where is the e2e whitelist stored?

ixdy:
Travis and Shippable will always fail, so there's no point in trying to trigger them.

Should fix #60.

@eparis @brendandburns @lavalamp

### DIFF
--- a/mungegithub/pulls/ping_ci.go
+++ b/mungegithub/pulls/ping_ci.go
@@ -40,6 +40,10 @@ func (PingCIMunger) MungePullRequest(client *github.Client, pr *github.PullReque
 	if !HasLabel(issue.Labels, "lgtm") {
 		return
 	}
+	if pr.Mergeable == nil || !*pr.Mergeable {
+		glog.Infof("skipping CI check for %d since mergeable is nil or false", *pr.Number)
+		return
+	}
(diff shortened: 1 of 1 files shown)


eparis:
while it would work around the example PRs loop, why isn't travis reporting success? https://travis-ci.org/kubernetes/kubernetes/builds/78763788 says it was successful...

ixdy:
That's for kubernetes/kubernetes#13570, a different PR?

The problematic one is kubernetes/kubernetes#13393.

ixdy:
(And on kubernetes/kubernetes#13570, it is reporting success as expected.)

eparis:
oh, duh, so the question I guess is, why the heck isn't travis paying any attention to 13393?  shouldn't it be failing rather than ignoring?

or is travis ignoring unmergable our intended

ixdy:
I'm not really sure. I'm guessing Travis just doesn't bother to waste cycles trying to merge an unmergeable PR. Shippable should probably be doing the same.

eparis:
having now read: http://docs.travis-ci.com/user/pull-requests/ this is the right fix I guess.

PR #62, opened 2015-09-04
Fix --once to actually run once instead of never running

eparis:


eparis:
@lavalamp easy enough for the next cut, doesn't actually matter....

lavalamp:
LGTM

PR #63, opened 2015-09-04
update build and .yaml for 0.4

lavalamp:
Also add a service to expose the queue status.

### DIFF
--- a/submit-queue/pod.yaml
+++ b/submit-queue/pod.yaml
@@ -3,9 +3,12 @@ kind: Pod
 metadata:
   name: submit-queue
   namespace: default
+  labels:
+    app: submit-queue
+    version: "0.4"
 spec:
   containers:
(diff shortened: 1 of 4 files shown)


lavalamp:
@eparis @ixdy

submit-queue/pod.yaml
```diff
@@ -3,9 +3,12 @@ kind: Pod
 metadata:
   name: submit-queue
   namespace: default
+  labels:
+    app: submit-queue
+    version: "0.4"
```
eparis: I'll have to think about how to generate this in the makefile, where we know the Tag

eparis:
#55 called something 0.5.  not sure if brendan actually is running something based on that...

lavalamp:
Huh, that is weird. Well, the 0.4 that's running now is one I built!

PR #64, opened 2015-09-04
clean up code; add more status messages

lavalamp:


lavalamp:
@brendandburns @eparis

submit-queue/github/github.go
```diff
@@ -308,16 +305,28 @@ func ForEachCandidatePRDo(client *github.Client, user, project string, fn PRFunc
...
 		if !HasLabels(issue.Labels, []string{"lgtm", "cla: yes"}) {
+			glog.V(2).Infof("Skipping %d - doesn't have requisite labels", *pr.Number)
 			continue
 		}
-		if !HasLabel(issue.Labels, config.WhitelistOverride) && !userSet.Has(*prs[ix].User.Login) {
-			glog.V(4).Infof("Dropping %d since %s isn't in whitelist and %s isn't present", *prs[ix].Number, *prs[ix].User.Login, config.WhitelistOverride)
-			if _, _, err := client.Issues.AddLabelsToIssue(user, project, *prs[ix].Number, []string{"needs-ok-to-merge"}); err != nil {
-				glog.Errorf("Failed to set 'needs-ok-to-merge' for %d", *prs[ix].Number)
+
+		pr, _, err = client.PullRequests.Get(user, project, *pr.Number)
```
ixdy: are you intentionally shadowing `pr` here?
I guess at least comment as to why you're doing this?

submit-queue/github/github.go
```diff
@@ -308,16 +305,28 @@ func ForEachCandidatePRDo(client *github.Client, user, project string, fn PRFunc
 
 		glog.V(8).Infof("%v", issue.Labels)
 		if !HasLabels(issue.Labels, []string{"lgtm", "cla: yes"}) {
+			glog.V(2).Infof("Skipping %d - doesn't have requisite labels", *pr.Number)
```
ixdy: seems like this may be very spammy for level 2?
also not sure that "requisite labels" is very meaningful without seeing the code.

submit-queue/submit-queue.go
```diff
@@ -131,11 +135,11 @@ func (e *e2eTester) runE2ETests(client *github_api.Client, pr *github_api.PullRe
...
 	// if there is a 'safe-to-merge' label, just merge it.
 	if len(*dontRequireE2E) > 0 && github.HasLabel(issue.Labels, *dontRequireE2E) {
-		e.msg(fmt.Sprintf("Merging %d since %s is set", *pr.Number, *dontRequireE2E))
+		e.msg("Merging %d since %s is set", *pr.Number, *dontRequireE2E)
 		return e.merge(client, org, project, pr)
 	}
 	// Ask for a fresh build
-	e.msg(fmt.Sprintf("Asking PR builder to build %d", *pr.Number))
+	e.msg("Asking PR builder to build %d", *pr.Number)
 	body := "@k8s-bot test this [testing build queue, sorry for the noise]"
```
ixdy: while you're here, maybe remove "[testing build queue, sorry for the noise]"?

submit-queue/submit-queue.go
```diff
@@ -152,42 +156,43 @@ func (e *e2eTester) runE2ETests(client *github_api.Client, pr *github_api.PullRe
...
-	if data, err := json.Marshal(e); err != nil {
-		res.WriteHeader(http.StatusInternalServerError)
-		res.Write([]byte(err.Error()))
-	} else {
-		res.WriteHeader(http.StatusOK)
-		res.Write(data)
-	}
+	e.locked(func() {
+		res.Header().Set("Content-type", "application/json")
+		if data, err := json.MarshalIndent(e, "", "\t"); err != nil {
```
ixdy: it probably doesn't really matter for this app, but don't we only need to lock the `json.MarshalIndent(e...)` bit? the rest of the HTTP stuff shouldn't need to be locked.

submit-queue/submit-queue.go
eparis: @thockin wanted that to make filtering easier/possible in e-mail....

ixdy: is he filtering on that text, or based on @k8s-merge-bot saying it?

eparis: couldn't say...

submit-queue/submit-queue.go
lavalamp: Just for you-- I'll fix it

lavalamp:
Comments addressed, and functionality improved.

lavalamp:
This is such a big improvement that I think I'll start it running without waiting for LGTM. @brendandburns

lavalamp:
I built 0.5; it's running at http://submit-queue.k8s.io/

Looks much better.

lavalamp:
We need to make the status show the whitelist, so we can verify that it's getting constructed correctly.

lavalamp:
I've pushed more changes to this branch, since the whitelist stuff was kinda broken. I'm about to start this up as 0.6.

eparis:
before you push a .6 can we try to run down what's up in https://github.com/kubernetes/kubernetes/pull/13187 ? message every 30 minutes?

lavalamp:
@eparis I fixed that already, actually-- noticed it would do that when I was making it remove the needs-ok-to-merge label.

lavalamp:
I updated the commit message to reflect that change.

lavalamp:
@brendandburns when it finds the "doesn't need e2e" label, it merges immediately. Shouldn't it rerun shippable, etc? It seems like that could break the build.

eparis:
LGTM

eparis:
separate thing, but maybe we should check whitelist.txt against the pulled information, because we know whitelist.txt is going to get out of date...

lavalamp:
Yeah, I was just thinking that. Maybe the generate command could do that.

lavalamp:
This has an LGTM so I'm going to merge it, just so the code that is running is checked in. @ixdy if you have any further requests I'll fix them in a follow-up.

PR #65, opened 2015-09-08
De-dup whitelist when committers list is generated

lavalamp:
@eparis This one is for you.


eparis:
Wow, it was already out of date! guess that wasn't a waste of time.

I don't love what you have to do/are doing to save comments, but it seems functionally correct. Do we really need to allow comments in the whitelist or can we simplify and both when we get there?

eparis:
If you like it this way, you can call it LGTM

lavalamp:
I can get rid of that if you want-- basically I just wanted a header on the committers one so people didn't accidentally edit it.

eparis:
I totally get the header in the committers, and a header at the start of whitelist telling how to clean it wouldn't be bad, but neither would require us to read in comments and print them back out.  We can just write that as the first line when we write out the files...   I think it'd be worth it to kill the comment handling code in this pr...

lavalamp:
I'll just take it out.

lavalamp:
Updated.

eparis:
LGTM

PR #67, opened 2015-09-09
OpenContrail playbook

pedro-r-marques:
This pull request adds support to provision OpenContrail via ansible. The support has been tested with Fedora 22.

The user is expected to define the following variables in group_vars/all.yml:

```
networking: opencontrail

opencontrail_public_subnet: <External subnet for LoadBalancer services>
opencontrail_private_subnet: <Internal subnet with connectivity to the kubelets>
opencontrail_gateway: <optional | gateway address on private_subnet>
opencontrail_interface: eth1
```

The provisioning assumes that the system has two interfaces: a management interface (e.g. on board 1G interface) and 1 data interface (e.g. 2x10G bond0 interface).

To configure the software gateway that provides connectivity between the external network and the private virtual networks a new type of host "gateway" is defined. Example inventory file:

```
[masters]
roque-kube-master

[etcd]
roque-kube-master

[nodes]
roque-node3 ipaddr=192.168.1.3/24
roque-node4 ipaddr=192.168.1.4/24

[gateways]
roque-gateway ipaddr=192.168.1.254/24
```
## Details
- OpenContrail uses kubelet to execute its userspace components. This implies running kubelet on the master, similarly to the approach taken by salt provisioning scripts which use kubelet to run the kubernetes api-server, etc.
- The OpenContrail kernel module is compiled on-demand for the target.
- OpenContrail control plane components are distributed as containers (via docker hub).

### DIFF
--- a/ansible/cluster.yml
+++ b/ansible/cluster.yml
@@ -25,9 +25,17 @@
     - nodes
   sudo: yes
   roles:
-    - flannel
+    - { role: flannel, when: networking == 'flannel' }
   tags:
     - flannel
+
+# install opencontrail
+- hosts: all
+  sudo: yes
+  roles:
+    - { role: opencontrail, when networking == 'opencontrail'}
+  tags:
+    - opencontrail
 
 # install kube master services
 - hosts: masters
@@ -53,3 +61,13 @@
     - node
   tags:
     - nodes
+
+# provision opencontrail once the services are operational
+- hosts:
+    - masters[0]
+    - nodes
+  sudo: yes
+  roles:
+    - { role: opencontrail-provision, when: networking == 'opencontrail' }
+  tags:
+    - opencontrail
--- a/ansible/roles/master/handlers/main.yml
+++ b/ansible/roles/master/handlers/main.yml
@@ -22,3 +22,6 @@
 
 - name: restart iptables
   service: name=iptables state=restarted
+
+- name: restart kubelet
+  service: name=kubelet state=restarted
--- a/ansible/roles/master/tasks/localBuildInstall.yml
+++ b/ansible/roles/master/tasks/localBuildInstall.yml
@@ -1,7 +1,7 @@
 ---
 - name: Copy master binaries
   copy:
-    src: ../../kubernetes/_output/local/go/bin/{{ item }}
+    src: "{{ localBuildOutput }}/go/bin/{{ item }}"
     dest: /usr/bin/
     mode: 755
   with_items:
@@ -9,6 +9,7 @@
     - kube-scheduler
     - kube-controller-manager
     - kubectl
+    - kubelet
   notify: restart daemons
 
 - name: Copy master service files
@@ -20,6 +21,7 @@
     - kube-apiserver.service
     - kube-scheduler.service
     - kube-controller-manager.service
+    - kubelet.service
   notify: reload systemd
 
 - name: Copy systemd tmpfile for apiserver
@@ -28,3 +30,8 @@
     dest: /etc/tmpfiles.d/
     mode: 644
   notify: reload systemd
+
+- name: Create the /var/lib/kubelet working directory
+  file:
+    path: /var/lib/kubelet
+    state: directory
--- a/ansible/roles/master/tasks/main.yml
+++ b/ansible/roles/master/tasks/main.yml
@@ -70,6 +70,14 @@
 - name: write the kubecfg (auth) file for kubectl
   template: src=kubectl.kubeconfig.j2 dest={{ kube_config_dir }}/kubectl.kubeconfig
 
+- name: write the config files for kubelet
+  template: src=kubelet.j2 dest={{ kube_config_dir }}/kubelet
+  notify:
+    - restart kubelet
+
+- name: Enable kubelet
+  service: name=kubelet enabled=yes state=started
+
 - include: firewalld.yml
   when: has_firewalld
 
--- a/ansible/roles/master/templates/kubelet.j2
+++ b/ansible/roles/master/templates/kubelet.j2
@@ -1,0 +1,17 @@
+###
+# kubernetes kubelet (node) config
+
+# The address for the info server to serve on (set to 0.0.0.0 or "" for all interfaces)
+KUBELET_ADDRESS="--address=0.0.0.0"
+
+# The port for the info server to serve on
+# KUBELET_PORT="--port=10250"
+
+# You may leave this blank to use the actual hostname
+KUBELET_HOSTNAME="--hostname-override={{ inventory_hostname }}"
+
+# location of the api-server
+KUBELET_API_SERVER="--api-servers=http://localhost:{{ kube_master_insecure_port }}"
+
+# Add your own!
+KUBELET_ARGS="--register-node=false --config={{ kube_manifest_dir }}"
--- a/ansible/.gitignore
+++ b/ansible/.gitignore
@@ -1,3 +1,4 @@
 myinventory
 inventory
 *.swp
+*~
--- a/ansible/group_vars/all.yml
+++ b/ansible/group_vars/all.yml
@@ -37,6 +37,18 @@ insecure_registrys:
 # addresses do not need to be routable and must just be an unused block of space.
 kube_service_addresses: 10.254.0.0/16
 
+# Network implementation (flannel|opencontrail)
+networking: flannel
+
+# External network
+# opencontrail_public_subnet: 10.1.4.0/24
+
+# Underlay network
+# opencontrail_private_subnet: 192.168.1.0/24
+
+# Data interface
+# opencontrail_interface: eth1
+
 # Flannel internal network (optional). When flannel is used, it will assign IP
 # addresses from this range to individual pods.
 # This network must be unused in your network infrastructure!
--- a/ansible/roles/kubernetes-addons/tasks/main.yml
+++ b/ansible/roles/kubernetes-addons/tasks/main.yml
@@ -65,6 +65,7 @@
   template: src=kube-addons.service.j2 dest=/etc/systemd/system/kube-addons.service
   notify:
     - reload and restart kube-addons
+  when: not (ansible_distribution == "Ubuntu" and ansible_distribution_major_version|int < 15)
 
 - name: Enable and start kube addons
   service: name=kube-addons.service enabled=yes state=started
--- a/ansible/roles/master/defaults/main.yml
+++ b/ansible/roles/master/defaults/main.yml
@@ -1,0 +1,3 @@
+kube_master_insecure_port: 8080
+
+localBuildOutput: ../../_output/local
--- a/ansible/roles/node/defaults/main.yml
+++ b/ansible/roles/node/defaults/main.yml
@@ -1,0 +1,1 @@
+localBuildOutput: ../../_output/local
--- a/ansible/roles/node/tasks/localBuildInstall.yml
+++ b/ansible/roles/node/tasks/localBuildInstall.yml
@@ -1,7 +1,7 @@
 ---
 - name: Copy node binaries
   copy:
-    src: ../../kubernetes/_output/local/go/bin/{{ item }}
+    src: "{{ localBuildOutput }}/go/bin/{{ item }}"
     dest: /usr/bin/
     mode: 755
   with_items:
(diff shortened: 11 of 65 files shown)


ansible/roles/master/meta/main.yml
```diff
@@ -1,4 +1,5 @@
 ---
 dependencies:
     - { role: common }
+    - { role: docker }
```
eparis: I have been trying to not require docker on the master. Why doid you add this?

ansible/roles/master/handlers/main.yml
```diff
@@ -22,3 +22,6 @@
 
 - name: restart iptables
   service: name=iptables state=restarted
+
+- name: restart kubelet
+  service: name=kubelet state=restarted
```
eparis: Why would there be a kubelet on the master, unless it was already a node, in which case, why would we need this?

ansible/roles/master/tasks/localBuildInstall.yml
```diff
@@ -9,6 +9,7 @@
     - kube-scheduler
     - kube-controller-manager
     - kubectl
+    - kubelet
```
eparis: You'll have to explain why the master must be most of a node (or maybe I'll get it when I read more)

eparis: (or I should read the summary instead of just commenting blindly like a fool, I'm going to delete all my previous comments)

eparis:
@rchopra @detiber

detiber:
@eparis only thing that stands out to me after a quick glance are the 'restart daemon' handlers that just call bin/true

ansible/roles/opencontrail-controller/tasks/config.yml
```diff
@@ -0,0 +1,12 @@
+---
+- name: Install templates
+  get_url:
+    url=https://raw.githubusercontent.com/Juniper/contrail-kubernetes/master/cluster/{{item}}
```
jeffbean: Should these manifests also be committed to the ansible role itself?

If they are changed outside of this repository this is the first instance where ansible is going outside of the kubernetes repo. Seems like triaging any issue around this might be troublesome later on.

ansible/roles/master/templates/kubelet.j2
```diff
@@ -0,0 +1,17 @@
...
+KUBELET_ADDRESS="--address=0.0.0.0"
+
+# The port for the info server to serve on
+# KUBELET_PORT="--port=10250"
+
+# You may leave this blank to use the actual hostname
+KUBELET_HOSTNAME="--hostname-override={{ inventory_hostname }}"
+
+# location of the api-server
+KUBELET_API_SERVER="--api-servers=https://{{ groups['masters'][0] }}:{{ kube_master_api_port }}"
```
jeffbean: Could the master address actually be `127.0.0.1` here?

ansible/roles/opencontrail-controller/handlers/main.yml
```diff
@@ -0,0 +1,6 @@
+---
+- name: restart contrail-control
+  command: docker kill $(docker ps | awk '/opencontrail\/control/ {print $1;}')
```
jeffbean: This handler assumes Docker is installed, but the role `opencontrail-controller` only depends on `kubernetes` so you can not assume Docker is installed.

ansible/roles/opencontrail-gateway/tasks/main.yml
```diff
@@ -0,0 +1,22 @@
...
+  include: redhat.yml
+  when: ansible_distribution == "Fedora"
+
+- name: Create vrouter agent service
+  copy: src=contrail-vrouter-agent.service dest=/etc/systemd/system
+  notify:
+    - reload systemd
+
+- name: Docker pull image
+  command: /usr/bin/docker pull opencontrail/vrouter-agent:2.20
```
jeffbean: Some tasks use  `docker` binary and some commands use full path `/usr/bin/docker`. These should be consistent across the roles.

ansible/roles/opencontrail-controller/tasks/config.yml
pedro-r-marques: Will do.

ansible/roles/master/templates/kubelet.j2
pedro-r-marques: Will do.

ansible/cluster.yml
```diff
@@ -25,10 +25,51 @@
...
   sudo: yes
   roles:
-    - flannel
+    - { role: flannel, when: networking == 'flannel' }
   tags:
     - flannel
 
+# install opencontrail
+- hosts:
+    - localhost
```
jeffbean: This tasks ends up putting a dependency on the running machine to be Linux and in particular certain versions that our `common` and `docker` roles support. If this kind of change is okay with @eparis, then the README  would need some serious changes for this kind of change. 

IMHO this kind of role is something I have tried to stay away from as much as possible.

pedro-r-marques: @jeffbean Would it preferable to run this task in one of the masters ?

jeffbean: If that could be done then that sounds like a better option for sure. :+1:

pedro-r-marques:
@jeffbean Latest update moves the kernel module compile to master; removes dependencies from github.com/Juniper/contrail-kubernetes repository and uses "docker" consistently.

alex-mohr:
@detiber was your quick glance an official lgtm?

ansible/cluster.yml
```diff
@@ -25,10 +25,51 @@
...
+  tags:
+    - opencontrail-kmod
+
+- hosts:
+    - masters
+  gather_facts: yes
+  sudo: yes
+  tags:
+    - opencontrail
+    - opencontrail-gateway
```
eparis: What does this do?

ansible/cluster.yml
```diff
@@ -53,3 +94,18 @@
...
+    - { role: opencontrail-master-provision, when: networking == 'opencontrail' }
+  tags:
+    - opencontrail-provision
+
+- hosts: nodes
+  sudo: yes
+  roles:
+    - { role: opencontrail-node-provision, when: networking == 'opencontrail' }
+  tags:
+    - opencontrail-provision
```
eparis: I would really like if you could find a way to embed this into your role. I know this is probably the 'right' way to do it, but I want to make these high level things look so nice and simple for future hackers.

Is there some way to do just a single opencontrail-install and opencontrail-provision at this level? If it's really going to make those roles terrible that's ok, but I don't like doubling this file size...

ansible/roles/common/tasks/fedora-install.yml
```diff
@@ -1,4 +1,9 @@
 ---
+- name: set package manager
+  set_fact:
+    ansible_pkg_mgr: dnf
+  when: ansible_distribution_major_version|int >= 22
```
eparis: This is a bug in ansible fixes in their development branch. Is this needed?  I'd rather not carry it...

ansible/roles/master/tasks/main.yml
```diff
@@ -70,6 +70,14 @@
 - name: write the kubecfg (auth) file for kubectl
   template: src=kubectl.kubeconfig.j2 dest={{ kube_config_dir }}/kubectl.kubeconfig
 
+- name: write the config files for kubelet
+  template: src=kubelet.j2 dest={{ kube_config_dir }}/kubelet
```
eparis: I'd rather 'hackily' use a src=../../node/templates/kubelet.j2 rather than duplicate the template....

ansible/roles/opencontrail-controller/handlers/main.yml
```diff
@@ -0,0 +1,13 @@
...
+  shell: bash -c "docker ps | awk '/opencontrail\/control/ {print $1;}'"
+  register: control_node_id
+  notify:
+    - kill contrail-control
+
+- name: kill contrail-control
+  command: docker kill {{ control_node_id.stdout }}
+  when: control_node_id.rc == 0 and control_node_id.stdout != ""
+
+- name: restart daemons
```
eparis: are these necessary?

ansible/roles/master/meta/main.yml
eparis: Do we have to put this here, since you do it in `opencontrial-*/meta/main.yml' ?

ansible/roles/opencontrail-gateway/templates/ifcfg-gateway.j2
```diff
@@ -0,0 +1,3 @@
+DEVICE={{ item }}
+TYPE=vhost
+MACADDR=00:00:5e:00:01:00
```
eparis: Stealing some random person's MAC address huh? http://www.coffer.com/mac_find/?string=00%3A00%3A5e

eparis:
couple of comments. Obviously not all need 'changes' but I want to understand.

ansible/cluster.yml
pedro-r-marques: The intent is to be able to use variables such as "ansible_<interface-name>" in order for the nodes to know the ip address of the master.

Example:
```
server={%for host in groups['masters']%}{{ hostvars[host]['ansible_' + opencontrail_interface]['ipv4']['address'] }} {% endfor %}
```

My understanding is that gather_facts has been disabled at the start of the playbook. So this overrides that behavior when running these specific tags.

ansible/cluster.yml
pedro-r-marques: There is an ordering issue that i'm not sure how to solve:
- Some opencontrail node components (e.g. the kubelet plugin) must be installed before the kubelet is started on the nodes;
- These provisioning tasks must execute after the master kubelet is running.

The kubelet on the nodes, to have the "opencontrail" plugin option requires the executable that is going to be called to exist. Ordering is something i really had trouble with. I would appreciate suggestions and/or better ways to do this. I agree that having so many top level roles is undesirable.

ansible/roles/common/tasks/fedora-install.yml
pedro-r-marques: No. I'll back this off... it is a bug that i need to patch in my local sandbox only.

ansible/roles/master/tasks/main.yml
pedro-r-marques: They are different. The master kubelet must not register itself and runs different options.

ansible/roles/master/meta/main.yml
pedro-r-marques: That is required because of the kubelet running on the master. Can i make this conditional ?

ansible/roles/opencontrail-gateway/templates/ifcfg-gateway.j2
pedro-r-marques: VRRP MAC address. These is a reserved mac address for a "virtual router". OpenContrail acts as a distributed virtual router so we present the same mac everywhere.

ansible/cluster.yml
eparis: gather_facts is only disabled for the pre-ansible role.  It runs before the next role (or it sure better be!)

ansible/roles/master/meta/main.yml
eparis: No, I'm asking, since you already pull in the docker role via the opencontrol roles, do you also need to pull it in here? I don't think you can make this conditional really.  I'm asking do we need it here, or was it taken care of in other roles?

ansible/roles/master/tasks/main.yml
eparis: makes sense enough.

ansible/cluster.yml
eparis: I'm just asking, can you do it in 2 top level roles?  Then do the kmod/gateway/controller/node stuff inside that one role?  In teh second top level role you'd have to do the master/node provision.

Two large roles might be horrible for you, if so, this is ok. I know this is the "ansible" way but I'd rather hide opencontrail for non-contrail users....

pedro-r-marques:
@eparis The most recent diff collapses the roles into 2 (install before kubelet is up and provision after the contrail API is running). I also removed a couple of issues that you have pointed out in your previous comment.

pedro-r-marques:
@detiber @eparis ping! any feedback on this review ?

PR #77, opened 2015-09-11
Remove unused fetchAllPRs() from submit-queue/github

eparis:


lavalamp:
Thanks for cleaning up my mess.

LGTM

PR #83, opened 2015-09-16
Blunderbuss: don't panic if less than 4 lines changed

eparis:
Basically the code does:
```
    log := math.Log10($NUM_LINES_CHANGED)
    logPlus := log + .5
    fileWeight := int64(logPlus)
```
So lets assume that $NUM_LINES_CHANGED == 3.
```
    log        == 0.4771212547196624
    logPlus    == 0.9771212547196624
    fileWeight == 0
```
This means that further on weightSum (and other things) will be calculated by
a multiple by fileWeight and thus will get the value of 0. Both

`100.0*float64(potentialOwners[cur])/float64(weightSum)`
  and
`selection := rand.Int63n(weightSum)`

Will cause a panic if weightSum == 0

I solve this by adding 1 to the fileWeight after it is truncated to
int64. So we can never get a value of 0.

### DIFF
--- a/mungegithub/pulls/blunderbuss.go
+++ b/mungegithub/pulls/blunderbuss.go
@@ -103,13 +103,13 @@ func (b *BlunderbussMunger) MungePullRequest(config *config.MungeConfig, pr *git
 		}
 		for _, file := range commit.Files {
 			fileWeight := int64(1)
-			if file.Changes != nil {
+			if file.Changes != nil && *file.Changes != 0 {
 				fileWeight = int64(*file.Changes)
 			}
 			// Judge file size on a log scale-- effectively this
 			// makes three buckets, we shouldn't have many 10k+
 			// line changes.
-			fileWeight = int64(math.Log10(float64(fileWeight)) + .5)
+			fileWeight = int64(math.Log10(float64(fileWeight))) + 1
 			fileOwners := b.config.FindOwners(*file.Filename)
 			if len(fileOwners) == 0 {
 				glog.Warningf("Couldn't find an owner for: %s", *file.Filename)


eparis:
@lavalamp since I think you wrote this part that is panic-ing....

lavalamp:
Oops. I should add one after truncation to int.

eparis:
@lavalamp ok, will fix it that way.

eparis:
@lavalamp updated that way.

eparis:
(although i'm panic-ing on 13901 now. debugging)

eparis:
Somehow it is getting
```
"cluster/docker-cluster/.gitignore" fileWeight: -9223372036854775807
```

lavalamp:
It must be overflowing an int64. Hm.

eparis:
*fileChanges == 0 WTF?

```
fileWeight := int64(1)
    if file.Changes != nil {
        fileWeight = int64(*file.Changes)
}
```

lavalamp:
Hm, maybe a rename?

eparis:
Create `cluster/baremetal/.gitignore`
https://github.com/ZJU-SEL/kubernetes/commit/bf2f19bcb10e1944fd9c60c6456c841aedc858b4
Mv `cluster/baremetal/.gitignore` to `cluster/docker-cluster/.gitignore`
https://github.com/ZJU-SEL/kubernetes/commit/0c620ac8d387a3cf19e9668fe5f44b484b59291e

`cluster/baremetal/.gitignore`  Changes:2 Additions:2 Deletions:0
`cluster/docker-cluster/.gitignore` Changes:0 Additions:0 Deletions:0

eparis:
Apparently log10(0) != panic()

How about we just skip any files where Changes = 0?

eparis:
re-running, but this fixes the 2 I've found so far...

mungegithub/pulls/blunderbuss.go
```diff
@@ -101,13 +101,13 @@ func (b *BlunderbussMunger) MungePullRequest(client *github.Client, pr *github.P
...
 			fileWeight := int64(1)
-			if file.Changes != nil {
+			if file.Changes != nil && *file.Changes != 0 {
 				fileWeight = int64(*file.Changes)
 			}
 			// Judge file size on a log scale-- effectively this
 			// makes three buckets, we shouldn't have many 10k+
 			// line changes.
-			fileWeight = int64(math.Log10(float64(fileWeight)) + .5)
+			fileWeight = int64(math.Log10(float64(fileWeight))+.5) + 1
```
lavalamp: Might as well remove the +.5...

mungegithub/pulls/blunderbuss.go
```diff
@@ -101,13 +101,13 @@ func (b *BlunderbussMunger) MungePullRequest(client *github.Client, pr *github.P
 		}
 		for _, file := range commit.Files {
 			fileWeight := int64(1)
-			if file.Changes != nil {
+			if file.Changes != nil && *file.Changes != 0 {
```
lavalamp: Nice catch.

eparis:
I had 1 clean run through the repo (with --dry-run)

lavalamp:
LGTM

lavalamp:
OK, I'll merge and restart this soon, then.

lavalamp:
#88

PR #84, opened 2015-09-16
Update service-loadbalance api with upstream

aledbf:


aledbf:
Closing. Is not required to update the api. I need to use `godep go build` instead of `go build`

lavalamp:
Actually this looks quite reasonable. godep go build probably works for you because the dependency is old.

lavalamp:
So we should probably update the dependency.

aledbf:
@lavalamp is there a rule/directive on how the dependencies can be updated?

lavalamp:
@aledbf I think this is still up to date: https://github.com/kubernetes/kubernetes/blob/b3bce3aa6088345f57c9ae8c6be5802b114a181d/docs/devel/development.md#godep-and-dependency-management

bprashanth:
I think @eparis (thanks!) is doing this https://github.com/kubernetes/contrib/pull/94

aledbf:
@lavalamp sorry for not read the development guide by complete.

Closing. #94 is merged.

PR #79, opened 2015-09-12
githubmunger: label PRs based on certain files touched

eparis:
Start with just proposals and changes to the API.

PR #87, opened 2015-09-16
Path Based Labeler: Use a config file for the mapping

eparis:
Instead of putting it directly in the go code.


eparis:
This is a followup to #79 which should make it easier to maintain.

eparis:
assigning brendan, since this really should go in right after #79

PR #89, opened 2015-09-16
Fix stale info problems in munger

lavalamp:


lavalamp:
@brendandburns @eparis

mungegithub/pulls/lgtm_after_commit.go
```diff
@@ -72,6 +72,11 @@ func (LGTMAfterCommitMunger) MungePullRequest(config *config.MungeConfig, pr *gi
 		return
 	}
 
+	// Don't remove the LGTM from merged PRs
+	if pr.Merged != nil && *pr.Merged {
```
eparis: This shouldn't happen, right?  We should abort way back up in the ForEachIssueGet() if the second Get() didn't give us an "open" issue...

lavalamp: Hm-- empirically it does happen, but I didn't verify that this would actually fix it.

github/github.go
```diff
@@ -654,10 +654,16 @@ func (config *GithubConfig) ForEachIssueDo(labels []string, fn PRFunction) error
...
 			continue
 		}
-		if err := fn(pr, issue); err != nil {
+		// re-get the issue, so it won't be stale.
+		currentIssue, _, err := config.client.Issues.Get(config.Org, config.Project, *issue.Number)
+		if err != nil {
+			glog.Errorf("Error getting issue %v: %v", *issue.Number, err)
+			continue
+		}
+		if err := fn(pr, currentIssue); err != nil {
```
eparis: Hmmmm, how about we change "fetchAllIssuesWithLables()" to return an []int. We then `Get` each issue at the top of this function and should "recheck" the "open" state here. I'd rather we didn't have an outdate `github.Issue` to confuse us down the road...

eparis:
It generally LGTM, but every munger would need the "skip if already closed" test. I'm also a little concerned that we tested things like `issue.User == nil` on the 'outdated` issue and then loaded a new one. I don't trust github not to send us a nil the second time. So I'd think we'd want those checks after we `re-get()` the issue.

lavalamp:
Yeah, those are all good ideas. I don't think this is urgent, so I may let
it sit for a bit today until I get to it.

lavalamp:
@eparis Updated. PTAL

lavalamp:
(I haven't built/pushed this yet, but I included the updated build file so I don't have to make another PR.)

lavalamp:
I took the liberty of building and starting this.

eparis:
guess we should merge it then  :)

PR #90, opened 2015-09-17
Don't pause if we spent the whole time running

eparis:
Today we expect the queue/mungers to run every "poll period". But in
fact we spend a rather substantial amount of time runnin and then pause
for a poll period of time. This PR will instead only sleep if our work
took less than poll period and will try to start the loop every poll
period.


alex-mohr:
Thanks Eric!

PR #92, opened 2015-09-17
Remove haproxy zombie processes left after a haproxy reload

aledbf:

PR #93, opened 2015-09-17
Add syslog server to servicelb image to get haproxy logs

aledbf:
Is not possible to use a file or device like `/dev/log`, but is possible to use a syslog server.
This allows the redirection of `haproxy` logs to stdout starting a local syslog server that just prints the content of the messages

```
Sep 15 17:45:57 backend-4 sh[25987]: servicelb [INFO] 172.17.144.1:58858 [15/Sep/2015:17:45:54.182] www app-demo/172.17.144.3:5000 3557/0/0/2/3559 200 139 - - ---- 2/2/0/1/0 0/0 "HEAD /v3/health-check HTTP/1.1"
```

### DIFF
--- a/service-loadbalancer/loadbalancer_log.go
+++ b/service-loadbalancer/loadbalancer_log.go
@@ -1,0 +1,69 @@
+/*
+Copyright 2015 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/golang/glog"
+	"github.com/ziutek/syslog"
+)
+
+type handler struct {
+	*syslog.BaseHandler
+}
+
+type syslogServer struct {
+	*syslog.Server
+}
+
+// newSyslogServer start a syslog server using a unix socket to listen for connections
+func newSyslogServer(path string) (*syslogServer, error) {
+	glog.Infof("Starting syslog server for haproxy using %v as socket", path)
+	// remove the socket file if exists
+	os.Remove(path)
+
+	server := &syslogServer{syslog.NewServer()}
+	server.AddHandler(newHandler())
+	err := server.Listen(path)
+	if err != nil {
+		return nil, err
+	}
+
+	return server, nil
+}
+
+func newHandler() *handler {
+	h := handler{syslog.NewBaseHandler(1000, nil, false)}
+	go h.mainLoop()
+	return &h
+}
+
+func (h *handler) mainLoop() {
+	for {
+		message := h.Get()
+		if message == nil {
+			break
+		}
+
+		fmt.Printf("servicelb [%s] %s%s\n", strings.ToUpper(message.Severity.String()), message.Tag, message.Content)
+	}
+
+	h.End()
+}
--- a/service-loadbalancer/service_loadbalancer.go
+++ b/service-loadbalancer/service_loadbalancer.go
@@ -128,11 +128,12 @@ type service struct {
 // loadBalancerConfig represents loadbalancer specific configuration. Eventually
 // kubernetes will have an api for l7 loadbalancing.
 type loadBalancerConfig struct {
-	Name      string `json:"name" description:"Name of the load balancer, eg: haproxy."`
-	ReloadCmd string `json:"reloadCmd" description:"command used to reload the load balancer."`
-	Config    string `json:"config" description:"path to loadbalancers configuration file."`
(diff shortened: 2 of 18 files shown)


service-loadbalancer/template.cfg
```diff
@@ -4,6 +4,10 @@ global
     daemon
     stats socket /tmp/haproxy
 
+    # log using a syslog socket
+    log /var/run/haproxy.log.socket local0 info
+    log /var/run/haproxy.log.socket local0 notice
```
bprashanth: What happens if we don't run the server on the other side of the socket but leave this in?

service-loadbalancer/service_loadbalancer.go
```diff
@@ -430,6 +430,12 @@ func main() {
 
 	var kubeClient *client.Client
 	var err error
+
+	_, err = newSyslogServer("/var/run/haproxy.log.socket")
```
bprashanth: Can you please add a -v flag that can turn this off in case it causes disk problems?

service-loadbalancer/loadbalancer_log.go
```diff
@@ -0,0 +1,69 @@
...
+
+type handler struct {
+	*syslog.BaseHandler
+}
+
+type syslogServer struct {
+	*syslog.Server
+}
+
+// NewSyslogServer start a syslog server using a unix socket to listen for connections
```
bprashanth: Please lower case New to match the function name

bprashanth:
Please move the godep to another commit as mentioned in https://github.com/kubernetes/contrib/pull/92#issuecomment-141303112

service-loadbalancer/template.cfg
aledbf: Nothing. Just no logs

service-loadbalancer/service_loadbalancer.go
aledbf: What do you mean?

bprashanth: Oh I meant basically add a -v or -syslog flag to (https://github.com/kubernetes/contrib/blob/master/service-loadbalancer/service_loadbalancer.go#L108) and only start the syslog server if it's defined. I don't mind a follow up pr for this if you'd rather get this one in first.

bprashanth:
Let me know if you'd rather do the -syslog flag in a follow up pr, LGTM otherwise

bprashanth:
@aledbf my logs have:
```
[ALERT] 264/015123 (61) : sendto logger #2 failed: No such file or directory (errno=2)
[ALERT] 264/015123 (61) : sendto logger #1 failed: No such file or directory (errno=2)
[ALERT] 264/015123 (61) : sendto logger #2 failed: No such file or directory (errno=2)
```
with syslog=false. Can we turn that off?

I also noticed with the alpine linux build, the errorfile lines cause haproxy to crash: https://github.com/kubernetes/contrib/blob/master/service-loadbalancer/template.cfg#L40

I think this is because the ubuntu installation created those directories by default, but the alpine install doesn't. Did you have a workaround locally? or did you just remove the errorfile lines because syslog is enough?

bprashanth:
by "logs" I mean stdout/output of kubectl logs podname

aledbf:
> with syslog=false. Can we turn that off?

No. But I can add the start condition in the template to remove that lines if syslog is not started

bprashanth:
> No. But I can add the start condition in the template to remove that lines if syslog is not started

Yeah logging the ALERT periodically is bound to be confusing to people debugging. Would be great if you can detect syslog and turn off the lines without passing a new variable into the template, but please do whatevers easier. 

WDYT about the errorfiles, should we bake them into the image or just remove them from the config and document that people can add them as required?

aledbf:
> WDYT about the errorfiles, should we bake them into the image or just remove them from the config and document that people can add them as required?

I think is a good idea to add the error files (what I did in the last commit) just to provide the same installation than in other distributions (like Ubuntu) and as a reference in case a user wants a custom one (using the provided ones in the images as a reference)

service-loadbalancer/template.cfg
```diff
@@ -4,6 +4,10 @@ global
     daemon
     stats socket /tmp/haproxy
 
+    {{ if eq .startSyslog "true" }}# log using a syslog socket
```
bprashanth: does having the if/end formatted this way give us better formatting of the final config? please have this if/end consistent with the ones below, eitherway (if this is better change those, or if it doesn't matter make this like the others.)

service-loadbalancer/service_loadbalancer.go
```diff
@@ -125,11 +128,12 @@ type service struct {
...
-	ReloadCmd string `json:"reloadCmd" description:"command used to reload the load balancer."`
-	Config    string `json:"config" description:"path to loadbalancers configuration file."`
-	Template  string `json:"template" description:"template for the load balancer config."`
-	Algorithm string `json:"algorithm" description:"loadbalancing algorithm."`
+	Name        string `json:"name" description:"Name of the load balancer, eg: haproxy."`
+	ReloadCmd   string `json:"reloadCmd" description:"command used to reload the load balancer."`
+	Config      string `json:"config" description:"path to loadbalancers configuration file."`
+	Template    string `json:"template" description:"template for the load balancer config."`
+	Algorithm   string `json:"algorithm" description:"loadbalancing algorithm."`
+	startSyslog bool
```
bprashanth: brief description string for the sake of following the pattern please

bprashanth:
Please add a few lines to the troubleshooting part of the README explaining the syslog flag and how one can use it to debug

PR #96, opened 2015-09-17
submit-queue: use ListCollaborators to get all users with push access

eparis:
I didn't think ListCollaborators worked, but it appears it does work. So
get rid of the crazy logic where we get all teams, get those teams with
push access, and then get the users of those teams.  Just get the users
directly from github.

I also use more util.StringSet to simplify the code. Wish it came with
an intersection. I did it myself but an Intersection() would have been
nice...


eparis:
@ixdy you're the one who showed me the ListCollaborators worked and that our committer list was different than yours in the e2e

ixdy:
Yeah, as long as you're using the "preview" version of the API you can get the proper list of contributors:

```shell
curl -fsS -H "Accept: application/vnd.github.ironman-preview+json" \
  "https://api.github.com/repos/kubernetes/kubernetes/collaborators?per_page=100&access_token=..." \
  | jq -r '.[] | select(.permissions.push) | .login' | sort
```

I'm not sure how you convince the Go API to use the preview API, though; as-is, this may not work.

eparis:
@ixdy shocking, it already does use the preview API!

eparis:
https://github.com/google/go-github/blob/4476aa45b5f735de27fd5e87fb3d9eaf831c9248/github/repos_collaborators.go#L25

alex-mohr:
@eparis needs rebase

eparis:
merge away!

PR #98, opened 2015-09-17
Possibly fix JSON complaining about channels?

### LINKED ISSUE #78: Submit queue ui is borken
/api says
```
json: unsupported type: <-chan struct {}
```
@lavalamp

lavalamp:
Maybe fix #78?

### DIFF
--- a/submit-queue/e2e.go
+++ b/submit-queue/e2e.go
@@ -30,9 +30,24 @@ import (
 	github_api "github.com/google/go-github/github"
 )
 
+// PRInfo explicitly collects the fields we wish to encode via JSON.
+type PRInfo struct {
+	Number *int    `json:"number"`
+	URL    *string `json:"html_url"`
+}
+
+func prInfo(pr *github_api.PullRequest) *PRInfo {
+	var out PRInfo
+	if pr != nil {
+		out.Number = pr.Number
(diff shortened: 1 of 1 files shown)


lavalamp:
@brendandburns

brendandburns:
LGTM, I don't understand where that channel is coming from, that object is serialized to JSON by go-github afaik...

brendandburns:
Also, see #99

submit-queue/e2e.go
```diff
@@ -30,9 +30,24 @@ import (
 	github_api "github.com/google/go-github/github"
 )
 
+// PRInfo explicitly collects the fields we wish to encode via JSON.
+type PRInfo struct {
+	Number *int    `json:"number"`
+	URL    *string `json:"html_url"`
```
eparis: I don't even know what the problems are here, but why not make this an `int` and `string` instead of `*int` and `*string` so you know the PRInfo can't mutate after it's created?

lavalamp: I can change it if you feel strongly, but I'm pretty sure everyone treats those as read-only; they're pointers only because they are optional.

eparis: I don't feel strongly at all.

PR #106, opened 2015-09-21
Re-add build status that was mistakenly removed.

brendandburns:
@lavalamp @eparis


eparis:
# k8s.io/contrib/submit-queue
submit-queue/submit-queue.go:185: unknown e2eTester field 'BuildStatus' in struct literal

brendandburns:
@eparis fixed.

PR #107, opened 2015-09-21
All my other other PRs combined.

eparis:
So many of them stepped on each other. So make a monster.


github/github.go
```diff
@@ -68,6 +67,57 @@ type GithubConfig struct {
 	DryRun bool
 
 	useMemoryCache bool
+
+	analytics analytics
+}
+
+type analytic int
```
brendandburns: why not use prometheus metrics (or any other metric type that we can consume w/ a server?)

PR #108, opened 2015-09-22
Add ansible playbook support for Ubuntu 14.04.

pedro-r-marques:
- Uses etcd from github release since it isn't available upstream.
- Upstream docker.io package is too old; override with apt repository from projectdocker.org.
- Use upstart for services since 14.04 doesn't yet support systemd.
- Ansible copy mode requires 0 prefix to be interpreted as octal. The issue seems to affect Ubuntu nodes but not Fedora.
  https://github.com/ansible/ansible-modules-core/issues/1798

### DIFF
--- a/ansible/roles/etcd/tasks/main.yml
+++ b/ansible/roles/etcd/tasks/main.yml
@@ -6,7 +6,16 @@
         state: latest
   notify:
     - restart etcd
-  when: not is_atomic
+  when: not is_atomic and ansible_distribution != "Ubuntu"
+
+- name: Force etcd_source_type to github when packages are not available
+  set_fact:
+    etcd_source_type: "github-release"
+  when: ansible_distribution == "Ubuntu"
+
+- name: Install from github
+  include: github-release.yml
+  when: etcd_source_type == "github-release"
 
 - name: Write etcd config file
   template: src=etcd.conf.j2 dest=/etc/etcd/etcd.conf
--- a/ansible/roles/node/handlers/main.yml
+++ b/ansible/roles/node/handlers/main.yml
@@ -3,6 +3,7 @@
   command: systemctl --system daemon-reload
   notify:
     - restart daemons
+  when: not (ansible_distribution == "Ubuntu" and ansible_distribution_major_version|int < 15)
 
 - name: restart daemons
   command: /bin/true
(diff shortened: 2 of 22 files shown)


ansible/roles/docker/tasks/debian-install.yml
```diff
@@ -1,7 +1,8 @@
 ---
 - name: DEBIAN | Make sure this is stretch or sid, jessies does not have docker
   fail: msg="Docker.io only available in sid and stretch, https://wiki.debian.org/Docker"
-  when: ansible_lsb.codename != "stretch" and ansible_lsb.codename != "sid"
+  when: ansible_distribution == "Debian" and
```
eparis: why do I need this?  aren't we only included is distro == debian?

ansible/roles/etcd/tasks/main.yml
```diff
@@ -6,7 +6,11 @@
         state: latest
   notify:
     - restart etcd
-  when: not is_atomic
+  when: not is_atomic and ansible_distribution != "Ubuntu"
+
+- name: Install from github
+  include: github-release.yml
+  when: ansible_distribution == "Ubuntu"
```
eparis: So I don't like that this is predicated on "Ubuntu".  Maybe a top level option on where to get etcd from?  And have it automatically selected if the user got it wrong?

ansible/roles/node/handlers/main.yml
```diff
@@ -3,6 +3,7 @@
   command: systemctl --system daemon-reload
   notify:
     - restart daemons
+  when: ansible_distribution != "Ubuntu"
```
eparis: Is there some better test here?  Because ubuntu is getting systemd 'soon', right?  Does init system get collected in the setup facts?

ansible/roles/docker/tasks/debian-install.yml
pedro-r-marques: Apologies. Leftover from when i was using the upstream docker.io package from ubuntu. Before i realized that it is incompatible with kubelet.

ansible/roles/etcd/tasks/main.yml
pedro-r-marques: Added etcd_source_type variable.

ansible/roles/node/handlers/main.yml
pedro-r-marques: Nope. "init" isn't an setup fact/var. I made the checks predicated to ubuntu less-than 15.x.
14.04 is not going to get systemd. 16.04 should default to systemd but that is still at least 6 months away.

PR #111, opened 2015-09-22
Fix a stupid bug in build stability logic.

### LINKED ISSUE #112: checkBuilds() returns true even when builds are failing?
https://github.com/kubernetes/contrib/blob/023e2c05df5899bf9734b0d6d21fad0b0617fb45/submit-queue/e2e.go#L93

As far as I can tell by looking at this code, this function returns true even if stable==false for some or all builds.

In theory this will cause PR's to be merged even if Jenkins e2e tests are failing.

@gmarek has reported this to be the case and @lavalamp has therefore disabled the merge bot.  

Need to verify that this theoretical bug is in fact manifesting.  The fix is pretty straightforward.

@brendandburns FYI.

brendandburns:
Fixes #112
I promise a unit test next, but I have a meeting...

### DIFF
--- a/submit-queue/e2e.go
+++ b/submit-queue/e2e.go
@@ -104,6 +104,7 @@ func (e *e2eTester) checkBuilds() (allStable bool) {
 			e.setBuildStatus(build, "Stable")
 		} else {
 			e.setBuildStatus(build, "Not Stable")
+			allStable = false
 		}
 	}
 	return allStable
--- a/submit-queue/e2e_test.go
+++ b/submit-queue/e2e_test.go
@@ -1,0 +1,98 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"k8s.io/contrib/submit-queue/jenkins"
+)
+
+type testHandler struct {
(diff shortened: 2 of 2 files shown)


submit-queue/e2e.go
```diff
@@ -104,6 +104,7 @@ func (e *e2eTester) checkBuilds() (allStable bool) {
 			e.setBuildStatus(build, "Stable")
 		} else {
 			e.setBuildStatus(build, "Not Stable")
+                        allStable = false
```
lavalamp: needs a gofmt? (extra space in front)

brendandburns:
ok, gofmt'd and unit test added.  ptal.

submit-queue/e2e_test.go
```diff
@@ -0,0 +1,98 @@
+package main
```
lavalamp: Needs the boilerplate. Time to set up the boilerplate checker, sigh...

lavalamp:
LGTM if you fix the missing boilerplate

gmarek:
I don't think the boilerplate was added. @eparis can you please create a PR that does this?

eparis:
https://github.com/kubernetes/contrib/pull/115

Issue #110, opened 2015-09-22
Service loadbalancer 404s after first request to http services


asim:
reqrep and keep-alive do not play nicely. The first request to a http service works fine but subsequent requests 404 because the request is not parsed. Haproxy merely passes the data through. Using "option http-server-close" fixes this issue. 

Issue noted here a few years back http://permalink.gmane.org/gmane.comp.web.haproxy/7670

bprashanth:
@asim please submit a pr if you have the time, if not I'll verify and report back (I'm pretty sure I've tried curling an http service multiple times, but I don't remember the specifis of reqrep and keep-alive). Perhaps @aledbf has run into this as well?

PR #119, opened 2015-09-23
Improve haproxy connection defaults

aledbf:


bprashanth:
Ah, so you're adding server-close probably fixes https://github.com/kubernetes/contrib/issues/110. Please add a brief description above the new options about why each one is helpful. LGTM otherwise.

PR #120, opened 2015-09-23
Reload HAProxy without impacting server states

aledbf:
~~Blocked until haproxy-1.6-rdev6~~

This will improve the time required to forward traffic after a reload [examples/seamless_reload.txt](http://git.haproxy.org/?p=haproxy.git;a=blob;f=examples/seamless_reload.txt;h=94df1bd6a8dbca9bf0f290d52b0c66a90b155009;hb=71503d32a74c65cf7aceb5db9ca216db2ccc453c)

### DIFF
--- a/service-loadbalancer/Dockerfile
+++ b/service-loadbalancer/Dockerfile
@@ -16,10 +16,13 @@
 FROM alpine:3.2
 MAINTAINER Prashanth B <beeps@google.com>
 
-RUN apk add -U haproxy bash curl && \
+COPY haproxy.apk /var/cache/apk/haproxy.apk
+
+RUN apk add -U pcre bash curl socat lua5.3 && \
+  apk add --allow-untrusted /var/cache/apk/haproxy.apk && \
   rm -rf /var/cache/apk/*
 
-RUN mkdir /etc/haproxy/errors 
+RUN mkdir -p /etc/haproxy/errors /var/state/haproxy
 RUN for ERROR_CODE in 400 403 404 408 500 502 503 504;do curl -sSL -o /etc/haproxy/errors/$ERROR_CODE.http \
 	https://raw.githubusercontent.com/haproxy/haproxy-1.5/master/examples/errorfiles/$ERROR_CODE.http;done
 
@@ -30,4 +33,5 @@ ADD template.cfg template.cfg
 ADD loadbalancer.json loadbalancer.json
 ADD haproxy_reload haproxy_reload
 ADD README.md README.md
+
 ENTRYPOINT ["/service_loadbalancer"]
--- a/service-loadbalancer/build/Dockerfile
+++ b/service-loadbalancer/build/Dockerfile
@@ -1,5 +1,3 @@
-#!/bin/bash
-
 # Copyright 2015 The Kubernetes Authors. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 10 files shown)


bprashanth:
this is great! 
@ramr @smarterclayton fyi

smarterclayton:
Very cool

ramr:
awesome, look fwd to trying it out.

aledbf:
ping @bprashanth

bprashanth:
Reading through the release notes it looks like haproxy will save the weights of the backends. This isn't useful today. How do we make it useful (https://github.com/kubernetes/contrib/issues/144?)? Can you give me a better example so I understand how to fully utilize this? 

Otherwise LGTM, please test/add a test for this to the basic e2e test I have in the main repo https://github.com/kubernetes/kubernetes/pull/14988

service-loadbalancer/Dockerfile
```diff
@@ -16,10 +16,12 @@
 FROM alpine:3.2
 MAINTAINER Prashanth B <beeps@google.com>
 
-RUN apk add -U haproxy bash curl && \
+RUN apk add -U pcre bash curl socat && \
+  wget -O /var/cache/apk/haproxy.apk https://github.com/aledbf/alpine-build-dev-haproxy/releases/download/v1.6-dev6/haproxy-1.6-rdev6.apk && \
```
bprashanth: can we download this from http://www.haproxy.org/? 
As a bonus if you're going to compile it from official, please enalbe the lua engine (http://blog.haproxy.com/2015/10/14/whats-new-in-haproxy-1-6/), I'm sure it'll come in handy ;)

service-loadbalancer/Dockerfile
```diff
@@ -16,10 +16,12 @@
 FROM alpine:3.2
 MAINTAINER Prashanth B <beeps@google.com>
 
-RUN apk add -U haproxy bash curl && \
+RUN apk add -U pcre bash curl socat lua5.3 && \
+  wget -O /var/cache/apk/haproxy.apk https://github.com/aledbf/alpine-build-dev-haproxy/releases/download/1.6-r0/haproxy-1.6-r0.apk && \
```
bprashanth: Can you just pull from haproxy.org and build with the right options (eg with nginx: https://github.com/bprashanth/Ingress/blob/master/examples/bench/ab/Dockerfile#L29). That way you can delete your repo/release without worrying about kubernetes/contrib ;)

aledbf: > Can you just pull from haproxy.org and build with the right options 

yes but this will add too many things not related to the kubernetes/contrib project (please check all the required files and packages required to compile the apk)

Just in case I'm using that repo for my alpine images, not just kubernetes/contrib

bprashanth: Hmm, we need to make it reproducible, perhaps you can upstream your build scripts into a /build subdir and modify the Makefile? Then it doesn't really matter where we host the tarball.

service-loadbalancer/build/Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM    alpine:3.2
+
```
bprashanth: can you please reduce the spaces in this file? (club parts off it into logically equivalent sections)

bprashanth:
Please update licenses and we can get this in

service-loadbalancer/build/src/APKBUILD
```diff
@@ -0,0 +1,41 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+pkgname=haproxy
+pkgver=1.6
+pkgrel=0
+pkgdesc="HAProxy is a free, very fast and reliable solution offering high availability, load balancing"
+url="http://http://www.haproxy.org/"
```
bprashanth: nit: please remove http://http://

service-loadbalancer/build/src/README
```diff
@@ -0,0 +1,2 @@
+
+This build is required because there is no official package for haproxy 1.6.x (yet)
```
bprashanth: Is the red character at the end a stray carriage return or something?

bprashanth:
LGTM but for the one comment nit. Please run the e2e with the new build, if you haven't already. Would be even better if you added one for the soft reload, but no real requirement right now, since it is a more advanced use case.

PR #121, opened 2015-09-25
WIP: fix binary src directory for local build install

pmorie:
@timothysc @jlebon @eparis @mikedanese looks like something that didn't get changed over in the move from kubernetes.

### DIFF
--- a/ansible/roles/master/tasks/localBuildInstall.yml
+++ b/ansible/roles/master/tasks/localBuildInstall.yml
@@ -1,7 +1,7 @@
 ---
 - name: Copy master binaries
   copy:
-    src: ../../_output/local/go/bin/{{ item }}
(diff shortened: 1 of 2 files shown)


jlebon:
Wouldn't it entirely depend on where people check out the kubernetes repo? Maybe a new parameter in `all.yml` instead would fit more cases.

ansible/roles/master/tasks/localBuildInstall.yml
```diff
@@ -1,7 +1,7 @@
 ---
 - name: Copy master binaries
   copy:
-    src: ../../_output/local/go/bin/{{ item }}
+    src: ../..kubernetes/_output/local/go/bin/{{ item }}
```
timothysc: ../k

PR #124, opened 2015-09-28
Default page for traffic that reach the lb not matching any acl

aledbf:


service-loadbalancer/index-template.http
```diff
@@ -0,0 +1 @@
+404 page not found
```
bprashanth: curious, why do you need this? doesn't haproxy already have something like:
```
$ curl http://haproxy/asdfsadf
<html><body><h1>503 Service Unavailable</h1>
No server is available to handle this request.
</body></html>
```

I like an explicit 404 page, just asking if you have some other concrete case.

service-loadbalancer/service_loadbalancer.go
```diff
@@ -110,6 +110,9 @@ var (
 
 	startSyslog = flags.Bool("syslog", false, `if set, it will start a syslog server
 		that will forward haproxy logs to stdout.`)
+
+	// content of the default index (to avoid read the file multiple times)
+	defaultIndex []byte
```
bprashanth: instead of doing this can you make this a type:
```go
type staticPageHandler struct {
  pagePath string
  pageContents []byte
}

// load loads the page into memory 
func (s *staticPageHandler)load(){}

// Get serves the page stored by load
func (s *staticPageHandler)Get(http.ResponseWriter, *http.Request) {}

// newStaticPageHandler returns a staticPageHandles with the contents of pagePath loaded and ready to serve
func newStaticPageHandler(pagePath string) (*staticPageHandler, error) {}
```

That way it's more obvious what we're doing and we can reuse it to serve non default index pages as well.

service-loadbalancer/service_loadbalancer.go
```diff
@@ -418,6 +421,13 @@ func healthzServer() {
 			}
 		}
 	})
+
+	// handler for not matched traffic
```
bprashanth: it's ok to put all handlers in one function, but please rename to something like registerHandlers (currently called healthzServer)

service-loadbalancer/service_loadbalancer.go
```diff
@@ -418,6 +421,13 @@ func healthzServer() {
...
+	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(404)
+		w.Write(getErrorIndex())
+	})
+
 	glog.Fatal(http.ListenAndServe(fmt.Sprintf(":%v", healthzPort), nil))
```
bprashanth: please rename healthzPort to lbApiPort or something

service-loadbalancer/service_loadbalancer.go
```diff
@@ -430,6 +440,17 @@ func dryRun(lbc *loadBalancerController) {
 	}
 }
 
+// getErrorIndex returns the default page to return for traffic
+// that reach the load balancer and does not match any acl
+func getErrorIndex() []byte {
```
bprashanth: I you accept my previous suggestion of making this a type, this function will move in there

service-loadbalancer/template.cfg
```diff
@@ -74,6 +74,7 @@ frontend httpfrontend
     acl url_{{$svc.Name}} path_beg /{{$svc.Name}}
     use_backend {{$svc.Name}} if url_{{$svc.Name}}
 {{end}}
+    default_backend default-backend
```
bprashanth: can we make this a global default_backend? That way we can also have a frontend specific default_backend that overrides the global. If we allow people to request different ports, they can also in theory give us their own 404 page. If they don't and nothing matches, we server the lb 404 page.

service-loadbalancer/index-template.http
aledbf: > 503 Service Unavailable

Well, 503 is not the "correct code" for a non existing service. If I see 503 it could means that there is no server for a particular backend. 

(If you don't think this is useful I can close this PR)

service-loadbalancer/service_loadbalancer.go
```diff
@@ -436,13 +467,19 @@ func main() {
...
 		glog.Infof("All tcp/https services will be ignored.")
 	}
-	go healthzServer()
-
-	proc.StartReaper()
 
 	var kubeClient *unversioned.Client
 	var err error
 
+	errorPage, err := newStaticPageHandler(*defStaticPage)
```
bprashanth: I just noticed a bug while reviewing, you're free to not fix it and I'll do so with a follow up pr. If you're ok fixing it, please move the flags.Parse from the first line of main() to after we're parsed the clientConfig, ie: 
```
clientConfig := kubectl_util.DefaultClientConfig(flags)
flags.Parse(os.Args)
```

This will allow use of the flag overrides defined by the client config (https://github.com/kubernetes/kubernetes/blob/master/pkg/client/unversioned/clientcmd/overrides.go#L105)

service-loadbalancer/service_loadbalancer.go
```diff
@@ -137,6 +140,30 @@ type loadBalancerConfig struct {
 	startSyslog bool   `description:"indicates if the load balancer uses syslog."`
 }
 
+type staticPageHandler struct {
+	pagePath     string
+	pageContents []byte
+}
+
+// Get serves the page stored by load
```
bprashanth: since you don't have a function called load, please modify comment or add a load function and do the readfile from there instead of newStaticPageHandler

service-loadbalancer/template.cfg
```diff
@@ -54,9 +54,17 @@ defaults
     # Maximum allowed time to wait for a new HTTP request to appear.
     timeout http-keep-alive 60s
 
+    # default traffic mode is http
+    mode http
+
+    # default default_backend. This allows custom default_backend in frontends
```
bprashanth: please clarify this comment by saying mode is overwritten in the tcp segment below

bprashanth:
LGTM but for the comment nits

service-loadbalancer/service_loadbalancer.go
```diff
@@ -110,6 +111,13 @@ var (
 
 	startSyslog = flags.Bool("syslog", false, `if set, it will start a syslog server
 		that will forward haproxy logs to stdout.`)
+
+	defStaticPage = flags.String("static-page", "/etc/haproxy/errors/404.http", `if set,
```
bprashanth: hmm, do we really need 2 flags? given this is a file on disk, no point opening it as a flag right? can we just check the errorPage, and if it isn't set or returns a non 200, always use the 404.http page? I've found that the number of states in the system roughly doubles with each new flag, because someone will try overloading it to mean something else down the line, so if we can get rid of this flag please do so.

aledbf: done

service-loadbalancer/service_loadbalancer.go
```diff
@@ -431,18 +484,33 @@ func dryRun(lbc *loadBalancerController) {
...
 		glog.Infof("All tcp/https services will be ignored.")
 	}
-	go healthzServer()
-
-	proc.StartReaper()
 
 	var kubeClient *unversioned.Client
 	var err error
 
+	defErrorPage, err := newStaticPageHandler("/etc/haproxy/errors/404.http")
```
bprashanth: optional nit: 
This is a little awkward because you're settings something in the struct, then you're loading a page and setting the same things outside. Would make threading it etc hard.  Since file:// is defined in the url spec http will handle it, eg (not sure if this code will compile):

```go
const defaultErrorPage = "file:///etc/haproxy/errors/404.http"

newStaticPageHandler(*errorPage, defaultErrorPage) *staticPageHandler {
  t := &http.Transport{}
  t.RegisterProtocol("file", http.NewFileTransport(http.Dir("/")))
  c := &http.Client{Transport: t}
  s := staticPageHandler{c}
  if err := s.loadUrl(*errorPage); err != nil {
    s.loadUrl(defaultErrorPage)
  }
}

func (s *staticPageHandler) loadErrorPage(url string) {
	res, err := s.c.Get(url)
	if err != nil {
		log.Fatalf("%v", err)
	}
	defer res.Body.Close()
	body, err := ioutil.ReadAll(res.Body)
	if err != nil {
		log.Fatalf("%v", err)
	}
	log.Printf("%v", string(body))
        // set page
}

newStaticPagehandler(*errorPage, defaultErrorPage)

```


However it's trivial enough that I would accept if you just add a TODO comment for now

bprashanth:
LGTM, please squash and I'll merge

PR #125, opened 2015-09-28
Add acl to route traffic using Host header using metadata from service

aledbf:
Example:

```
aledbf$ kubectl get services
NAME           LABELS                                    SELECTOR           IP(S)        PORT(S)
example-go     run=example-go                            run=example-go     10.3.0.36    8080/TCP
kubernetes     component=apiserver,provider=kubernetes   <none>             10.3.0.1     443/TCP
```

```
aledbf$ kubectl label services --overwrite --selector="run=example-go" lbHost="www.k8s.io"
```

```
aledbf$ kubectl get services
NAME           LABELS                                    SELECTOR           IP(S)        PORT(S)
example-go     lbHost=www.k8s.io,run=example-go          run=example-go     10.3.0.36    8080/TCP
kubernetes     component=apiserver,provider=kubernetes   <none>             10.3.0.1     443/TCP
```

### DIFF
--- a/service-loadbalancer/service_loadbalancer.go
+++ b/service-loadbalancer/service_loadbalancer.go
@@ -47,6 +47,7 @@ const (
 	reloadQPS    = 10.0
 	resyncPeriod = 10 * time.Second
 	healthzPort  = 8081
+	lbHostKey    = "serviceloadbalancer/lb.host"
 )
 
 var (
@@ -272,7 +273,7 @@ func (lbc *loadBalancerController) getServices() (httpSvc []service, tcpSvc []se
 				Name: getServiceNameForLBRule(&s, servicePort.Port),
 				Ep:   ep,
 			}
-			if val, ok := s.Labels["lbHost"]; ok {
+			if val, ok := s.Labels[lbHostKey]; ok {
 				newSvc.Host = val
 			}
 			if port, ok := lbc.tcpServices[sName]; ok && port == servicePort.Port {
--- a/service-loadbalancer/service_loadbalancer.go
+++ b/service-loadbalancer/service_loadbalancer.go
@@ -124,6 +124,10 @@ type service struct {
 	// for this service. For http, it's always :80, for each tcp service it
 	// is the service port of any service matching a name in the tcpServices set.
 	FrontendPort int
+
+	// Host if not empty it will add a new haproxy acl to route traffic using the
+	// host header inside the http request. It only applies to http traffic.
+	Host string
 }
 
 // loadBalancerConfig represents loadbalancer specific configuration. Eventually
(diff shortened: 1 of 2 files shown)


service-loadbalancer/service_loadbalancer.go
```diff
@@ -124,6 +124,10 @@ type service struct {
 	// for this service. For http, it's always :80, for each tcp service it
 	// is the service port of any service matching a name in the tcpServices set.
 	FrontendPort int
+
+	// Host if not empty it will add a new haproxy acl to route traffic using the
```
bprashanth: please add a comment here about what gets precedence, I believe the url wins over host header right?

service-loadbalancer/service_loadbalancer.go
```diff
@@ -268,6 +272,9 @@ func (lbc *loadBalancerController) getServices() (httpSvc []service, tcpSvc []se
 				Name: getServiceNameForLBRule(&s, servicePort.Port),
 				Ep:   ep,
 			}
+			if val, ok := s.Labels["lbHost"]; ok {
```
bprashanth: please make this key more formal, and define it as a const. eg:
```
const lbHostKey =  "serviceloadbalancer/lb.host",
```

Eg how we do it for mirror pods: https://github.com/kubernetes/kubernetes/blob/e5670c104d1c8cdc5bd31f472b3ca806c5b2df35/pkg/api/pod_example.json#L12. becuase then we're free to define a `serviceloadbalancer/svcname.url` too.

service-loadbalancer/template.cfg
```diff
@@ -73,6 +73,10 @@ frontend httpfrontend
 {{range $i, $svc := .services.http}}
     acl url_{{$svc.Name}} path_beg /{{$svc.Name}}
     use_backend {{$svc.Name}} if url_{{$svc.Name}}
+    {{ if $svc.Host }}
```
bprashanth: please either indent this or the if block above so they're both the same

bprashanth:
As we start adding more features it would be great to get some simple e2e verification tests of some form (even if not in kube through the ginkgo framework, maybe just in this repo as a bash script that creates the rc against an existing cluster and does curl? Just so I don't break your changes if I refactor)

bprashanth:
Y'know what would be really cool, if we allowed arbitrary sticky sessions based a server id cookie haproxy sets after the first request goes through. just throwing it out there as an idea.

service-loadbalancer/service_loadbalancer.go
aledbf: > I believe the url wins over host header right?

This adds a new backend (i.e. both backends are available)

aledbf:
> if we allowed arbitrary sticky sessions based a... just throwing it out there as an idea.

Is in the list :) 

*I want to add:*
- sticky sessions
- balance algorithm per service
- checks using the livenessProbe defined

service-loadbalancer/service_loadbalancer.go
bprashanth: I mean if I have a single requests that matches url and host header, is my backend invoked twice?

aledbf: No, it will use the first matched acl

aledbf: > I mean if I have a single requests that matches url and host header, is my backend invoked twice?

I changed the template to simplify the matching. It uses multiple acls and one `use_backend` directive with an `or` condition. The order is `url_acl` and then `host_acl`.

bprashanth:
Great thanks!

PR #130, opened 2015-09-30
add podmaster support for etcd tls

roboll:
add support for podmaster to consume tls secured etcd.

cc. @brendandburns (since you are listed as the maintainer in the dockerfile for podmaster)

### DIFF
--- a/pod-master/podmaster.go
+++ b/pod-master/podmaster.go
@@ -36,6 +36,9 @@ import (
 
 type config struct {
 	etcdServers string
+	etcdCA      string
+	etcdCert    string
+	etcdKey     string
 	key         string
 	whoami      string
 	ttl         uint64
@@ -138,6 +141,9 @@ func copyFile(src, dest string) error {
 
 func initFlags(c *config) {
 	pflag.StringVar(&c.etcdServers, "etcd-servers", "", "The comma-seprated list of etcd servers to use")
+	pflag.StringVar(&c.etcdCA, "etcd-ca", "", "The CA file for etcd tls.")
+	pflag.StringVar(&c.etcdCert, "etcd-cert", "", "The cert file for etcd tls.")
+	pflag.StringVar(&c.etcdKey, "etcd-key", "", "The key file for etcd tls.")
 	pflag.StringVar(&c.key, "key", "", "The key to use for the lock")
 	pflag.StringVar(&c.whoami, "whoami", "", "The name to use for the reservation.  If empty use os.Hostname")
 	pflag.Uint64Var(&c.ttl, "ttl-secs", 30, "The time to live for the lock.")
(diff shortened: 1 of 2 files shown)


pod-master/podmaster.go
```diff
@@ -176,7 +186,16 @@ func main() {
 	validateFlags(&c)
 
 	machines := strings.Split(c.etcdServers, ",")
-	etcdClient := etcd.NewClient(machines)
+	var etcdClient *etcd.Client
+	if c.etcdCert != "" && c.etcdKey != "" && c.etcdCA != "" {
```
brendandburns: Just test one here, you've already validated that if one is empty they all need to be empty.

pod-master/podmaster.go
```diff
@@ -159,6 +165,10 @@ func validateFlags(c *config) {
 	if len(c.dest) == 0 {
 		glog.Fatalf("--dest-file=<some-file> is required")
 	}
+	//either all are populated, or none are populated
```
brendandburns: please move this into the validateFlags function (the ability to return early will clean up this condition)

brendandburns: oops, silly git diff made this look like it wasn't in there already...

pod-master/podmaster.go
```diff
@@ -159,6 +165,10 @@ func validateFlags(c *config) {
...
+	if !((c.etcdCert != "" && c.etcdCA != "" && c.etcdKey != "") || (c.etcdCert == "" && c.etcdCA == "" && c.etcdKey == "")) {
```
brendandburns: I think that:

```go
if len(c.etcdCert) != 0 || len(c.etcdCA) != 0 || len(c.etcdKey) != 0 {
  if len(c.etcdCert) == 0 || len(c.etcdCA) == 0 || len(c.etcdKey) == 0 {
    glog.Fatalf(...)
  }
}
```

is easier to read.

pod-master/podmaster.go
```diff
@@ -138,6 +141,9 @@ func copyFile(src, dest string) error {
 
 func initFlags(c *config) {
 	pflag.StringVar(&c.etcdServers, "etcd-servers", "", "The comma-seprated list of etcd servers to use")
+	pflag.StringVar(&c.etcdCA, "etcd-ca", "", "The CA file for etcd tls.")
```
brendandburns: add these to https://github.com/kubernetes/contrib/blob/master/hack/verify-flags/known-flags.txt

brendandburns:
Thanks for the PR!  small comments.

roboll:
thanks for the comments @brendandburns, everything should be addressed.

glerchundi:
WDYT about using etcd-config in order to be aligned with the apiserver?

roboll:
yeah, i think thats a better idea. good call.

glerchundi:
yep.

It seems like kubernetes core is going to use the new etcd client `github.com/coreos/etcd/client` (not the deprecated one: `github.com/coreos/go-etcd`) which doesn't support reading config from a custom formatted file. Although, in order to be backward compatible maybe that parsing should be made by kubernetes itself...

/ref kubernetes/kubernetes#11962

rwehner:
This PR appears to address Issue https://github.com/kubernetes/contrib/issues/133. Just wanted to make someone reference between the two in case someone (like me) comes across the issue but somehow fails to find the this PR quickly.

I'm looking forward to using this when merged. thanks!

glerchundi:
@rwehner @roboll until this is merged and available for all, i created a public image which uses `etcd-config` parameter as `kube-apiserver` does: `quay.io/saltosystems/podmaster.

biwwy:
I had to do etcd-config file version for my own use, so if anybody wants to use etcd-config from a file, there is a version of pod-master here: https://github.com/biwwy/contrib/commit/79664e46d38cb03d38ca6425e7cc57c0bd2b38e0

roboll:
revisiting this, i poked around all the relevant issues and didn't find any details leading me to believe this was going down the wrong path. 

updating podmaster to use the new etcd config is a little more intense, and probably not as high a priority as this.

@brendandburns is this good for merge?

roboll:
going to close this per https://github.com/kubernetes/contrib/issues/133#issuecomment-163371391

PR #131, opened 2015-09-30
Allow custom default algorithm in service-loadbalancer

aledbf:
Allow to choose a different algorithm to the default `roundrobin` using labels in the service.

_Is possible to choose between:_
- roundrobin (default)
- leastconn
- first
- source

HAProxy provides other [algorithms](https://cbonte.github.io/haproxy-dconv/configuration-1.5.html#4.2-balance) but is required to use additional parameters (if someone wants support for those please open an issue or just a comment in this PR)

To change the default just set the value in the label serviceloadbalancer/lb.algorithm

```
kubectl annotate svc example-go serviceloadbalancer/lb.algorithm="leastconn"
```

**If the specified string it does not exists in the list it will use the default.**

### DIFF
--- a/service-loadbalancer/service_loadbalancer.go
+++ b/service-loadbalancer/service_loadbalancer.go
@@ -47,6 +47,7 @@ const (
 	reloadQPS        = 10.0
 	resyncPeriod     = 10 * time.Second
 	lbApiPort        = 8081
+	lbAlgorithmKey   = "serviceloadbalancer/lb.algorithm"
 	lbHostKey        = "serviceloadbalancer/lb.host"
 	defaultErrorPage = "file:///etc/haproxy/errors/404.http"
 )
@@ -59,6 +60,14 @@ var (
 
 	// Error used to indicate that a sync is deferred because the controller isn't ready yet
 	errDeferredSync = fmt.Errorf("deferring sync till endpoints controller has synced")
+
+	// See https://cbonte.github.io/haproxy-dconv/configuration-1.5.html#4.2-balance
+	// In brief:
+	//  * roundrobin: backend with the highest weight (how is this set?) receives new connection
+	//  * leastconn: backend with least connections receives new connection
+	//  * first: first server sorted by server id, with an available slot receives connection
+	//  * source: connection given to backend based on hash of source ip
+	supportedAlgorithms = []string{"roundrobin", "leastconn", "first", "source"}
 
 	config = flags.String("cfg", "loadbalancer.json", `path to load balancer json config.
 		Note that this is *not* the path to the configuration file for the load balancer
@@ -116,6 +125,9 @@ var (
 	errorPage = flags.String("error-page", "", `if set, it will try to load the content
 		as a web page and use the content as error page. Is required that the URL returns
 		200 as a status code`)
+
+	lbDefAlgorithm = flags.String("balance-algorithm", "roundrobin", `if set, it allows a custom 
+		default balance algorithm.`)
 )
 
 // service encapsulates a single backend entry in the load balancer config.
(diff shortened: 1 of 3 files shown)


aledbf:
ping @bprashanth

service-loadbalancer/service_loadbalancer.go
```diff
@@ -60,6 +61,8 @@ var (
 	// Error used to indicate that a sync is deferred because the controller isn't ready yet
 	errDeferredSync = fmt.Errorf("deferring sync till endpoints controller has synced")
 
+	supportedAlgorithms = []string{"roundrobin", "leastconn", "first", "source"}
```
bprashanth: please add a comment above with a brief oneline explanation of each, eg:
```
// See https://cbonte.github.io/haproxy-dconv/configuration-1.5.html#4.2-balance
// In brief:
//  * roundrobin: backend with the highest weight (how is this set?) receives new connection
//  * leastconn: backend with least connections receives new connection
//  * first: first server sorted by server id, with an available slot receives connection
//  * source: connection given to backend based on hash of source ip
```

service-loadbalancer/service_loadbalancer.go
```diff
@@ -322,9 +338,19 @@ func (lbc *loadBalancerController) getServices() (httpSvc []service, tcpSvc []se
 				Name: getServiceNameForLBRule(&s, servicePort.Port),
 				Ep:   ep,
 			}
+
 			if val, ok := s.Labels[lbHostKey]; ok {
```
bprashanth: I want to encapsulate all the annotations logic so it's clear how someone would add support for a new annotation by looking at the code. 
Suggest something like: 
```
type serviceAnnotations map[string]string

func (s serviceAnnotations)getAlgorithm() {
    return s[lbAlgorightmsKey]
}
...
s.Labels.(serviceAnnotations).getAlgorightm()
```

Also why are we using labels instead of annotations (I'm not against it, just asking)? See [apiconvention](https://github.com/kubernetes/kubernetes/blob/11c31aeca7ec8931150ba8f6ceab86c38155fb9c/docs/devel/api-conventions.md#label-selector-and-annotation-conventions ) on this
if you still feel labels are the right thing, then this is ok, but elaborate on your reasoning in a comment.

```
Annotations have very different intended usage from labels. We expect them to 
be primarily generated and consumed by tooling and system extensions. I'm inclined
to generalize annotations to permit them to directly store arbitrary json. Rigid names 
and name prefixes make sense, since they are analogous to API fields.
```

If you want to keep them in labels because they're easier to play around with for now, but plan to move them to annotation soon, then please add a TODO.

service-loadbalancer/template.cfg
```diff
@@ -61,6 +61,9 @@ defaults
     # default default_backend. This allows custom default_backend in frontends
     default_backend default-backend
 
+    # default load balancing algorithm
```
bprashanth: This files is getting more complex. Can we add a unittest that puts in some fake data, and checks the output? In theory, it would be better to have an integration test that actually starts up haproxy and makes sure it doesn't crash, but anything is fine for now. you can do it in a follow up pr, or if you're pressed for time, just file a bug and we'll get to it somethime post kubernetes 1.1

bprashanth:
Sorry for the delay, got caught up in some kubernetes 1.1 release work

service-loadbalancer/service_loadbalancer.go
aledbf: > Also why are we using labels instead of annotations 

No reason. I will change the labels to annotation.

bprashanth:
LGTM, please update the README, you can do so in another pr since I think one of the other features you're adding will also need a readme update.

bprashanth:
@aledbf I had some comments in the previous pr that you didn't address, please send a follow up or respond when you have time. I figured we'd get this one in as is and address the nits in a follow up.

PR #132, opened 2015-10-01
Gce L7s through Kubernetes Ingress

bprashanth:
Depends on https://github.com/kubernetes/kubernetes/pull/14459 
Probably just want to wait on the README for a Tl;Dr.


service-loadbalancer/gce/README.md
```diff
@@ -0,0 +1,18 @@
+# GCE Load Balancers
+
+
+## Disclaimer:
+- This is a **work in progress**.
+- For fault tolerant load balancing of ingress traffic, you need:
+  1. Multiple hosts running load balancers
```
thockin: I'm confused - why do we need hosts running anything for GCE load-balancers?

Am I misunderstanding what this is?

bprashanth: heh, no i just copied that over from the other serviceloadbalancer README. I'm writing e2es atm, figured that would somehow validate this enough that you won't have to do some hardcore review.

thockin: OK, so what do you need me to look at? :)

bprashanth: The code is good to go

bprashanth: Well, checking it in is blocked on https://github.com/kubernetes/contrib/pull/142, but otherwise

bprashanth:
@thockin finally done with the README. Travis failing because I need to add the GCE compute godep.

thockin:
Which commit has a README?

bprashanth:
Yes, I didn't think you'd have time for a full review. Perhaps someone from @kubernetes/goog-gke ?
I need to bump up godeps before I checkin, but doing so now makes the diff unreadable.

brendandburns:
@bprashanth you can include the godeps as a separate commit (that's best practice anyway) that way we can just look at the code commit, not the godeps commit.

service-loadbalancer/gce/Makefile
```diff
@@ -0,0 +1,17 @@
+all: push
+
+# 0.0 shouldn't clobber any released builds
+TAG = 0.0
+PREFIX = bprashanth/gcelb
```
brendandburns: please use gcr.io/google_containers as the prefix.

service-loadbalancer/gce/backends.go
```diff
@@ -0,0 +1,212 @@
...
+limitations under the License.
+*/
+
+package main
+
+import (
+	"fmt"
+	"strconv"
+
+	compute "google.golang.org/api/compute/v1"
```
brendandburns: move this down with the github import.

service-loadbalancer/gce/backends.go
```diff
@@ -0,0 +1,212 @@
...
+
+	"k8s.io/kubernetes/pkg/util/sets"
+
+	"github.com/golang/glog"
+)
+
+type Backends struct {
```
brendandburns: Godoc (or make private)

service-loadbalancer/gce/backends.go
```diff
@@ -0,0 +1,212 @@
...
+	cloud          BackendServices
+	instanceGroups InstanceGroups
+	pool           *poolStore
+	defaultIg      *compute.InstanceGroup
+	defaultBackend *compute.BackendService
+	defaultHc      *compute.HttpHealthCheck
+}
+
+func portKey(port int64) string {
+	return fmt.Sprintf("%v", port)
```
brendandburns: why not `%d`?

service-loadbalancer/gce/backends.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+func beName(port int64) string {
+	return fmt.Sprintf("%v-%v", backendPrefix, port)
```
brendandburns: as above.

service-loadbalancer/gce/backends.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+func NewBackendPool(
```
brendandburns: GoDoc

service-loadbalancer/gce/backends.go
```diff
@@ -0,0 +1,212 @@
...
+	defaultHc *compute.HttpHealthCheck,
+	instanceGroups InstanceGroups) (BackendPool, error) {
+	backends := &Backends{
+		cloud:          cloud,
+		instanceGroups: instanceGroups,
+		pool:           newPoolStore(),
+		defaultIg:      defaultIg,
+		defaultHc:      defaultHc,
+	}
+	err := backends.Add(defaultBackendNodePort)
```
brendandburns:
```go
if err := ... ; err != nil {
...
}
```

service-loadbalancer/gce/backends.go
```diff
@@ -0,0 +1,212 @@
...
+	if err != nil {
+		return nil, err
+	}
+	backends.defaultBackend, err = backends.Get(defaultBackendNodePort)
```
brendandburns: why not just

```go
return backends.Get(defaultBackendNodePort)
```

service-loadbalancer/gce/cluster_manager.go
```diff
@@ -0,0 +1,158 @@
...
+	if err := c.l7Pool.Shutdown(); err != nil {
+		return err
+	}
+	if err := c.backendPool.Shutdown(); err != nil {
+		return err
+	}
+	if err := c.instancePool.Shutdown(); err != nil {
+		return err
+	}
+	if err := c.healthChecker.Delete(defaultHttpHealthCheck); err != nil {
```
brendandburns: just:
```go
return c.healthChecker.Delete(...)
```

service-loadbalancer/gce/cluster_manager.go
```diff
@@ -0,0 +1,158 @@
...
+		return err
+	}
+	// TODO: Delete default health check, or is it actually used by the
+	// cluster instance group?
+	return nil
+}
+
+func defaultInstanceGroupName(clusterName string) string {
```
brendandburns: nit: I don't really see the win of these functions.

why not delete these functions and just use

```go
strings.Join([]string{instanceGroupPrefix, clusterName}, "-")
``

for example at the call sites.  I think that's actually clearer to read.

File: service-loadbalancer/gce/cluster_manager.go [hunk 12]
```diff
@@ -0,0 +1,158 @@
...
+	// every urlmap, even if the user doesn't specify one.
+	// This is the backend that gets requests if no paths match.
+	// Note that this backend doesn't actually occupy a port
+	// on the instance group.
+	//
+	// Default Health Check: The default backend used by an
+	// Ingress that doesn't specify it.
+
+	defaultIgName := defaultInstanceGroupName(name)
+	cluster.instancePool, err = NewNodePool(
```
brendandburns: inline `err != nil`

service-loadbalancer/gce/cluster_manager.go
```diff
@@ -0,0 +1,158 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO: We're roud tripping for a resource we just created.
+	defaultIg, err := cluster.instancePool.Get(defaultIgName)
+	if err != nil {
+		return nil, err
+	}
+	cluster.healthChecker, err = NewHealthChecker(
```
brendandburns: here too

service-loadbalancer/gce/cluster_manager.go
```diff
@@ -0,0 +1,158 @@
...
+		cloud, defaultHttpHealthCheck)
+	if err != nil {
+		return nil, err
+	}
+	defaultHc, err := cluster.healthChecker.Get(defaultHttpHealthCheck)
+	if err != nil {
+		return nil, err
+	}
+	cluster.backendPool, err = NewBackendPool(
```
brendandburns: move all the parameters onto one line.

service-loadbalancer/gce/cluster_manager.go
```diff
@@ -0,0 +1,158 @@
...
+		cloud,
+		defaultBackendNodePort,
+		defaultIg,
+		defaultHc,
+		cloud)
+	if err != nil {
```
brendandburns: inline the `err != nil`

service-loadbalancer/gce/cluster_manager.go
```diff
@@ -0,0 +1,158 @@
...
+		return nil, err
+	}
+	cluster.defaultBackendNodePort = defaultBackendNodePort
+	// TODO: Don't cast, the problem here is the default backend doesn't have
+	// a port and the interface only allows backend access via port.
+	cluster.l7Pool = NewLoadBalancerPool(
```
brendandburns: all on one line (this isn't c++ ;)

service-loadbalancer/gce/controller.go
```diff
@@ -0,0 +1,295 @@
+/*
+Copyright 2014 The Kubernetes Authors All rights reserved.
```
brendandburns: it's 2015...

service-loadbalancer/gce/controller.go
```diff
@@ -0,0 +1,295 @@
...
+	eventBroadcaster := record.NewBroadcaster()
+	eventBroadcaster.StartLogging(glog.Infof)
+	eventBroadcaster.StartRecordingToSink(kubeClient.Events(""))
+
+	lbc := loadBalancerController{
+		client:         kubeClient,
+		clusterManager: clusterManager,
+		stopCh:         make(chan struct{}),
+		recorder: eventBroadcaster.NewRecorder(
+			api.EventSource{Component: "loadbalancer-controller"}),
```
brendandburns: move onto the previous line.

service-loadbalancer/gce/controller.go
```diff
@@ -0,0 +1,295 @@
...
+	lbc.ingQueue = NewTaskQueue(lbc.sync)
+
+	// Ingress watch handlers
+	pathHandlers := framework.ResourceEventHandlerFuncs{
+		AddFunc:    lbc.ingQueue.enqueue,
+		DeleteFunc: lbc.ingQueue.enqueue,
+		UpdateFunc: func(old, cur interface{}) {
+			if !reflect.DeepEqual(old, cur) {
+				glog.Infof("Ingress %v changed, syncing",
+					cur.(*experimental.Ingress).Name)
```
brendandburns: here too.

service-loadbalancer/gce/controller.go
```diff
@@ -0,0 +1,295 @@
...
+			}
+			lbc.ingQueue.enqueue(cur)
+		},
+	}
+	lbc.ingLister.Store, lbc.ingController = framework.NewInformer(
+		&cache.ListWatch{
+			ListFunc: func() (runtime.Object, error) {
```
brendandburns: ugh, this line breaking is ugly.  Can you extract this out as a separate function, instead of using an anonymous func

service-loadbalancer/gce/controller.go
```diff
@@ -0,0 +1,295 @@
...
+				return lbc.client.Experimental().Ingress(api.NamespaceAll).List(
+					labels.Everything(), fields.Everything())
+			},
+			WatchFunc: func(rv string) (watch.Interface, error) {
```
brendandburns: here too...

service-loadbalancer/gce/controller.go
```diff
@@ -0,0 +1,295 @@
...
+
+	lbc.svcLister.Store, lbc.svcController = framework.NewInformer(
+		cache.NewListWatchFromClient(
+			lbc.client, "services", api.NamespaceAll, fields.Everything()),
+		&api.Service{}, resyncPeriod, svcHandlers)
+
+	nodeHandlers := framework.ResourceEventHandlerFuncs{
+		AddFunc:    lbc.nodeQueue.enqueue,
+		DeleteFunc: lbc.nodeQueue.enqueue,
+		// Nodes are updated every 10s and we don't care, so no update handler.
```
brendandburns: why don't we care?  What if we miss an add?

service-loadbalancer/gce/controller.go
```diff
@@ -0,0 +1,295 @@
...
+		lbc.ingQueue.requeue(key, err)
+		return
+	}
+
+	ing := *obj.(*experimental.Ingress)
+	if urlMap, err := lbc.tr.toUrlMap(&ing); err != nil {
+		lbc.ingQueue.requeue(key, err)
+	} else if err := l7.UpdateUrlMap(urlMap); err != nil {
+		lbc.ingQueue.requeue(key, err)
+	} else if updateLbIp(
```
brendandburns: one line.

service-loadbalancer/gce/loadbalancer.go
```diff
@@ -0,0 +1,458 @@
...
+
+// gceUrlMap is a nested map of hostname->path regex->backend
+type gceUrlMap map[string]map[string]*compute.BackendService
+
+// getDefaultBackend performs a destructive read and returns the default
+// backend of the urlmap.
+func (g gceUrlMap) getDefaultBackend() *compute.BackendService {
+	var d *compute.BackendService
+	if h, ok := g[defaultBackendKey]; ok {
+		if d, ok = h[defaultBackendKey]; ok {
```
brendandburns: ugh, this shadowing `ok` is evil and should be avoided.

service-loadbalancer/gce/main.go
```diff
@@ -0,0 +1,164 @@
...
+	exitCode := 0
+	if err := lbc.Stop(); err != nil {
+		glog.Infof("Error during shutdown %v", err)
+		exitCode = 1
+	}
+	glog.Infof("Exiting with %v", exitCode)
+	os.Exit(exitCode)
+}
+
+func main() {
```
brendandburns: give an example of how you call this.

brendandburns:
first round...

bprashanth:
Bah, pushed godeps before I noticed you'd commented. Will re-up.

service-loadbalancer/gce/controller.go
bprashanth: We don't trust the event that's sent on the wire for this reason, the Informer will call the add handler if the thing it got isn't in local store (https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/framework/controller.go#L246). If we get some add/delete/drop update pattern that snubs us the relist will eventually kick in, whenever.

service-loadbalancer/gce/backends.go
bprashanth: I'm reusing err below that:
```
err := blah
if err != nil {
    return err
}
b.default, err = backend.Get()
if err != nil {
    return err
}
```
I mean I need to either create a new var or reuse, so no difference?

service-loadbalancer/gce/backends.go
bprashanth: Hmm, Get just gets the backend, I want to assign to the defaultBackend member before returning the object from the constructor

service-loadbalancer/gce/cluster_manager.go
bprashanth: They're just global consts but with fmt.Sprintf. I want to use the same string in multiple places and don't want to update multiple places if there's a breaking use case for that default.

bprashanth:
PTAL. 

Only thing I left as is are the line breaks. Can I continue limiting to 80 chars? You know I'm going to do this when you're not around right ;). Can still change if you really want. 

Also added a section to the README about paths.

brendandburns:
LGTM.  Good enough for the first cut, though 80 chars + Go makes me a sad panda.

brendandburns:
waiting on @thockin to merge in case he wants to take a look.

brendandburns:
@thockin ping on this.  Shall I just merge it?

thockin:
Looking now

thockin:
I read the README, but only skimmed the code, picked nits (mostly abbreviations and acronyms and naming).  Thanks for the good comments.

I'm OK to commit this when nits are satisfied.

bprashanth:
I can't change the interfaces, they mirror things in cloudprovider/gce or GCE's compute libarary. I'll do the rest, update the godep (experimental->extensions so this won't work on a HEAD cluster) and upload a new patch.

bprashanth:
PTAL, I think I uncovered a limitation that will need addressing in a follow up pr. Apparently GCE doesn't allow you to add more than 10 ports on a single Instance group so we can't share instance groups as aggressively (but instance groups are also limited by quota). Grr.

thockin:
I don't understand the "mirror things in cloudprovider/gce or GCE's compute libarary" - you mean they implement interfaces that happen to have bad names (if so comment that) or you just want them to be named the same?

bprashanth:
The former. Tldr it's because of compute.UrlMap and compute.HttpTargetProxy etc (https://github.com/kubernetes/kubernetes/blob/master/pkg/cloudprovider/providers/gce/gce.go#L900). Added comment: https://github.com/bprashanth/contrib/commit/2af0bc35f40455b354a0e06a7f720c8265204cf9#diff-0a671d4cc2db6b7adac7f4b4cfd594d4R47.

thockin:
LGTM

PR #134, opened 2015-10-01
Show container ip and port instead of service name in the backends

aledbf:
The reason for this change is show the ip address of the container running in the backend and help in case is required to know information about the container.

Instead of

```
backend example-go:8080
    option      httplog
    balance roundrobin
    server example-go:8080_0 10.2.21.11:8080
    server example-go:8080_1 10.2.21.8:8080
```

it use this

```
backend example-go:8080
    option      httplog
    balance roundrobin
    server 10.2.21.11:8080 10.2.21.11:8080
    server 10.2.21.8:8080 10.2.21.8:8080
```


bprashanth:
LGTM

PR #136, opened 2015-10-01
sevice-loadbalancer: First take at a sv lb diagram

jayunit100:
Heres a first take at a diagram for describing how this works for users.
Also hoping it will help clarify details which i may be missing.

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/136)

### DIFF
--- a/service-loadbalancer/README.md
+++ b/service-loadbalancer/README.md
@@ -15,6 +15,35 @@ This module will not help with the latter
 
 ## Overview
 
+The image below roughly describes the service-loadbalancer alternative (compared with using NodePorts).
+
+```
+                          +----------------------+                     
+                          | haproxy load balancer|                     
+                          +---------+------------+                     
+                                    |                                  
+                +-------------------+---------------+                  
+                |                                   |                  
+   +------------+----------------+     +-----------+--------------------+
+   |kubelet role=loadbalancer (1)|     |kubelet role=loadbalancer (2)   |
+   +----+------------------------+     +--------------------------------+
+        |                                                       |
+        |                                                       |
+        | --tcpServices=postgresql:3333                         | --tcpServices=postgresql:3333  
+        |                                                       |    
+        |               +-----------------------------+         |    
+        |               |   service for postgresql    |         |    
+        |               +-----------------------------|         |    
+        |                 | targetPort=2222,port=3333 |         |    
+        |           +------+-----------+    +----------------+  |    
+        |           |                  |    |                |  |    
+        +---------->| k1mysql          |    | k2mysql        | <-  
+                    | container        |    | container      |
+                    | Port=2222        |    | Port=2222      |          
+                    +------------------+    +----------------+          
+Direct load balanced connections to containers, without need for NodePort on every kubelet.
+```
+
 ### Ingress
 
 There are 2 ways to expose a service to ingress traffic in the current kubernetes service model:


jayunit100:
@bprashanth , does this look accurate to you ?

bprashanth:
It's a little confusing, haproxy can directly proxy to service endpoints, it doesn't need services with nodeports though your diagram seems to indicate it does. I can make more sensible recommendations when I have some more time. 

In somewhat related news: 
1. Are you up for testing servicelb? 
2. We should: mv service-loadbalancer haproxy; mkdir service-loadbalancers; mv haproxy service-loadbalancers.

@aledbf might also have some thoughts

jayunit100:
Yes, up for testing, happy to put those changes in that you mentioned also

jayunit100:
@bprashanth look closer :) the serviceloadbalancer is on the RIGHT, the OLD way is on the left... sorry i shuda labelled em

now makes more sense :) ?

bprashanth:
Oh, I missed that. Well I can't really comment on this, can I ask for an ascii rendition instead? that way we can go back and forth. Doesn't have to be fancy.

jayunit100:
ok will do graphviz ascii

jayunit100:
Ok @bprashanth , https://github.com/kubernetes/contrib/pull/136/files Updated with diagram generated from the ASCII text .

aledbf:
@jayunit100 what are you trying to you in the graph?

aledbf:
@jayunit100 I think the graph could be hard to follow.

I'm assuming the traffic reaches the same node with `role=loadbalancer`

`User -> service (VIP) -> kube-proxy -> go LB -> flannel -> pod/s`
with
`User -> haproxy VIP -> haproxy -> flannel -> pod/s`

Using haproxy to forward the traffic to the service removes functionality like controlling the algorithm to be used.

jayunit100:
Yes, @aledbf  you are correct, but  
- ..As per the docs `HAProxy is the only working implementation` , right? , so this diagram shows how a concrete use of service loadbalancers works.  
- I've updated it to clarify that :) .... 

does this make a little more sense?

jayunit100:
@aledbf ping.. does it make more sense now that i've updated the diagram to clarify the "loadbalancer" vs the old way of doing things?  

I think we are all on the same page, but the diagram may have been misleading originally.

jayunit100:
Hi folks, I've updated the README and figure as well to clarify things.

service-loadbalancer/service-loadbalancer.dot
```diff
@@ -0,0 +1,26 @@
...
+        user->haproxyLB [label="Service Loadbalancer: Pluggable + pod-aware loadbalancing alternative\n which forwards directly to endpoints."];
+
+} 
+        subgraph cluster_a { 
+        label="Nodeports";
+         haproxy->k1nodeport_8001;
+         haproxy->k2nodeport_8001;
+         k1nodeport_8001->service;
+         k2nodeport_8001->service;
+         service->k1PostgreSQL [label=targetPort_80];
```
jayunit100: nit.  possibly this could be **outside** the nodeport box?    just a thought I'd put in here as  part of a self-review im doing.

jayunit100:
Hi @bprashanth since this is just a doc update, 
- Any thoughts on merging this as a first pass doc on how it works?  
- Or do you think its misleading to people?   
I think it could really help to jumpstart the conversation around it ...

bprashanth:
I would still prefer raw ascii :)
I can cat the file, view it in terminal, copy/paste diagram into github comment etc. I'd only resort to graphing tools when the design is set in stone. Even the kubernetes architecture diagram is inaccurate, and no one has updated it. However I'll take anything with boxes and straight lines, for the sake of consistency with the rest of our docs.

I would also not shove both those before/after views of the world into the same diagram. Ideally something like:
Short explanation about what the first diagram is (without using specific terms like haproxy and postgres)
ascii
A short explanation about why the setup is sub optimal.
Segway into new setup with serviceloadbalacer
ascii 

And finally, I would number the steps like we do in services.md (https://github.com/kubernetes/kubernetes/blob/f00e0b4b18cbed8442c6b76134b9a6503cd29646/docs/user-guide/services.md#ips-and-vips)

jayunit100:
Ok.. I see what you mean. So, how do you diagram things in ascii? If you give me an example I can take a look. I've never done an ascii diagram before... Or are you suggesting no diagram? I'm also happy to close the PR if you think it's not a good idea......

jayunit100:
Ah ! I found http://asciiflow.com which I can try

bprashanth:
Ascii flow looks useful. No real format for ascii, that's why it's useful, just do something like https://lwn.net/Articles/620087/ or what I did in the proposal https://github.com/kubernetes/kubernetes/pull/12827/files

jayunit100:
Ok @bprashanth hows this ... 
- pure ascii . 
- removed the "before" vs "after" diagram style

service-loadbalancer/README.md
```diff
@@ -15,6 +15,35 @@ This module will not help with the latter
 
 ## Overview
 
+The image below roughly describes the service-loadbalancer alternative (compared with using NodePorts).
```
bprashanth: This sentence is confusing. Suggest removing and moving this diagram down to the section where we discuss tcp (since it's actually tcp specific).

service-loadbalancer/README.md
```diff
@@ -15,6 +15,35 @@ This module will not help with the latter
...
+
+```
+                          +----------------------+                     
+                          | haproxy load balancer|
```
bprashanth: this box is actually service loadbalancer, not haproxy. labeling it service loadbalancer implies some lb is being used, but the reverse is not true.

service-loadbalancer/README.md
```diff
@@ -15,6 +15,35 @@ This module will not help with the latter
...
+
+```
+                          +----------------------+                     
+                          | haproxy load balancer|                     
+                          +---------+------------+                     
+                                    |                                  
+                +-------------------+---------------+                  
+                |                                   |                  
+   +------------+----------------+     +-----------+--------------------+
+   |kubelet role=loadbalancer (1)|     |kubelet role=loadbalancer (2)   |
```
bprashanth: make this node role=loadbalancer since one cannot assign labels to the kubelet

service-loadbalancer/README.md
```diff
@@ -15,6 +15,35 @@ This module will not help with the latter
...
+   +----+------------------------+     +--------------------------------+
+        |                                                       |
+        |                                                       |
+        | --tcpServices=postgresql:3333                         | --tcpServices=postgresql:3333
```
bprashanth: you started the loadbalancer with --tcpServices right? so why not put this up there in the haproxy box?

service-loadbalancer/README.md
```diff
@@ -15,6 +15,35 @@ This module will not help with the latter
...
+        |                                                       |
+        | --tcpServices=postgresql:3333                         | --tcpServices=postgresql:3333  
+        |                                                       |    
+        |               +-----------------------------+         |    
+        |               |   service for postgresql    |         |    
+        |               +-----------------------------|         |    
+        |                      |  targetPort=2222,port=3333     |    
+        |           +------+-----------+    +----------------+  |    
+        |           |                  |    |                |  |    
+        +---------->| k1postgresql     |    | k2postgresql   | <-
```
bprashanth: Since the tcp example is actually mysql it would help if you reworded this example to suit the example

service-loadbalancer/README.md
```diff
@@ -15,6 +15,35 @@ This module will not help with the latter
...
+        |               |   service for postgresql    |         |    
+        |               +-----------------------------|         |    
+        |                      |  targetPort=2222,port=3333     |    
+        |           +------+-----------+    +----------------+  |    
+        |           |                  |    |                |  |    
+        +---------->| k1postgresql     |    | k2postgresql   | <-  
+                    | container        |    | container      |
+                    | Port=2222        |    | Port=2222      |          
+                    +------------------+    +----------------+          
+Direct load balanced connections  to kubelets, without need for NodePort on every kubelet.
```
bprashanth: The kubelet is nowhere in this picture actually, the loadbalancer opens up a tcp connection to the endpoint

service-loadbalancer/README.md
jayunit100: ah interesting i didn't realize the distinction

service-loadbalancer/README.md
jayunit100: This is a good point, by endpoint, you mean the containerPort, correct?  Since that container is running in a conceptual "pod", that is why its shown this way.  Maybe I'm missing something?

bprashanth: I meant the container can keep serving traffic even if the kubelet is dead

bprashanth:
Bump @jayunit100

jayunit100:
Hiya beeps.will look at it this week and update the diagram for sure !  got pulled in a few other efforts

jayunit100:
I've updated with the above comments.  Sorry about the delay ; have been pulled away on other stuff.
Am happy to close or handoff the patch to someone else if anyone wants to take over

test-foxish:
recomputing cla status...

test-foxish:
recomputing cla status...

jayunit100:
closing

PR #142, opened 2015-10-02
Update godep

bprashanth:
@eparis @lavalamp can one of you sanity check? Updating to grab: https://github.com/kubernetes/kubernetes/pull/14459

I essentially followed https://github.com/kubernetes/kubernetes/blob/92f21b3fe3399ae6e1c29979649ba7e64d6d096e/docs/devel/development.md#using-godep to checkout a clean kubernetes, godep restore to populate that temp GOPATH, cloned contrib, rm -rf Godeps, and godep save ./... in contrib to get the kubernetes godep I just restored.

Second commit changes go build to godep go build.


lavalamp:
Not sure about the second commit. There are more makefiles, no? Is godep smart enough to look for Godeps in a parent directory?

Godeps/Godeps.json
```diff
@@ -1,6 +1,6 @@
 {
 	"ImportPath": "k8s.io/contrib",
-	"GoVersion": "go1.4.2",
+	"GoVersion": "go1.5.1",
```
karlkfi: we haven't updated the rest of the project to 1.5 yet, so we should probably do this with 1.4.2 to make sure it works.

bprashanth:
> Not sure about the second commit. There are more makefiles, no? Is godep smart enough to look for Godeps in a parent directory?

The other makefiles use godep. I didn't touch the for-demos makefiles, but all the others I entered the dir and ran make.

Godeps/Godeps.json
bprashanth: fair enough and nice catch, godep just picked up the go from my system

bprashanth:
Let's do this? Assigning to Karl based on his interest

lavalamp:
oh, cool. LGTM then

bprashanth:
@davidopp on call fyi, self merging since contrib is wild wild west anyway and there's no real testing

karlkfi:
Ha. Definitely Wild West. 

That said, if we break out to more repos there's even less barrier to change ;)

lavalamp:
I want to make a standard set of tools for all of our repos... contrib
doesn't get enough attention :(

PR #143, opened 2015-10-05
Add support for etcd config file

glerchundi:
Mimics kube-apiserver behaviour and allows to access a TLS secured etcd too.

Signed-off-by: Gorka Lerchundi Osa <glertxundi@gmail.com>

### DIFF
--- a/pod-master/podmaster.go
+++ b/pod-master/podmaster.go
@@ -35,14 +35,15 @@ import (
 )
 
 type config struct {
-	etcdServers string
-	key         string
-	whoami      string
-	ttl         uint64
-	src         string
-	dest        string
-	sleep       time.Duration
-	lastLease   time.Time
+	etcdServers    string
+	etcdConfigFile string
+	key            string
+	whoami         string
+	ttl            uint64
+	src            string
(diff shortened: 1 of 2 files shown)


pod-master/podmaster.go
```diff
@@ -35,14 +35,15 @@ import (
...
-	etcdServers string
-	key         string
-	whoami      string
-	ttl         uint64
-	src         string
-	dest        string
-	sleep       time.Duration
-	lastLease   time.Time
+	etcdServers    string
+	etcdConfigFile string
```
wojtek-t: My feeling was that we are going to get rid of it in apiserver.

@timothysc @lavalamp - am I right?

glerchundi:
I signed it!

pod-master/podmaster.go
glerchundi: Ok, If you all agree I'll close this PR.

Whatever you guys decide it's ok but having a tls capable podmaster is absolutely necessary :-)

timothysc: Yes, it's removed in the new client patches.

PR #145, opened 2015-10-06
Add a simple tool for calculating a CSV of perf. data from Jenkins.

brendandburns:
@wojtek-t

### DIFF
--- a/perfdash/perfdash.go
+++ b/perfdash/perfdash.go
@@ -1,0 +1,149 @@
+/*
+Copyright 2015 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"bufio"
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	// TODO: move this somewhere central
+	"k8s.io/contrib/submit-queue/jenkins"
+	"k8s.io/kubernetes/pkg/util/sets"
+)
+
+// LatencyData represents the latency data for a set of RESTful API calls
+type LatencyData struct {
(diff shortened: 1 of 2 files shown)


perfdash/perfdash.go
```diff
@@ -0,0 +1,125 @@
...
+		defer reader.Close()
+		scanner := bufio.NewScanner(reader)
+		buff := &bytes.Buffer{}
+		inLatency := false
+
+		hist := resourceToHistogram{}
+		found := false
+		for scanner.Scan() {
+			line := scanner.Text()
+			// TODO: This is brittle, we should emit a tail delimiter too
```
wojtek-t: +1 :)

perfdash/perfdash.go
```diff
@@ -0,0 +1,125 @@
...
+			if strings.Contains(line, "INFO") || strings.Contains(line, "STEP") || strings.Contains(line, "Failure") {
+				if inLatency {
+					obj := LatencyData{}
+					if err := json.Unmarshal(buff.Bytes(), &obj); err != nil {
+						fmt.Printf("error parsing JSON in build %d: %v %s\n", build.Number, err, buff.String())
+						return
```
wojtek-t: Do we want to return here?
I think that in case of corrupted data in some build, we should break everything - we should just break the computation for this particular build.

perfdash/perfdash.go
```diff
@@ -0,0 +1,125 @@
...
+	fmt.Println(strings.Join(header, ","))
+	
+	for build, data := range buildLatency {
+		line := []string{fmt.Sprintf("%d", build)}
+		for _, rsrc := range resources.List() {
+			podData := data[rsrc]
+			line = append(line, fmt.Sprintf("%g", findMethod("GET", "Perc50", podData)))
+			line = append(line, fmt.Sprintf("%g", findMethod("GET", "Perc90", podData)))
+			line = append(line, fmt.Sprintf("%g", findMethod("GET", "Perc99", podData)))
+		}
```
wojtek-t: IIUC - we are going to add more (like POST, PUT, ...) in the future, right?

wojtek-t:
Looks good as a first step :)

perfdash/perfdash.go
brendandburns: done.

perfdash/perfdash.go
brendandburns: done.

wojtek-t:
"Verifying hack/../hack/verify-golint.sh FAILED"

@brendandburns Can you please fix it?

Other than that LGTM - feel free to self-apply the label to not block on waiting for me.

brendandburns:
fixed. merging.

PR #148, opened 2015-10-08
ignore errors for nodes without a running firewalld

jasonbrooks:
If firewalld is installed on a fedora or centos system, the ansible scripts attempt to configure it, whether it's running or not. Most of the tasks for configuring firewalld are set to ignore errors, which allows the provisioning to continue on hosts that have firewalld installed, but where firewalld is not running. This PR adds that same `ignore_errors: yes` to the node firewalld tasks.

cc @eparis


elemoine:
With 9a36b5fcbc19567682cbee5aac0bf92713305d21 should this PR be closed?

PR #155, opened 2015-10-09
Recheck long running e2e tests are stable before merge

eparis:
Since we rerun the github e2e tests (the fast ones) async, the time
between the check for stability of the slow google internal e2e tests
and when we are ready to merge could be quite large. Recheck the state
of the slow e2e tests before we merge.


eparis:
@lavalamp @wojtek-t this should (greatly reduce) the race

lavalamp:
LGTM

lavalamp:
(I haven't restarted the bot yet-- it may be a while before I have a chance)

PR #157, opened 2015-10-09
Flush the async github e2e retester on slow e2e failure

eparis:
As soon as we notice that the slow e2e tests are failing, flush the
queue of PRs waiting to rerun the fast e2e tests.


eparis:
@lavalamp

lavalamp:
LGTM

PR #162, opened 2015-10-12
Add support for session affinity using service.spec.sessionAffinity or cookies

aledbf:
Use the attribute `sessionAffinity` from the service spec to enable the use a stickiness table in haproxy
http://cbonte.github.io/haproxy-dconv/configuration-1.5.html#stick-table

Example: 
- stick-table
  `sessionAffinity=ClientIP`
- cookies
  `sessionAffinity=ClientIP` and annotation `serviceloadbalancer/lb.cookie-sticky-session=true`

### DIFF
--- a/service-loadbalancer/service_loadbalancer.go
+++ b/service-loadbalancer/service_loadbalancer.go
@@ -25,6 +25,7 @@ import (
 	"os"
 	"os/exec"
 	"reflect"
+	"sort"
 	"strconv"
 	"strings"
 	"text/template"
@@ -166,6 +167,18 @@ type service struct {
 	// The name of the cookie is SERVERID
 	// This only can be used in http services
 	CookieStickySession bool
+}
+
+type serviceByName []service
+
+func (s serviceByName) Len() int {
+	return len(s)
+}
+func (s serviceByName) Swap(i, j int) {
+	s[i], s[j] = s[j], s[i]
+}
+func (s serviceByName) Less(i, j int) bool {
+	return s[i].Name < s[j].Name
 }
 
 // loadBalancerConfig represents loadbalancer specific configuration. Eventually
@@ -417,6 +430,10 @@ func (lbc *loadBalancerController) getServices() (httpSvc []service, tcpSvc []se
 			glog.Infof("Found service: %+v", newSvc)
 		}
 	}
+
+	sort.Sort(serviceByName(httpSvc))
+	sort.Sort(serviceByName(tcpSvc))
+
 	return
 }
 
--- a/service-loadbalancer/service_loadbalancer.go
+++ b/service-loadbalancer/service_loadbalancer.go
@@ -44,12 +44,13 @@ import (
 )
 
 const (
-	reloadQPS        = 10.0
-	resyncPeriod     = 10 * time.Second
-	lbApiPort        = 8081
-	lbAlgorithmKey   = "serviceloadbalancer/lb.algorithm"
-	lbHostKey        = "serviceloadbalancer/lb.host"
-	defaultErrorPage = "file:///etc/haproxy/errors/404.http"
+	reloadQPS                = 10.0
+	resyncPeriod             = 10 * time.Second
+	lbApiPort                = 8081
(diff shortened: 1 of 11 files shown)


aledbf:
ping @bprashanth

service-loadbalancer/service_loadbalancer.go
```diff
@@ -149,6 +149,10 @@ type service struct {
 
 	// Algorithm
 	Algorithm string
+
+	// Indicates if the service must use sticky sessions
+	// http://cbonte.github.io/haproxy-dconv/configuration-1.5.html#stick-table
+	SessionAffinity bool
```
bprashanth: please update the README about how to use/set this, also document the other annotations we've recently added if they're not already there

service-loadbalancer/service_loadbalancer.go
```diff
@@ -370,6 +374,12 @@ func (lbc *loadBalancerController) getServices() (httpSvc []service, tcpSvc []se
 				}
 			}
 
+			// By default sticky session is disabled
+			newSvc.SessionAffinity = false
+			if s.Spec.SessionAffinity != "None" {
```
bprashanth: can we get cookie based stickyness as well?

aledbf: > can we get cookie based stickyness as well?

Yes is possible. I will add the key `serviceloadbalancer/lb.cookie-sticky-session`
If the service uses `SessionAffinity` and the annotation `serviceloadbalancer/lb.cookie-sticky-session` it will use cookies.

service-loadbalancer/service_loadbalancer.go
```diff
@@ -149,6 +150,17 @@ type service struct {
...
+
+	// Indicates if the service must use sticky sessions
+	// http://cbonte.github.io/haproxy-dconv/configuration-1.5.html#stick-table
+	// Enabled using the attribute service.spec.sessionAffinity
+	// https://github.com/kubernetes/kubernetes/blob/master/docs/user-guide/services.md#virtual-ips-and-service-proxies
+	SessionAffinity bool
+
+	// CookieStickySession use a cookie to enable sticky sessions.
+	// The name of the cookie is SERVERID
+	// This only can be used in http services
```
bprashanth: Please include a comment like (assuming my comment is accurate, i haven't double checked):
```
If SessionAffinity is set and without CookieStickySession, requests are routed to a backend based on client ip. If both SessionAffinity and CookieStickSession are set, a SERVERID cookie is inserted by the loadbalancer and used to route subsequent requests. If neither is set, requests are routed based on the algorithm.
```

As an aside, what happens when a backend is unhealthy, is it automatically purged from the stick table and all SERVERID cookies are ignored?

service-loadbalancer/service_loadbalancer.go
bprashanth: Please make this
```
s.Spec.SessionAffiity != api.SessionAffinityNone
```

bprashanth:
LGTM sans nits. 2 suggestions:
1. Please write a unittest like this for the config, since it's getting more and more complex: https://github.com/bprashanth/Ingress/blob/master/controllers/nginx/template_test.go
2. Please add a subsection to the README explaning how one would use the new annotations

If you're doing those, you can do it in another pr if you want to keep this one small.

service-loadbalancer/service_loadbalancer.go
aledbf: If the backend is unhealthy haproxy will choose other server from the backend

service-loadbalancer/service_loadbalancer.go
```diff
@@ -149,6 +151,34 @@ type service struct {
...
+	CookieStickySession bool
+}
+
+type serviceByName []service
+
+func (s serviceByName) Len() int {
```
bprashanth: hmm, there might be a way to take this a step further and not reload haproxy if nothing has changed (hash?). Currently if i update a service or endpoint with something unrelated we'll still whack haproxy.

bprashanth:
LGTM

PR #165, opened 2015-10-13
Start to collect and export more information about the e2e queue

eparis:
From a user PoV it should have no changes, but it should put more data in the API which can used to develop the web interface.


eparis:
@lavalamp can you push a new build after I merge this?

lavalamp:
I've built and pushed.

PR #168, opened 2015-10-13
Display github e2e queue in submit queue web page

eparis:


wojtek-t:
Was this already pushed?

eparis:
no. was working on fixing the deadlock and test after merge bug before i ask for another build

PR #171, opened 2015-10-14
Recheck merged and mergeability before rerunning e2e

eparis:
Because humans merge things by hand and the queue can then get hung


eparis:
@lavalamp probably worth another build and kick

eparis:
It'll get moving again, but right now it's hung up on https://github.com/kubernetes/kubernetes/pull/14871 being merged by hand.

eparis:
Hey, the timeout worked!  But now it's being dumb on the next thing dawn merged by hand. /me slaps bot.

lavalamp:
OK, will push

lavalamp:
Pushed.

BTW I know contrib hasn't been getting much attention but we should probably at least pretend to do a little bit more review before self merging :)

PR #174, opened 2015-10-14
Ansible: fix typo

gouyang:
Signed-off-by: Guohua Ouyang <gouyang@redhat.com>


eparis:
I'm shocked this ever worked, but i'm 99% sure it did/does.  Did you have a problem that this fixes?

gouyang:
I didn't see any problem so far, it seems once "--service-account-private-key-file" is added in /etc/kubernetes/controller-manager, it will create the service account "default" automatically.

It should be a bug of k8s, right?

eparis:
So I think this is correct, but maybe the flag names are terrible?  The controller manager gets the `private-key-file` and the apiserver gets the `key-file`. Which is really the public half of the private key. This is what we are doing...

gouyang:
@eparis are you meaning using "server.crt" is correct?

eparis:
Yes, I think so...

PR #176, opened 2015-10-14
Update Google Internal E2E Status More Often

eparis:
Currently, the submit queue web page only updates the google internal e2e
when it is processing PRs. This means if no PRs gets to that point in the
submit queue logic we will never update the web page. But even if there
are PRs which get to that point, the loop is every 30 minutes so its a
long delay.

Also, the web interface showed 'red' if it failed but never would go
back green if the tests started passing. So fix the javascript there.


eparis:
@wojtek-t

PR #177, opened 2015-10-14
Clear the submit queue any time we hit a non-stable e2e

eparis:
We were clearning the submit queue when we found google internal e2e was
unstable during the 30 minute loop, but not when we found e2e failed
after a github build. Flush e2e any time we found an unstable build.


eparis:
@wojtek-t

wojtek-t:
LGTM

PR #180, opened 2015-10-14
SubmitQueue: Web Page More Mobile Friendly

eparis:
We did some static offsets and uses plain html elements in a number of
places. These looked great on a big monitor but didn't scale well to a
mobile device.  Using all material design elements looks good on an
iPhone and has little visual impack on big displays.


eparis:
http://plnkr.co/edit/M9xY3oM8e9VChnPXJsFt?p=preview

Is a good way to see the web site that will come out.

PR #175, opened 2015-10-14
Add zmerlynn to review rotation for examples/

davidopp:
@mikedanese @zmerlynn

PR #181, opened 2015-10-15
Add zmerlynn to review rotation for examples

eparis:


eparis:
same as #175 except it can be merged  :)

PR #183, opened 2015-10-16
Ansible fixes

cgwalters:
See commits for detail.  /cc @pedro-r-marques for regressions.

### DIFF
--- a/ansible/cluster.yml
+++ b/ansible/cluster.yml
@@ -33,7 +33,7 @@
 - hosts: all
   sudo: yes
   roles:
-    - { role: opencontrail, when networking == 'opencontrail'}
+    - { role: opencontrail, when: networking == 'opencontrail'}
   tags:
     - opencontrail
 
--- a/ansible/roles/master/tasks/main.yml
+++ b/ansible/roles/master/tasks/main.yml
@@ -70,13 +70,17 @@
 - name: write the kubecfg (auth) file for kubectl
   template: src=kubectl.kubeconfig.j2 dest={{ kube_config_dir }}/kubectl.kubeconfig
 
+# Enable kubelet on master only when OpenContrail is in use; see
+# https://github.com/kubernetes/contrib/pull/183
 - name: write the config files for kubelet
   template: src=kubelet.j2 dest={{ kube_config_dir }}/kubelet
   notify:
     - restart kubelet
+  when: networking == 'opencontrail'
 
 - name: Enable kubelet
   service: name=kubelet enabled=yes state=started
+  when: networking == 'opencontrail'
 
 - include: firewalld.yml
   when: has_firewalld
--- a/ansible/roles/master/tasks/packageManagerInstall.yml
+++ b/ansible/roles/master/tasks/packageManagerInstall.yml
@@ -1,4 +1,8 @@
 ---
+- name: Init the did_install fact
+  set_fact:
+    did_install: false
+
 - name: Set fact for Atomic Host package install
   set_fact:
     did_install: true


cgwalters:
@googlebot I'm a Red Hat employee.

cgwalters:
This time with `@redhat.com` as author.

cgwalters:
I signed it!

eparis:
@googlebot he signed it.

jayunit100:
Looks like maybe we should merge this..  I tested it, it seems to be a step forward ?  But not sure what opencontrail is really supposed to do.  At least this patch makes it **optional**...

- I  think right now as-is, opencontrail is installed no matter what, 
- Additionally, it seems like it might be a little buggy... 
-  this patch fixed this for me... now opencontrail isnt complainaing,  I think because the `:when` clause disables it ...  

In any case, the patch helped me

eparis:
commits 1 and 2 I'm ready to merge now. commit #3 I'm going to wait for @pedro-r-marques 

What actual problem did you get @cgwalters ?

pedro-r-marques:
@eparis @cgwalters Enabling kubelet on the master should have been conditional.
Can we have a var 'kube_master_run_kubelet' that would be set in order to enable this  ?

pedro-r-marques:
Please see #184 as a proposal to address the issue as whether to run kubelet on the master node.

cgwalters:
I pushed an alternative approach :arrow_up:  which just does it by default iff OpenContrail, replacing the last patch.

eparis:
It's better, it's what @pedro-r-marques but I won't merge either until I hear what was actually wrong with running the kublet on the master. I actually LIKE having the kubelet on the master. It means you can use static pod manifests consistently...

cgwalters:
Okay, I created another fresh environment without the tip commit here, and here's what I ran into:

```
TASK: [master | Enable kubelet] *********************************************** 
failed: [walters-os1priv-rhelah-kubetest-master] => {"failed": true}
msg: Error when trying to enable kubelet: rc=1 Failed to execute operation: Access denied


FATAL: all hosts have already failed -- aborting
```

It's simple, we're not installing the `kubernetes-kubelet` package.

cgwalters:
I actually didn't know what static pods are, I found https://github.com/kubernetes/kubernetes/blob/master/docs/admin/static-pods.md  - this one is deprecated, is there another one?

pedro-r-marques:
Colin,
The kubelet is being installed in the build from source option, but not the install from .rpms option.

  Pedro.

On Oct 19, 2015, at 2:53 PM, Colin Walters <notifications@github.com<mailto:notifications@github.com>> wrote:


Okay, I created another fresh environment without the tip commit here, and here's what I ran into:

TASK: [master | Enable kubelet] ***********************************************
failed: [walters-os1priv-rhelah-kubetest-master] => {"failed": true}
msg: Error when trying to enable kubelet: rc=1 Failed to execute operation: Access denied


FATAL: all hosts have already failed -- aborting


It's simple, we're not installing the kubernetes-kubelet package.

eparis:
@cgwalters yes, that's the one. and while it's deprecated, its not going anywhere until the daemon controller is a first class citizen. So it's all we have.

eparis:
@pedro-r-marques I'm going to merge this PR, and close yours. Will you please send another PR which installs kubernetes-node on the master if contrail is enabled?

cgwalters:
I was working on that this morning - basically merging the master/node package bits into the `kubernetes/` role.  But ran into flannel failing, debugging that now.

cgwalters:
@eparis Isn't that exactly what https://github.com/kubernetes/contrib/commit/401711b6afe75c269ba174435865e4968c61bbe5 that you merged does?

Anyways I followed up with https://github.com/kubernetes/contrib/pull/188 and after that, we can revert that commit.

PR #184, opened 2015-10-19
Introduce setting as to whether to run the kubelet on a master.

pedro-r-marques:
Kubelet should be an optional setting.

### DIFF
--- a/ansible/cluster.yml
+++ b/ansible/cluster.yml
@@ -44,6 +44,8 @@
     - master
   tags:
     - masters
+  vars:
+    kube_master_run_kubelet: networking == 'opencontrail'
 
 # launch addons, like dns
 - hosts: masters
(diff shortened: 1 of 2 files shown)


eparis:
I also might be fine if we forced a kubelet on the master, assuming it was 'non-schedulable' unless the the master was also a node...

pedro-r-marques:
@eparis The kubelet configuration has the option: "--register-node=false"; this makes it so that the kubelet doesn't register itself as a node. This is set in roles/master/templates/kubelet.j2.

eparis:
ok, i'll wait to merge until we hear the actual problem @cgwalters to decide if we should merge this or just leave it as it is...

eparis:
@pedro-r-marques I'm going to close this PR as it was a part of @cgwalters fix.  Will you please send another PR which installs kubernetes-node on the master if contrail is enabled?

pedro-r-marques:
Will do.

On Oct 20, 2015, at 8:23 AM, Eric Paris <notifications@github.com<mailto:notifications@github.com>> wrote:


@pedro-r-marques<https://github.com/pedro-r-marques> I'm going to close this PR as it was a part of @cgwalters<https://github.com/cgwalters> fix. Will you please send another PR which installs kubernetes-node on the master if contrail is enabled?

PR #186, opened 2015-10-19
Update kubernetes godep

eparis:
Which in turn pulls in others

and removes all of _test.go


bprashanth:
Can you please update the top level contrib README with basic instructions of revving the godep? Just sometime simple like: cd contrib, rm -rf Godep, cd... etc (or just point to the docs in the main repo). I find a lot of contributors miss that this repo even requires an update.

PR #188, opened 2015-10-20
ansible: Unify pkg installs for master/node

cgwalters:
To support static pods, also install kubernetes-node on the master.
That makes the package code identical between `master/` and `node/`
roles, so deduplicate into the dependent `kubernetes/` role.

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/188)


eparis:
just getting off my lazy butt after 7 months. If you rebase, I will merge.

ingvagabund:
Closing this in favor of #1444, both master and node are independent of their installation so one can choose rpm based master vs. github-release node and other combinations.

PR #189, opened 2015-10-20
Fix bug in queuing logic of serviceloadbalancer

bprashanth:
We should be marking all keys done when we're finished processing them, even the ones we requeue. Otherwise we might leave a zombie key in the processing queue that halts progress.


bprashanth:
@aledbf can you please review?

brendandburns:
lgtm.

aledbf:
Code LGTM

PR #200, opened 2015-10-24
Simple 404 server

bprashanth:
A dumb 404 server for use as a loadbalancer controllers default backend.

### DIFF
--- a/404-server/server.go
+++ b/404-server/server.go
@@ -1,0 +1,38 @@
+/*
+Copyright 2015 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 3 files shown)


bprashanth:
Used in https://github.com/kubernetes/kubernetes/pull/16232

404-server/server.go
```diff
@@ -0,0 +1,38 @@
...
+	"fmt"
+	"net/http"
+)
+
+var port = flag.Int("port", 8080, "Port number to serve default backend 404 page.")
+
+func main() {
+	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusNotFound)
+		fmt.Fprint(w, "default backend - 404")
```
thockin: Avenues for growth - add a defined volume mount point and let the user provide a template file and images.

PR #201, opened 2015-10-24
Glbc fixes round 2

bprashanth:
~~Still a work in progress.~~


bprashanth:
This is good to go

Needs: https://github.com/kubernetes/kubernetes/pull/16142 and https://github.com/kubernetes/kubernetes/pull/16221 in the godep, after which i need to re-push the image.

This is the image for: https://github.com/kubernetes/kubernetes/pull/16232

PR #202, opened 2015-10-26
Update kubernetes godep

janetkuo:
cc @bprashanth


eparis:
How was this generated?

eparis:
(and I guess why?)

bprashanth:
@eparis pretty much https://github.com/kubernetes/contrib i think
To pull in some changes to Ingress resource, but also ends up pulling a few other changes to related packages apparently. Would you prefer keeping the other packages as they were? I figured always keeping kube and contrib in sync with every roll makes sense?

bprashanth:
LGTM. Will wait for @eparis before merging.

PR #206, opened 2015-10-26
[wip] Haproxy ingress

bprashanth:
See first commit for actual changes to existing code, everything else is diff confusion. This pr:
* Adds a --use-ingress flag to use ingress
* Adds a --watch-all flag to watch all namespaces instead of just one. This only makes sense with ingress, as services might have name collisions across namespace.
* Move service-loadbalancer to Ingress/controller

@aledbf I'm totally OK with you taking over this pr if you have the bandwidth ;)


yuvipanda:
This should probably be closed in favor of #223

foxish:
@bprashanth Can we close this?

PR #209, opened 2015-10-28
submit queue: Do not update 'status' if PR already in queue

eparis:
If a PR is on the merge queue and the munger runs again, it will
(likely) determine that the PR needs to be added to the queue. The
problem is that we also put this determination in the history. So we log
every 10 minutes that a PR is being 'enqueued' even though it was always
on the queue. This makes the log less useful.


lavalamp:
LGTM

lavalamp:
feel free to self merge when you next push

PR #210, opened 2015-10-28
Add multiple namespace support in service loadbalancer

### LINKED ISSUE #203: Service loadbalancer should support multiple namespaces

aledbf:
Closes #203

### DIFF
--- a/service-loadbalancer/service_loadbalancer.go
+++ b/service-loadbalancer/service_loadbalancer.go
@@ -171,9 +171,11 @@ type service struct {
 	// This only can be used in http services
 	CookieStickySession bool
 
-	// Namespace is the name of the namespace where the service running.
-	// If is empty means that is running in the "default" namespace
-	Namespace string
+	// URLs are the paths used to route the traffic to the backend, like "<namespace>/<svc name>:<port>".
+	// It requires at least one URL
+	// If the namespace is "default" it will not be used.
+	// If the port if is equals to 80 it will not be used.
+	URLs []string
 }
 
 type serviceByName []service
@@ -358,14 +360,35 @@ func (lbc *loadBalancerController) getEndpoints(
 	return
 }
 
-// encapsulates all the hacky convenience type name modifications for lb rules.
-// - :80 services don't need a :80 postfix
-// - default ns should be accessible without /ns/name (when we have /ns support)
-func getServiceNameForLBRule(s *api.Service, servicePort int) string {
-	if servicePort == 80 {
-		return s.Name
+// LBRule encapsulates a rule object compatible with Ingress to allow custom routing rules
+type LBRule struct {
+	ACL  string
+	URLs []string
+}
+
+type RuleGetter interface {
+	getRules(s *api.Service, servicePort int) *LBRule
+}
+
+func getLBRules(lb RuleGetter, s *api.Service, servicePort int) *LBRule {
+	return lb.getRules(s, servicePort)
+}
+
+func (lbc *loadBalancerController) getRules(s *api.Service, servicePort int) *LBRule {
+	prefix := ""
+	if !strings.EqualFold(s.Namespace, api.NamespaceDefault) {
+		prefix = "/" + s.Namespace
 	}
-	return fmt.Sprintf("%v:%v", s.Name, servicePort)
+
+	port := ""
+	if servicePort != 80 {
+		port = ":" + strconv.Itoa(servicePort)
+	}
+
+	return &LBRule{
+		ACL:  fmt.Sprintf("%v_%v%v", s.Namespace, s.Name, port),
+		URLs: []string{fmt.Sprintf("%v/%v%v", prefix, s.Name, port)},
+	}
 }
 
 // getServices returns a list of services and their endpoints.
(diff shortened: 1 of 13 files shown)


service-loadbalancer/service_loadbalancer.go
```diff
@@ -635,7 +651,11 @@ func main() {
 		glog.Fatalf("unexpected error: %v", err)
 	}
 	if !specified {
-		namespace = "default"
+		namespace = api.NamespaceDefault
```
bprashanth: to make this ordering clearer (i.e that allNamespaces always wins) please make it an if/elseif/else where in the "specified" case you just log the namespace.

service-loadbalancer/service_loadbalancer.go
```diff
@@ -413,6 +423,12 @@ func (lbc *loadBalancerController) getServices() (httpSvc []service, tcpSvc []se
 				newSvc.SessionAffinity = true
 			}
 
+			if strings.EqualFold(s.Namespace, api.NamespaceDefault) {
```
bprashanth: Do you actually need the service name for anything or can we just call the field `path`? I'd prefer having a single function called getServicePathForLBRule that takes a Service+port and returns the path for it. This may/may-not include namespace port etc based on some if statements. That makes it easier to integrate into Ingress as well (https://github.com/bprashanth/contrib/commit/dca689f7e6f805d54c90717814a86b8f0a696c04#diff-75fc7741fc92747587a186ae42240c0eR427).

service-loadbalancer/service_loadbalancer.go
```diff
@@ -387,9 +420,17 @@ func (lbc *loadBalancerController) getServices() (httpSvc []service, tcpSvc []se
 					sName, servicePort)
 				continue
 			}
+
+			rules := getLBRules(lbc, &s, servicePort.Port)
```
bprashanth: please pass in servicePort instead of servicePort.Port, since we also have named ports.

service-loadbalancer/service_loadbalancer.go
```diff
@@ -348,14 +360,35 @@ func (lbc *loadBalancerController) getEndpoints(
...
+type LBRule struct {
+	ACL  string
+	URLs []string
+}
+
+type RuleGetter interface {
+	getRules(s *api.Service, servicePort int) *LBRule
+}
+
+func getLBRules(lb RuleGetter, s *api.Service, servicePort int) *LBRule {
```
bprashanth: do we really need this function if it's just passthrough? i.e something like:
```go
// implements ruleGetter
type serviceRuleGetter struct {
    loadbalancercontroller
}

type loadBalancerController struct {
   ruleGetter RuleGetter
}

func getServices() {
  lbc.ruleGetter.getRule(svc, svcport)
}

func newLoadBalancerController() {
  lbc := loadbalancercontroller {}
  lbc.ruleGetter = serviceRuleGetter{lbc}
  return lbc
}
```

service-loadbalancer/service_loadbalancer.go
```diff
@@ -348,14 +360,35 @@ func (lbc *loadBalancerController) getEndpoints(
...
-	if servicePort == 80 {
-		return s.Name
+// LBRule encapsulates a rule object compatible with Ingress to allow custom routing rules
+type LBRule struct {
+	ACL  string
+	URLs []string
+}
+
+type RuleGetter interface {
+	getRules(s *api.Service, servicePort int) *LBRule
```
bprashanth: Either make this a private interface or make the getRules method public. 
Also suggest:
// RuleGetter translates a service:port combination to a loadbalancer rule. Eg: foo, 8080 becomes acl rule ...

service-loadbalancer/service_loadbalancer.go
```diff
@@ -348,14 +360,35 @@ func (lbc *loadBalancerController) getEndpoints(
...
 
-// encapsulates all the hacky convenience type name modifications for lb rules.
-// - :80 services don't need a :80 postfix
-// - default ns should be accessible without /ns/name (when we have /ns support)
-func getServiceNameForLBRule(s *api.Service, servicePort int) string {
-	if servicePort == 80 {
-		return s.Name
+// LBRule encapsulates a rule object compatible with Ingress to allow custom routing rules
+type LBRule struct {
+	ACL  string
```
bprashanth: Please document ACL and URLs, it's ok to say see documentation for service urls for URLs. Please also be specific about what ACL is or just link to haproxy docs. Also comment on weather regexes are allowed in URLs.

service-loadbalancer/service_loadbalancer.go
```diff
@@ -326,6 +335,9 @@ func (lbc *loadBalancerController) getEndpoints(
 	// the target port are capable of service traffic for it.
 	for _, ss := range ep.Subsets {
 		for _, epPort := range ss.Ports {
+			if epPort.Protocol == api.ProtocolUDP {
```
bprashanth: is this needed? i don't remember how our conversation about a similar guard around the service port in the calling function ended

service-loadbalancer/service_loadbalancer.go
```diff
@@ -167,6 +170,12 @@ type service struct {
 	// The name of the cookie is SERVERID
 	// This only can be used in http services
 	CookieStickySession bool
+
+	// URLs are the paths used to route the traffic to the backend, like "<namespace>/<svc name>:<port>".
```
bprashanth: suggest a rewording:
// URLs is a list of paths used to route traffic to the set of endpoints associated with a Service.
// Each URL takes the form: `<namespace>/<svc name>:<port>`

Add the comment about omissiong of the default namespace and port 80 above the implementation of getLBRules, since different rulegetters might do different things.

bprashanth:
I think travis is sad because you're missing golint on some file (hack/verify-golint)

service-loadbalancer/service_loadbalancer.go
aledbf: yes because without this a service using TCP and UDP result in two elements in the array with the same ip:port.

silasbw:
@aledbf I'm interested in this. Why did you close it? Are you planning on picking it back up anytime soon?

PR #211, opened 2015-10-29
Avoid the listening port conflict of etcd.

morganwu277:
Usually we use `2379` as the default port for communicating with other etcd instances, we should use `4001` port to let local etcdctl access  etcd contents.


morganwu277:
I signed it!

morganwu277:
YES, my email configuration was wrong. Sorry.

PR #213, opened 2015-10-31
Handle paging when getting commits

eparis:
The change after LGTM hook has been trigger incorrectly. See:

https://github.com/kubernetes/kubernetes/pull/13146
https://github.com/kubernetes/kubernetes/pull/15994

This handles paging on commits (which we shouldn't have more than 100
commits, but that is a code path this munger uses) and adds a touch more
debug info.

### DIFF
--- a/mungegithub/github/github.go
+++ b/mungegithub/github/github.go
@@ -563,15 +563,12 @@ func (obj *MungeObject) IsPR() bool {
 
 // GetEvents returns a list of all events for a given pr.
 func (obj *MungeObject) GetEvents() ([]github.IssueEvent, error) {
-	if obj.events != nil {
-		return obj.events, nil
-	}
 	config := obj.config
 	prNum := *obj.Issue.Number
 	events := []github.IssueEvent{}
 	page := 1
 	for {
-		eventPage, response, err := config.client.Issues.ListIssueEvents(config.Org, config.Project, prNum, &github.ListOptions{Page: page})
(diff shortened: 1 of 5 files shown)


mungegithub/github/github.go
```diff
@@ -637,12 +634,14 @@ func (obj *MungeObject) GetStatus(requiredContexts []string) (string, error) {
 		glog.Errorf("pr.Head is nil in GetStatus for PR# %d", *pr.Number)
 		return "failure", nil
 	}
+	// TODO If we have more than 100 statuses we need to deal with paging.
 	combinedStatus, response, err := config.client.Repositories.GetCombinedStatus(config.Org, config.Project, *pr.Head.SHA, &github.ListOptions{})
```
lavalamp: Is setting PerPage to 100 needed or useful?

mungegithub/mungers/lgtm_after_commit.go
```diff
@@ -62,6 +62,7 @@ func (LGTMAfterCommitMunger) Munge(obj *github.MungeObject) {
 	}
 
 	if lastModified.After(*lgtmTime) {
+		glog.Infof("lgtm:%s  lastModified:%s", lgtmTime.String(), lastModified.String())
```
lavalamp: Print the PR #, or the logs will be confusing.

lavalamp:
Feel free to self-lgtm once you fix nits :)

lavalamp:
Did those commits actually break the travis tests? :)

eparis:
no, -tip is broken, not the tests.

lavalamp:
gotcha.

PR #218, opened 2015-11-05
1) Avoid the listening port conflict of etcd ; 2) kube-controller-manager.service should be started after kube-apiserver.service

morganwu277:
1. Usually we use 2379 as the default port for communicating with other etcd instances, we should use 4001 port to let local etcdctl access etcd contents for preventing port conflicts.
2. kube-controller-manager.service should be started after kube-apiserver.service. Otherwize, we'll meet

```
Nov 04 01:31:12 centos7-node-221 kube-controller-manager[963]: F1104 01:31:12.757305     963 controllermanager.go:292] Failed to get api versions from server: Get https://centos7-node-221:443/api: dial tcp 192.168.1.221:443: connection refused
Nov 04 01:31:12 centos7-node-221 kube-controller-manager[963]: E1104 01:31:12.759007     963 nodecontroller.go:154] Error monitoring node status: Get https://centos7-node-221:443/api/v1/nodes: dial tcp 192.168.1.221:443: connection refused
```

after minions reboot.


brendandburns:
@eparis because ansible

gouyang:
> Usually we use 2379 as the default port for communicating with other etcd instances, we should use 4001 port to let local etcdctl access etcd contents for preventing port conflicts.

I think your hostname is resolved as 127.0.0.1, so it reports port in use. For most cases, the hostname is resolved as the host's IP, I don't see port conflict in this case. And if you change port to 4001, you should advertise it as well.

@morganwu277 Could you check your host entry and try to remove "127.0.0.1  $your_hostname" in /etc/hosts, and then etcd again?

morganwu277:
@gouyang Actually no, my hostname is resolved as the host's IP. 
But because 
etcd_listen_client_urls: "{{ etcd_url_scheme }}://0.0.0.0:{{ etcd_client_port }} configuration is set in kubernetes/contrib/ansible/roles/etcd/defaults/main.yaml 
so 0.0.0.0:2379 in the default case, this is why I said port conflict. We should use 127.0.0.1:4001 as candidate local communication port.

eparis:
personally I think etcd_listen_*_url should just be 0.0.0.0 and do away with all of this 127 non-sense. Any objections? Also, no to the controller-manager after api-server. These may not even be on the same computer (someday) and even if they start in the 'wrong' order the system must work. If it doesn't eventually work we need to fix kube, not work around it here.

morganwu277:
@eparis YES, reasonable. Different  components should be designed as loose-coupling.

gouyang:
> But because 
> etcd_listen_client_urls: "{{ etcd_url_scheme }}://0.0.0.0:{{ etcd_client_port }} configuration is set in kubernetes/contrib/ansible/roles/etcd/defaults/main.yaml 
> so 0.0.0.0:2379 in the default case, this is why I said port conflict. We should use 127.0.0.1:4001 as candidate local communication port.

it's not 0.0.0.0 so far, if it's 0.0.0.0, then no need `127.0.0.1:4001`.  

@eparis etcd_listen_*_url should just be 0.0.0.0 is acceptable to me.

PR #219, opened 2015-11-08
Increasity verbosity of l7 controller, improve /healthz

bprashanth:
Add a bunch of events, surfaced broken cloud client via /healthz


bprashanth:
Now includes a better /healthz (as advertised in https://github.com/kubernetes/kubernetes/issues/16985)

freehan:
LGTM

PR #221, opened 2015-11-09
Adds a flag to service_loadbalancer which allows to add entries in ha…

ghost:
…proxy config file only for services with a given label.


bprashanth:
Since https://github.com/kubernetes/contrib/pull/222 has this as the first commit, can we just keep that and close this? Please reopen if I missed something

PR #222, opened 2015-11-09
Removing last service from service-loadbalancer

ghost:
Small change in service_loadbalancer. Now after last service is removed from Kubernetes it will be also removed from configuration file.

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/222)

### DIFF
--- a/service-loadbalancer/service_loadbalancer.go
+++ b/service-loadbalancer/service_loadbalancer.go
@@ -462,9 +462,6 @@ func (lbc *loadBalancerController) sync(dryRun bool) error {
 		return errDeferredSync
 	}
 	httpSvc, tcpSvc := lbc.getServices()
-	if len(httpSvc) == 0 && len(tcpSvc) == 0 {
-		return nil
-	}
 	if err := lbc.cfg.write(
 		map[string][]service{
 			"http": httpSvc,
--- a/service-loadbalancer/service_loadbalancer.go
+++ b/service-loadbalancer/service_loadbalancer.go
@@ -79,6 +79,10 @@ var (
 	dry = flags.Bool("dry", false, `if set, a single dry run of configuration
 		parsing is executed. Results written to stdout.`)
 
+	nameLabel = flags.String("httpName", "", `if this flag is set value of the service label indicated
+		by this flag will be used to generate http endpoint name in load balancer configuration. All
+		services which do not have the correct label set will be ignored.`)
+
 	cluster = flags.Bool("use-kubernetes-cluster-service", true, `If true, use the built in kubernetes
 		cluster for creating the client`)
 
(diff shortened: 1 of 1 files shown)


bprashanth:
@aledbf feel free to review if you have time before I do

service-loadbalancer/service_loadbalancer.go
```diff
@@ -352,6 +356,12 @@ func (lbc *loadBalancerController) getEndpoints(
 // - :80 services don't need a :80 postfix
 // - default ns should be accessible without /ns/name (when we have /ns support)
 func getServiceNameForLBRule(s *api.Service, servicePort int) string {
+	if *nameLabel != "" {
```
bprashanth: This is actually one of the reasons we need Ingress, so we can let users choose urls to their Services, but I'm OK just allowing this. However, I think you should put the new string into the service Annotations like we did with a couple of other options (https://github.com/kubernetes/contrib/blob/master/service-loadbalancer/service_loadbalancer.go#L52). then you can also set/get it via the serviceAnnotations type @aledbf put in.

service-loadbalancer/service_loadbalancer.go
```diff
@@ -367,6 +377,14 @@ func (lbc *loadBalancerController) getServices() (httpSvc []service, tcpSvc []se
 			glog.Infof("Ignoring service %v, it already has a loadbalancer", s.Name)
 			continue
 		}
+
+		if *nameLabel != "" {
```
bprashanth: please store nameLabel in loadBalancerController like we do the other cmd line flags instead of accessing it as a global

service-loadbalancer/service_loadbalancer.go
```diff
@@ -367,6 +377,14 @@ func (lbc *loadBalancerController) getServices() (httpSvc []service, tcpSvc []se
...
+			if _, ok := s.Labels[*nameLabel]; !ok {
```
bprashanth: can you avoid doing this in 2 places by having getServiceLB return an error and skipping? or move it out of that method and keep all the lbc.nameLabel logic here.

service-loadbalancer/service_loadbalancer.go
```diff
@@ -444,9 +462,6 @@ func (lbc *loadBalancerController) sync(dryRun bool) error {
 		return errDeferredSync
 	}
 	httpSvc, tcpSvc := lbc.getServices()
-	if len(httpSvc) == 0 && len(tcpSvc) == 0 {
```
bprashanth: aha

test-foxish:
recomputing cla status...

test-foxish:
recomputing cla status...

PR #224, opened 2015-11-10
Create custom haproxy image to publish in gcr.io

aledbf:
This will allow `FROM gcr.io/google_containers/haproxy` in #223

### DIFF
--- a/Ingress/images/haproxy/Dockerfile
+++ b/Ingress/images/haproxy/Dockerfile
@@ -1,0 +1,31 @@
+# Copyright 2015 The Kubernetes Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+FROM alpine:3.2
+MAINTAINER Prashanth B <beeps@google.com>
+
+COPY haproxy.apk /var/cache/apk/haproxy.apk
+
+RUN apk add -U pcre bash curl socat lua5.3 && \
+  apk add --allow-untrusted /var/cache/apk/haproxy.apk && \
+  rm -rf /var/cache/apk/*
+
+RUN mkdir -p /etc/haproxy/errors /var/state/haproxy
+RUN for ERROR_CODE in 400 403 404 408 500 502 503 504;do curl -sSL -o /etc/haproxy/errors/$ERROR_CODE.http \
+	https://raw.githubusercontent.com/haproxy/haproxy-1.5/master/examples/errorfiles/$ERROR_CODE.http;done
+
+ADD haproxy.cfg /etc/haproxy/haproxy.cfg
+
+CMD ["/usr/sbin/haproxy", "-db", "-f", "/etc/haproxy/haproxy.cfg"]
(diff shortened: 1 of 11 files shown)


Ingress/images/haproxy/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+  apk add --allow-untrusted /var/cache/apk/haproxy.apk && \
+  rm -rf /var/cache/apk/*
+
+RUN mkdir -p /etc/haproxy/errors /var/state/haproxy
+RUN for ERROR_CODE in 400 403 404 408 500 502 503 504;do curl -sSL -o /etc/haproxy/errors/$ERROR_CODE.http \
+	https://raw.githubusercontent.com/haproxy/haproxy-1.5/master/examples/errorfiles/$ERROR_CODE.http;done
+
+ADD haproxy.cfg /etc/haproxy/haproxy.cfg
+
+ENTRYPOINT ["/usr/sbin/haproxy", "-f", "/etc/haproxy/haproxy.cfg"]
```
bprashanth: You have 2 options that will make this more widely usable:
1. specify entrypoint "haproxy" and -args as in my example below
2. specify entire cmdline and args in pod

I prefer 2 because I can then start it with something like docker run -it /bin/bash and play around with haproxy inside,  but both allow me to start haproxy with different args which is the real goal.

Ingress/images/haproxy/build/README.md
```diff
@@ -0,0 +1,4 @@
+
+Build haproxy 1.6 package for alpine 3.2
```
bprashanth: please add 1 or 2 basic usage examples here, eg: 
Start via docker with docker run ...
Start as kubernetes pod with:
```
apiVersion: v1
kind: Pod
metadata:
  name: simple
spec:
  containers:
  - image: bprashanth/haproxy:0.5
    imagePullPolicy: Always
    name: haproxy
    command:
    - /usr/sbin/haproxy
    args:
    - -db
    - -f
    - /etc/haproxy/haproxy.cfg
    ports:
    - containerPort: 80
```

Ingress/images/haproxy/haproxy.cfg
```diff
@@ -0,0 +1,36 @@
...
+
+defaults
+	log	global
+	mode	http
+	option	httplog
+	option	dontlognull
+    timeout connect 5000
+    timeout client 50000
+    timeout server 50000
+	errorfile 400 /etc/haproxy/errors/400.http
```
bprashanth: please fix the indent

bprashanth:
LGTM but for nits, will push it out when you've fixed ENTRYPOINT

Ingress/images/haproxy/haproxy.cfg
```diff
@@ -0,0 +1,36 @@
...
+global
+    daemon
+    stats socket /tmp/haproxy
+    server-state-file global
+    server-state-base /var/state/haproxy/
+
+defaults
+    log global
+    mode  http
+    option  httplog
```
bprashanth: can we replace this with dontlog-normal? so I don't get 
```
[WARNING] 313/183747 (1) : config : log format ignored for proxy 'stats' since it has no log address.
[WARNING] 313/183747 (1) : config : log format ignored for frontend 'httpfrontend' since it has no log address.
```

Ingress/images/haproxy/pod.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: simple
+spec:
+  containers:
+  - image: gcr.io/google_containers/haproxy:0.0.1
```
bprashanth: why 0.0.1? can we just make it 0.1 or 0.5 or were you following some other pattern (in which case please add a comment here and in the Makefile). Most of our gcr images are just a simple decimal.

aledbf:
ping @bprashanth

bprashanth:
gcr.io/google_containers/haproxy:0.1 is now live! thanks for your work. Can you please rebase so travis is green and I'll merge.

PR #226, opened 2015-11-10
Initial add of the scale-demo

brendandburns:
@bgrant0607 for api object style
@kelseyhightower for demo guide

PR #234, opened 2015-11-13
fix the syntax error

gouyang:
Signed-off-by: Guohua Ouyang gouyang@redhat.com


bprashanth:
Fyi I think the travis failure is because https://github.com/kubernetes/contrib/pull/226 changed go version, will confirm tomorrow morning

PR #235, opened 2015-11-13
Disable tip on travis

bprashanth:
Was changed in https://github.com/kubernetes/contrib/pull/226 and I suspect it's what's breaking Travis. This pr is a test to see if it fixes things.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,6 @@ matrix:
   include:
     - go: 1.4
     - go: 1.5
-    - go: tip
 
 install:
   - mkdir -p $HOME/gopath/src/k8s.io


eparis:
@bprashanth ? That file doesn't DO anything...  At least I sure don't think so. It's just the version of go the person had when they updated the godeps....

bprashanth:
isn't it the godep.json travis consults for actions like godep restore etc (specifically godep go install ./...)? It's just a shot in the dark about why we're seeing the weird travis failures that I haven't debugged

eparis:
I think it's the golang -tip is just busted somehow....

bprashanth:
@eparis disable tip? hah

eparis:
At least you got a green CI run...

bprashanth:
Alright assigning to you, we don't care about tip right? looks like in main repo we only have 1.4

PR #239, opened 2015-11-14
ansible/make-ca-cert: Clean up verbosity, error on no-public-ip

cgwalters:
In a configuration I was testing in OpenStack, there was no
public floating IP attached to the master.  This caused `make-ca-cert.sh`
to fail in a very obscure way while doing a `cp` of a nonexistent
cert, with `stderr` redirected to `/dev/null`.

Now apparently there's a reason for directing `easyrsa` `stderr`
to `/dev/null` - OpenSSL is just verbose.

But for the other things, e.g. curl has a `-s` flag.  And AFAICS
there was never any reason to hide errors from `cp`.

Explicitly using the `v` flag to tar but also redirecting to
`/dev/null` was clearly a left hand not knowing what the right hand is
doing sort of thing.  Just drop both.

Anyways, we fix the real problem here by cleanly erroring out if we
fail to get the public IP from any of the cluod providers, before we
even try to make a cert.

### DIFF
--- a/ansible/roles/kubernetes/files/make-ca-cert.sh
+++ b/ansible/roles/kubernetes/files/make-ca-cert.sh
@@ -52,14 +52,23 @@ cert_group="${CERT_GROUP:="kube-cert"}"
 # TODO: Add support for discovery on other providers?
 if [ "$cert_ip" == "_use_gce_external_ip_" ]; then
   cert_ip=$(curl -s -H Metadata-Flavor:Google http://metadata.google.internal./computeMetadata/v1/instance/network-interfaces/0/access-configs/0/external-ip)
+  if test -z "${cert_ip}"; then
+      echo "Failed to retrieve external IP" 1>&2; exit 1
+  fi
 fi
 
 if [ "$cert_ip" == "_use_aws_external_ip_" ]; then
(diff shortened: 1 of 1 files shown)


ansible/roles/kubernetes/files/make-ca-cert.sh
```diff
@@ -87,26 +96,25 @@ service_ip=$(echo "${octets[*]}" | sed 's/ /./g')
 # Determine appropriete subject alt names
 sans="IP:${cert_ip},IP:${service_ip},DNS:kubernetes,DNS:kubernetes.default,DNS:kubernetes.default.svc,DNS:kubernetes.default.svc.${dns_domain},DNS:${master_name}"
 
-curl -L -O https://storage.googleapis.com/kubernetes-release/easy-rsa/easy-rsa.tar.gz > /dev/null 2>&1
-tar xzf easy-rsa.tar.gz > /dev/null
```
eparis: that was `xzf` not `xVf`. Does this work?

cgwalters: Right you are, sir.  It did actually still work because modern GNU tar at least will autodetect compression formats.

cgwalters:
Updated patch :arrow_up:

PR #241, opened 2015-11-14
SubmitQueue: Add ability to filter on PRs and Historic decisions in URL

eparis:
http://submit-queue.k8s.io/#?prDisplay=eparis&historyDisplay=17235

will open the web page already filtered with the PR tab showing those
oppened by eparis and the hsitory tab will show only pr 17235 (no matter
who openned it)


brendandburns:
LGTM

PR #242, opened 2015-11-15
ansible: make apiserver port easy to configure

gouyang:
Atomic host cannot listen to port 443 currently, so it should be able to configure the api port.

Signed-off-by: Guohua Ouyang gouyang@redhat.com

### DIFF
--- a/ansible/group_vars/all.yml
+++ b/ansible/group_vars/all.yml
@@ -31,6 +31,9 @@ insecure_registrys:
 #http_proxy: "http://proxy.example.com:3128"
 #https_proxy: "http://proxy.example.com:3128"
 
+# The port that the Kubernetes apiserver component listens on.
+kube_master_api_port: 443
+
 # Kubernetes internal network for services.
 # Kubernetes services will get fake IP addresses from this range.
 # This range must not conflict with anything in your infrastructure. These


eparis:
you should just be able to set it to 6443 in all.yaml without any changes to default.yaml in your local configuration...

gouyang:
@eparis updated, leave default.yaml untouched.

cgwalters:
See also https://bugzilla.redhat.com/show_bug.cgi?id=1275379#c8

eparis:
@gouyang if that works, I don't think we need a PR at all, just set it in your all.yaml...

gouyang:
@eparis I think the point of this PR is making kube_master_api_port visible in all.yaml. Once a user met this error then he could know how to make changes.

cgwalters:
The real fix is making low ports Just Work, but I'm (somewhat weakly) :+1: on this patch, just to make the variable more visible.

jasonbrooks:
In the [bug](https://bugzilla.redhat.com/show_bug.cgi?id=1275379#c8) that @cgwalters mentioned above, there's discussion of handling this with systemd socket activation. I spent some time trying to make that work ([my notes here](https://gist.github.com/jasonbrooks/cbffdb0b00fe4cdc491c#gistcomment-1589956)), but haven't gotten it working. The systemd docs note that a daemon must know how to accept a socket from systemd, or know how to work w/ inetd-style socket passing -- that may be the issue?

We could do a systemd drop-in to make kube-apiserver run as root just on non-fedora atomics (fedora's kube sets this capability during the RPM build, so no problem there), but if that's unpalatable, making the value easier to change is at least somewhat better.

eparis:
The value is already easy to change, just set it in all.yaml....

jasonbrooks:
Ah, I understand now.

PR #245, opened 2015-11-15
kube-ui.yml references a hard coded proxy

jSherz:
The other tasks in this role have proxy environment variables passed through from ansible. This file has hard coded values for a proxy (Squid?) at RedHat. Should this be the case?


gouyang:
It's introduced in #233, sorry for that.  @eparis

PR #246, opened 2015-11-16
Increase poll interval for getting Node Port of default backend.

bprashanth:
No point burning through a core if the default backend isn't up yet


bprashanth:
Really https://github.com/kubernetes/kubernetes/pull/17302, but Really 100 micro is too short.

brendandburns:
LGTM.

PR #248, opened 2015-11-16
Add kubernetes-kubemark-gce as a blocker to the submit-queue

gmarek:
cc @quinton-hoole @fgrzadkowski @davidopp 

It looks completely stable now, so I believe we can add a 100-node kubemark test as a blocker to the submit queue.


wojtek-t:
LGTM (from my perspective)

eparis:
I can probably get this in production later today.

brendandburns:
Mergez!

Issue #249, opened 2015-11-16
Ingress: resolver option in nginx-alpha controller causes bad gateway


maklemenz:
[controller.go](https://github.com/kubernetes/contrib/blob/60008fc849d04582e35d12edcecf025fa6ac07e7/Ingress/controllers/nginx-alpha/controller.go#L45) is setting the dns resolver to localhost. I don't know how this is handled in GCE, but on my kubernetes-cluster installed on openstack I don't have a dns server inside each pod. The names don't get resolved and the result is a _bad gateway_ after a long time.

Removing [this line](https://github.com/kubernetes/contrib/blob/60008fc849d04582e35d12edcecf025fa6ac07e7/Ingress/controllers/nginx-alpha/controller.go#L45) would cause nginx to use the OS to resolve the names, hence would use _/etc/resolv.conf_ which is set by kubernetes.

maklemenz:
Pull request is merged, so I close this one.

PR #250, opened 2015-11-16
refs #249: Ingress: resolver option in nginx-alpha controller causes bad gateway

ghost:
#249

Resolver-option removed from Ingress nginx-alpha controller because it causes dns problems


maklemenz:
edited cla to contain both emails.

bprashanth:
LGTM, thanks.

PR #252, opened 2015-11-16
SubmitQueue: Set as 'Status' in github for every PR

eparis:
As written this will change almost every PR on
https://github.com/kubernetes/kubernetes/pulls to show a yellow dot
instead of a green check (red X will stay red X). That may be a major
issue for some viewers of that page.

At the bottom of github we have statuses for CI, cla, etc. This adds a
status, to all PRs for the submit queue. It will add the same message as
seen below a PR at http://submit-queue.k8s.io in the github UI. These
messages are (hopefully) the correct next step for someone trying to get
their PR merged.

The 'details' link will link back to the submit-queue web page for the
given PR. The PR tab of the web page will not hae any additional useful
information, although the history tab might, if the PR every got greens
across the board.

### DIFF
--- a/mungegithub/github/github.go
+++ b/mungegithub/github/github.go
@@ -668,7 +668,7 @@ func (obj *MungeObject) SetStatus(state, url, description, context string) error
 	config := obj.config
 	status := &github.RepoStatus{
 		State:       &state,
-		URL:         &url,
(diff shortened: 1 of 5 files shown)


brendandburns:
LGTM.  Travis is grumpy.  kicked it.

Issue #198, opened 2015-10-22
Update ansible directions (1) for mac people (2) for openstack urls


jayunit100:
I had to make some changes to run this from my mac with a local repo build.

**first, we may want to show how to push local binaries - its tricky to get the right path, b/c you need to use the dockerized build and specify amd64 iirc**

You need to use `run.sh` to build dockerized linux binaries.  Then, you have to do something like this...

```
  3 --- a/ansible/roles/master/defaults/main.yml
  4 +++ b/ansible/roles/master/defaults/main.yml
  5 @@ -1,3 +1,3 @@
  6  kube_master_insecure_port: 8080
  7 
  8 -localBuildOutput: ../../_output/local
  9 \ No newline at end of file
 10 +localBuildOutput: ~/Development/kubernetes/_output/dockerized/bin/linux
```

and

```
 13 --- a/ansible/roles/master/tasks/localBuildInstall.yml
 14 +++ b/ansible/roles/master/tasks/localBuildInstall.yml
 16  ---
 17  - name: Copy master binaries
 18    copy:
 19 -    src: "{{ localBuildOutput }}/go/bin/{{ item }}"
 20 +    src: "{{ localBuildOutput }}/amd64/{{ item }}"
 21      dest: /usr/bin/
 22      mode: 0755
 23    with_items:
```

**also, i think we might want to update vagrant example file for openstack**

and finally in my main.yml, this is what it looks like.  Mainly, the os_auth_url was different... So we should update the example vagrant file 

```
 24 diff --git a/ansible/roles/node/defaults/main.yml b/ansible/roles/node/defaults/main.yml
 29 -localBuildOutput: ../../_output/local
 30 +localBuildOutput: ~/Development/kubernetes/_output/dockerized/bin/linux
 31 os_username: jvyas
 32 os_password: 123123123123
 33 os_tenant: "ENG CTO Office"
 34 os_auth_url: "http://os1-public.osop.rhcloud.com:5000/v2.0/tokens"
 35 os_region_name: "OS1Public"
 36 os_ssh_key_name: "JPeerindex"
 37 os_flavor: "m1.small"
 38 os_image: "_OS1_Fedora-Cloud-Base-22-20150521.x86_64.qcow2"
 39 os_security_groups:
 40   - "default"
 41 os_floating_ip_pool: "os1_public"
```

jayunit100:
fyi @eparis , mostly just some notes, will refine later, might help someone - will try to pr at some point.

[...]

PR #255, opened 2015-11-16
Allow the user to specify the full path for local build install.

pedro-r-marques:
Local build path will be different when using docker build; or when
downloading pre-built binaries from github. Let the user specify the
full path.


pedro-r-marques:
@eparis do you mind reviewing this PR ?

eparis:
@rrati @jayunit100

jayunit100:
ah yes, fixes  my earlier issue , maybe guess we can close  #198 now

PR #256, opened 2015-11-16
Make http_proxy configuration work across both RedHat and Debian

pedro-r-marques:
The format for proxy variables is different across RedHat (existing) and Debian based systems (as tested with the apt repository provided by docker).
Make the docker install a separate play, rather than being a dependency so that it is easier to verify.

### DIFF
--- a/ansible/cluster.yml
+++ b/ansible/cluster.yml
@@ -18,6 +18,14 @@
   tags:
     - etcd
 
+- hosts: all
+  sudo: yes
+  roles:
+    - common
+    - docker
+  tags:
+    - docker
+
 # install flannel
 - hosts:
     - etcd
--- a/ansible/roles/docker/tasks/apt-docker-install.yml
+++ b/ansible/roles/docker/tasks/apt-docker-install.yml
@@ -1,12 +1,14 @@
 ---
 - name: Install GPG key
-  command: apt-key adv --keyserver hkp://p80.pool.sks-keyservers.net:80 --recv-keys 58118E89F3A912897C070ADBF76221572C52609D
+  command: apt-key adv --keyserver hkp://p80.pool.sks-keyservers.net:80 {% if http_proxy is defined -%}--keyserver-options http-proxy={{ http_proxy }} {% endif -%} --recv-keys 58118E89F3A912897C070ADBF76221572C52609D
 
 - name: Configure apt repository
   template: src="docker.list.j2" dest="/etc/apt/sources.list.d/docker.list"
+  register: add_apt_source_docker
 
-- name: Update respository information
+- name: Update repository information
   command: apt-get update
+  when: add_apt_source_docker.changed
 
 - name: Install docker engine
   action: "{{ ansible_pkg_mgr }}"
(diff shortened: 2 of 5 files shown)


eparis:
On debian, how does the $docker_config_dir/docker file get used? Sourced by some startup script? How does it get used if docker is launched with systemd? I'm admittedly a bit baffled that the docker daemon itself would be different and that I would want lower case env vars ever, but I didn't really look...

pedro-r-marques:
@eparis 
snippet of the upstart script used by the docker package:
```
script
        # modify these in /etc/default/$UPSTART_JOB (/etc/default/docker)
        DOCKER=/usr/bin/$UPSTART_JOB
        DOCKER_OPTS=
        if [ -f /etc/default/$UPSTART_JOB ]; then
                . /etc/default/$UPSTART_JOB
        fi
        exec "$DOCKER" daemon $DOCKER_OPTS
end script
```
It sources the config file; while systemd uses EnvironmentFile.

eparis:
@pedro-r-marques thank you so much!  Sadly, I'm still a little dense.

Sourcing this file means that docker is going to launch with an env var `http_proxy=1.2.3.4`. On the non-debian side it will get launched with an env var `HTTP_PROXY=1.2.3.4`.

Does docker pay attention to `http_proxy` instead of `HTTP_PROXY`?

pedro-r-marques:
@eparis lowercase "http_proxy" is the canonical variable that all unix utils tend to use (e.g. curl); the golang  HTTP code for instance respects that. I typically config this var on /etc/environment.
I'm not sure how docker handles capital HTTP_PROXY on systemd systems... i wouldn't be surprised if this ends up being treated the same way as lowercase http_proxy by docker/golang http code.

cgwalters:
```
$ git describe --tags --always
go1.5beta3-1054-g6bf794a
golang/go [git master]
$ git grep HTTP_PROXY
src/net/http/proxy_test.go:     for _, v := range []string{"HTTP_PROXY", "http_proxy", "NO_PROXY", "no_proxy"} {
src/net/http/transport.go:// as directed by the $HTTP_PROXY and $NO_PROXY (or $http_proxy and
src/net/http/transport.go:// HTTP_PROXY, HTTPS_PROXY and NO_PROXY (or the lowercase versions
src/net/http/transport.go:// thereof). HTTPS_PROXY takes precedence over HTTP_PROXY for https
src/net/http/transport.go:              names: []string{"HTTP_PROXY", "http_proxy"},
src/net/http/transport_test.go: env      string // HTTP_PROXY
src/net/http/transport_test.go:         os.Setenv("HTTP_PROXY", tt.env)
```

In other words, yes, golang special cases this.  Agreed that `http_proxy` lowercase is canonical, and we could switch the other parts of the playbook to it.

eparis:
I have seen the light! I guess I'd like to see a switch from HTTP_PROXY to http_proxy across the code base if the old way was the wrong way.

Now enlighten me about the docker as it's own play. You didn't remove the dependency that I see, so if it getting run twice?

pedro-r-marques:
@eparis Having something as an independent play make it easier to test and verify. And while dependencies where supposed to run only once per playbook (according to docs); they run multiple times on my setup. I think the right way to do this is to remove the docker dependencies from the existing plays... but my grasp of ansible is very limited.

eparis:
I'm good/fine making it explicit, but lets pull out the dependencies. So   change the HTTP_PROXY to http_proxy and remove the dependancies from roles/node/meta/main.yml and roles/opencontrail/meta/main.yml and I'll commit.

pedro-r-marques:
@eparis done.

PR #264, opened 2015-11-18
Ansible: etcd: listen on 0.0.0.0

eparis:
The etcd code listens on ansible_fqdn and then hard codes in 127.0.0.1.
We've had trouble where ansible_fqdn maps to localhost.localdomain and
now we are trying to bind on the same port twice.

Stop the madness, just listen on 0.0.0.0


cgwalters:
Worth comparing to what openshift-ansible does in https://github.com/openshift/openshift-ansible/blob/master/roles/etcd/defaults/main.yaml#L2

Seems like they bind to the interface, not ansible_fqdn.  Which should work fine with also binding to `127.0.0.1` as no one is likely to be trying to run this playbook against `127.0.0.1`.

cgwalters:
I'm :+1:  on this patch as is too.

PR #266, opened 2015-11-19
Enable cross-namespace ingress in nginx-alpha ingress controller.

krancour:
This allows you to run one ingress controller rc and fan out to services in multiple namespaces.


bprashanth:
Thanks, LGTM

PR #272, opened 2015-11-20
Cluster tagging

bprashanth:
Tag all resources with user specified cluster name. Useful to match resources with the controller that created them in e2e tests.

### DIFF
--- a/Ingress/controllers/gce/backends.go
+++ b/Ingress/controllers/gce/backends.go
@@ -40,7 +40,13 @@ func portKey(port int64) string {
 }
 
 func beName(port int64) string {
-	return fmt.Sprintf("%v-%d", backendPrefix, port)
+	// TODO: Pipe the clusterName through, for now it saves code churn to just
+	// grab it globally, especially since we haven't decided how to handle
+	// namespace conflicts in the Ubernetes context.
+	if *clusterName == "" {
+		return fmt.Sprintf("%v-%d", backendPrefix, port)
+	}
+	return truncate(fmt.Sprintf("%v-%d%v%v", backendPrefix, port, clusterNameDelimiter, *clusterName))
 }
 
 // NewBackendPool returns a new backend pool.
--- a/Ingress/controllers/gce/main.go
+++ b/Ingress/controllers/gce/main.go
@@ -26,6 +26,7 @@ import (
 	"time"
 
 	flag "github.com/spf13/pflag"
+	"k8s.io/kubernetes/pkg/api"
 	client "k8s.io/kubernetes/pkg/client/unversioned"
 	kubectl_util "k8s.io/kubernetes/pkg/kubectl/cmd/util"
 
@@ -43,6 +44,13 @@ const (
 	// lbApiPort is the port on which the loadbalancer controller serves a
 	// minimal api (/healthz, /delete-all-and-quit etc).
 	lbApiPort = 8081
+
+	// A delimiter used for clarity in naming GCE resources.
+	clusterNameDelimiter = "--"
+
+	// Arbitrarily chosen alphanumeric character to use in constructing resource
+	// names, eg: to avoid cases where we end up with a name ending in '-'.
+	alphaNumericChar = "0"
 )
 
 var (
@@ -56,7 +64,7 @@ var (
 		printed to stdout and no changes are made to your cluster. This flag is for
 		testing.`)
 
-	clusterName = flags.String("gce-cluster-name", "default-cluster-name",
+	clusterName = flags.String("cluster-uid", "",
 		`Optional, used to tag cluster wide, shared loadbalancer resources such
 		 as instance groups. Use this flag if you'd like to continue using the
 		 same resources across a pod restart. Note that this does not need to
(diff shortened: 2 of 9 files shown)


bprashanth:
Needs to get pushed out as gcr.io/google_containers/glbc:0.6

bprashanth:
Let me know if you're not comfortable doing the review, I figured it's just a golang review so why not

Ingress/controllers/gce/backends.go
```diff
@@ -40,7 +40,13 @@ func portKey(port int64) string {
...
 
 func beName(port int64) string {
-	return fmt.Sprintf("%v-%d", backendPrefix, port)
+	// TODO: Pipe the clusterName through, for now it saves code churn to just
+	// grab it globally, especially since we haven't decided how to handle
+	// namespace conflicts in the Ubernetes context.
+	if *clusterName == "" {
+		return fmt.Sprintf("%v-%d", backendPrefix, port)
+	}
+	return fmt.Sprintf("%v-%d%v%v", backendPrefix, port, clusterNameDelimiter, *clusterName)
```
jlowdermilk: Be careful about character limits. I suspect backends have a much shorter name length cap then cluster names. We probably need to either truncate or use the cluster hash.

bprashanth: good catch, it's 64 charcters. I'm just going to log a warning and truncate for now. Default e2e namespaces are well within that though.

bprashanth:
PTAL

Ingress/controllers/gce/main.go
```diff
@@ -56,7 +60,7 @@ var (
 		printed to stdout and no changes are made to your cluster. This flag is for
 		testing.`)
 
-	clusterName = flags.String("gce-cluster-name", "default-cluster-name",
+	clusterName = flags.String("gce-cluster-name", "",
```
jlowdermilk: Just curious, where is this passed through?

Ingress/controllers/gce/backends.go
```diff
@@ -40,7 +40,17 @@ func portKey(port int64) string {
...
+	// grab it globally, especially since we haven't decided how to handle
+	// namespace conflicts in the Ubernetes context.
+	if *clusterName == "" {
+		return fmt.Sprintf("%v-%d", backendPrefix, port)
+	}
+	taggedName := fmt.Sprintf("%v-%d%v%v", backendPrefix, port, clusterNameDelimiter, *clusterName)
+	if len(taggedName) > nameLenLimit {
+		taggedName = taggedName[:nameLenLimit]
+	}
+	return taggedName
```
jlowdermilk: This is definitely better; at least we won't break on long clusters. My concern is that it won't be uniquely identifying for cluster names that share prefixes, given how short the limit is. We use shared name prefixes on a lot of e2e cluster names, and I imagine that's not uncommon practice. What do you think of using truncated clusterName + hash of clusterName? that would make it semi-human readable for short cluster names while still providing uniqueness for test/debugging purposes.

Ingress/controllers/gce/main.go
bprashanth: do you mean what specifies the --gce-cluster-name argument? the pod in the test: 
https://github.com/kubernetes/kubernetes/pull/17614/files#diff-ec4b96325ef379121837e76bb579c833R257

It's actually just a uid used to tag cluster resoureces and I'm passing the test namespace through, so I suppose I should rename it to cluster-uid or something. I had initially name it gce-cluster-name because I though that's how people would use it.

Ingress/controllers/gce/backends.go
bprashanth: I actually just pass the test namespace so it should be unique eitherway, and contains a similar hash. I'd assumed if users specify colliding cluster names that's their problem? Not against your hash idea though, can I do it in a follow-up?

Ingress/controllers/gce/main.go
jlowdermilk: Yeah, you should probably rename it. I thought it was the actual cluster name passed through via the apiserver.

Ingress/controllers/gce/backends.go
jlowdermilk: See comment below: I thought this was the actual cluster name retrieved via the apiserver/some other authoritative channel. If it's an arbitrary cluster-uid, this seems reasonable, at least for testing purposes.

That being said, if the cluster name is being passed through on creation to the kube-system pod in GKE/GCE, we probably still want something better than truncation. If we only truncate there is no easy way to determine whether a particular backend was created by `my-test-kubernetes-cluster1` or `my-test-kubernetes-cluster2`.

bprashanth: I'm currently taking the simpler approach of asking users to modify the rc so it has a unique name if they want to avoid collisions. I'm probably going to acquire a uid from the master before it leaves beta.

Changed gce-cluster-name to cluster-uid

jlowdermilk:
Two remaining nits: tag should be `v0.5.1`, and you may want to change the `clusterName` global to `clusterUID. Otherwise looks good.

jlowdermilk:
lgtm

PR #275, opened 2015-11-21
Add ansible network-service-install tag instead of flannel

galthaus:
As additional network layers are added to the ansible playbook, it would be good to have a general tag to install and eventually configure the network layer so that a group variable can drive the networking choice and configuration, but the tag will ensure that one of the items is installed.


galthaus:
This is incomplete.  I didn't realize that the opencontrail patch had been added.  I'll resubmit.

Issue #276, opened 2015-11-21
Ansible: feature of network-service tags instead of flannel/opencontrail


galthaus:
It would be nice to have a couple tags for the ansible playbook that represented installing and configuring the network service layer.  Instead of having opencontrail and flannel, we should have network-service-install and network-service-config and let the group_var networking control which service is installed.

This allows for more generalized execution of the playbook from other yml files or orchestration engines.

PR #277, opened 2015-11-21
Ansible: Add network service tags

galthaus:
Addresses - #276

### DIFF
--- a/ansible/cluster.yml
+++ b/ansible/cluster.yml
@@ -35,7 +35,7 @@
   roles:
     - { role: flannel, when: networking == 'flannel' }
   tags:
-    - flannel
+    - network-service-install
 
 # install opencontrail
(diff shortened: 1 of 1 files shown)


eparis:
@pedro-r-marques your thoughts?

pedro-r-marques:
@eparis If i understand correctly, Greg is trying to have the same install workflow independently of the network provider. Thus his workflow engine [https://github.com/digitalrebar] can invoke the same sequence of steps (in kubernetes ansible) regardless of provider. The workflow tool is most likely invoking plays as independent steps (which i believe makes sense); rather than the full playbook.

eparis:
@pedro-r-marques and thus you are fine with this change?

pedro-r-marques:
@eparis yes.

Issue #278, opened 2015-11-21
Ansible: master-only with opencontrail fails missing kubelet


galthaus:
When running the play book with opencontrail as networking and setting up only a master node, the playbook will fail configuring the master because the opencontrail components need kubelet to manage local opencontrail containers.

This is not a problem is the master is a node.

PR #279, opened 2015-11-21
Make sure to install kubelet on the master (if node not installed).

galthaus:
This applies to systems that have opencontrail networking enabled.

This is a fix for #278


eparis:
@pedro-r-marques

pedro-r-marques:
lgtm

PR #282, opened 2015-11-23
[WIP][mungegithub] OWNERS implementation

eparis:
Not enforced, just adds/removes a label. No way to know who needs to respond at any given moment. I'm sure 1,000 other problems....

### DIFF
--- a/mungegithub/mungers/owner-approval.go
+++ b/mungegithub/mungers/owner-approval.go
@@ -1,0 +1,225 @@
+/*
+Copyright 2015 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package mungers
+
+import (
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+
+	"k8s.io/contrib/mungegithub/github"
+	"k8s.io/kubernetes/pkg/util/sets"
+	"k8s.io/kubernetes/pkg/util/yaml"
+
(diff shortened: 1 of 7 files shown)


mungegithub/mungers/owner-approval.go
```diff
@@ -0,0 +1,226 @@
...
+	o.owners[path] = sets.NewString(c.Owners...)
+	return nil
+}
+
+func (o *OwnerApproval) updateOwnerMap() error {
+	cmd := exec.Command("git", "pull")
+	cmd.Dir = o.kubernetesDir
+	err := cmd.Run()
+	if err != nil {
+		glog.Fatalf("%v", err)
```
eparis: This probably shouldn't be fatal.

davidopp:
@brendandburns

eparis:
So this requires at least:

- More testings
- Enforcement in the merge bot
- Some display of who needs to approve a given PR
- Some display of what users need to approve

Those last two are going to take some thought. I'd expect them to be extensions of the submit queue web site, maybe with a new github status? but maybe not...  I haven't thought out the usability UI stuff...

thockin:
maybe related, maybe not.

We need a way to denote maintainers - specifically who do you contact when
something like a volume driver needs to be touched.  Do we want to overload
OWNERS to denote this or should we create a different idea here?

davidopp:
I think it would be better to put that kind of information in comments in the relevant source file(s). People might not be aware of the OWNERS mechanism.

thockin:
I don't think it belongs I'm the code.  It needs to be more obvious, I
think.  I am OK with a top-level maintainers file or a hierarchy of them,
but that smacks of OWNERS to me.

eparis:
I think the 'OWNERS' file is a good place for this. In this WIP code the file is yaml/json with an 'owners' list and I don't see a reason there shouldn't be a 'maintainers' list in there as well. In fact I think the blunderbuss should get info from the in tree file instead of from the blunderbuss.yaml in contrib. This data should live somewhere inside the kubernetes tree, and this WIP doesn't prevent that...

davidopp:
ping?

davidopp:
@eparis I noticed you closed this PR -- are you planning to replace it with another one?

eparis:
WTH?

eparis:
I guess I accidentally deleted the branch and that closed the PR, WHOOPS!  See
https://github.com/kubernetes/contrib/pull/474

PR #284, opened 2015-11-24
Update Readme to reflect new network-service tags

galthaus:
Change references #276


brendandburns:
LGTM.

@eparis any thoughts?

PR #285, opened 2015-11-24
Submit Queue: Update static committers list

eparis:
Couple new people with pull access. Couple people moved from pull to
push. One person moved from push to pull. First time for that...


eparis:
self merge since this is automatic and is just writing down what the bot already does dynamically today.

Issue #197, opened 2015-10-21
submit queue should ensure CI results are fresh


ixdy:
The submit queue automatically re-runs Jenkins before merging, but it's possible for the Shippable and Travis results to be very stale. Because Jenkins doesn't run all of the same verifications and checks currently, it's possible for things to break. 

Motivating example: https://github.com/kubernetes/kubernetes/pull/15608 was auto-merged, but it immediately broke the build, since some new docs had been added since its last CI run.

The GitHub API indicates when commit statuses were updated (see https://github.com/google/go-github/blob/master/github/repos_statuses.go#L34); it should hopefully be fairly straightforward to require these to be no more than N days old or retrigger CI if they are stale.

@brendandburns @eparis

ixdy:
Another breaking merge due to stale test results: kubernetes/kubernetes/pull/15674

eparis:
Right now the bot is only 're-testing' e2e. Even if the unit test is being re-run, it is being ignored. It (obviously) does not care at all how old shippable/travis/jenkin-unit-test are.

We can force a re-test of the jenkins-unit test (if it is fast enough) like we do the e2e test (going to make the bot more complex but I'm sure I can figure it out).

Or I can just make another munger which re-tests jenkins-unit-test every however many days and leaves it 'independent' of the submit queue. That seems a lot easier to me from a code PoV. Every PR that is 'mergeable' and has test more than 2 days old (1 day old?) gets a re-test triggered....

That I can code in an hour or so....

ixdy:
Does the bot only look at the e2e result after triggering? I assumed it looked at all commit statuses, though I haven't looked at the code.

Right now saying "@k8s-bot test this" actually triggers both unit + e2e tests, so if it's looking at all commit statuses it's already re-checking verification/unit tests in the common case.

eparis:
e2e only. (That can be changed quite easily actually)

But it looks like unit test also passed, so while I can fix the bot to check for both e2e and unit to go pending and come back green, didn't that happen in this case?

eparis:
ok, now I get it. That was an e2e-not-required.

eparis:
This is fixed by https://github.com/kubernetes/contrib/pull/287

PR #287, opened 2015-11-25
Submit Queue: Retest PRs every 48 hours

eparis:
We don't actually test every PR every 48 hours. Only those which seem close to merge. They have LGTM, their tests look good, etc.  If this were to run right this instant, it would hit against 4 PRs.


eparis:
fixes https://github.com/kubernetes/contrib/issues/197

PR #288, opened 2015-11-25
Add nginx 1.9.x image

aledbf:
```
gcr.io/google_containers/nginx-slim 0.0          038a045ad0bc   15 seconds ago  22.66 MB
gcr.io/google_containers/nginx      latest       9f4cd86c1aec   8 months ago    425.7 MB
```

PR #290, opened 2015-11-26
Reduce echoheaders image size using nginx-slim image and lua

aledbf:
requires #288


Ingress/echoheaders/nginx.conf
```diff
@@ -0,0 +1,45 @@
+events {
+    worker_connections 1024;
+}
+
+http {
+    default_type 'text/plain';
+    client_max_body_size 10m;
+
+    server {
+        listen 8080 default_server reuseport;
```
bprashanth: can you add oneline comments above nginx config terms (like reuseport, client_max_body_size etc)? Mostly just describe why you're setting it/what happens when we exceed it or it isn't set.

bprashanth:
LGTM but for the comment nit

aledbf:
@bprashanth done

bprashanth:
LGTM

PR #291, opened 2015-11-26
Ansible tasks to install latest version of Docker

furikake:
I ran the Ansible playbook on CentOS 7 and it installed Docker 1.8 rather than 1.9.  I'm assuming that someone that wants a `localBuild` rather than `packageManager` for Kubernetes probably wants the latest for Docker too.  I certainly did.

I'm downloading the script locally first so that I don't have to use the `validate_certs=False` to work around SNI and Python.

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/291)

### DIFF
--- a/ansible/roles/docker/tasks/getdocker-install.yml
+++ b/ansible/roles/docker/tasks/getdocker-install.yml
@@ -1,0 +1,26 @@
+---
+- name: get.docker.com | Create temp dir
+  local_action: command mktemp -d -t docker.XXXXXXXXXX
+  register: docker_temp_dir
+  sudo: no
+
+- name: get.docker.com | Create remote temp dir
+  command: mktemp -d -t docker.XXXXXXXXXX
+  register: remote_docker_temp_dir
+
+- name: get.docker.com | Download Docker shell script locally to workaround SNI 
+  local_action: get_url url=https://get.docker.com dest="{{ docker_temp_dir.stdout }}/docker-install.sh"
+  sudo: no
+
+- name: get.docker.com | Copy the Docker installer to remote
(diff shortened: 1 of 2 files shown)


cgwalters:
I don't use `localBuild` myself, but I'm not sure that follows...not everyone hacking on kubernetes wants to install whatever is from `docker.com`.

Maybe as a separate playbook?

furikake:
Good point and thanks for the feedback.  

Would something similar to the etcd playbook suit?  `etcd/defaults/main.yml` defines:

```
# etcd_source_type: package | github-release
etcd_source_type: package
```

There could be a similar variable in  `docker/defaults/main.yml` to define:

```
# docker_source_type: package | get-docker-com
docker_source_type: package
```

This would be defaulted to the existing behavior of getting it from the distro's package manager but allow you to explicitly change it to install the latest via get.docker.com shell script.

eparis:
sounds good to me!

eparis:
Sorry to ignore you for so long. First, lets create a `docker_source_type` or something like that. Given the kubernetes doesn't support docker 1.10 right now, if I were to install from local build I wouldn't want it to pull 1.10...

Second, if I'm doing a large deployment that upload to all my nodes could be quite painful. So I don't like the 'pull local first'. What was the problem there? Would it be better to make the install use the shell module with: `curl -fsSL https://get.docker.com/ | sh`

eparis:
#456 is also allowing for more recent versions. They defined a 'source type'. Who-ever gets merged first will have to do it the same way as the other...

ingvagabund:
Implemented in https://github.com/kubernetes/contrib/pull/1610

PR #292, opened 2015-11-28
Changes for Spinning up VMs on OS X, Updating Centos boxes, flannel

jayunit100:
Here are some thanskgiving updates for 
- spinning up the VMs in vagrant on OSX. 
- IP address detect for vbox
- adding hints: docker example paths.
- adding hints: on how to get rid of the IP address option in easyrsa.
- flannel eth options in jinja w eth1 for vbox 
- Python selinux install directives
- Fixing the Centos base image

cc @eparis @justinsb

### DIFF
--- a/ansible/roles/common/tasks/fedora-install.yml
+++ b/ansible/roles/common/tasks/fedora-install.yml
@@ -5,3 +5,15 @@
     name: python-firewall
     state: latest
   when: ansible_distribution_major_version|int >= 22
+
+- name: Generic | Install Python
+  action: "{{ ansible_pkg_mgr }}"
+  args: 
+    name: python
+    state: latest
+
+- name: Generic | Install SELinux Python Bindings
+  action: "{{ ansible_pkg_mgr }}"
+  args:
+    name: libselinux-python
+    state: latest
--- a/ansible/roles/master/defaults/main.yml
+++ b/ansible/roles/master/defaults/main.yml
@@ -1,3 +1,3 @@
 kube_master_insecure_port: 8080
 
-localBuildOutput: ../../_output/local/go/bin
+localBuildOutput: ../../_output/local/go/bin # localBuildOutput: ../../kubernetes/_output/dockerized/bin/linux/amd64/ (docker builds)
--- a/ansible/roles/node/tasks/localBuildInstall.yml
+++ b/ansible/roles/node/tasks/localBuildInstall.yml
@@ -1,7 +1,7 @@
 ---
 - name: Copy node binaries
   copy:
-    src: "{{ localBuildOutput }}/{{ item }}"
+    src: "{{ localBuildOutput }}/{{ item }}" #docker builds, use: kubernetes/_output/dockerized/bin/linux/amd64/{{ item }}
     dest: /usr/bin/
     mode: 0755
   with_items:
--- a/ansible/vagrant/Vagrantfile
+++ b/ansible/vagrant/Vagrantfile
@@ -6,6 +6,7 @@ require "yaml"
 ### RUN: vagrant plugin uninstall vagrant-openstack-plugin"
 ### Then RUN: "vagrant plugin install vagrant-openstack-provider"
 require 'vagrant-openstack-provider'
+require 'vagrant-hostmanager'
 
 $num_nodes = (ENV['NUM_NODES'] || 2).to_i
 ansible_tags = ENV['ANSIBLE_TAGS']
@@ -14,6 +15,20 @@ VAGRANTFILE_API_VERSION = "2"
 
 # Openstack providers are best used with latest versions.
 Vagrant.require_version ">= 1.7"
+
+# for vbox, we avoid private networks: Let the vbox VM set the IP and just write it out at runtime.
+def getip(vm, cached_addresses, name)
+	
+	if cached_addresses[vm.name].nil?
+		if hostname = (vm.ssh_info && vm.ssh_info[:host])
+		        vm.communicate.execute("/usr/sbin/ip addr show eth1 | grep 'inet ' | xargs | cut -f 2 -d ' '| cut -f 1 -d '/' 2>&1") do |type, contents|
+		        cached_addresses[name] = contents.split("\n").first[/(\d+\.\d+\.\d+\.\d+)/, 1]
+	           end
+	        end
+         end
+### Since we aren't gauranteed to have calculated the result, just return whatevers in the cache.
+cached_addresses[name]
+end
 
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   # By default, Vagrant 1.7+ automatically inserts a different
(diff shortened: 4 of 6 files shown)


ansible/roles/common/tasks/fedora-install.yml
```diff
@@ -5,3 +5,15 @@
     name: python-firewall
     state: latest
   when: ansible_distribution_major_version|int >= 22
+
+- name: Generic | Install Python
```
jayunit100: iirc needed for f23 and forward.  Should be no-ops on centos.

ansible/vagrant/Vagrantfile
```diff
@@ -59,13 +74,17 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     os.server_name = n.vm.hostname
   end
 
-  def set_vbox(vb, config)
-    config.vm.box = "chef/centos-7.0"
+  def set_vbox(vb, config, cached_addresses_in)
+    config.vm.box = "hfm4/centos7"
```
jayunit100: looks like centos-7.0 has dissapeared, so updated the box.

ansible/vagrant/Vagrantfile
```diff
@@ -6,6 +6,7 @@ require "yaml"
 ### RUN: vagrant plugin uninstall vagrant-openstack-plugin"
 ### Then RUN: "vagrant plugin install vagrant-openstack-provider"
 require 'vagrant-openstack-provider'
+require 'vagrant-hostmanager'
```
jayunit100: I was using this for vbox, updates etc/hosts dynamically.    not sure if it needs to be in req unless it works for other VMs to ( i think it could be useful for os-provisioner also?)

ansible/roles/common/tasks/fedora-install.yml
```diff
@@ -5,3 +5,15 @@
...
+  action: "{{ ansible_pkg_mgr }}"
+  args: 
+    name: python
+    state: latest
+
+- name: Generic | Install SELinux Python Bindings
```
jayunit100: This also was needed on f23 base images.  Centos seems to not-need it

ansible/roles/common/tasks/fedora-install.yml
eparis: This already lives in the 'pre-ansible' role and doesn't belong here.

ansible/roles/common/tasks/fedora-install.yml
eparis: Same with this.

ansible/roles/node/tasks/localBuildInstall.yml
```diff
@@ -1,7 +1,7 @@
 ---
 - name: Copy node binaries
   copy:
-    src: "{{ localBuildOutput }}/{{ item }}"
+    src: "{{ localBuildOutput }}/{{ item }}" #docker builds, use: kubernetes/_output/dockerized/bin/linux/amd64/{{ item }}
```
eparis: I don't like this comment. It suggests that someone might want to make a change here. Which would be a bad idea, given we use localBuildOutput in more than this one location....

ansible/roles/master/defaults/main.yml
```diff
@@ -1,3 +1,3 @@
 kube_master_insecure_port: 8080
 
-localBuildOutput: ../../_output/local/go/bin
\ No newline at end of file
+localBuildOutput: ../../_output/local/go/bin # localBuildOutput: ../../kubernetes/_output/dockerized/bin/linux/amd64/ (docker builds)
```
eparis: Should this not default to `../../kubernetes/.....`    instead of just `../../_output`?

ansible/vagrant/Vagrantfile
eparis: Isn't that what https://github.com/kubernetes/contrib/blob/master/ansible/vagrant/Vagrantfile#L123 is all about? Doing this in a reproducible way across systems. Instead of doing some vagrant specific crap, like the salt stack is full of?

eparis:
Trying to get my head on straight. This PR does 2 real things.

1) update the centos 7 box because it disappeared.
2) switch from using https://github.com/kubernetes/contrib/blob/master/ansible/vagrant/vagrant-ansible.yml to using some custom vagrant plugin thing.

I dislike number 2. What was wrong with what we were doing? That is what should be fixed. I do not want more code where magic is done by the 'infrastructure'. Do it in ansible. Do it clearly. Don't do it like the salt stack  :)

jayunit100:
@eparis really its a code drop of what i had to do to get it working in OS X, vagrant.    

Im totally open to keeping any or none of the changes.  here is a bulleted list for you .... 

- centos update
- install parts for f23 selinux stuffs .  I also tested on f23 and found some selinux missing pylibs errors.
- sets up host networking.  i dont think etc hosts was working in current setup (at least, not on vbox)
- enables dynamic ips rather than static ones.   again, another vbox issue more than anything else.  
- updates the path to the install w comments,  its tricky to get that right imo.  
- took me a while to  get easyrsa working (fails due to the 64byte limit)...  the comment could save someone alot of time imo.
- Most important of all is it adds the `--iface` eth1 flag and this setup tests/confirms it works.

jayunit100:
@eparis which of the above bullets should we keep/throw away  ? i can close the PR to .  I will have a few hrs to look at this over the wknd.

ansible/roles/common/tasks/fedora-install.yml
jayunit100: hmm ok.  we can dump it.  not triggered automatically maybe.

ansible/roles/node/tasks/localBuildInstall.yml
jayunit100: agree its kinda hacky for OSx  simply cause all the paths are different and dir structure doesnt match.  not sure what the right solution is honestly to make it cross platform.  I wonder wether upstream can just  have one path for build output... so this isnt an issue

jayunit100:
Moved here, closing http://jayunit100.blogspot.com/2015/12/extending-contribansible-for-osx.html @eparis ill ping you later in the wk

eparis:
Things I want:
- Update centos box
- If ansible requires are missing (selinux) fix in 'pre-ansible' role
- Figure out why the ansible /etc/hosts was wrong, do not use the vagrant module to do it
- The flannel iface change (both to config file and to vbox vagrant block)
- Update default binary path to assume that contrib and kubernetes are both in $GOPATH/src/github/kubernetes/, but leave a comment pointing also to $GOPATH/src/k8s.io/kubernetes, just in case.

- Comment on easyrsa 64b limit is also ok, but what should be the real answer?

Test again and see what else we want/need.

jayunit100:
why not just use kubernetes.blah.org as the RSA key cname ? can we make  that "just work" ?  we dont really need to append all the ip addresses, do we? maybe md5 them ?

jayunit100:
per conversation w eparis in irc.
- will reimpl this wknd w/ the bullets above
- we also need to add deployments/experimental into the mix.

PR #294, opened 2015-11-29
[terraform-aws] add terraform aws provisioning scripts

tmc:
Mirrors upstream kube-up.sh behavior.


terraform-aws/LICENSE
```diff
@@ -0,0 +1,14 @@
+Copyright (c) 2015, Travis Cline <travis.cline@gmail.com>
```
brendandburns: I think we want everything under kubernetes to be covered by the Kubernetes apache license and copyright the kubernetes owners.

If that's not ok, this probably belongs in a different repo.

tmc: That's totally ok.

bgrant0607:
cc @mikedanese @justinsb

foxish:
@bgrant0607 can this be closed?

zmerlynn:
This doesn't look very current, and both [`kops`](https://github.com/kubernetes/kops) and [`kube-aws`](https://coreos.com/kubernetes/docs/latest/kubernetes-on-aws.html) can output Terraform and are maintained (and there are probably others).

mikedanese:
If we want this we should accept it into github.com/kubernetes/kube-deploy anyway and not contrib.

PR #295, opened 2015-11-29
Delay starting services until network is actually up

pesho:
Some services may attempt to start before a network interface is up, and then fail badly. A simple master or node server reboot is enough to reproduce - it will fail to bring up all required services more often than not (tested on Fedora 23). The services in question are:

- flanneld and kube-apiserver (on masters)
- flanneld and kube-proxy (on minion nodes)

The systemd unit files could have been updated directly to fix this, but as flanneld is installed from a distro package, a less invasive approach was taken instead: A dummy systemd target unit is introduced, which starts after network-online.target, but before the fragile services.

### DIFF
--- a/ansible/roles/master/files/delay-master-services.target
+++ b/ansible/roles/master/files/delay-master-services.target
@@ -1,0 +1,10 @@
+[Unit]
+Description=Delay Kubernetes services until network is actually online
+Wants=network-online.target
+After=network-online.target
+Before=flanneld.service
+Before=kube-apiserver.service
+
+[Install]
+WantedBy=flanneld.service
+WantedBy=kube-apiserver.service
--- a/ansible/roles/master/tasks/main.yml
+++ b/ansible/roles/master/tasks/main.yml
@@ -82,6 +82,12 @@
   service: name=kubelet enabled=yes state=started
   when: networking == 'opencontrail'
 
+- name: write the delay-master-services target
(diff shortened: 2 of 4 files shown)


eparis:
I do not understand api-server v flannel at all. Why would it matter?  I can at least understand why there might be problems with proxy v flannel, but I'd want to know what those problems were. As the proxy should likely be made more resilient if something is wrong...

Can we get logs of your failures?

pesho:
@eparis sure, here are the relevant logs:

https://gist.github.com/pesho/ecf29aca5f0f00bcbcf2

Briefly, this is what happens:

- `flanneld` fails to start before the network is up, because it can't get the IP address of the default interface. It will not attempt to retry.
- `docker` will not start because it requires `flanneld`.
- `kubelet` will not start because it requires `docker`.
- `kube-proxy` fails to start before the network is up, because it can't resolve the master's IP. It will then attempt to respawn too quickly and systemd will auto-disable it.
- `kube-apiserver` fails to start before the network is up, because it can't select an IP. Just like `kube-proxy` it will then attempt to respawn too quickly and systemd will auto-disable it.

I'd welcome proper fixes for each of these services as well, of course. This PR is just a quick, minimal fix, which resolves all these issues, and has minimum side-effects.

pesho:
Rebased, which fixed the Travis failure (it was an unstable test unrelated to this patch).

PR #298, opened 2015-11-30
Merge Bot: Fix infinite loop possible with missing CI tests

eparis:
Number of cleanups, but the main fix is in the last commit. After we stop paying attention to shippable altogether we can simplify the `ping_ci` and `ok-to-test` mungers. ping_ci was able to go into an infinite loop. Ping CI seemed to be able to trigger a bug in jenkins where only the e2e was getting run, instead of both e2e and unit. So make sure we get both of them.


brendandburns:
LGTM

PR #299, opened 2015-12-01
Rewrite relnotes tool using mungegithub/github

eparis:
Simplifies the code readability.

### DIFF
--- a/mungegithub/github/github.go
+++ b/mungegithub/github/github.go
@@ -396,6 +396,37 @@ func (config *Config) GetObject(num int) (*MungeObject, error) {
 	return obj, nil
 }
 
+// MergedAt will return the time the object was merged. If this is not a PR
+// is this was not merged, or if we can't tell, we return nil
+func (obj *MungeObject) MergedAt() *time.Time {
+	merged, err := obj.IsMerged()
+	if err != nil {
+		return nil
+	}
+	if !merged {
+		return nil
+	}
+	pr, err := obj.GetPR()
+	if err != nil {
+		return nil
+	}
+	return pr.MergedAt
+}
+
+// IsForBranch return true if the object is a PR for a branch with the given
+// name. It return false if it is not a pr, it isn't against the given branch,
+// or we can't tell
+func (obj *MungeObject) IsForBranch(branch string) bool {
+	pr, err := obj.GetPR()
+	if err != nil {
+		return false
+	}
+	if pr.Base != nil && pr.Base.Ref != nil && *pr.Base.Ref == branch {
+		return true
+	}
+	return false
+}
+
 // LastModifiedTime returns the time the last commit was made
 // BUG: this should probably return the last time a git push happened or something like that.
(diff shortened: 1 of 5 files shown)


mungegithub/github/github.go
```diff
@@ -396,6 +396,37 @@ func (config *Config) GetObject(num int) (*MungeObject, error) {
 	return obj, nil
 }
 
+// MergedAt will return the time the object was merged. If this is not a PR
+// is this was not merged, or if we can't tell, we return nil
```
ihmccreery: s/is/if

release-notes/release-notes.go
```diff
@@ -18,130 +18,82 @@ package main
...
-	}
+	config := &github.Config{}
+	cmd := &cobra.Command{
+		Use: "release-notes --last-release-pr=NUM --current-release-pr=NUM",
+		RunE: func(_ *cobra.Command, _ []string) error {
+			if err := config.PreExecute(); err != nil {
+				return err
+			}
+			if last == 0 || current == 0 {
+				return fmt.Errorf("Must set (at least) --last-releae-pr and --current-release-pr")
```
ihmccreery: s/releae/release

eparis:
@ihmccreery is the new code more readable maintainable?  I guess that's the real question.

release-notes/release-notes.go
```diff
@@ -18,130 +18,82 @@ package main
...
-	if len(token) > 0 {
-		tc = oauth2.NewClient(
-			oauth2.NoContext,
-			oauth2.StaticTokenSource(
-				&oauth2.Token{AccessToken: token}),
-		)
-	}
-	client := github.NewClient(tc)
+			buffer := &bytes.Buffer{}
+			config.ForEachIssueDo("closed", []string{"release-note"}, func(obj *github.MungeObject) error {
```
ihmccreery: Hm.  I'd like to keep the possibility of listing all PRs, not just those with the release-note label.  With this implementation, isn't that going to require pulling all ~20,000 issues (minus the open ones)?  That seems excessive to me.

In the current implementation, we pull PRs based on when they were last updated, and stop pulling when we see an issue that was last updated before the `last-release-pr` was merged, (and thus couldn't have been merged after).  Could we do something like that here?

(It requires making the `ForEachIssueDo` abstraction a bit more complicated ... I'm not sure if it's worth it.)

ihmccreery:
> @ihmccreery is the new code more readable maintainable? I guess that's the real question.

Yes, that is the real question.  See my last comment.  The raw GitHub API certainly has some serious limitations, but I'm not sure this abstraction actually solves any of them enough to make it worth the layer.

release-notes/release-notes.go
eparis: Yes it will, although they are pulled 100 at a time, so it's not that bad.

The abstraction has a build in min-pr-number and max-pr-number. We can add some sort of time based controls without making thigs too much uglier...

bgrant0607:
@david-mcmahon Is this still relevant, or could it be closed?

PR #306, opened 2015-12-03
Add vacation checking to blunderbuss

lavalamp:
A bunch of us will be out of the office for a stretch soon. It'd be nice if PRs were routed to people who _aren't_ out. What's the best way to tell the blunderbuss munger that we're out of the office?

1) Have people make a gist named "kubernetes-vacation" with a human readable format indicating dates that they're unavailable.
2) Same as above, but people make their own ooo repo.
3) Have people add the string " - OOO" to their location in github while they're out. (This was easy, so it's in this PR right now)
4) Keep a file in the main repo that people just update.
5) Other ideas?

Some people don't want their vacation schedules to be public.

### DIFF
--- a/mungegithub/mungers/blunderbuss.go
+++ b/mungegithub/mungers/blunderbuss.go
@@ -21,6 +21,7 @@ import (
 	"math/rand"
 	"os"
 	"strings"
+	"time"
 
 	"k8s.io/contrib/mungegithub/github"
 	"k8s.io/kubernetes/pkg/util/yaml"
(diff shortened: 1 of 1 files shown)


lavalamp:
@eparis

davidopp:
Can we just turn off blunderbuss for two weeks and tell the oncall to assign PRs like we did in the pre-blunderbuss days? I suspect the oncall load will be a lot lighter than normal, so this doesn't seem like too much extra burden.

eparis:
It is easy to turn off blunderbuss...

lavalamp:
Another idea: A private repo under the kubernetes org.

lavalamp:
@eparis: What do you think about merging this as-is? We can revert it in January. If people are worried about disclosing their vacation status, they don't have to use it. But I get a lot of PRs and I'd like to make use of this.

eparis:
We can turn it on this year, but its complete lack of discover-ability, the likelihood that we will set it and never remember to unset it, etc makes me not like this implementation.  So I'm ok with merge/revert and keep rolling around ideas for next year, or just turn off blunderbuss on the 18th and ask on call to do it until the 1st. I won't complain about either way.

eparis:
we're through christmas. I'm just gunna close....

PR #309, opened 2015-12-04
Update haproxy and add checks for services

aledbf:


service-loadbalancer/service_loadbalancer.go
```diff
@@ -140,6 +140,8 @@ type service struct {
 	Name string
 	Ep   []string
 
+	BackendPort int
```
bprashanth: please add a comment, something like: 
```
Kubernetes endpoint port. The application must serve a 200 page on this port.
```

service-loadbalancer/service_loadbalancer.go
```diff
@@ -388,8 +390,9 @@ func (lbc *loadBalancerController) getServices() (httpSvc []service, tcpSvc []se
 				continue
 			}
 			newSvc := service{
-				Name: getServiceNameForLBRule(&s, servicePort.Port),
-				Ep:   ep,
+				Name:        getServiceNameForLBRule(&s, servicePort.Port),
+				Ep:          ep,
+				BackendPort: servicePort.Port,
```
bprashanth: You're going to use this for health checks on the endpoint right? don't you want the endpoint port? I mean, you can have an nginx service with port 80, and targetport 8080 going to an endpoint running on port 8080.

service-loadbalancer/haproxy_reload
```diff
@@ -23,6 +23,4 @@
 # -s soft reload, wait for pids to finish handling requests
 # -f send pids a resume signal if reload of new config fails
 
-socat /tmp/haproxy - <<< "show servers state" > /var/state/haproxy/global
```
bprashanth: what was the problem here?

service-loadbalancer/template.cfg
```diff
@@ -114,14 +110,14 @@ backend {{$svc.Name}}
     # http://cbonte.github.io/haproxy-dconv/configuration-1.5.html#stick-table
     stick-table type ip size 100k expire 30m
     stick on src
-    {{range $j, $ep := $svc.Ep}}server {{$ep}} {{$ep}}
+    {{range $j, $ep := $svc.Ep}}server {{$ep}} {{$ep}} check port {{$svc.BackendPort}} inter 5
```
bprashanth: WDYT about doing what you once suggested (make this go straight to the liveness probe of the pod)? just a suggestion, ok to punt

bprashanth:
Ah i think you need to update the template in unittests. While you're doing that can you please remove comments from the unittest templates, so we can modify comments in the main one without breaking tests.

service-loadbalancer/haproxy_reload
aledbf: If haproxy is reloaded too frequently the contents of that file is not correct

service-loadbalancer/template.cfg
aledbf: > make this go straight to the liveness probe of the pod

some apps could not define one

aledbf:
ping @bprashanth

service-loadbalancer/haproxy_reload
bprashanth: Is that an haproxy bug? or do they document a min reloads per second? can we just limit the reloads per second to work around this?

service-loadbalancer/service_loadbalancer.go
```diff
@@ -388,8 +391,9 @@ func (lbc *loadBalancerController) getServices() (httpSvc []service, tcpSvc []se
 				continue
 			}
 			newSvc := service{
-				Name: getServiceNameForLBRule(&s, servicePort.Port),
-				Ep:   ep,
+				Name:        getServiceNameForLBRule(&s, servicePort.Port),
+				Ep:          ep,
+				BackendPort: servicePort.TargetPort.IntVal,
```
bprashanth: Named ports are a thing, so lets just get this right here even. Please isolate the port switch case in getEndpoints to something like getTargetPort(servicePort) and call it from both places.

service-loadbalancer/template.cfg
bprashanth: Yeah in that case we'd default to the current 'ep:backendPort/'.

aledbf: can we leave that for another PR? :)

bprashanth:
LGTM. Would be great if you could add a test case for this under: 
https://github.com/kubernetes/kubernetes/blob/ffdfc68d11153b3ca019b84ebce8b867387bf42f/test/e2e/serviceloadbalancers.go

Sounds like all that's needed to trigger it is a failing health check on a secondary replica, so maybe just bump the replicas on the test rc and confirm all endpoints with rr algorithm.

PR #310, opened 2015-12-04
Alphabetize blunderbuss.yml for easy human reading

eparis:
simply `mungegithub blunderbuss-normalize`


lavalamp:
Thanks! sorry, I just didn't see the previous plugin definition. :)

lavalamp:
LGTM

PR #313, opened 2015-12-04
Put client_max_body_size in server config if found in ingress annotation

moserke:


aledbf:
@moserke please check this https://github.com/kubernetes/contrib/pull/280 
The idea is to allow the customization of the configuration without the use of annotations.

bprashanth:
@moserke thanks for the pr. WDYT about just doing https://github.com/kubernetes/contrib/pull/280? feel free to review that one and put in feature requests. If you'd really like just this one feature in the nginx-toy I'm fine with that as well (though I doubt that's the case?)

moserke:
This was actually inadvertently opened.  Meant to PR onto our forked master branch.  It's a dirty solution we did to get us moving past a small problem.  Not ready for wide prime time.  :-)

bprashanth:
@moserke well, comments on https://github.com/kubernetes/contrib/pull/280 are still much appreciated. Especially if you can shoehorn your use cases in there.

Issue #314, opened 2015-12-06
GLBC should recover gracefully from default backend mangling


bprashanth:
If you go behind the controller's back and delete the default backend, it doesn't recreate (if fact it does weird things by trying to create a urlmap with a non-existent backend). It does for everything else. This should get fixed.

PR #315, opened 2015-12-06
Sync default backend if user deletes it.

bprashanth:
Simple change that does as the name suggests.


bprashanth:
https://github.com/kubernetes/contrib/issues/314

freehan:
LGTM

PR #317, opened 2015-12-07
Fixups for release-notes.go: end-of-branch issues, API rate limiting, and release-note filter flag

ikehz:
There were some issues with finding PRs and falling off the end of the branch, and the API rate limit, which I've resolved here.

This also adds a flag to allow a user to not filter on the `release-note` label, to list all PRs between two.

### DIFF
--- a/release-notes/release-notes.go
+++ b/release-notes/release-notes.go
@@ -97,6 +97,10 @@ func main() {
 		}
 		unmerged := 0
 		merged := 0
+		if len(results) == 0 {
+			done = true
+			break
+		}
 		for ix := range results {
 			result := &results[ix]
 			// Skip Closed but not Merged PRs
@@ -130,6 +134,7 @@ func main() {
 			// Check to see if it has the release-note label.
 			fmt.Printf(".")
 			labels, _, err := client.Issues.ListLabelsByIssue("kubernetes", "kubernetes", *pr.Number, &github.ListOptions{})
+			time.Sleep(5 * time.Second)
 			if err != nil {
 				fmt.Printf("Error contacting github: %v", err)
 				os.Exit(1)
--- a/release-notes/release-notes.go
+++ b/release-notes/release-notes.go
@@ -30,10 +30,11 @@ import (
 )
 
 var (
-	base    string
-	last    int
-	current int
-	token   string
+	base          string
+	last          int
+	current       int
+	token         string
+	relnoteFilter bool
 )
 
 type byMerged []*github.PullRequest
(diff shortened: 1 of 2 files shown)


release-notes/release-notes.go
```diff
@@ -127,16 +133,21 @@ func main() {
...
-			for _, label := range labels {
-				if *label.Name == "release-note" {
-					fmt.Fprintf(buffer, "   * %s (#%d, @%s)\n", *pr.Title, *pr.Number, *pr.User.Login)
+			if !relnoteFilter {
+				fmt.Fprintf(buffer, "   * %s (#%d, @%s)\n", *pr.Title, *pr.Number, *pr.User.Login)
+			} else {
+				// Check to see if it has the release-note label.
+				fmt.Printf(".")
+				labels, _, err := client.Issues.ListLabelsByIssue("kubernetes", "kubernetes", *pr.Number, &github.ListOptions{})
+				time.Sleep(5 * time.Second)
```
roberthbailey: debugging device that you left in?

release-notes/release-notes.go
```diff
@@ -127,16 +133,21 @@ func main() {
...
-				os.Exit(1)
-			}
-			for _, label := range labels {
-				if *label.Name == "release-note" {
-					fmt.Fprintf(buffer, "   * %s (#%d, @%s)\n", *pr.Title, *pr.Number, *pr.User.Login)
+			if !relnoteFilter {
+				fmt.Fprintf(buffer, "   * %s (#%d, @%s)\n", *pr.Title, *pr.Number, *pr.User.Login)
+			} else {
+				// Check to see if it has the release-note label.
+				fmt.Printf(".")
```
roberthbailey: You should print the dot ahead of the if/else branch since it would be nice to see progress in both modes.

release-notes/release-notes.go
roberthbailey: or is this is the "api rate limit" part of the PR?

release-notes/release-notes.go
ihmccreery: With `!relnoteFilter`, there's no need for progress because there's not API calling; it's instantaneous.

release-notes/release-notes.go
ihmccreery: Yeah, the latter.  Added a comment.

ihmccreery:
@roberthbailey PTAL.

roberthbailey:
There's probably a better way to do this than adding sleeps, but this is fine for now. Travis is failing though.

eparis:
maybe we can take a look at https://github.com/kubernetes/contrib/pull/299 instead of making this more complex?

ihmccreery:
Hm.  See my comment:

> Hm.  I'd like to keep the possibility of listing all PRs, not just those with the release-note label.  With this implementation, isn't that going to require pulling all ~20,000 issues (minus the open ones)?  That seems excessive to me.

> In the current implementation, we pull PRs based on when they were last updated, and stop pulling when we see an issue that was last updated before the `last-release-pr` was merged, (and thus couldn't have been merged after).  Could we do something like that here?

> (It requires making the `ForEachIssueDo` abstraction a bit more complicated ... I'm not sure if it's worth it.)

ihmccreery:
@eparis I'm going to go ahead and merge this, just so we have something that works until #299 gets in.

PR #319, opened 2015-12-08
add myself to blunderbuss for pkg/client/leaderelection

mikedanese:
I'd like to field these initially

### DIFF
--- a/mungegithub/blunderbuss.yml
+++ b/mungegithub/blunderbuss.yml
@@ -123,6 +123,9 @@ prefixMap:
     - lavalamp
     - smarterclayton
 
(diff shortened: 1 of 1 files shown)


lavalamp:
LGTM

lavalamp:
Won't take affect until we build & restart-- @eparis has been doing that, I think.

mikedanese:
I can kick it once the PR goes in

eparis:
That's fine, you should be able to 'make' from head, upload the image to gcr, and then do a rolling update to rc.yaml.

eparis:
(rc.yaml in your local tree will be updated when you do the make)

PR #327, opened 2015-12-11
service-loadbalancer: HAProxy: Disallow concurrent reload attempts, should fix #321

### LINKED ISSUE #321: Haproxy reload not correctly in service loadbalance
we encounter the problem just like this [haproxy-issue-48](https://github.com/haproxy/haproxy/issues/48), and post a new [issue](https://github.com/haproxy/haproxy/issues/56). The old haproxy process will influence the new one that make the service visiting not stable. Until now I cannot find a solid solution and we make a little change when haproxy reload by ```-sf```. if the old haproxy process still running, then kill it by ```kill -9 <pid>```. It seems work fine but I don't know it would or not cause some bad affect when a lot of service create or modified. Did you have any suggestions or solutions that have no side effects?

pesho:
The script-file itself is used as a mutex, idea borrowed from https://github.com/mesosphere/marathon-lb/commit/598ff09b0c80349418c38c44808a616e77efec74.

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/327)

### DIFF
--- a/service-loadbalancer/haproxy_reload
+++ b/service-loadbalancer/haproxy_reload
@@ -23,6 +23,12 @@
 # -s soft reload, wait for pids to finish handling requests
 # -f send pids a resume signal if reload of new config fails
 
+exec 200<$0
+# Block till we can acquire an exclusive file lock:
+flock 200
+
 socat /tmp/haproxy - <<< "show servers state" > /var/state/haproxy/global
 
 haproxy -f /etc/haproxy/haproxy.cfg -p /var/run/haproxy.pid -D -sf $(cat /var/run/haproxy.pid)
+
+flock -u 200


bprashanth:
IIUC the problem is that haproxy doesn't reload synchronously, one tells it to reload, and it waits for keep-alive connections to shutdown first. If telling it to reload repeatedly while it's waiting for keep-alives is bad, this will not help right? (because haproxy can still be waiting on a keep-alive after we've released the flock).

Or have I misunderstood the real problem here?

service-loadbalancer/haproxy_reload
```diff
@@ -23,6 +23,11 @@
 # -s soft reload, wait for pids to finish handling requests
 # -f send pids a resume signal if reload of new config fails
 
+exec 200<$0
+flock 200
```
bprashanth: If the file handle is already locked does this exit or wait?

pesho: It waits. Adding the `-n` parameter would cause it to exit.

pesho:
> IIUC the problem is that haproxy doesn't reload synchronously, one tells it to reload, and it waits for keep-alive connections to shutdown first. If telling it to reload repeatedly while it's waiting for keep-alives is bad, this will not help right? (because haproxy can still be waiting on a keep-alive after we've released the flock).
> 
> Or have I misunderstood the real problem here?

No, waiting for the keep-alive connections is not a problem. It's ok to have several soft-stopped processes still serving keep-alive connections.

The problem this targets is if a second reload attempt occurs while the new process is being setup, but it still hasn't sent the SIGUSR1 signal to the old one.

pesho:
For context, please see https://github.com/haproxy/haproxy/issues/48

service-loadbalancer/haproxy_reload
bprashanth: Can you please add this as a comment? Something like:
```
Block till we can acquire an exclusive file lock.
```

bprashanth:
LGTM but for the one nit

service-loadbalancer/haproxy_reload
pesho: @bprashanth, added.

test-foxish:
recomputing cla status...

pesho:
@k8s-merge-robot I signed it!

bprashanth:
@pesho sorry I dropped the ball on this, is it still an issue? (I think we released a few slb images after this).

pesho:
@bprashanth not 100% sure myself, as I've moved to using Nginx as loadbalancer in the meantime.

In any case, this should be a desirable change, as concurrent Haproxy reload attempts are known to cause trouble. Looking at the Go sources I don't see anything that prevents them, so this PR is probably still relevant (there is a rate limiter used in the Go binary, but as far as I understand it does not actually prevent concurrent executions).

bprashanth:
forcing travis rerun

grodrigues3:

- **service-loadbalancer**

cc @eparis 

PR #328, opened 2015-12-11
Avoid simultaneous haproxy reloads

aledbf:


aledbf:
@ZhuPeng please test with this update

ZhuPeng:
@aledbf Not work, and encounter the same problem. I think the wrong place was haproxy, not the service loadbalance reload policy.

aledbf:
closing. This doesn't solve the haproxy reload issue.

Issue #329, opened 2015-12-11
Fails on Fedora23 for apiserver_runtime_config undefined variable


MrMitch17:
TASK: [master | write the config file for the api server] ********************\* 

fatal: [x.x.x.x] => {'msg': "AnsibleUndefinedVariable: One or more undefined variables: 'apiserver_runtime_config' is undefined", 'failed': True}
ansible/roles/master/templates/apiserver.j2:26

Appears it's possibly the same issue as #323

eparis:
@jayunit100

jayunit100:
might require a similar patch,  let me see

jayunit100:
@MrMitch17  hi ! 
For now, uncomment `group_vars/all.yml ` that should be enough for it to work. 
Meanwhile ill submit a quick patch!

jayunit100:
Hi @MrMitch17 is this now all working for you?  I tested it on os x but i think ansible behaviour can vary a little esp . since I make a few changes to the files before running.  Just let us know.  thanks again for rapidly reporting these.

PR #331, opened 2015-12-14
API Server opts use | for nil check

jayunit100:
@eparis I think this is what you had in mind for this? It also fixes the other screw up i made with the pyhton style `if nil` check on apiserver


jayunit100:
NOT TESTED.  I'll test as soon as you confirm that this is the way you wanted the args to look.

jayunit100:
fixes #329

eparis:
I like it....

jayunit100:
k ran a quick test  looks like its working to me

PR #332, opened 2015-12-15
use alpine baseimage for glbc

MDrollette:
33mb image versus 225mb

### DIFF
--- a/Ingress/controllers/gce/Dockerfile
+++ b/Ingress/controllers/gce/Dockerfile
@@ -12,23 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# TODO: use radial/busyboxplus:curl or alping instead
-FROM ubuntu:14.04
+FROM alpine:3.2
 MAINTAINER Prashanth B <beeps@google.com>
 
-# so apt-get doesn't complain
-ENV DEBIAN_FRONTEND=noninteractive
-RUN sed -i 's/^exit 101/exit 0/' /usr/sbin/policy-rc.d
-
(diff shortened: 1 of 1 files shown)


brendandburns:
Last time we tried Alpine it had problems w/ our DNS @thockin do we know if this is still true?

thockin:
Alpine does have DNS problems.  Does this image need to do DNS lookups at
all?  Like does it use `kubernetes.default` as a DNS name, even?

MDrollette:
Is there a preferred minimal baseimage that does DNS properly?

thockin:
busybox is what we mostly use.  It's doesn't have a nice package manager
like Alpine, but it has a shell and basic utilities.

@bprashanth, I feel like there was a reason you didn't use busybox as a
base?

bprashanth:
Sorry for the delay. I think it was because certificat trust stores came pre-installed in the ubuntu image (or it was easy enough to apt-get when I needed them). I'm planning to swap the base image for something smaller very soon. @MDrollette if you want to swap it for busybox now I can figure out the roots later (we might get away with mounting them from host, I'm open to ideas; IIRC go checks standard locations by default: https://code.google.com/p/go/issues/detail?id=6391).

bprashanth:
Also we use a custom alpine image with certs for the service-loadbalancer (https://github.com/kubernetes/contrib/blob/master/service-loadbalancer/Dockerfile), so I think that should be fine. I'll run this pr against E2Es and confirm.

lleszczu:
alpine:edge is using new musl version, which supports `search` directive in resolv.conf.

PR #333, opened 2015-12-15
Update nginx and replace lua5.1 with luajit 

aledbf:
The reason to replace lua with luajit is the support of [yield/resume in coroutines](https://github.com/openresty/lua-nginx-module#lua-coroutine-yieldingresuming)


bprashanth:
LGTM, will merge when travis is green

aledbf:
ping @bprashanth

PR #336, opened 2015-12-17
service-loadbalancer: When set targetPort by string name and didnot have a port name, will found no endpoints in that service

ZhuPeng:
The EndpointPorts's Name attr was corresponds to the ServicePort.Name. If we set EndpointPort Name, then we must set the same name to the corresponds ServiePort.Name on test file.
https://github.com/kubernetes/kubernetes/blob/master/pkg/api/types.go#L1420-L1431

We use a string name to represent TargetPort, which might set in Pod spec, and we do not set a ServicePort.Name or the TargetPort different with ServicePortName
https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/fluentd-elasticsearch/es-service.yaml#L12-L14
If I create the above service in kubernetes and it works well. But this service did not found any endpoints in service-loadbalancer.

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/336)

### DIFF
--- a/service-loadbalancer/service_loadbalancer.go
+++ b/service-loadbalancer/service_loadbalancer.go
@@ -341,7 +341,10 @@ func (lbc *loadBalancerController) getEndpoints(
 					targetPort = epPort.Port
 				}
 			case util.IntstrString:
-				if epPort.Name == servicePort.TargetPort.StrVal {
+				// two situations:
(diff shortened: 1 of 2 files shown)


test-foxish:
recomputing cla status...

grodrigues3:

- **service-loadbalancer**

cc @bprashanth 

PR #339, opened 2015-12-18
Handle salt templated monitoring definitions

### LINKED ISSUE #337: Heapster Deployment Broken
The heapster deployment seems to be broken because in the Ansible playbooks because of a recent-ish change to the controller definition to allow for dynamic configurations. I suspect this is usually handled by Salt, but this does not occur during an Ansible based deployment. Here is the relevant commit that seems to introduce this issue:

https://github.com/kubernetes/kubernetes/commit/eee87558e06e07ee62362c61c3d0b49e29de4fdf

Removing the templating from the top of the service file seems to allow for creation as expected.

eparis:
Which were added https://github.com/kubernetes/kubernetes/commit/eee87558e06e07ee62362c61c3d0b49e29de4fdf

And fixes #337

### DIFF
--- a/ansible/roles/kubernetes-addons/defaults/main.yml
+++ b/ansible/roles/kubernetes-addons/defaults/main.yml
@@ -1,4 +1,5 @@
 kube_addons_dir: "{{ kube_config_dir }}/addons"
 
-
 local_temp_addon_dir: /tmp/kubernetes/addons
+
+heapster_memory: (200 + length(groups['nodes']) + 12)|string + "Mi"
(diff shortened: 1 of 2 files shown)


eparis:
@rsmitty Can you test this? I have not.

rsmitty:
Sure, I can try this out today.

rsmitty:
Looks almost perfect. There's a rogue `with_items:` on line 10 of cluster-monitoring.yml. Once I removed that, it seems to deploy as expected. 

Unsure about the Travis CI check failing though. Maybe it's the same reason? I'm not familiar with what's happening on the build side.

rsmitty:
I believe this is good to go now.

PR #343, opened 2015-12-30
validate sha1 sum for src file & dest file

dalanlan:
kill TODO

>// TODO: use the master election utility once it is merged in.

What does a master election utility exactly mean, if i may ask?


mqliang:
> What does a master election utility exactly mean, if i may ask?

see https://github.com/kubernetes/kubernetes/pull/6995 and https://github.com/kubernetes/kubernetes/pull/16830

PR #353, opened 2016-01-06
Add election code and examples

brendandburns:


mikedanese:
Can we use the leader election client?

https://godoc.org/k8s.io/kubernetes/pkg/client/leaderelection

brendandburns:
oh, sure, I missed that you'd added that...  I'll refactor this in terms of that.

brendandburns:
@mikedanese 
Switched to use the leaderelection stuff.  Had to wrap it up a bit to get a cleaner API, we should migrate that code back into the leaderelection package.

ptal,
thanks!
--brendan

election/example/main.go
```diff
@@ -0,0 +1,99 @@
...
+	fn := func(str string) {
+		leader.Name = str
+		fmt.Printf("%s is the leader\n", leader.Name)
+	}
+
+	e, err := election.NewElection(*name, *id, *namespace, *ttl, fn, kubeClient)
+	if err != nil {
+		glog.Fatalf("failed to create election: %v", err)
+	}
+	go e.Run()
```
mikedanese: the e.Run() should be warpped in an Until Forever. A single election client does one cycle of candiate->aquire master->lose master and is expected to compelete.

election/lib/election.go
```diff
@@ -0,0 +1,104 @@
...
+		OnStartedLeading: func(stop <-chan struct{}) {
+			callback(id)
+		},
+		OnStoppedLeading: func() {
+			leader, _, err := getCurrentLeader(electionId, namespace, c)
+			if err != nil {
+				glog.Errorf("failed to get leader: %v", err)
+				return
+			}
+			callback(leader)
```
mikedanese: if we are a leader and the network partitions, we want to stop leading but we may not be able to observe a new leader if we can't talk to the apiserver. What do we want to report? callback("")? We shouldn't just keep leading.

election/example/main.go
brendandburns: done.

election/lib/election.go
brendandburns: done.

brendandburns:
Comments addressed, please re-check.

Thanks!
--brendan

mikedanese:
LGTM, I might try to use this to write an e2e test for the leader election client.

mikedanese:
Looks like you need to upgrade the godeps for k8s

mikedanese:
Also do you mean to check in the binary?

brendandburns:
@mikedanese this has been rebased on top of the godeps update elsewhere.  ptal.

thanks!
--brendan

mikedanese:
Some boilerplate failures.

brendandburns:
fixed (I hope)

ptal
thanks!

mikedanese:
LGTM

imesh:
@brendandburns Do we have any documentation on how the leader election works? Is it using etcd internally? Thanks!

aledbf:
@imesh please check this comment https://github.com/kubernetes/kubernetes/blob/master/pkg/client/leaderelection/leaderelection.go#L17

imesh:
Thanks @aledbf!

pdeva:
> @imesh please check this comment https://github.com/kubernetes/kubernetes/blob/master/pkg/client/leaderelection/leaderelection.go#L17

@aledbf the link is broken 😔

samchenraising:
> > @imesh please check this comment https://github.com/kubernetes/kubernetes/blob/master/pkg/client/leaderelection/leaderelection.go#L17
> 
> @aledbf the link is broken 😔

@pdeva, you can browse to the repo from [that day in time](https://github.com/kubernetes/kubernetes/blob/e7d01097dcd2b23c571006792fb2ed8cb85553ef/pkg/client/leaderelection/leaderelection.go#L17)

PR #361, opened 2016-01-08
Add keepalived controller to provide virtual IP addresses

aledbf:


aledbf:
ping @bprashanth

bprashanth:
Nice, I'll try to get to this over the weekend

keepalived-vip/README.md
```diff
@@ -0,0 +1,374 @@
...
+need to set up firewall rules for the service port(s) (tcp:30302) to serve traffic.
+
+See http://releases.k8s.io/HEAD/docs/user-guide/services-firewalls.md for more details.
+service "echoheaders" created
+```
+
+Next add the required annotation to expose the service using a local IP
+
+```
+$ kubectl annotate svc echoheaders "k8s.io/public-vip=10.4.0.50"
```
bprashanth: It would be really cool if you could give an ingress an public vip like this as well. The Ingress was born out of the need to fanout a single ip to various services, either based on hostname or url regex.

keepalived-vip/README.md
```diff
@@ -0,0 +1,374 @@
...
+This just works. What's the issue then? 
+
+The issue is that it does not provide High Availability because beforehand is required to know the IP addresss of the node where is running and in case of a failure the pod can be be moved to a different node. Here is where ipvs could help. 
+The idea is to define an IP address per service to expose it outside the Kubernetes cluster and use vrrp to announce this "mapping" in the local network.
+With 2 or more instance of the pod running in the cluster is possible to provide high availabity using a single IP address.
+
+##### What is the difference between this and [service-loadbalancer](https://github.com/kubernetes/contrib/tree/master/service-loadbalancer) or [nginx-alpha](https://github.com/kubernetes/contrib/tree/master/Ingress/controllers/nginx-alpha) to expose one or more services?
+
+This should be considered a complement, not a replacement for HAProxy or nginx. The goal using keepalived is to provide high availability and to bring certainty about how an exposed service can be reached. For instance keepalived can use used to expose the service-loadbalancer or nginx ingress controller in the LAN using one IP address.
+
```
bprashanth: Please also add a section clarifying the difference between this and simply giving service-loadbalancer a vip via a nodeport service (the main difference would probably be that the vrrp vip is discoverable).

bprashanth:
While reading i realized that Unicast would be a lot more useful because we can then deploy this to all clouds as well, by using the notify script to reconfigure the static ip (http://kaivanov.blogspot.com/2015/02/keepalived-using-unicast-track-and.html). How hard would that be to integrate right now (noticed it's in your TODO section currently)?

aledbf:
> How hard would that be to integrate right now (noticed it's in your TODO section currently)?

Is not hard, just add a new parameter to choose between unicast/multicast and in case of unicast add the nodes addresses in keepalived.conf

aledbf:
@bprashanth optional unicast support ready

keepalived-vip/README.md
```diff
@@ -0,0 +1,373 @@
...
+Now the creation of the daemonset
+```
+$ kubectl create -f vip-daemonset.yaml
+daemonset "kube-keepalived-vip" created
+$ kubectl get daemonset
+NAME                  CONTAINER(S)          IMAGE(S)                         SELECTOR                        NODE-SELECTOR
+kube-keepalived-vip   kube-keepalived-vip   gcr.io/google_containers/kube-keepalived-vip:0.1   name in (kube-keepalived-vip)   type=worker
+```
+
+**Note: the daemonset yaml file contains a node selector. This is not required**
```
bprashanth: can you clarify this line?

keepalived-vip/controller.go
```diff
@@ -0,0 +1,253 @@
...
+}
+
+// getEndpoints returns a list of <endpoint ip>:<port> for a given service/target port combination.
+func (ipvsc *ipvsControllerController) getEndpoints(
+	s *api.Service, servicePort *api.ServicePort) (endpoints []service) {
+	ep, err := ipvsc.epLister.GetServiceEndpoints(s)
+	if err != nil {
+		return
+	}
+
```
bprashanth: suggestion for future cleanup, we're duplicating the same loops in many service related controllers, we should isolate into a library interface if possible.

keepalived-vip/controller.go
```diff
@@ -0,0 +1,253 @@
...
+	if err != nil {
+		return err
+	}
+
+	err = ipvsc.keepalived.Reload()
+	if err != nil {
+		return err
+	}
+
+	ipvsc.reloadRateLimiter.Accept()
```
bprashanth: This takes tokens out of the tokenbucket and blocks if none are available, so if you're calling sync from multiple threads wouldn't you want them to block at the head of this function instead?

keepalived-vip/keepalived.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	conf := make(map[string]interface{})
+	conf["iface"] = k.iface
+	conf["myIP"] = k.ip
+	conf["netmask"] = k.netmask
+	conf["svcs"] = svcs
+	conf["nodes"] = k.neighbors
+	conf["priority"] = k.priority
+	conf["authPass"] = k.getSha()
```
bprashanth: most of the fields are obvious except for this one. Suggest a one line comment somewhere, especially explaining why it's the sha of nodenames.

keepalived-vip/main.go
```diff
@@ -0,0 +1,101 @@
...
+	if !specified {
+		namespace = ""
+	}
+
+	err = loadIPVModule()
+	if err != nil {
+		glog.Fatalf("Terminating execution: %v", err)
+	}
+
+	err = changeSysctl()
```
bprashanth: Please add a comment about this, or a link to some explanation. Either here or above the function def is fine (or above sysctlAdjustments).

keepalived-vip/utils.go
```diff
@@ -0,0 +1,219 @@
...
+	"k8s.io/kubernetes/pkg/util/node"
+	"k8s.io/kubernetes/pkg/util/sysctl"
+)
+
+const (
+	vethRegex = "^veth.*"
+)
+
+var (
+	invalidIfaces = []string{"lo", "docker0", "flannel.1"}
```
bprashanth: Why not make these flags with defaults (or a list of invalid interfaces names as a single flag), since the docker interface can be setup as cbr0 by the kubelet?

keepalived-vip/utils.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+
+	return &nodeInfo{
+		iface:   interfaceByIP(ip),
+		ip:      ip,
+		netmask: maskForLocalIP(ip),
+	}, nil
+}
+
+// myIP returns
```
bprashanth: please complete the comment

keepalived-vip/utils.go
```diff
@@ -0,0 +1,219 @@
...
+			return ip, nil
+		}
+	}
+
+	glog.Errorf("error getting local IP: %v", err)
+	return "0.0.0.0", err
+}
+
+// netInterfaces returns a slice containing the local network interfaces
+// excluding lo, docker0, flannel.1 and veth interfaces.
```
bprashanth: what happens if you don't ignore these interfaces, rather, what happens if i have rkt.0 or lxcbr0 or something?

bprashanth:
mostly nits, looks good otherwise

bprashanth:
Thanks

PR #362, opened 2016-01-08
Don't comment "Labelling this PR as size/X", that's redundant.

spxtr:
I haven't tested this. The comment doesn't tell us anything new.  The only reason I can think of for it to exist is if you want an email notification or something, but that seems like a stretch.

### DIFF
--- a/mungegithub/mungers/size.go
+++ b/mungegithub/mungers/size.go
@@ -18,7 +18,6 @@ package mungers
 
 import (
 	"bufio"
-	"fmt"
 	"os"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


thockin:
Email is exactly why it comments.  Many of the project leads depend on
email.

spxtr:
Why do project leads need an email telling them how big their PRs are?

Feel free to close this. It's not a big deal at all.

thockin:
The fact that github sends me an email means I can label it in my inbox, which makes it easier for me to triage and review.  In other words, XL PRs turn red in my inbox, and don't get reviewed as readily :)

thockin:
Thanks, though.  Appreciate the PR.

PR #363, opened 2016-01-08
go2docker: remove path assumptions for portability

raggi:
The linux amd64 package does not conform to the layout previously expected by
the code. The new code relies on introspection of the installed go runtime
rather than assumptions about the package layout.

Put another way: this makes go2docker work with the standard go linux package installed using the standard instructions.

### DIFF
--- a/go2docker/go2docker.go
+++ b/go2docker/go2docker.go
@@ -69,17 +69,17 @@ func main() {
 		args = flag.Args()
 	}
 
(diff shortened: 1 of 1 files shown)


raggi:
FYI: this was tested on both OSX (homebrew) and Linux from both a debian package and the binary tarball.

bgrant0607:
LGTM, thanks.

PR #377, opened 2016-01-13
Remove unused dependency

aledbf:


keepalived-vip/main.go
```diff
@@ -85,8 +84,6 @@ func main() {
 		glog.Fatalf("Terminating execution: %v", err)
 	}
 
-	proc.StartReaper()
```
bprashanth: so this doesn't risk the same zombie leaking problem?

aledbf: @bprashanth no because we send the `SIGHUP` signal and keepalived reloads the configuration file.

bprashanth: Ah, while we're doing this I noticed the public methods in keepalived.go don't have godocs, could you please populate them (and comment above Reload what you just said)

bprashanth:
LGTM but for the comment nit

aledbf:
ping @bprashanth

bprashanth:
ah didn't realize you had a new commit. LGTM.

PR #389, opened 2016-01-19
Show valid options

aledbf:
Replace error

```
$ ./hack/for-go-proj.sh
./hack/for-go-proj.sh: line 25: 1: unbound variable
```

with

```
$ ./hack/for-go-proj.sh
missing subcommand: [build|install|test]
```


mikedanese:
Thanks!

PR #391, opened 2016-01-20
Update Travis to Go 1.5.3

ixdy:
@ihmccreery @mikedanese @eparis


ihmccreery:
LGTM; Travis still seems borked.

ixdy:
I think one of the tests is just flaky. Probably more-so in Go 1.5?

PR #392, opened 2016-01-21
Comparator for resource usage.

gmarek:
cc @fgrzadkowski @wojtek-t

### DIFF
--- a/compare/src/compare-resources.go
+++ b/compare/src/compare-resources.go
@@ -1,0 +1,231 @@
+/*
+Copyright 2015 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package src
+
+import (
+	"fmt"
+	"math"
+	"os"
+	"sort"
+	"strings"
+	"text/tabwriter"
+
+	"k8s.io/kubernetes/test/e2e"
+
+	"github.com/golang/glog"
+)
+
+const (
+	// ResourceUsageVarianceAllowedPercent specifies how much variance we allow between results.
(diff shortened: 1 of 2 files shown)


gmarek:
Another flake.

compare/src/compare-resources.go
```diff
@@ -0,0 +1,231 @@
...
+	// ResourceUsageVarianceAllowedPercent bigger than the smaller one. I.e. if X < Y, then Y is too
+	// big if Y > X + X*(ResourceUsageVarianceAllowedPercent/100).
+	ResourceUsageVarianceAllowedPercent = 50
+	// To avoid false negatives we assume that minimal CPU usage is 5% and memory 50MB
+	minCPU = 0.05
+	minMem = int64(50 * 1024 * 1024)
+)
+
+type percentileUsageData struct {
+	perc    string
```
mwielgus: why string?

mwielgus: and why abbreviated?

mwielgus:
Not sure if i understand the purpose of this PR - lets do a short vc tomorrow.

compare/src/compare-resources.go
gmarek: Don't get me started on 'why string'... It was an int, as it should be, but JSON encoder can't handle int keys in map. Hence it's just a serialized int, so I won't need it to cast it.

And it's abbreviated because I'm lazy - I can change that.

mwielgus: Yes, please make it percentile and add a comment why it is string.

compare/src/compare-resources.go
```diff
@@ -0,0 +1,231 @@
...
+// identified by the container name.
+type ViolatingResourceUsageData map[string]ViolatingResourceUsageDataPair
+
+func writeViolatingResourceData(minCPU, maxCPU, baselineMinCPU, baselineMaxCPU, allowedVariance float64, enableOutputColoring bool, writer *tabwriter.Writer) {
+	fmt.Fprint(writer, "(")
+	changeColorFloat64AndWrite(minCPU, baselineMinCPU, allowedVariance, enableOutputColoring, writer)
+	fmt.Fprint(writer, ",")
+	changeColorFloat64AndWrite(maxCPU, baselineMaxCPU, allowedVariance, enableOutputColoring, writer)
+	fmt.Fprint(writer, ")")
+	return
```
mwielgus: return not needed

mwielgus:
Just a few nitpicks - otherwise LGTM

gmarek:
@mwielgus PTAL

mwielgus:
LGTM

PR #398, opened 2016-01-24
Set file permissions so another container can read them

paulbakker:
When used on Kubernetes as a side-car, the other container should have read permissions to the cloned files. This is not necessarily the case without explicitly setting it.

I ran into this problem in combination with an Nginx container and a shared volume.


bgrant0607:
LGTM, thanks.

Issue #367, opened 2016-01-11
GCE Loadbalancer controller doesn't report events anymore


bprashanth:
From logs:
```
E0111 00:56:27.615692       1 event.go:252] Unsupported event type: 'ADD'
...
E0111 00:57:39.362128       1 event.go:252] Unsupported event type: 'CREATE'
```

Maybe https://github.com/kubernetes/contrib/pull/358

brendandburns:
@bprashanth it's possible, please check:

https://github.com/brendandburns/contrib/commit/cdccdfdafe0a537530433b7467a6de7ac9497512

PR #399, opened 2016-01-25
Refactor L7 modules and use a single instance group

bprashanth:
First commit is part of an overdue refactor, second adds all ports to a single IG.


bprashanth:
closes https://github.com/kubernetes/contrib/issues/367

thockin:
LGTM

PR #400, opened 2016-01-25
[WIP] Use unbound as caching DNS resolver

aledbf:
This two gist contains the benchmark between current implementation (skydns) and unbound
- https://gist.github.com/aledbf/bc38682df4e0d0daf4c3
- https://gist.github.com/aledbf/c8212143b88b1aa45a35

From this benchmark I see some interesting differences:
- skydns latency is 5 times the latency of unbound
- skydns uses 100% of the CPU (unbound just 34%) during the test
- ~~skydns loses 7.69% of the requests~~  https://github.com/kubernetes/contrib/pull/400#issuecomment-175382313
- skydns QPS is 4 times lower than unbound


aledbf:
ping @bprashanth

bprashanth:
Thanks for the benchmark! @thockin @ArtfulCoder fyi

> skydns loses 7.69% of the requests

Eww, do we know why?

dns-resolver/rootfs/etc/unbound/unbound.tmpl
```diff
@@ -0,0 +1,111 @@
...
+  so-reuseport: yes
+
+  statistics-interval: 0
+  extended-statistics: yes
+  # set to yes if graphing tool needs it
+  statistics-cumulative: no 
+        
+forward-zone:
+      name: "#DNS_DOMAIN#"
+      forward-addr: #DNS_SERVICE_IP#
```
bprashanth: don't you just need to hit localhost:54 for this? or have i misunderstood

ArtfulCoder:
I had profiled DNS performance and documented my findings here: https://github.com/kubernetes/kubernetes/issues/14394

Before we get into the results, could you elaborate on your setup.

1. Basic question: did you verify every dns request from your data.txt and ensure that you get a response.

2.  Can you run the perf by eliminating skydns and the cluster-specific dns queries ? 
Sometimes, skydns is just being the messenger and if it gets a failure from the recursed dns-server, it propagates that to the perf test. 

3. Were you hitting skydns using a cluster-ip or were you hitting the actual pod ip of skydns ? If you used the cluster-ip, did you use kube-proxy in iptables mode or userspace mode ?

4. Are you using the latest skydns version ?

ArtfulCoder:
5. Can you run the perf test for cluster-dns queries only seperately, then a perf test for non-cluster-dns queries only ?

aledbf:
Before we get into the results, could you elaborate on your setup.

> Basic question: did you verify every dns request from your data.txt and ensure that you get a response.

yes. I used one of the tests in kubernetes/kubernetes#14394 to create data.txt.

> Can you run the perf by eliminating skydns and the cluster-specific dns queries ? 
Sometimes, skydns is just being the messenger and if it gets a failure from the recursed dns-server, it propagates that to the perf test.

ok, I will create a different scripts to simulate this conditions.

> Were you hitting skydns using a cluster-ip or were you hitting the actual pod ip of skydns ? If you used the cluster-ip, did you use kube-proxy in iptables mode or userspace mode ?

127.0.0.1:54. No kube-proxy involved. I'm running dnsperf from inside the pod

> Are you using the latest skydns version ?

Yes, same version located in current dns addon (`2015-10-13-8c72f8c`)

aledbf:
@ArtfulCoder here is the comparison. Now it runs two different test, one with internal and other with external hosts

https://gist.github.com/aledbf/06a2679356840f730306#file-unbound-md
https://gist.github.com/aledbf/06a2679356840f730306#file-skydns-md

2 - 9jt1qbm
1 - 5un2wjt

thockin:
Thanks.  At least this shows no real loss of queries :)

That a cache is faster is not very surprising - SkyDNS does not actively cache, if I recall.  3x faster is surprising, but not unrealistic.  The real question is whether this matters and what we should do with this info.  I'll review the PR.

dns-resolver/README.md
```diff
@@ -0,0 +1,20 @@
+# dns-resolver
+==============
+
+This module is a replacement of the [dns addon](https://github.com/kubernetes/kubernetes/tree/master/cluster/addons/dns). The principal difference is the use of [unbound](https://www.unbound.net) as caching DNS resolver that forwards queries for the kubernetes domain to skydns and uses an external DNS server for other queries (like in skydns). One interesting feature of unbound is the extraction of [statistics](https://www.unbound.net/documentation/howto_statistics.html). The image built with this module uses [muning](https://github.com/munin-monitoring/munin) and a web server to access this stats in a graphical way.
```
thockin: s/muning/munin

thockin: How is it a replacement for skydns if it forwards to skydns?

dns-resolver/rootfs/build.sh
```diff
@@ -0,0 +1,92 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+export DEBIAN_FRONTEND=noninteractive
+
+apt-get update && apt-get install -y --no-install-recommends \
```
thockin: You really should not be doing this at run time.  First, it's very slow, so any restarts will be extended outages.  Second it's the opposite of hermetic.  You should do this at build time and commit the image that results.

aledbf: > You really should not be doing this at run time.

this script is only executed during `docker build`
```
$ kubectl logs -f kube-dns-v12-5vums -c unbound --namespace=kube-system
[s6-init] making user provided files available at /var/run/s6/etc...exited 0.
[s6-init] ensuring user provided files have correct perms...exited 0.
[fix-attrs.d] applying ownership & permissions fixes...
[fix-attrs.d] done.
[cont-init.d] executing container initialization scripts...
[cont-init.d] done.
[services.d] starting services
[services.d] done.
```

dns-resolver/README.md
aledbf: > How is it a replacement for skydns if it forwards to skydns?

In this PR I forward the queries for the local domain
(I'm adding a second commit that replaces skydns and etcd containers)

thockin:
This is a pretty expensive way to get performance, and it's not clear that performance matters.  Wouldn't it be much cheaper to teach skydns to cache (or cache "better")?  It looks like some caching features were added - we should rebase our image.

That said, it's a good approach, but I think your build is pretty fundamentally broken.  You can't fetch and compile at runtime, and we don't want an image with make and GCC and that stuff in it - too huge!  It's interesting to have an optional "more expensive but faster" option for people who care.  Another alternative might be to make it act as a machine-local cache of the main DNS service, but I think that will need some other work (which we want to do anyway).

The munin looks cute, but it's more dependencies and more memory - is it really worthwhile?

aledbf:
> The munin looks cute,

the reason for that interface is to get information about the dns usage. Right now is not possible to get this information using skydns.

> That said, it's a good approach, but I think your build is pretty fundamentally broken. 

I will change the approach.

@thockin thank you for the review.

dns-resolver/rootfs/build.sh
thockin: doh, I saw RUN and read CMD, sorry.

dns-resolver/README.md
thockin: Replaces skydns?  I'll just wait to see that, I guess :)

thockin:
I hope we'll see another variant of this?

aledbf:
> I hope we'll see another variant of this?

Absolutely. The new commit replaces debian with alpine to reduce the size of the image.
Now is half of the size.

ArtfulCoder:
when we go etcd-less, we should see a perf jump for internal-only queries even without unbound.

The unbound effort would continue to help with external queries.

Most of the external queries are tried with internal domain suffixes first, And would get nxdomain response faster in a etcd-less world.

aledbf:
> when we go etcd-less, ...

The next in this PR will contain a binary similar to `kube2sky` with the support of different backend dns servers: skydns and nsd for now.

dns-resolver/rootfs/etc/unbound/unbound.tmpl
```diff
@@ -0,0 +1,114 @@
...
+  # set to yes if graphing tool needs it
+  statistics-cumulative: no 
+        
+forward-zone:
+  name: "#DNS_DOMAIN#."
+  forward-addr: 127.0.0.1@54
+
+forward-zone:
+  name: "."
+  forward-addr: 8.8.8.8
```
ArtfulCoder: this would have to be set at runtime and must match the nameservers from the resolv.conf of the host

dns-resolver/rootfs/etc/unbound/unbound.tmpl
```diff
@@ -0,0 +1,114 @@
...
+
+  # Faster UDP with multithreading
+  so-reuseport: yes
+
+  statistics-interval: 0
+  extended-statistics: yes
+  # set to yes if graphing tool needs it
+  statistics-cumulative: no 
+        
+forward-zone:
```
ArtfulCoder: what happens if a forward-zone is unavailable ? 
i.e. unbound gets time-out on the first forward-zone. Does it fallback to the second one which has the catch-all "." ?

dns-resolver/rootfs/etc/unbound/unbound.tmpl
```diff
@@ -0,0 +1,114 @@
...
+  # http://tools.ietf.org/html/draft-vixie-dnsext-dns0x20-00
+  # While upper and lower case letters are allowed in domain names, no significance
+  # is attached to the case. That is, two names with the same spelling but
+  # different case are to be treated as if identical. This means calomel.org is the
+  # same as CaLoMeL.Org which is the same as CALOMEL.ORG.
+  use-caps-for-id: yes
+
+  # the time to live (TTL) value lower bound, in seconds. Default 0.
+  # If more than an hour could easily give trouble due to stale data.
+  cache-min-ttl: 3600
```
ArtfulCoder: Can we set a per-forward-one ttl ?
1 hour could mean that we are still serving DNS A records, which have been long deleted..

ArtfulCoder:
1. Is it possible to trigger a cache invalidate on unbound process, via a signal ?

2. After we adjust the timeout, we should do the perf test again and verify the gains (especially if we go down from 3600 to say, 5-10)

dns-resolver/rootfs/etc/unbound/unbound.tmpl
aledbf: If the cache contains the host from the query (not expired) it will return the content from the cache. Any other case the query is forwarded to the "." zone.

dns-resolver/rootfs/etc/unbound/unbound.tmpl
aledbf: Yes. I will update the value to the same value than skydns

aledbf:
> Is it possible to trigger a cache invalidate on unbound process, via a signal ?

Yes: `unbound-control reload` (or kill -HUP)

> After we adjust the timeout, we should do the perf test again and verify the gains (especially if we go down from 3600 to say, 5-10)

ok

> Tim asked unbound respects the TTL on the response it gets from the forward zone.

Yes if `cache-min-ttl` is set to 0.

dns-resolver/rootfs/etc/unbound/unbound.tmpl
ArtfulCoder: I think we should set the TTL to be 0 so that the TTL on the response is always respected.

dns-resolver/rootfs/etc/unbound/unbound.tmpl
ArtfulCoder: I think you missed my question, or I didnt follow your answer..
we set up two forward-zones
we have a cache miss
the first zone timesout.
does unbound go to the second zone ?

ArtfulCoder:
I think the correct comparison of numbers would be with setting cache-min-ttl to 0 instead of 3600...
I still expect unbound to come with better results..

aledbf:
@ArtfulCoder this is the result with the last commit:
- no skydns or etcd
- unbound as cache
- nsd as authoritative dns server
- cache-min-ttl

```
./dns-test.sh 53 internal
...

[Status] Command line: dnsperf -s 127.0.0.1 -d ./dns.txt -l 300 -c 1000 -p 53
[Status] Sending queries (to 127.0.0.1)
[Status] Started at: Wed Feb  3 17:15:47 2016
[Status] Stopping after 300.000000 seconds
[Status] Testing complete (time limit)

Statistics:

  Queries sent:         14054866
  Queries completed:    14054866 (100.00%)
  Queries lost:         0 (0.00%)

  Response codes:       NOERROR 14054866 (100.00%)
  Average packet size:  request 55, response 71
  Run time (s):         300.000504
  Queries per second:   46849.474626

  Average Latency (s):  0.000650 (min 0.000000, max 4154504685.551608)
  Latency StdDev (s):   0.001186
```

```
./dns-test.sh 53 external
...

[Status] Command line: dnsperf -s 127.0.0.1 -d ./dns.txt -l 300 -c 1000 -p 53
[Status] Sending queries (to 127.0.0.1)
[Status] Started at: Wed Feb  3 17:21:18 2016
[Status] Stopping after 300.000000 seconds
[Status] Testing complete (time limit)

Statistics:

  Queries sent:         14885021
  Queries completed:    14885021 (100.00%)
  Queries lost:         0 (0.00%)

  Response codes:       NOERROR 14885021 (100.00%)
  Average packet size:  request 36, response 87
  Run time (s):         300.007994
  Queries per second:   49615.414581

  Average Latency (s):  0.000599 (min 0.000000, max 4154504685.551615)
  Latency StdDev (s):   0.001021
```

aledbf:
> Is it possible to trigger a cache invalidate on unbound process, via a signal ?

If a pod/service is removed the record are removed from unbound

Issue #395, opened 2016-01-21
TASK: [kubernetes-addons] fails for heapster-controller.yaml.j2


jojimt:
Seeing this error since the last hour or so:
TASK: [kubernetes-addons | MONITORING | Download monitoring files from Kubernetes repo] **\* 
ok: [k8master -> 127.0.0.1] => (item=grafana-service.yaml)
changed: [k8master -> 127.0.0.1] => (item=heapster-controller.yaml)
ok: [k8master -> 127.0.0.1] => (item=heapster-service.yaml)
ok: [k8master -> 127.0.0.1] => (item=influxdb-grafana-controller.yaml)
ok: [k8master -> 127.0.0.1] => (item=influxdb-service.yaml)

TASK: [kubernetes-addons | MONITORING | Convert pillar vars to ansible vars for monitoring files] **\* 
ok: [k8master -> 127.0.0.1] => (item=grafana-service.yaml)
changed: [k8master -> 127.0.0.1] => (item=heapster-controller.yaml)
ok: [k8master -> 127.0.0.1] => (item=heapster-service.yaml)
ok: [k8master -> 127.0.0.1] => (item=influxdb-grafana-controller.yaml)
ok: [k8master -> 127.0.0.1] => (item=influxdb-service.yaml)

TASK: [kubernetes-addons | MONITORING | Install template from converted saltfile] **\* 
ok: [k8master] => (item=grafana-service.yaml)
fatal: [k8master] => {'msg': "AnsibleError: file: /tmp/kubernetes/addons/cluster-monitoring/heapster-controller.yaml.j2, line number: 2, error: no test named 'None'", 'failed': True}
fatal: [k8master] => {'msg': 'One or more items failed.', 'failed': True, 'changed': False, 'results': [{'group': 'root', 'uid': 0, 'changed': False, 'owner': 'root', 'item': 'grafana-service.yaml', 'state': 'file', 'gid': 0, 'secontext': 'system_u:object_r:etc_t:s0', 'mode': '0755', 'invocation': {'module_name': 'template', 'module_args': ''}, 'path': '/etc/kubernetes/addons/cluster-monitoring/grafana-service.yaml', 'size': 431}, {'msg': "AnsibleError: file: /tmp/kubernetes/addons/cluster-monitoring/heapster-controller.yaml.j2, line number: 2, error: no test named 'None'", 'failed': True}]}

FATAL: all hosts have already failed -- aborting

Any ideas/suggestions?

[...]

PR #402, opened 2016-01-26
Resolved ansible errors in cluster-monitoring setup

patrickeasters:
Due to changes made in commit [3da8d80](https://github.com/kubernetes/kubernetes/commit/3da8d80187b12db4e505bd2be5ef41dc819f0506]) in the kubernetes repo, this ansible playbook fails when trying to convert the pillar vars to ansible vars. This resolves that by changing the regex to match the new logic in the template.

I tested this in my lab and am able to setup cluster monitoring via the ansible role successfully now.

This also is reported in issue [395](https://github.com/kubernetes/contrib/issues/395)

### DIFF
--- a/ansible/roles/kubernetes-addons/tasks/cluster-monitoring.yml
+++ b/ansible/roles/kubernetes-addons/tasks/cluster-monitoring.yml
@@ -21,11 +21,25 @@
     - influxdb-grafana-controller.yaml
     - influxdb-service.yaml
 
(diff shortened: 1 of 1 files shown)


patrickeasters:
I signed it!

Sorry, realized I had the wrong email set on my commit.

eparis:
I'm going to blindly believe that it works, because 'not working' is better than it is today   :)

PR #405, opened 2016-01-26
[mungegithub] Helpers to list and delete comments

eparis:


brendandburns:
@eparis unit tests are failing...

PR #408, opened 2016-01-27
[submit queue] Order by priority label then by PR number

eparis:


mungegithub/mungers/submit-queue.go
```diff
@@ -520,15 +520,41 @@ func (sq *SubmitQueue) flushGithubE2EQueue(reason string) {
...
+
+	aPrio := a.Priority()
+	bPrio := b.Priority()
+	if aPrio < bPrio {
+		return true
+	} else if aPrio > bPrio {
+		return false
+	}
+
+	if *a.Issue.Number < *b.Issue.Number {
```
yujuhong: nit: return `*a.Issue.Number < *b.Issue.Number`

mungegithub/github/github.go
```diff
@@ -539,6 +540,26 @@ func (obj *MungeObject) RemoveLabel(label string) error {
 	return nil
 }
 
+// Priority returns the priority an issue was labeled with.
+// The labels must take the form 'priority/P?[0-9]+'
+// or math.MaxInt32 if unset
+func (obj *MungeObject) Priority() int {
+	priority := math.MaxInt32
```
yujuhong: I was thinking whether a default priority such as p2 makes sense. This spares most people from applying priority labels explicitly.

pmorie: +1

mungegithub/mungers/submit-queue_test.go
```diff
@@ -140,6 +141,82 @@ func FailJenkins() jenkins.Job {
...
+			name: "With a priority label",
+			issues: []github.Issue{
+				*github_test.Issue(whitelistUser, 2, []string{"priority/P1"}, true),
+				*github_test.Issue(whitelistUser, 3, []string{"priority/P1"}, true),
+				*github_test.Issue(whitelistUser, 4, []string{"priority/P0"}, true),
+				*github_test.Issue(whitelistUser, 5, nil, true),
+			},
+			expected: []int{4, 2, 3, 5},
+		},
+	}
```
pmorie: add some tests for trash labels

pmorie: add tests for p2/p3

mungegithub/github/github.go
eparis: I really don't like that idea.  I don't want to punish someone for taking the time to approprietely prioritize something. I'd rather 'punish' the people who are lazy and provide no transparency...

eparis:
Added more tests.  Decided to put 'unlabeled' as P3.  Not P2.

PR #409, opened 2016-01-27
Add Jimmi Dyson to the whitelist.

vishh:


eparis:
He used to have commit and was removed when his commit privs were removed. Next time I restart submit-queue we'll get picked back up.

vishh:
Ohh ok :)

PR #412, opened 2016-01-28
[submit queue] fix flaky test

eparis:


eparis:
don't care for review. I fixed it!

PR #413, opened 2016-01-28
[submit queue] Add comment when PR requires rebase

pmorie:
@eparis

### DIFF
--- a/mungegithub/mungers/needs_rebase.go
+++ b/mungegithub/mungers/needs_rebase.go
@@ -61,5 +61,10 @@ func (NeedsRebaseMunger) Munge(obj *github.MungeObject) {
 	}
 	if !mergeable && !obj.HasLabel(needsRebase) {
 		obj.AddLabels([]string{needsRebase})
+
+		body := "PR needs rebase"
+		if err := obj.WriteComment(body); err != nil {
+			return
+		}
 	}
 }


mungegithub/mungers/needs_rebase.go
```diff
@@ -60,6 +60,12 @@ func (NeedsRebaseMunger) Munge(obj *github.MungeObject) {
 		obj.RemoveLabel(needsRebase)
 	}
 	if !mergeable && !obj.HasLabel(needsRebase) {
+		glog.Infof("PR: %d needs rebase", *obj.Issue.Number)
```
eparis: I'd rather not spam the logs with this, since as the person looking at the logs for problems, this doesn't tell me anything.

pmorie: Fine, you monster

eparis:
What would you think about a (future) PR that removes comments the bot made which no longer make sense?

pmorie:
I'm cool with that.

eparis:
@spxtr @thockin 

you two have been the most interested in e-mail spam....
yay/nay?

pmorie:
My rationale for this is that you don't get an email when a label is added.  It would help me to get these, as there have been a couple times where a PR needed rebase and I didn't catch it until later.  I would be cool with driving this behavior off a label.

spxtr:
If it's not too much of a hassle, yay.

PR #416, opened 2016-01-28
[submit queue] Leave queue intact when e2e tests are failing

eparis:
This means that you can always see the queue in the web interface, we'll
always merge in order, etc.

You'll also get a single message in the history when the queue
breaks/repairs, instead of a flood....


eparis:
@pmorie review me already!

PR #425, opened 2016-01-29
[mungegithub] Include the #IGNORE when asking merge bot to retest

eparis:
We know what we are doing and never do it because of a flake.


eparis:
@brendandburns @ixdy  ?

ixdy:
I guess this is fine, though it feels pretty kludgy.

eparis:
we can also just make the k8s-bot not enforce the issue requirement for k8s-merge-bot....

PR #445, opened 2016-02-02
[mungegithub] Properly ignore retest requests from robots

eparis:
We had a loop to test if the commit was from a robot. But on finding a
message from a robot we would break that test loop instead of the main
loop. So we would still remove the comment.

Convert to kubernetes sets to make it easy to read/understand.


eparis:
@brendandburns

PR #446, opened 2016-02-02
[mungegithub] Enable the retest request without issue report complainer

eparis:


eparis:
@brendandburns likely to push it live tomorrow...

eparis:
I went through today and deleted all of the 'bad' retest request comments. (Without removing LGTM)

PR #447, opened 2016-02-02
[mungegithub] retest munger complaint message update for clarity

eparis:


eparis:
@brendandburns @lavalamp 
https://github.com/kubernetes/kubernetes/pull/19874#issuecomment-178345665
Is the current message.
- it isn't clear what the user did wrong
- the \<number\> is messed up because markdown eats the <>
- it was showing all test flakes, not just open test flakes.

lavalamp:
LGTM

PR #449, opened 2016-02-02
[submit queue] Add jsafrane to submit queue whitelist

eparis:


eparis:
@jsafrane

eparis:
Jan works for RH on our storage team and having to add 'ok-to-merge' is just a PITA.

bgrant0607:
@eparis Did you regenerate the list, or just add this by hand?

eparis:
This one was by hand. the list is generated with a combination of those people whom have explicit read/write access to the repo and some people who are manually added to the list.

It would be a lot easier for us if we could give 'read' access to the main kube repo to certain teams, then we at Red Hat wouldn't have to manage as many places, and that might be a better solution than adding Jan manually....

Or maybe I should just add people to both the RH team and to the contributors team...

bgrant0607:
I thought the list was generated from kubernetes-collaborators?

bgrant0607:
BTW, we're moving towards eliminating the goog-* github teams, in favor of sig and component teams (e.g., kubectl).

PR #450, opened 2016-02-02
[submit queue] Refresh issues in the submit queue

eparis:
This actually has 2 side effects.
- If LGTM was removed from an issue after it was queued it could still merge
- If labels which affected the queue position were changed after it was queued the position would not be updated
    
Historically #1 wasn't a huge problem since the queue was constantly cleared by e2e flakes/failures and so we got fresh data. Now we aren't clearing the queue constantly the issue data could get way out of date.

### DIFF
--- a/mungegithub/Dockerfile
+++ b/mungegithub/Dockerfile
@@ -16,15 +16,15 @@ FROM google/debian:wheezy
 MAINTAINER Brendan Burns <bburns@google.com>
 RUN apt-get update
 RUN apt-get install -y -qq ca-certificates
-ADD mungegithub /mungegithub
-ADD blunderbuss.yml /blunderbuss.yml
 ADD path-label.txt /path-label.txt
 ADD generated-files.txt /generated-files.txt
+ADD blunderbuss.yml /blunderbuss.yml
 # User lists for submit-queue and 'needs-ok-to-merge'
 ADD committers.txt /committers.txt
 ADD whitelist.txt /whitelist.txt
 # Submit queue web interface
 ADD www /www
+ADD mungegithub /mungegithub
 EXPOSE 8080
 ENTRYPOINT ["/mungegithub"]
 CMD ["--dry-run", "--token-file=/token"]
--- a/mungegithub/github/github.go
+++ b/mungegithub/github/github.go
@@ -377,12 +377,50 @@ func (config *Config) SetClient(client *github.Client) {
 	config.client = client
 }
 
-// GetObject will return an object (with only the issue filled in)
-func (config *Config) GetObject(num int) (*MungeObject, error) {
+func (config *Config) getPR(num int) (*github.PullRequest, error) {
+	pr, response, err := config.client.PullRequests.Get(config.Org, config.Project, num)
+	config.analytics.GetPR.Call(config, response)
+	if err != nil {
+		glog.Errorf("Error getting PR# %d: %v", num, err)
+		return nil, err
+	}
+	return pr, nil
+}
+
+func (config *Config) getIssue(num int) (*github.Issue, error) {
 	issue, resp, err := config.client.Issues.Get(config.Org, config.Project, num)
 	config.analytics.GetIssue.Call(config, resp)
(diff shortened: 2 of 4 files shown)


mungegithub/Dockerfile
```diff
@@ -16,15 +16,15 @@ FROM google/debian:wheezy
...
-ADD blunderbuss.yml /blunderbuss.yml
 ADD path-label.txt /path-label.txt
 ADD generated-files.txt /generated-files.txt
+ADD blunderbuss.yml /blunderbuss.yml
 # User lists for submit-queue and 'needs-ok-to-merge'
 ADD committers.txt /committers.txt
 ADD whitelist.txt /whitelist.txt
 # Submit queue web interface
 ADD www /www
+ADD mungegithub /mungegithub
```
pmorie: why did this change?

eparis: I guess it slipped in, it means that I can build the container over and over and over changing only the binary and docker will cache all of the other stuff.  Putting 'mungegithub' at the top meant that it had to add /www when the mungegithub changed and it was slowing me down, man.

pmorie: Ah

mungegithub/github/github.go
```diff
@@ -377,12 +377,50 @@ func (config *Config) SetClient(client *github.Client) {
...
 	config.analytics.GetIssue.Call(config, resp)
 	if err != nil {
-		glog.Errorf("GetObject: %v", err)
+		glog.Errorf("getIssue: %v", err)
+		return nil, err
+	}
+	return issue, nil
+}
+
+// RefreshIssue will refresh the Issue (and PR if this is a PR)
```
pmorie: nit: Refresh

mungegithub/github/github.go
```diff
@@ -932,7 +956,12 @@ func (obj *MungeObject) GetPR() (*github.PullRequest, error) {
 	if !obj.IsPR() {
 		return nil, fmt.Errorf("Issue: %d is not a PR", *obj.Issue.Number)
 	}
-	return obj.RefreshPR()
+	pr, err := obj.config.getPR(*obj.Issue.Number)
```
pmorie: this is the same as the part of `Refresh

mungegithub/github/github.go
```diff
@@ -1145,7 +1174,12 @@ func (obj *MungeObject) IsMergeable() (bool, error) {
 		glog.V(4).Infof("Waiting for mergeability on %q %d", *pr.Title, *pr.Number)
 		// TODO: determine what a good empirical setting for this is.
 		time.Sleep(2 * time.Second)
-		pr, err = obj.RefreshPR()
+		err := obj.Refresh()
+		if err != nil {
+			glog.Errorf("Unable to refresh PR# %d: %v", prNum, err)
+			return false, err
+		}
+		pr, err = obj.GetPR()
```
pmorie: Do you still need the distinct check for PRs?

mungegithub/mungers/submit-queue.go
```diff
@@ -506,9 +506,13 @@ func (sq *SubmitQueue) Munge(obj *github.MungeObject) {
...
 	sq.Lock()
 	if _, ok := sq.githubE2EQueue[*obj.Issue.Number]; !ok {
-		sq.githubE2EQueue[*obj.Issue.Number] = obj
 		added = true
 	}
+	// Add this most-recent object in place of the existing object. It will
+	// have more up2date information. Even though we explicitly refresh the
+	// PR information before do anything with it, this allow things like the
+	// queue order to change dynamically as labels are added/removed.
+	sq.githubE2EQueue[*obj.Issue.Number] = obj
```
pmorie: Add a test that should fail if you use the old value?

mungegithub/github/github.go
eparis: Yes. So the next caller to GetPR() won't hit the Github API server again. Only an explicit call to Refresh() should cause us to hit it again.

PR #456, opened 2016-02-03
Ansible: Allow ability to install from Docker yum repos

### LINKED ISSUE #455: Ansible: Allow for use of Docker yum repositories
This will allow for newer versions of Docker engine to be installed (CentOS 7 only has v1.8 in the repos). I'll add a pull request for this functionality.

rsmitty:
Fixes #455. This will allow for installation of newer versions of docker engine by editing the group_var for docker provider.

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/456)


ansible/roles/docker/tasks/main.yml
```diff
@@ -5,6 +5,9 @@
 - include: apt-docker-install.yml
   when: ansible_distribution == "Ubuntu"
 
+- include: use-docker-repos.yml
+  when: docker_provider == "docker"
```
eparis: Shouldn't this be gated on ansible_pkg_mgr == yum or dnf? I don't want to run this on a apt/dpkg based system. Update that and I will merge.

eparis:
https://github.com/kubernetes/contrib/pull/291 is also playing in this space. I requested they use a 'docker source type' to determine if they should do what they are doing. Hopefully they will follow your lead, but if they address comments first this will need a rebase.

ansible/roles/docker/tasks/main.yml
ingvagabund: I would prefer ``docker_source_type == custom_repository``. As currently, there are ``etcd_source_type``, ``flannel_source_type`` and upcoming ``kube_source_type``, it make sense to continue in the naming schema. At the same time, there are various ways how to install a component. Such as:

* package manager
* github release
* local build
* distribution rpm
* custom repository

I have already a usecase for choosing the custom repository. So this would work for both of us. At the same time it would be beneficial to make the repository file configurable. E.g.

```ini
[dockerrepo]
name=Docker Repository
baseurl={{ repository_baseurl }}
enabled=1
gpgcheck={% if repository_gpgkey == "" %}0{% else %}1{% endif %}
gpgkey={{ repository_gpgkey }}
```

```yaml
# Docker custom repository
repository_baseurl: https://yum.dockerproject.org/repo/main/centos/$releasever/
repository_gpgkey: https://yum.dockerproject.org/gpg
```

ingvagabund:
Implemented in https://github.com/kubernetes/contrib/pull/1610

Issue #458, opened 2016-02-03
Ansible: Support OverlayFS for Docker Storage


rsmitty:
Upon continuing to try deploying against CentOS 7 hosts, I've encountered another error which does not pop up when using Atomic. When the Docker service is launched for the first time the "docker-storage-setup" command is run. This command fails because it attempts to use the devicemapper driver and LVM to create a Docker volume group. The CentOS cloud image (in OpenStack and I presume other clouds) does not use LVM for the root volume and is just a single ext4 partition at /. 

However, this can be worked around by using the OverlayFS driver instead and seems to be what is actually the default on the Atomic OSes. An interesting writeup can be found [here](http://www.projectatomic.io/blog/2015/06/notes-on-fedora-centos-and-docker-storage-drivers/).

I'll submit a pull request with a change for this.

PR #459, opened 2016-02-03
support overlayfs for non-atomic hosts

rsmitty:
This will resolve issue #458. I'm also willing to add a group_var for "docker_storage_driver" or similar if that's something folks think would be useful as part of this. But it seems like this would be a required change anyways for any Fedora or CentOS non-atomic image in OpenStack, and I presume any other cloud using the provided "Generic Cloud" images from CentOS/Fedora.

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/459)

### DIFF
--- a/ansible/roles/docker/files/docker-storage
+++ b/ansible/roles/docker/files/docker-storage
@@ -1,0 +1,1 @@
+DOCKER_STORAGE_OPTIONS= -s overlay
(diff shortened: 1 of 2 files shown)


eparis:
This is forcing the use of a non-POSIX compliant filesystem? nak.

I would take a PR which lays down a template where DOCKER_STORAGE_OPTIONS= is set by an ansible variable but ONLY if the variable is set. Care must be taken not to overwrite the storage options someone did outside of ansible.

test-foxish:
recomputing cla status...

grodrigues3:

- **ansible**

cc @bprashanth 

PR #464, opened 2016-02-05
Disable ping-ci until travis comes back

eparis:


eparis:
@mikedanese I'm not 100% sure that setting --required-context="" would work properly, and I need to get this fixed since I broke the bot...

mikedanese:
Ya it would create a list of length one with one empty string.

mikedanese:
LGTM

PR #465, opened 2016-02-05
Fix truncated password warning in keepalived

wienczny:
Fix truncated password warning in keepalived by slicing the password in kube-keeplived-vip


bprashanth:
@aledbf since he has more context here

aledbf:
Code LGTM

PR #467, opened 2016-02-07
Initial support for federated builder status

justinsb:
--jenkins-jobs can now include a `name=src` entry.

name is the name of the job.
src specifies the source for the output of builders

Limitations of initial implementation:
- only gs:// URLs are supported as src (Google Cloud Storage).
- all `name=src` entries are treated as non-gating, and
  will not contribute to the overall build status.

For example:
--jenkins-jobs=kubernetes-e2e-aws=gs://kubernetes-buildlogs-aws/logs/kubernetes-e2e-aws

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/467)


justinsb:
More for discussion at the moment, I believe.  But should work!
cc @ihmccreery

ihmccreery:
cc @kubernetes/goog-testing @kubernetes/sig-testing

justinsb:
Reworked this so that it is much cleaner: it now uses a set of JSON files in config/ which determine the builder jobs & GCS buckets to crawl.  Also in #520 I then parse the JUnit output, so we can see exactly which tests fail, amongst all sort of other goodies.

How can we get this actually running?  It would be great to be able to display AWS test results also...

eparis:
submit-queue_test.go doesn't build. To get it live? Get it committed and ask me, @lavalamp, @mikedanese to kick off a new submit queue. I'm out until tuesday, but if CI is clean I can turn it up then.

justinsb:
Will fix e2e.  I have an idea though... what would it take to bring up "submit-queue-alpha.k8s.io" or "submit-queue-preview.k8s.io" in dry-run mode?

justinsb:
Will fix tests - sorry. I have a though though... what would it take to bring up "submit-queue-alpha.k8s.io" or "submit-queue-preview.k8s.io" in dry-run mode?

This PR is really just a UI change (well, the status that the mergebot uses for decision making is also reflected in the UI I believe, so we can test it in dry-run mode). And we likely have some other UI changes to come, because it isn't clear exactly how to display federated builds. Or the names of the failing tests.

What I'm thinking is that we could have a preview environment that automatically deploys from master. We can then experiment with UI changes, speed-ups etc, without risking breaking the merge queue.

I am happy to work on this... would it be a Jenkins job? (Do we have a jenkins job that does the deploy?)

lavalamp:
It's not hard at all.

justinsb:
Sorry @lavalamp - I hit "Close and comment" in the middle of writing my comment by mistake.  Not sure if you saw the more complete idea of auto-deploying the dry-run version from master?  Do you think that's a good idea, and what work can I do to make it happen (should I write a batch file we can call from Jenkins?)

lavalamp:
Oh, auto deploying in --dry-run might be interesting, but we might have concerns about running random code. @ixdy do you have any thoughts on that?

justinsb:
Only _merged_ random code :-)  I do think it would be exciting if we could figure this out... would set us down an interesting road of dogfooding!

eparis:
It needs to use a different secret/github-oauth token...

justinsb:
@eparis what if it didn't have a token at all...?

lavalamp:
It needs a token even just for dry-run. @eparis wants it separate so it
doesn't count against our ratelimit (I assume).

eparis:
It only gets 60 API calls per hour without a token. I don't think the code really runs at all without a valid token....

eparis:
And yes, I would want a second token for the dry-run version as even at steady state I doubt the 5000 API calls per hour is adequate...

gmarek:
@eparis - do we have any estimated time for this PR to get in (and others from this set)? If it's not soon I'll need to hack something temporary around

foxish:
What do we plan on doing with this PR? ping @justinsb @eparis

test-foxish:
recomputing cla status...

eparis:
closing as I don't think these are coming back in this form.

PR #468, opened 2016-02-07
GZIP responses

justinsb:
We might as well gzip the JSON responses; the history is 50KB+ and the
list of users is about 15KB.


eparis:
merged #593 instead.

PR #469, opened 2016-02-07
Use a RWLock for the submit queue

justinsb:
By doing this, we should be able to serve multiple readers concurrently.


justinsb:
On second thoughts, because we just buffer it into memory anyway, I don't think this matters in the slightest.  Closing.

PR #474, opened 2016-02-09
WIP: [mungegithub] OWNERS implementation

eparis:
This PR is missing some key usability features.
- There is no way to know who needs to approve a given PR
- There is no way for me, as an approver, to know what PRs I need to approve.

I believe that before we turn this on, we REALLY need to solve those usability issues.

### DIFF
--- a/mungegithub/mungers/owner-approval.go
+++ b/mungegithub/mungers/owner-approval.go
@@ -1,0 +1,126 @@
+/*
+Copyright 2015 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package mungers
+
+import (
+	"strings"
+
+	"k8s.io/contrib/mungegithub/features"
+	"k8s.io/contrib/mungegithub/github"
+	"k8s.io/kubernetes/pkg/util/sets"
+
+	"github.com/golang/glog"
+	"github.com/spf13/cobra"
+)
+
+const (
+	ownerApproval = "has-owner-approval"
+)
+
+// OwnerApproval it a github munger which labels PRs based on 'owner' approval.
+// Owners are listed in files (called OWNERS) in the git tree. An owner owns
+// everything beneath the directory. So a person who 'owns' the root directory
+// can approve everything
+type OwnerApproval struct {
+	features *features.Features
+}
+
+func init() {
+	RegisterMungerOrDie(&OwnerApproval{})
+}
+
+// Name is the name usable in --pr-mungers
+func (o *OwnerApproval) Name() string { return "owner-approval" }
+
+// RequiredFeatures specifies that we need the features which loads assignees and owners
+func (o *OwnerApproval) RequiredFeatures() []string { return []string{features.RepoFeatureName} }
+
+// Initialize does just that
+func (o *OwnerApproval) Initialize(config *github.Config, features *features.Features) error {
+	o.features = features
+	return nil
+}
+
+// EachLoop is called at the start of every munge loop
+func (o *OwnerApproval) EachLoop() error { return nil }
+
+// AddFlags will add any request flags to the cobra `cmd`
+func (o *OwnerApproval) AddFlags(cmd *cobra.Command, config *github.Config) {}
+
+// Munge is the workhorse the will actually make updates to the PR
+func (o *OwnerApproval) Munge(obj *github.MungeObject) {
+	if !obj.IsPR() {
+		return
+	}
+	commits, err := obj.GetCommits()
+	if err != nil {
+		return
+	}
+
+	neededApprovals := map[string]sets.String{}
+	for _, c := range commits {
+		for _, f := range c.Files {
+			file := *f.Filename
+			neededApprovals[file] = o.features.Repos.Owners(file)
+		}
+	}
+
+	lastModified := obj.LastModifiedTime()
+	if lastModified == nil {
+		return
+	}
+
+	comments, err := obj.ListComments()
+	if err != nil {
+		return
+	}
+
+	approvalsGiven := sets.String{}
+	for _, comment := range comments {
+		if lastModified.After(*comment.UpdatedAt) {
+			continue
+		}
+		lines := strings.Split(*comment.Body, "\n")
+		for _, line := range lines {
+			line = strings.TrimPrefix(line, "@k8s-merge-bot")
+			line = strings.TrimSpace(line)
+			line = strings.ToLower(line)
+			switch line {
+			case "i approve":
+				approvalsGiven.Insert(*comment.User.Login)
+			case "approved":
+				approvalsGiven.Insert(*comment.User.Login)
+			}
+		}
+	}
+
+	missingApprovals := sets.NewString()
+	for _, needed := range neededApprovals {
+		intersection := needed.Intersection(approvalsGiven)
+		if intersection.Len() != 0 {
+			// Someone who approved covered this area
+			continue
+		}
+		missingApprovals = missingApprovals.Union(needed)
+	}
+
+	if missingApprovals.Len() == 0 && !obj.HasLabel(ownerApproval) {
+		obj.AddLabel(ownerApproval)
+	} else if missingApprovals.Len() != 0 && obj.HasLabel(ownerApproval) {
+		obj.RemoveLabel(ownerApproval)
+	}
+}
--- a/mungegithub/mungers/owner-approval_test.go
+++ b/mungegithub/mungers/owner-approval_test.go
@@ -1,0 +1,180 @@
+/*
+Copyright 2015 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package mungers
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"runtime"
+	"strings"
+	"testing"
+	"time"
+
+	github_features "k8s.io/contrib/mungegithub/features"
+	github_util "k8s.io/contrib/mungegithub/github"
+	github_test "k8s.io/contrib/mungegithub/github/testing"
+	"k8s.io/kubernetes/pkg/util/sets"
+
+	"github.com/golang/glog"
+	"github.com/google/go-github/github"
+)
+
+var (
+	_ = fmt.Printf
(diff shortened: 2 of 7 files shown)


thockin:
Is this ready for review or targetted discussion?  can you nominate a reviewer?  I'm keen to get this moving, but I am also buried in PRs..

eparis:
This code is ready for review.  @pmorie for some reason has started looking more and more in this area.  But I see this as completely blocked on UI/usability and have no interest in seeing it merged until this owners stuff is discoverable and manageable by both PR authors and owners. And I don't know when I'll get around to playing with UI stuff myself....

pmorie:
> @pmorie for some reason has started looking more and more in this area

I think the reason was you started asking me for reviews :)

mungegithub/mungers/owner-approval.go
```diff
@@ -0,0 +1,225 @@
...
+
+	"github.com/golang/glog"
+	"github.com/spf13/cobra"
+)
+
+const (
+	ownerApproval = "has-owner-approval"
+)
+
+type assignmentConfig struct {
```
thockin: if I interpret correctly, this is the type that the OWNERS file is going to unmarshal into, yes?

Can we make it be a struct so we have room to grow?

```
type OwnersFile struct {
    Owners []string `json:owners`
 }
```

This then leaves me room for adding maintainers or other metadata.

thockin: I'm tired.  I just asked you for exactly what you did, modulo JSON tags.

thockin:
This could do with a short doc explaining the semantics.  We are sort of lacking in docs for all of the gh bot stuff..

bgrant0607:
What is the status of this?

bgrant0607:
What does this PR do?

eparis:
This PR adds or removes a label called `has-owner-approval` to every PR. It looks for `approved` or `I approve` on any line in any comment on the PR.

- If `bgrant` is in the OWNERS file in / his approval would be enough for any PR to get the label.

- If user `pkg-user` is in pkg/OWNERS
  user `pkg-api-user` is in pkg/api/OWNERS
  users `pkg-client1-user` and `pkg-client2-user` are in pkg/client/OWNERS

  and a PR changes `pkg/api/endpoints/util.go` and `pkg/client/typed/generated/core/v1/endpoints.go`

Any of the following combinations of people could approve the PR.
- `bgrant`
- `pkg-user`
- `pkg-api-user` and `pkg-client1-user`
- `pkg-api-user` and `pkg-client2-user`

We can EASILY make the submit queue require the 'has-owner-approval' label. If it gets added by hand, it will be removed.

------------------------------------------------

Now I am unwilling to merge, and I feel like I've written this 10 times, but can't find it, due to UI/UX/usability. We absolutely need 2 things before i think we should turn this on.

1) We need some way for an OWNER to know all of the PRs for which (s)he should be giving approval. I also think we need a way to know which of those you are the 'leaf' approver and which of those you are a 'trunk' approver. In above example I'd call bgrant a 'trunk' approver and pkg-api-user and pkg-client-user as 'lead' approvers. I don't want the bgrant list to have every single PR even though he can approve every single one. But I would want it to show every PR which touches say README.md, which has to be approved by a / OWNER.

2) We need some way for a PR Author to see the list of OWNERS who need to approve their PR. They should see the set of all 'leaf' approvers and should be able to see the 'trunk' approvers as well if they so choose. Without knowledge of who you are waiting on, processes become stuck. Yes, you can troll around in the repo and look at the OWNERS files, but if that something we should hoist on an unsuspecting community member?

Even if I had the time, I'm not sure I have the web design skills to put those 2 UIs together into anything reasonable, but without them, this whole process because quite opaque to developers and almost impossible for owners....

bgrant0607:
Thanks, @eparis.

The logic about who can approve makes sense to me -- approvers from all touched directories are required.

I think LGTM from an OWNER should count as approval. I don't think we need to introduce different text. Ideally, LGTM from OWNERs would apply the LGTM label automatically, as well. With this, I think we could get rid of "ok-to-merge", also: if the required OWNERS say it's good, just merge it already. Maybe the lgtm label could mean LGTM from someone trusted and has-owner-approval could be applied once all required owners have given LGTM.

I'd like to drive PR assignment (blunderbuss) from OWENRS, also. The blunderbuss list is currently too small, and many PRs languish with overloaded assignees without being reassigned. 

Maybe we should even automatically reassign any PR to which the assignee doesn't respond within some amount of time (a day, a week?). 

The auto-assigner could reassign from a pool of owners that hasn't yet approved in the case that multiple approvers are required, favoring leaf owners, in general.

We could maybe provide support to reassign to a trunk approver, using ESCALATE or somesuch. Manual reassignment should work, also, at least until the reassignment timeout.

For UX, we could post a message to the PR to describe the workflow and the escalation procedure, and list the leaf and trunk owners. The PR author could notify other owners or escalate if automatic reassignment wasn't working.

Distinguishing between partially lgtm and fully approved would be useful for listing PRs in limbo.

eparis:
`I'd like to drive PR assignment (blunderbuss) from OWENRS, also. The blunderbuss list is currently too small, and many PRs languish with overloaded assignees without being reassigned.`

I should probably do that separately. I intended that as an enhancement on this, but it is actually reasonable to do first and makes all of the code for this smaller/easier...

thockin:
Whats the status on this?  Is this PR still alive, or has it been supplanted by others?  I see OWNERS files in tree..

eparis:
This PR is still alive-ish. I don't know how to write a decent UI for OWNERS.  The OWNERS files in the kube tree only currently supports 'assignees'. Which is what the old blunderbuss uses. This PR expands that to also have an owners concept. But I don't want to see it merged until we have a good UI for both owners to know what they need to approve and for PR authors to know where to get approval.

bgrant had some ideas (auto-reassign the PR to another owner) but I think that gives a rather 'black box' view for everyone...

So the mechanics are here in this PR, but I don't know if I have to time to create a good story for both owners and authors. So I'd love someone else who has time to pick up the ball there...

bgrant0607:
Dashboard to view PRs that need to be approved, parameterized by approver.
and
List of de-duped approver lists posted to each PR.

bgrant0607:
For tests and docs it would be useful to be able to specify file-level ownership.

cc @hongchaodeng @lavalamp

lavalamp:
@bgrant0607 actually for tests, ownership of just parts of files is what we want. Or maybe a different mechanism. @fejta may be able to find people to do this. I'm about to send an email to the testing sig about this.

bgrant0607:
@lavalamp Ok, though I would like to be able to use whatever mechanism to automatically assign PRs and issues.

lavalamp:
Yes, the whole idea would be auto-flaky test issue assignment. Worst case,
we can split out tests owned by different people/groups into separate files.

asalkeld:
This might be a solution to the UI problem: http://gerrithub.io/

davidopp:
OK, it is my new mission in life to get some version of this merged ASAP. This feature has become pretty critical, and I think we should do the absolute minimum to get it merged.

First, @eparis , thank you very much for doing this.

The semantics you proposed look reasonable. 

I think the issue @bgrant0607 brought up about reusing LGTM is an interesting one. A scenario we need to consider is when someone wants to indicate that the PR looks fine to them, but they want someone else to also do a review (and thus it should not merge until the other person has looked at it). Under Brian's suggestion, if the first reviewer is also a valid approver, then they cannot use the phrase "LGTM" to indicate their satisfaction with the PR, because that will trigger a premature merge. There are obviously two solutions here:
(1) do what Brian suggested, where LGTM really means approval, and the first reviewer in my scenario would use some English phrasing that did not contain LGTM in it
(2) use your original suggestion, where the meaning of LGTM does not change, and we introduce a new phrase for approval
I'm not sure which one is better. I don't really object to Brian's suggestion, but wanted to at least mention the possible downside.

I think we should get rid of ok-to-merge completely. I don't think it has ever been useful. (I think it was intended to denote "I didn't check this PR's logic, but the PR at least seems non-malicious" or something like that.)

I think we should also get rid of the LGTM label. The useful label will be `has-owner-approval` once this feature is implemented.

Regarding the usability enhancements: I disagree that either of them is needed for the initial version of this feature. I think we should make it the PR assignee's responsibility to shepherd the PRs assigned to them through the process. So for example, if they are not a suitable approver, then they are responsible for reassigning the PR to a suitable approver (either instead of, or after, they have done a review). If this is done properly, then the user will be "in the loop" and doesn't need to understand the full policy (though I certainly don't object to having the bot automatically add a link to the policy when a PR is created). I am not a big fan of Brian's suggestion "we could post a message to the PR to describe the workflow and the escalation procedure, and list the leaf and trunk owners." The thing I worry about is that users will see that and immediately @-mention everyone in that (presumably very long) list, which will make being @-mentioned a completely useless signal, which would be very bad. Also it will go stale unless we edit it each time OWNERS files change.

To be clear, I agree it would be useful to have tools to show approvers a list of PRs that they could approve, and (somewhat less so, since it's not hard to figure out once you know the policy) to show PR authors who is in the candidate approver pool. But I think that, for now, it's OK to lean on the PR's assignee to shepherd the PR through the approval process (and to reassign when they can't). Now that the auto-assigner is integrated with OWNERs this seems reasonable.

WDYT?

bgrant0607:
See also #869

bgrant0607:
There's discussion about ok-to-merge in https://github.com/kubernetes/kubernetes/issues/24228

davidopp:
ping @eparis -- we'd really like to get this in... Could you comment on my last comment in this issue?

bgrant0607:
@davidopp I'm also in favor of doing MVP, and iterating from there.

I think the minimum is
* Add approvers to OWNERS files
* Require approvers of all directories touched to approve -- "I approve" is fine. We can change it later if desired 
* Apply owners-approve label when all have approved 
* Gate submit queue on approval instead of lgtm
* Eliminate lgtm label

As a separate task, someone can work on assigning approvers after the initial reviewer comments lgtm.

fejta:
What causes the PR to wait for a review if it is authored by an owner?

bgrant0607:
@fejta Honor system? Nothing stops anyone with write access from applying `lgtm` to their own PRs today. An author shouldn't comment "I approve" on their own PR. That seems pretty easy to enforce, though.

fejta:
If I'm an owner I expect to be able to send the review to a non owner and
get it merged after they lgtm... but I guess in this system either I'll
have to self approve or else find another owner to do the review?

bgrant0607:
@fejta 

Previous workflow:
* a reviewer, who didn't necessarily have write access (and therefore label access), would be assigned
* the reviewer would eventually type "LGTM" into a comment
* the author would (often) squash commits and/or rebase
* the reviewer would then apply the `lgtm` label if they had label access, or would notify someone in kubernetes-maintainers to do that if they did not; the author could self-apply `lgtm` if they had label access

Proposed workflow:
* a reviewer, who isn't necessarily an approver, would be assigned
* the reviewer would eventually type "LGTM" into a comment
* the author would (often) squash commits and/or rebase
* the reviewer would then type `I approve` if they were an approver, or would notify an approver to do that if they were not; the author could self-approve if they were an approver

It's pretty much the same, except that it bypasses github ACLs.

We've already taken direct merge powers away from most people in kubernetes-maintainers.

fejta:
New flow sgtm. Thanks!

davidopp:
@eparis What do you think of @bgrant0607 's suggestions here
https://github.com/kubernetes/contrib/pull/474#issuecomment-226069426

? Do you think you can update this PR to do it? I am willing to take the first item he mentioned ("Add approvers to OWNERS files")

bgrant0607:
I created a tracking issue for finishing this: #1389

mungegithub/mungers/owner-approval_test.go
```diff
@@ -0,0 +1,180 @@
+/*
+Copyright 2015 The Kubernetes Authors All rights reserved.
```
Kashomon: nit: 2016? (here and above)

bgrant0607:
@hongchaodeng is working on OWNERS, but I don't think this particular PR is going anywhere, so closing.

PR #475, opened 2016-02-09
Fix perfdash

gmarek:
This change should make perfdash usable as a place that gathers performance data from various sources. It extracts information from output of (mainly) density.go test, but it's able to use load.go results as well.

I'll try to get someone who knows JS to update the page, so there will be possible to have multiple tabs for different sources (e.g. Google, CoreOS, etc.). After it's done, if you want to publish your performance test data, you'll need to publish test outputs somewhere in the net, and add a scraper for it to this code.

Perfdash is running at http://perf-dash.k8s.io/

cc @hongchaodeng @timothysc @spiffxp

### DIFF
--- a/hack/verify-flags/known-flags.txt
+++ b/hack/verify-flags/known-flags.txt
@@ -45,6 +45,7 @@ shame-cc
 shame-from
 shame-report-cmd
 source-file
+start-from
 stats-port
 sync-period
 tcp-services
--- a/perfdash/Dockerfile
+++ b/perfdash/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM google/debian:wheezy
+FROM google/debian:jessie
 MAINTAINER Brendan Burns <bburns@google.com>
 RUN apt-get update
 RUN apt-get install -y -qq ca-certificates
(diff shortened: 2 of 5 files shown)


wojtek-t:
LGTM

gmarek:
Added a flag to known-flags...

jeremyeder:
Is perf-dash installable locally?

gmarek:
As always, the answer is: it depends. It is, but it won't have access to our jenkins, hence you'll need to write your own source.

jeremyeder:
"the answer is it depends". spoken like a true perf guy ;)

spiffxp:
>>it won't have access to our jenkins

ack, what about the google cloud storage buckets where jenkins dumps all of its results?

gmarek:
Yeah, that we can do. Actually contrib/compare does exactly that. I just didn't feel like it's needed to rewrite whole perfdash backend, but if few people think it is, I can do it easily enough.

spiffxp:
Discussed during @kubernetes/sig-testing meeting today, sounds like there's interest in moving away from jenkins as source-of-truth for results, and towards buckets long-term, for things like perf-dash, submit-queue, etc.  Opens up the door to federated test results a little more easily

gmarek:
OK, I'll add it to my list, though not with the very high priority. If someone wants to take it over, I'll gladly give some pointers.

PR #476, opened 2016-02-09
Add a drop down in perfdash to select test kind

gmarek:


perfdash/www/script.js
```diff
@@ -21,7 +21,10 @@ var PerfDashApp = function(http, scope) {
     this.scope = scope;
     this.selectedResource = "pods";
     this.selectedVerb = "GET";
+    this.testNames = [];
+    this.testName = "Density";
```
bryk: You can initialize it with null and get first available value after line 39. With `this.testNames[0]`.

bryk:
This code has some serious flaws, but is no worse than it was, so LGTM. I'd be really happy to have somebody clean up and document the frontend code :)

gmarek:
Adding label per @bryk comment.

PR #480, opened 2016-02-10
pod-master: Enable https connection to etcd servers

victorgp:
I know the podmaster will be deprecated, but in the meantime, until that happens there are a lot of us that needs the podmaster to connect securely to etcd.


victorgp:
I signed it!

mikedanese:
Pod master has been deprecated and removed.

PR #482, opened 2016-02-11
Perfdash uses GCS as a source

gmarek:
@spiffxp @jeremyeder @kubernetes/sig-scalability

### DIFF
--- a/perfdash/perfdash.go
+++ b/perfdash/perfdash.go
@@ -29,7 +29,6 @@ import (
 	"time"
 
 	// TODO: move this somewhere central
-	"k8s.io/contrib/mungegithub/mungers/jenkins"
 	"k8s.io/kubernetes/pkg/util/sets"
 )
 
@@ -73,10 +72,17 @@ func (buildLatency *TestToBuildData) ServeHTTP(res http.ResponseWriter, req *htt
 	res.Write(data)
 }
 
+// states of parsing machine
 const (
 	scanning   = iota
 	inTest     = iota
 	processing = iota
+)
+
+// constants to use for downloading data.
+const (
+	urlPrefix = "https://storage.googleapis.com/kubernetes-jenkins/logs/kubernetes-e2e-gce-scalability/"
+	urlSuffix = "/build-log.txt"
 )
 
 var descriptionToName = map[string]string{
@@ -107,7 +113,7 @@ func parseTestOutput(scanner *bufio.Scanner, buildNumber int, resources sets.Str
 			if strings.Contains(line, "INFO") || strings.Contains(line, "STEP") || strings.Contains(line, "Failure") || strings.Contains(line, "[AfterEach]") {
 				obj := LatencyData{}
 				if err := json.Unmarshal(buff.Bytes(), &obj); err != nil {
-					fmt.Printf("error parsing JSON in build %d: %v %s\n", buildNumber, err, buff.String())
+					fmt.Fprintf(os.Stderr, "error parsing JSON in build %d: %v %s\n", buildNumber, err, buff.String())
 					// reset state and try again with more input
 					state = scanning
 					continue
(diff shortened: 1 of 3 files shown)


perfdash/perfdash.go
```diff
@@ -134,37 +140,94 @@ func parseTestOutput(scanner *bufio.Scanner, buildNumber int, resources sets.Str
...
-		if build.Number < *startFrom {
-			continue
-		}
-		reader, err := client.GetConsoleLog(job, build.Number)
+	for ; buildNumber < lastBuildNo; buildNumber++ {
+		testDataResponse, err := http.Get(fmt.Sprintf("%v%v%v", urlPrefix, buildNumber, urlSuffix))
 		if err != nil {
-			fmt.Printf("error getting logs: %v\n", err)
-			continue
+			panic(err)
```
wojtek-t: I think that doing panic() is risky. I think you should log error and continue here

perfdash/perfdash.go
```diff
@@ -134,37 +140,94 @@ func parseTestOutput(scanner *bufio.Scanner, buildNumber int, resources sets.Str
...
+	for response.StatusCode == 200 {
+		inc *= 2
+		start = end
+		end += inc
+		response, err = http.Get(fmt.Sprintf("%v%v%v", urlPrefix, end, urlSuffix))
+		if err != nil {
+			panic(err)
+		}
+	}
+	// We know that data exists for start and does not exist for end. Now binary search.
```
wojtek-t: This comment is true before the for, but it's not the invariant of the loop. Maybe make it clear.

perfdash/perfdash.go
```diff
@@ -134,37 +140,94 @@ func parseTestOutput(scanner *bufio.Scanner, buildNumber int, resources sets.Str
...
+	for start+1 < end {
```
wojtek-t: I think it would be simpler to do:

for start < end {
  candidate = (start + end) / 2
  if response == 200 {
    start = candidate+1
  } else {
    end = candidate 
  }
}

Then you are guaranteed to have start=end at the end, and don't need the rest of the code.

gmarek:
@wojtek-t PTAL

perfdash/perfdash.go
```diff
@@ -134,37 +140,85 @@ func parseTestOutput(scanner *bufio.Scanner, buildNumber int, resources sets.Str
...
+func getLatencyDataFormGCS() (TestToBuildData, sets.String, sets.String, error) {
 	buildLatency := TestToBuildData{}
 	resources := sets.NewString()
 	methods := sets.NewString()
 
-	queue, err := client.GetJob(job)
-	if err != nil {
-		return buildLatency, resources, methods, err
+	buildNumber := *startFrom
+	lastBuildNo := binSearch()
```
ixdy: just read https://storage.googleapis.com/kubernetes-jenkins/logs/kubernetes-e2e-gce-scalability/latest-build.txt

gmarek: Thanks!

gmarek:
cc @hongchaodeng

perfdash/perfdash.go
wojtek-t: That was easy :) Thanks @ixdy

perfdash/perfdash.go
```diff
@@ -107,7 +113,7 @@ func parseTestOutput(scanner *bufio.Scanner, buildNumber int, resources sets.Str
 			if strings.Contains(line, "INFO") || strings.Contains(line, "STEP") || strings.Contains(line, "Failure") || strings.Contains(line, "[AfterEach]") {
 				obj := LatencyData{}
 				if err := json.Unmarshal(buff.Bytes(), &obj); err != nil {
-					fmt.Printf("error parsing JSON in build %d: %v %s\n", buildNumber, err, buff.String())
+					fmt.Errorf("error parsing JSON in build %d: %v %s\n", buildNumber, err, buff.String())
```
wojtek-t: fmt.Errorf() doesn't log anything (it just creates an error, which you are dropping here) - so this change is incorrect

perfdash/perfdash.go
```diff
@@ -134,37 +140,51 @@ func parseTestOutput(scanner *bufio.Scanner, buildNumber int, resources sets.Str
...
 	buildLatency := TestToBuildData{}
 	resources := sets.NewString()
 	methods := sets.NewString()
 
-	queue, err := client.GetJob(job)
+	buildNumber := *startFrom
+	latestBuildResponse, err := http.Get(fmt.Sprintf("%vlatest-build.txt", urlPrefix))
 	if err != nil {
-		return buildLatency, resources, methods, err
+		panic(err)
```
wojtek-t: please just return err here

perfdash/perfdash.go
```diff
@@ -134,37 +140,51 @@ func parseTestOutput(scanner *bufio.Scanner, buildNumber int, resources sets.Str
...
+	}
+	latestBuildBody := latestBuildResponse.Body
+	defer latestBuildBody.Close()
+	latestBuildBodyScanner := bufio.NewScanner(latestBuildBody)
+	latestBuildBodyScanner.Scan()
+	var lastBuildNo int
+	fmt.Sscanf(latestBuildBodyScanner.Text(), "BUILD_NUMBER=%d", &lastBuildNo)
+
```
wojtek-t: remove empty line

wojtek-t:
Few more comments.

perfdash/perfdash.go
```diff
@@ -134,37 +140,51 @@ func parseTestOutput(scanner *bufio.Scanner, buildNumber int, resources sets.Str
...
 	if err != nil {
-		return buildLatency, resources, methods, err
+		panic(err)
+	}
+	latestBuildBody := latestBuildResponse.Body
+	defer latestBuildBody.Close()
+	latestBuildBodyScanner := bufio.NewScanner(latestBuildBody)
+	latestBuildBodyScanner.Scan()
+	var lastBuildNo int
+	fmt.Sscanf(latestBuildBodyScanner.Text(), "BUILD_NUMBER=%d", &lastBuildNo)
```
ixdy: I feel like a bad person, since I just recommended using this file, but I'm thinking of changing the format so the file contains only the build number and nothing else (#21180). It would make this logic even cleaner, though, since you'd just need to read the content.

wojtek-t: I think we can change it once the change you desribed is done.

perfdash/perfdash.go
gmarek: Done.

perfdash/perfdash.go
gmarek: Done.

perfdash/perfdash.go
gmarek: Done.

wojtek-t:
LGTM

PR #485, opened 2016-02-15
Replacing the virtualbox vm.box with the official centos box

atosatto:
The `chef/centos-7.0` box is no more available on Atlas.
This PR replace it with the official centos 7 box.


atosatto:
Another option could be to provide the explicit location of the vagrant box with the `config.vm.box_url` setting. @eparis which solution do you believe would be most effective?

PR #490, opened 2016-02-17
Allow nginx Ingress controller run as DaemonSet

### LINKED ISSUE #357: [Nginx Ingress Loadbalancer] Support daemonset
Should the loadbalancer pods be created by a daemonset instead of a rc thus we can be able to ensure that the pods will always be present on, say edge nodes?

aledbf:
fixes #357

requires #491


aledbf:
ping @bprashanth

bprashanth:
Guessing this just needs a rebase? (we updated godeps already right)

PR #472, opened 2016-02-08
Remove the ciphers option from the httpsfrontend

silasbw:
ciphers requires an argument and the following string (no-sslv3) isn't a
valid cipher, so haproxy was failing to start. The template config still
selects ciphers via ssl-default-bind-ciphers in global.

PR #492, opened 2016-02-17
Update haproxy image and service-loadbalancer

aledbf:
- removes the custom build of haproxy
- service-loadbalancer uses haproxy image (gcr.io/google_containers/haproxy)
- closes #472


bprashanth:
I can't rebuild an image till we've revved the godep to include the fix to the namespace watching issue. Would you happen to have the cycles to do that? (current kube should be ok)

aledbf:
@bprashanth done

bprashanth:
Thanks. LGTM but for the nit. 
Travis is complaining, probably because of the godep.

aledbf:
> Is edge like latest?

Is the version with the new musl version that solves the issue with dns search.

bprashanth:
I meant does it translate to some X.Y version that we can use instead of edge? Otherwise if someone is running this in prod and wants to know if they're vuln to CVE-2016, and I ask for their version, and they reply with edge, I'll probably have to ask them to redeploy to latest weather or not they are.

aledbf:
@bprashanth I can use 3.3 but I don't know if the fix will land in that version. 
Can we push this? I can open another PR to change the version once/if is ported to 3.3 or 3.4 is released

bprashanth:
LGTM thanks

PR #493, opened 2016-02-17
Add support for private repositories in git-sync

aledbf:
This adds support for http authentication, shallow clone using depth flag and one time execution (exit 0 after initial clone)

ref kubernetes/kubernetes/issues/16247

### DIFF
--- a/git-sync/Dockerfile
+++ b/git-sync/Dockerfile
@@ -1,4 +1,14 @@
-FROM golang:1.4-onbuild
-VOLUME ["/git"]
+FROM alpine:edge
+
 ENV GIT_SYNC_DEST /git
-ENTRYPOINT ["/go/bin/git-sync"]
+
+RUN apk add -U bash git && \
+  rm -rf /var/cache/apk/*
+
+VOLUME ["/git"]
+
+COPY git-sync /git-sync
+
+RUN touch /.netrc && chmod 777 /.netrc
+
+ENTRYPOINT ["/git-sync"]
--- a/git-sync/Dockerfile
+++ b/git-sync/Dockerfile
@@ -1,14 +1,16 @@
-FROM alpine:edge
+FROM ubuntu:15.10
 
 ENV GIT_SYNC_DEST /git
-
-RUN apk add -U bash git && \
-  rm -rf /var/cache/apk/*
-
 VOLUME ["/git"]
 
-COPY git-sync /git-sync
+ENV DEBIAN_FRONTEND noninteractive
+RUN apt-get update && \
+  apt-get dist-upgrade -y && \
+  apt-get install -y git ca-certificates --no-install-recommends && \
+  apt-get clean -y && \
+  rm -Rf /usr/share/man /usr/share/doc && \
+  rm -rf /var/lib/apt/lists/*
 
-RUN touch /.netrc && chmod 777 /.netrc
+COPY git-sync /git-sync
 
 ENTRYPOINT ["/git-sync"]
--- a/git-sync/Dockerfile
+++ b/git-sync/Dockerfile
@@ -5,12 +5,12 @@ VOLUME ["/git"]
 
 ENV DEBIAN_FRONTEND noninteractive
 RUN apt-get update && \
-  apt-get dist-upgrade -y && \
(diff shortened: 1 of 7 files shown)


bprashanth:
@kubernetes/sig-storage review?

git-sync/main.go
```diff
@@ -66,8 +84,24 @@ func main() {
 	if _, err := exec.LookPath("git"); err != nil {
 		log.Fatalf("required git executable not found: %v", err)
 	}
+
+	if *flUsername != "" && *flPassword != "" {
+		log.Println("creating .netrc file (required for https authentication)")
+		if err := createNetrcFile(*flUsername, *flPassword, *flRepo); err != nil {
```
rootfs: `netrc` exposes plaintext password, how about git credential cache?

git-sync/main.go
```diff
@@ -43,6 +49,18 @@ func envString(key, def string) string {
 	return def
 }
 
+func envBool(key string, def bool) bool {
+	if env := os.Getenv(key); env != "" {
+		res, err := strconv.ParseBool(env)
+		if err != nil {
+			return def
+		}
+
```
markturansky: dead line

git-sync/main.go
```diff
@@ -66,8 +84,24 @@ func main() {
 	if _, err := exec.LookPath("git"); err != nil {
 		log.Fatalf("required git executable not found: %v", err)
 	}
+
+	if *flUsername != "" && *flPassword != "" {
```
markturansky: I'm curious.  Why pointers?  is `len(flUsername)==0` or `flUsername==""` insufficient?

git-sync/main.go
```diff
@@ -66,8 +84,24 @@ func main() {
...
+		if err := syncRepo(*flRepo, *flDest, *flBranch, *flRev, *flDepth); err != nil {
+			log.Fatalf("error syncing repo: %v", err)
+		}
+
+		os.Exit(0)
+	}
+
 	for {
-		if err := syncRepo(*flRepo, *flDest, *flBranch, *flRev); err != nil {
+		if err := syncRepo(*flRepo, *flDest, *flBranch, *flRev, *flDepth); err != nil {
```
markturansky: same with all of these.  Does passing `flRepo`, for example, not work?  Why pointers?

git-sync/Dockerfile
```diff
@@ -1,4 +1,14 @@
-FROM golang:1.4-onbuild
-VOLUME ["/git"]
+FROM alpine:edge
```
thockin: I don't think we can use Alpine for containers like this - DNS in Alpine is currently broken (doesn't respect search paths) which means this will not work for anyone who wants to access a git server in your own cluster.

git-sync/main.go
```diff
@@ -66,8 +84,24 @@ func main() {
...
 	}
+
+	if *flUsername != "" && *flPassword != "" {
+		log.Println("creating .netrc file (required for https authentication)")
+		if err := createNetrcFile(*flUsername, *flPassword, *flRepo); err != nil {
+			log.Fatalf("error creating .netrc file: %v", err)
+		}
+	}
+
+	if *flOneTime {
```
thockin: rather than a new block, can you test this flag inside the for loop and exit(0), rather than duplicating it into a whole block like this?

git-sync/main.go
thockin: flags are pointers - this is dereferencing them to values.

markturansky: Thanks for the clarification :)

git-sync/main.go
aledbf: @rootfs done

git-sync/Dockerfile
aledbf: @thockin changed to `ubuntu:15.10`

git-sync/main.go
aledbf: @thockin done

bprashanth:
looks like comments are addressed, anyone LGTM?

aledbf:
ping

rootfs:
LGTM

PR #502, opened 2016-02-20
Add DNS to requirements and troubleshooting tips for 1.2.0a7

deftdawg:
Update nginx-third-party docs to include requirement for working DNS and troubleshooting tips for 1.2.0-alpha7

FYI @aledbf

### DIFF
--- a/Ingress/controllers/nginx-third-party/README.md
+++ b/Ingress/controllers/nginx-third-party/README.md
@@ -16,7 +16,7 @@ This is a nginx Ingress controller that uses [ConfigMap](https://github.com/kube
 
 ## Requirements
 - default backend [404-server](https://github.com/kubernetes/contrib/tree/master/404-server) (or a custom compatible image)
-
+- DNS must be operational and able to resolve default-http-backend.default.svc.cluster.local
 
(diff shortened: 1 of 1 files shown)


bprashanth:
@jumpkick thanks for the clarification, I've seen people complain about this. I'm thinking if there's a better way to handle this (in addition to the docs). @aledbf in case you have ideas.

jumpkick:
@bprashanth I wish I knew a good way to feed this back in that it can help others.  Much of the documentation is out of date and the shear volume of components and documentation can make it difficult to sort through the problems to a working installation.  Maybe if there was a per release wiki page people could add workarounds for a particular release that they had to do to get working configurations, that could help others targeting the same release.

bprashanth:
The ideal way to surface something like this would be through an event, if that's possible. Something that shows up in kubectl describe rc nginx-rc-name, or describe service. In this case of course that's hard because we can't talk to the apiserver, so the next best thing would be kubectl logs (log to stdout of the container).

Lets merge this for now, I'll give it some thought once https://github.com/kubernetes/kubernetes/pull/21504. Thanks for helping out.

PR #504, opened 2016-02-22
Add HTTPS to the GCE ingress controller

bprashanth:
Adds support for HTTPS, and 3 modes:
1. len(ing.Spec.TLS) == 0: pure http
2. len(ing.Spec.TLS) > 0: http on :80 + https on :443 with ing.Spec.TLS[0]
3. ing.Annotation[allowHTTP] == false: don't setup anything on :80

Second and third commits have real code. 

@kubernetes/goog-cluster feel free to jump in and take over as reviewers, assigning to Tim as defacto reviewer.


bprashanth:
Assignig to @freehan because Tim is ooo

freehan:
Overall it is good. It is just the L7 and L7RuntimeInfo abstraction seems to be trapped between gcp and k8s. And I cannot think of a better way to refactor it. So LGTM...

bprashanth:
> Overall it is good. It is just the L7 and L7RuntimeInfo

Yeah this is because GCE doens't have an L7. It's a lot easier to reason about an L7 as an L7 object. This object contains all the resources that make up the conceptual GCE L7, and hook them up with the Kubernetes resources. 

Hopefully GCE will just get an L7 object.

bprashanth:
@freehan added the requested comments: https://github.com/bprashanth/contrib/commit/a2f82d5f48fe470db0a259888b29582699c3fb9f#diff-0ca32fb01d30b7ffd4f37d77adcce2e5R252

bprashanth:
@freehan thanks for the review

PR #509, opened 2016-02-23
Update image tags to reflect pushed images.

bprashanth:
@aledbf fyi.


bprashanth:
Self merging because these are just image updates I just pushed out.

PR #512, opened 2016-02-25
Check if the dns is working properly. Skip invalid TCP services

### LINKED ISSUE #511: [Ingress nginx-third-party] nginx does not get initial configuration if nginx-third-party rc is annotated with a TCP Service which does not exist
If the nginx-third-party Ingress controller is configured to expose a TCP Service and that Service does not exist, the initial configuration for nginx is not created in time, and the Pod ends up in a restart loop. 

To reproduce:
1. Deploy http-default-backend `kubectl create -f $contrib/Ingress/controllers/nginx-third-party/examples/default-backend.yaml && kubectl expose --port 8080 rc default-http-backend`
2. Deploy nginx-third-party: `kubectl create -f $contrib/Ingress/controllers/nginx-third-party/examples/rc-tcp.yaml`
3. Annotate with tcpservices annotation: `kubectl annotate rc nginx-ingress-3rdpartycfg "nginx-ingress.kubernetes.io/tcpservices=default/bogus:1337"`
4. Kill the nginx-third-party Pods to force the new configuration: `k delete po -l name=nginx-ingress-lb`

Now you'll see something like this in the logs:
```
I0225 13:14:32.982404       1 utils.go:177] Waiting for default/bogus
...
E0225 13:14:37.910683       1 utils.go:65] Requeuing default/some-service, err Post http://127.0.0.1:8080/update-ingress: dial tcp 127.0.0.1:8080: getsockopt: connection refused
E0225 13:14:37.911002       1 utils.go:65] Requeuing default/some-other-service, err Post http://127.0.0.1:8080/update-ingress: dial tcp 127.0.0.1:8080: getsockopt: connection refused
```

From what I can understand, this is because [the code path that should write the initial config](https://github.com/kubernetes/contrib/blob/ae06390f72b0d5893a1fe318af51e6ee867da39b/Ingress/controllers/nginx-third-party/controller.go#L220) ends up spinning in [this wait.Poll loop](https://github.com/kubernetes/contrib/blob/ae06390f72b0d5893a1fe318af51e6ee867da39b/Ingress/controllers/nginx-third-party/utils.go#L178). This means that nginx doesn't start, so nothing is listening on localhost:8080, hence the connection refused errors. This continues untill the Pod is killed by Kubernetes due to failing health checks, as nginx is supposed to handle these as well, but it doesn't since it is not running.

I'm not really sure how to fix this, but I think it's a bad thing that one misconfigured TCP Service should end up breaking the Ingress controller for all Services.

Verified on Kubernetes 1.1.2/custom setup with `gcr.io/google_containers/nginx-third-party:0.3`.

aledbf:
fixes #511

### DIFF
--- a/ingress/controllers/nginx-third-party/main.go
+++ b/ingress/controllers/nginx-third-party/main.go
@@ -63,14 +63,24 @@ func main() {
 		glog.Fatalf("Please specify --default-backend")
 	}
 
+	glog.Info("Checking if DNS is working")
+	ip, err := checkDNS(*defaultSvc)
+	if err != nil {
+		glog.Fatalf("Please check if the DNS addon is working properly.\n%v", err)
+	}
+	glog.Infof("IP address of '%v' service: %s", *defaultSvc, ip)
+
 	kubeClient, err := unversioned.NewInCluster()
 	if err != nil {
 		glog.Fatalf("failed to create client: %v", err)
 	}
 
 	lbInfo, _ := getLBDetails(kubeClient)
-	defSvc := getService(kubeClient, *defaultSvc)
-	defError := getService(kubeClient, *customErrorSvc)
+	defSvc, err := getService(kubeClient, *defaultSvc)
+	if err != nil {
+		glog.Fatalf("no default backend service found: %v", err)
+	}
+	defError, _ := getService(kubeClient, *customErrorSvc)
 
 	// Start loadbalancer controller
 	lbc, err := NewLoadBalancerController(kubeClient, *resyncPeriod, defSvc, defError, *watchNamespace, lbInfo)
--- a/ingress/controllers/nginx-third-party/utils.go
+++ b/ingress/controllers/nginx-third-party/utils.go
@@ -19,6 +19,7 @@ package main
 import (
 	"encoding/json"
 	"fmt"
+	"net"
 	"os"
 	"strconv"
 	"strings"
@@ -34,6 +35,11 @@ import (
 	"k8s.io/kubernetes/pkg/client/unversioned"
 	"k8s.io/kubernetes/pkg/util/wait"
 	"k8s.io/kubernetes/pkg/util/workqueue"
+)
+
+const (
+	httpPort  = "80"
+	httpsPort = "443"
 )
 
 var (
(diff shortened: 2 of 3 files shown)


aledbf:
ping @bprashanth

Ingress/controllers/nginx-third-party/main.go
```diff
@@ -62,14 +62,24 @@ func main() {
 		glog.Fatalf("Please specify --default-backend")
 	}
 
+	glog.Info("Checking if DNS is working")
+	ip, err := checkDNS(*defaultSvc)
```
bprashanth: can we run this without dns? Or clarify why we need dns in the failure message. In theory we can just serve a default backend in this pod right?

aledbf: > can we run this without dns?

Yes but then we need to specify the ip of the service. Right now it just try to resolve `svcName.namespace` to an [IP address](https://github.com/kubernetes/contrib/pull/512/files#diff-e6b8f3022373cce5f55cdfce8ad4516cR234) (maybe I need to remove the word add-on, it will work with any dns capable to resolve name->ip)

Just in case dns is a requirement [check here](https://github.com/kubernetes/contrib/blob/master/Ingress/controllers/nginx-third-party/nginx.tmpl#L237) 
I'm just trying to avoid to start nginx and not be able to serve content because

thecanadianbaker:
@bprashanth can we get a merge on this and #519 so work can start on issue #543 without collisions?

bprashanth:
Ack, sorry for the downtime was working on 1.2 stuff

ingress/controllers/nginx-third-party/utils.go
```diff
@@ -206,7 +201,18 @@ func getTcpServices(kubeClient *unversioned.Client, tcpServices string) []nginx.
...
 		if len(namePort) == 2 {
-			tcpSvc := getService(kubeClient, namePort[0])
+			tcpSvc, err := getService(kubeClient, namePort[0])
+			if err != nil {
+				glog.Errorf("%s", err)
+				continue
+			}
+
+			// the exposed TCP service cannot use 80 or 443 as ports
+			if namePort[1] == "80" || namePort[1] == "443" {
```
bprashanth: nit, suggest httpPort httpsPort const and using them here and in the error message

ingress/controllers/nginx-third-party/main.go
```diff
@@ -62,14 +62,24 @@ func main() {
 		glog.Fatalf("Please specify --default-backend")
 	}
 
+	glog.Info("Checking if DNS is working")
+	ip, err := checkDNS(*defaultSvc)
```
bprashanth: Why is DNS a requirement exactly, you can lookup the endpoints of a Service through apiserver right? or am I missing something

aledbf: > or am I missing something

nginx is not using the IP of the services and without a valid server to resolve the addresses it will fail to return content from the services

aledbf: @bprashanth I will try to remove this requirement in another PR creating a lookup table (dns name, svc ip)

bprashanth: SG, but why do you even need IP of a service? you can directly map svcName -> endpoints no? or is this for something else

bprashanth:
LGTM, merging, we can discuss the dns requirement in the follow up pr

PR #513, opened 2016-02-26
keepalived-vip: Add support for custom templates, routes and routing rules

wienczny:
This PR adds support for custom templates by adding a flag to set a custom template file.
It also supports defining routes and routing rules using additional annotations.
The next release of keepalived is going to add support for routing rules.

New annotations are:
- k8s.io/public-vip-netmask 
- k8s.io/public-vip-route

This looks like this:

```
k8s.io/public-vip: "11.22.33.44"
k8s.io/public-vip-netmask: "24"
k8s.io/public-vip-route: "11.22.33.254"
```

Add flags for:
- path to keepalived binary
- path to template filename
- path to keepalived.conf
- path to keepalived.pid
- template generation debug (does not start keepalived)
- to enable ipvs conntrack (implemented it before it was upstream and added to flag to keep upstream behavior by default)

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/513)

### DIFF
--- a/keepalived-vip/Dockerfile
+++ b/keepalived-vip/Dockerfile
@@ -15,10 +15,11 @@
 FROM alpine:3.3
 
 RUN echo "http://dl-4.alpinelinux.org/alpine/edge/testing" >> /etc/apk/repositories && \
-  apk add -U --repository http://dl-3.alpinelinux.org/alpine/edge/testing/ --allow-untrusted \
+  apk add -U --repository http://dl-3.alpinelinux.org/alpine/edge/testing/ --allow-untrusted --no-cache \
   bash curl ipvsadm iproute2 keepalived && \
   rm -rf /var/cache/apk/* /etc/keepalived/keepalived.conf
   
+#COPY keepalived.conf.template /etc/keepalived/keepalived.conf.template
 COPY kube-keepalived-vip /kube-keepalived-vip
 
-ENTRYPOINT ["/kube-keepalived-vip"]
+ENTRYPOINT ["/kube-keepalived-vip", ""]
--- a/keepalived-vip/Dockerfile
+++ b/keepalived-vip/Dockerfile
@@ -22,4 +22,4 @@ RUN echo "http://dl-4.alpinelinux.org/alpine/edge/testing" >> /etc/apk/repositor
 #COPY keepalived.conf.template /etc/keepalived/keepalived.conf.template
 COPY kube-keepalived-vip /kube-keepalived-vip
 
-ENTRYPOINT ["/kube-keepalived-vip", ""]
+ENTRYPOINT ["/kube-keepalived-vip", "--keepalivedPidFilename=/keepalived.pid"]
--- a/keepalived-vip/keepalived.go
+++ b/keepalived-vip/keepalived.go
@@ -68,54 +68,73 @@ vrrp_instance vips {
   }
 }
 
-{{ range $i, $svc := .svcs }}
-virtual_server {{ $svc.Ip }} {{ $svc.Port }} {
+{{ range $i, $svc := .svcs }}# Start {{$svc.Name}}
+{{ range $j, $port := $svc.Ports }}# Start {{$port.Port}} {{$port.Protocol}}
+virtual_server {{ $svc.IP }} {{ $port.Port }} {
   delay_loop 5
   lvs_sched wlc
   lvs_method NAT
   persistence_timeout 1800
-  protocol TCP
+  protocol {{ $port.Protocol }}
 
-  {{ range $j, $backend := $svc.Backends }}
-  real_server {{ $backend.Ip }} {{ $backend.Port }} {
+  {{ range $k, $backend := $port.Backends }}
+  real_server {{ $backend.IP }} {{ $backend.Port }} {
     weight 1
     TCP_CHECK {
       connect_port {{ $backend.Port }}
       connect_timeout 3
     }
   }
-{{ end }}
-}    
+  {{ end }}
+}
+# End {{$port.Port}} {{$port.Protocol}}
+{{ end }}# End {{$svc.Name}}
 {{ end }}
 `
 )
 
 type keepalived struct {
-	iface      string
-	ip         string
-	netmask    int
-	priority   int
-	nodes      []string
-	neighbors  []string
-	useUnicast bool
-	password   string
-	started    bool
-	vips       []string
+	iface          string
+	ip             string
+	netmask        int
+	priority       int
+	nodes          []string
+	neighbors      []string
+	useUnicast     bool
+	password       string
+	vips           []string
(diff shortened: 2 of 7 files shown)


bprashanth:
@aledbf

keepalived-vip/keepalived.go
```diff
@@ -68,54 +68,73 @@ vrrp_instance vips {
   }
 }
 
-{{ range $i, $svc := .svcs }}
```
aledbf: Good idea the removal of the template of the code. Please remove the the template from the go code to avoid unnecessary duplication. The external template should check for the content to be generated (like only adding virtual_rules if there's elements in `.routes`)

keepalived-vip/main.go
```diff
@@ -39,15 +40,22 @@ var (
...
 
+	enableIPVSConntrack = flags.Bool("enable-ipvs-conntrack", true,
+		`enable ipvs conntrack in net/ipv4/vs/conntrack`)
+
 	// sysctl changes required by keepalived
 	sysctlAdjustments = map[string]int{
 		// allows processes to bind() to non-local IP addresses
 		"net/ipv4/ip_nonlocal_bind": 1,
-		// enable connection tracking for LVS connections
-		"net/ipv4/vs/conntrack": 1,
```
aledbf: This is not optional. Without connection tracking is not possible to receive the response from the backend. 
With the change in network addresses is not possible to use DR.

keepalived-vip/utils.go
```diff
@@ -59,6 +60,16 @@ func getNodeInfo(nodes []string) (*nodeInfo, error) {
 	}, nil
 }
 
+// getEffectiveIP returns the forcedIP if it was given on the cmdline.
+// If forcedIP is not set it will return the node address
+func getEffectiveIP(nodes []string, forcedIP string) (string, error) {
```
aledbf: can you clarify what you're trying...? (Is not possible to assign just one IP (this can be running as DaemonSet)

keepalived-vip/Dockerfile
```diff
@@ -15,10 +15,11 @@
...
-  apk add -U --repository http://dl-3.alpinelinux.org/alpine/edge/testing/ --allow-untrusted \
+  apk add -U --repository http://dl-3.alpinelinux.org/alpine/edge/testing/ --allow-untrusted --no-cache \
   bash curl ipvsadm iproute2 keepalived && \
   rm -rf /var/cache/apk/* /etc/keepalived/keepalived.conf
   
+#COPY keepalived.conf.template /etc/keepalived/keepalived.conf.template
 COPY kube-keepalived-vip /kube-keepalived-vip
 
-ENTRYPOINT ["/kube-keepalived-vip"]
+ENTRYPOINT ["/kube-keepalived-vip", "--keepalivedPidFilename=/keepalived.pid"]
```
aledbf: can we just expose flags via -args through yaml instead?

keepalived-vip/keepalived.go
```diff
@@ -68,54 +68,73 @@ vrrp_instance vips {
...
 
+var (
+	keepalivedPath = flags.String("keepalived", "keepalived",
+		`Path to keepalived`)
+	keepalivedTemplateFilename = flags.String("keepalived-conf-template",
+		"/etc/keepalived/keepalived.conf.template", `Path to keepalived.conf.template`)
+	keepalivedConfFilename = flags.String("keepalived-conf", "/etc/keepalived/keepalived.conf",
+		`Path to keepalived.conf`)
+	keepalivedPidFilename = flags.String("keepalived-pid", "/run/keepalived.pid",
+		`Path to keepalived.conf`)
```
aledbf: why not keep `keepalivedPath`, `keepalivedTemplateFilename`, `keepalivedConfFilename` and  `keepalivedPidFilename` "hidden"? what use case do you have in mind with this flags?

keepalived-vip/main.go
wienczny: I has not been included when I started and I wanted to keep the original behavior as close as possible. I had to enable it for myself though. Thats why I added a flag to enable it on demand.

keepalived-vip/utils.go
wienczny: I'm running this neither as DaemonSet nor inside a Docker container. Because the interface could not be resolved I added a flag to specify the ip of the used interface.

keepalived-vip/Dockerfile
wienczny: What do you meab by "through yaml"?

keepalived-vip/keepalived.go
wienczny: I've got to use a custom build of keepalived with support routing rules. This is the reason I added the keepalivedPath to specify the custom path to my keepalived binary. keepalivedTemplateFilename, keepalivedConfFilename and keepalivedPidFilename are useful for testing multiple configurations and Docker setups where /etc/keepalived and /run do not exist.
I think those flags have sane defaults and a user is not required to change them. He just should be able to change them if required.

wienczny:
The combination of an external template and kubernetes client might be useful in different cases. I could be something like https://github.com/kelseyhightower/confd for kubernetes.

keepalived-vip/Dockerfile
aledbf: like here https://github.com/kubernetes/contrib/blob/master/keepalived-vip/vip-daemonset.yaml#L25

keepalived-vip/keepalived.go
wienczny: There already is a check for .routes from line 32 to 38. 
It currently always generates both routes and rules if they are available. We could add a switch for rules though. I did not add that switch for because the template can now changed by the user.

keepalived-vip/Dockerfile
wienczny: We can do that but I like my Docker containers to be startable with as few additional configuration parameters as possible. This simplifies testing and improves reusability. There is no problem when you change the configuration parameters using the yaml if you really need to. There just should be a working default in the docker image.

bgrant0607:
cc @bprashanth

eparis:
I this still alive? It seems @wienczny is pushing updates. can @aledbf review?

aledbf:
@wienczny can you update the PR?

wienczny:
The changes to the configuration (annotations -> configmap) are a great burden for this PR. There are several problems:
- The use of ConfigMaps requires a central administration. Multiple namespaces are not supported.
- The new configuration format is less flexible. You can't easily add additional values without breaking it.
- There can only be one service for an IP

Some changes have already been implemented in the meantime. I think a start from scratch is easier than getting this PR to merge.

test-foxish:
recomputing cla status...

grodrigues3:

- **hack**
- **keepalived-vip**

cc @bprashanth 

PR #517, opened 2016-02-26
blunderbuss updates

lavalamp:


dchen1107:
lgtm

dchen1107:
Can you update by adding pkg/kubelet to include node team? Or I can handle it in a separate pr?

lavalamp:
I don't think we can add teams, just individuals.

(I mean anything is possible given time, but as it works today...)

dchen1107:
I mean everyone in node team. But I guess it should be ok now since they are routing to us all the time anyway.

PR #521, opened 2016-02-28
Munger: Don't lock while building users

justinsb:
If we don't have a github token, we never build the users, and thus we
can't develop the UI etc.

### DIFF
--- a/mungegithub/mungers/submit-queue.go
+++ b/mungegithub/mungers/submit-queue.go
@@ -26,8 +26,6 @@ import (
 	"sync"
 	"time"
 
-	"k8s.io/kubernetes/pkg/util/sets"
-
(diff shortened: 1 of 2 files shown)


justinsb:
I think this is important as otherwise development is hard, and I end up making mistakes like https://github.com/kubernetes/kubernetes/issues/22136

eparis:
hack/verify-go-lint.sh failed...

eparis:
and now submit-queue_test.go doesn't build...

eparis:
Without any token, what does this enable? It gets the web site up?

justinsb:
Exactly... All in service of bringing up a dev environment with no token.
Going to revisit all of these this afternoon..

eparis:
May be superceeded by #1108

PR #524, opened 2016-03-01
Remove ihmccreery from blunderbuss for test/

ikehz:
Really, cc @kubernetes/goog-testing and @kubernetes/sig-testing should be thinking about better ways to target reviews to the teams who maintain the tests, rather than @ixdy.


ixdy:
LGTM

:cries:

ihmccreery:
Like I said, you might consider making it a priority to factor the tests better so that those teams responsible for them actually maintain them, instead of going through you.

PR #526, opened 2016-03-01
Add a 'kind/old-docs' label to PRs which touch files moving out of tree

eparis:
... as we split the docs to their own repo


eparis:
@bgrant0607

eparis:
@johndmulhausen

bgrant0607:
Awesome! LGTM!

Side note: I noticed is that docs/proposals are labeled kind/design but docs/design are not so labeled. They probably should be.

bgrant0607:
It doesn't look like Travis checks anything that this could break, so I'm going to merge.

bgrant0607:
@eparis I assume someone on our side needs to rebuild, push, and deploy the munger?

eparis:
I can do that too, let me fix docs/design first.

PR #527, opened 2016-03-01
Default to --dry-run

eparis:
Must use --dry-run=false if you want to make changes to github.

Helps make sure developers don't accidentally make changes to github.


eparis:
@justinsb

eparis:
Maybe I should just do away with the name `dry-run` and think of a better flag name.  `write-mode` ? Something like that?

justinsb:
This LGTM - thanks so much.

I think dry-run is a good name for it.  I know it's not normal for dry-run to be the default, but I think it's fine.

PR #529, opened 2016-03-02
serve_hostname debug logging after receiving SIGTERM

pwittrock:


mikedanese:
@pwittrock Do we use this in a test?

bprashanth:
Yeah, please move it to test/images

for-demos/serve_hostname/serve_hostname.go
```diff
@@ -88,8 +90,16 @@ func main() {
...
-		go log.Fatal(http.ListenAndServe(fmt.Sprintf(":%d", *port), nil))
+		go func() {
+			log.Fatal(http.ListenAndServe(fmt.Sprintf(":%d", *port), nil))
+		}()
 	}
 
-	select {}
+	// Write to stdout after receiving SIGTERM
+	signals := make(chan os.Signal)
+	signal.Notify(signals, syscall.SIGTERM)
```
bprashanth: Please catch handle SIGINT too

mikedanese:
Ya, should be in main repo.

https://github.com/kubernetes/kubernetes/tree/master/test/images

Also this needs a version bump in the makefile and should be pushed to gcr

for-demos/serve_hostname/serve_hostname.go
```diff
@@ -88,8 +90,16 @@ func main() {
 			log.Printf("HTTP request from %s", r.RemoteAddr)
 			fmt.Fprintf(w, "%s", hostname)
 		})
-		go log.Fatal(http.ListenAndServe(fmt.Sprintf(":%d", *port), nil))
+		go func() {
+			log.Fatal(http.ListenAndServe(fmt.Sprintf(":%d", *port), nil))
```
bprashanth: do you have a reason for creating the closure?

for-demos/serve_hostname/serve_hostname.go
```diff
@@ -88,8 +90,16 @@ func main() {
...
+		}()
 	}
 
-	select {}
+	// Write to stdout after receiving SIGTERM
+	signals := make(chan os.Signal)
+	signal.Notify(signals, syscall.SIGTERM)
+	select {
+	case sig := <- signals:
+		log.Panicf("Terminating after receiving signal: %s.\n", sig.String())
```
bprashanth: please add a comment about why you're choosing panicf

bprashanth:
Please bump the version and make sure that this is in line with the code in the latest gcr.io/google_containers/serve_hostname image (1.4 i think), in case we forgot and pushed a version out in between.

for-demos/serve_hostname/serve_hostname.go
bprashanth: Do you need to invoke String() or will %s do that for you?

for-demos/serve_hostname/serve_hostname.go
pwittrock: When go log.Fatal(http.ListenAndServe...), ListenAndServe is invoked synchronously (blocking) so the signal code is never reached.  The closure makes the whole thing async, which I am guessing was the origin intent.  I can add this as a comment.

for-demos/serve_hostname/serve_hostname.go
pwittrock: Done

for-demos/serve_hostname/serve_hostname.go
pwittrock: %s seems to do it for us

pwittrock: added comment

pwittrock:
RE: moving this to the main repo.  @thockin suggested in the future that we phase out using this for tests, and use `netexec` in its place.  We will still want to keep this around for demos.  In this case, is moving it to the main repo still the correct thing to do?

pwittrock:
bumped the version to 1.4. Is there a way to query gcr.io to see what tags exist for an image?

thockin:
you have to log in to GCR to look at tags - 1.3 is latest

pwittrock:
Thanks

bprashanth:
> We will still want to keep this around for demos. In this case, is moving it to the main repo still the correct thing to do?

I think so, unless you're planning to switch out netexec right now. It should just be a lift-and-shift. If you'd like to defer that to another pr that's fine too. You will have to update the test to use the 1.4 image anyway.

pwittrock:
Sounds good.  Why don't we get this PR looking like we want and submit it, then I will move the code between repos without making any changes to it.  Afterward I can build the new image: `make containerf && make push` and finally update the tests to use the new tag.

pwittrock:
PTAL

bprashanth:
SGTM

for-demos/serve_hostname/serve_hostname.go
```diff
@@ -88,8 +90,18 @@ func main() {
 			log.Printf("HTTP request from %s", r.RemoteAddr)
 			fmt.Fprintf(w, "%s", hostname)
 		})
-		go log.Fatal(http.ListenAndServe(fmt.Sprintf(":%d", *port), nil))
+		go func() {
+			// Run in a closure so http.ListenAndServe doesn't block
```
bprashanth: this commend applies for "run via go" too right? I mean why specifically create a new func() thingy

for-demos/serve_hostname/serve_hostname.go
```diff
@@ -88,8 +90,18 @@ func main() {
...
 	}
 
-	select {}
+	// Write to stdout after receiving SIGTERM and SIGINT to help with debugging kubernetes issue #21605
+	signals := make(chan os.Signal)
+	signal.Notify(signals, syscall.SIGTERM, syscall.SIGINT)
+	select {
+	case sig := <- signals:
+		// Keep behavior consistent with how the signal is handled by default (default is to panic)
+		log.Panicf("Terminating after receiving signal: %s.\n", sig)
```
bprashanth: can't you just pull from the sig chan directly instead of selecting, since there's just a single case?

for-demos/serve_hostname/serve_hostname.go
pwittrock: run via go applies to the first function, but not the arguments.  In this case the http listener is run and blocks (not run via go).  log.Fatal is then run in a go routine with the http listener result passed as an argument.

See this [example old](https://play.golang.org/p/DgVLa-PP2w) - deadlocks
See this [example new](https://play.golang.org/p/-3Mgr2po0N) - passes

for-demos/serve_hostname/serve_hostname.go
pwittrock: Yeah works.  Done.

for-demos/serve_hostname/serve_hostname.go
bprashanth: Ah you need to evaluate the arguments to execute the fatalf and the closure explicitly requires lazy evaluation.

for-demos/serve_hostname/serve_hostname.go
```diff
@@ -88,8 +90,16 @@ func main() {
...
+		}()
 	}
 
-	select {}
+	// Write to stdout after receiving SIGTERM and SIGINT to help with debugging kubernetes issue #21605
+	signals := make(chan os.Signal)
+	signal.Notify(signals, syscall.SIGTERM, syscall.SIGINT)
+	sig := <- signals
+	// Keep behavior consistent with how the signal is handled by default (default is to panic)
+	log.Panicf("Terminating after receiving signal: %s.\n", sig)
```
bprashanth: nit: you don't need the sig variable right?

pwittrock: Not needed, I could put the read in the log function, but I think this structuring makes the intention more clear that we are blocking and reading from a channel.

bprashanth:
LGTM

bprashanth:
Will merge on green

pwittrock:
Let me make one change

pwittrock:
forgot to run gofmt

pwittrock:
Updated with gofmt.

pwittrock:
PTAL

bprashanth:
LGTM, merging

pwittrock:
danke

PR #531, opened 2016-03-03
Add @fejta and @spxtr to blunderbuss config for test/

ixdy:
Largely `test/` PRs have mostly involve triaging to proper owners, but I'd like not to have to do that all myself.

cc @fejta @spxtr and also @kubernetes/sig-testing in case anyone else feels they should be on this

### DIFF
--- a/mungegithub/blunderbuss.yml
+++ b/mungegithub/blunderbuss.yml
@@ -233,4 +233,6 @@ prefixMap:
     - ihmccreery
 
   test:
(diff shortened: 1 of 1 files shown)


eparis:
https://github.com/kubernetes/kubernetes/pull/22393

And here I was hoping to kill this file in the next couple days. I'll merge here and rebase that PR, but then hopefully I can stop merging these things here.

PR #533, opened 2016-03-03
Use current milestone after 'priorities' but before 'pr number'

eparis:
This means a p0 with v1.3 will come before a p1 with v1.2. It's hard to decide how priority vs milestone relate. Is a p0 with no milestone before or after a p3 with v1.2? My algorithm says the p0 always comes first.

Priority first
Then due date of milestone
Then issue number

### DIFF
--- a/mungegithub/github/github.go
+++ b/mungegithub/github/github.go
@@ -23,6 +23,7 @@ import (
 	"io/ioutil"
 	"math"
 	"net/http"
+	"regexp"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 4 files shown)


eparis:
@lavalamp @bgrant0607 This will hopefully keep most of the v1.2 PRs (without priority) towards the front of the queue for the last 1.5 days.

mungegithub/github/github.go
```diff
@@ -48,6 +49,11 @@ const (
 	headerRateReset     = "X-RateLimit-Reset"
 )
 
+var (
+	releaseMilestoneRE = regexp.MustCompile(`^v[\d]+.[\d]$`)
```
bgrant0607: I guess the milestone string doesn't contain `milestone-`?

bgrant0607:
LGTM AFAICT

mungegithub/github/github.go
eparis: It is the name of the milestone it will match again, so in kubernetes it will today only (likely) see 3 possible values:
`v1.2` (matches), `v1.2-candidate` (will not match), `next-candidate` (will not match).  (It could actually get closed milestones, but that won't happen much/ever)

PR #539, opened 2016-03-03
submit queue: Enable the path block merge mungers

eparis:
This will block PRs which change docs from automatically merging.


eparis:
This should not be merged/enabled until the docs freeze, with the 1.2 cut, Friday around 11AM Pacific

Issue #406, opened 2016-01-26
Ansible: Add CoreOS Support


danehans:
I would like to add CoreOS support to the contrib/ansible project. I have a working implementation at [1]. Keep in mind [1] is very rough and still needs a lot of work before submitting a PR. It would be helpful to discuss design decisions with the project maintainer before submitting PRs. Your feedback would be greatly appreciated.

[1] https://github.com/danehans/contrib/tree/coreos

danehans:
@eparis Can this be closed now that coreos is merged?

PR #548, opened 2016-03-07
Adds CoreOS Support

danehans:
Adds initial support for running Kubernetes on CoreOS. This commit
adds :
1. Tasks and files to install pypy to CoreOS.
   ansible_python_interpreter was added to group_vars/all.yml to
   reference the Python installation path.
2. Roles were updated to reference systemd unit files or dropins
   that are specific to CoreOS.
3. Conditional logic added throughout based on the
   ansible_lsb.id == "CoreOS" system fact.

Closes Issue: https://github.com/kubernetes/contrib/issues/406


ansible/roles/kubernetes/tasks/main.yml
```diff
@@ -2,22 +2,65 @@
...
+    bin_dir: "/opt/bin"
+    kube_script_dir: "/opt/bin/kubernetes"
+  when: ansible_lsb.id == "CoreOS" and kube_script_dir == "/usr/libexec/kubernetes"
+
 - name: Update {{ kube_script_dir }} if this is atomic
   set_fact:
     kube_script_dir: "/usr/local/libexec/kubernetes"
   when: is_atomic and kube_script_dir == "/usr/libexec/kubernetes"
 
+- name: Determine if Kubernetes download directory exists
```
eparis: Why are you checking if things exist and creating them if they don't? Seems like you are trying to make the declarative nature of ansible imperative...

ansible/roles/master/templates/kube-apiserver-dropin.j2
```diff
@@ -0,0 +1,17 @@
+[Unit]
+After=etcd2.service
+
+[Service]
+User=root
+ExecStart=
```
eparis: Why 2 `ExecStart`s

eparis:
Some nits: and I dislike the double flannel and double etcd installs....   What can we do to reduce that duplication?

ansible/roles/kubernetes/tasks/main.yml
danehans: I will remove.

ansible/roles/master/templates/kube-apiserver-dropin.j2
danehans: According to CoreOS, the 1st empty ExecStart is needed when you use a dropin that uses ExecStart.I confirmed that it's needed. I will add a comment so people are aware of the issue.

danehans:
@eparis thanks for the review. I responded to your line notes and will update my PR accordingly. 

Regarding the double flannel/etcd installs, the PR does not install etcd, it only manages the existing etcd service that comes with CoreOS. CoreOS advises not changing the etcd version of a particular release. I am however using etcd2 instead of etcd. I can go back and test etcd instead of etcd2, but CoreOS recommends the use of etcd2 over etcd. I think etcd2 support should be added throughout the roles as a separate PR. Please let me know if you require that I change from etcd2 > etcd to merge the PR.

The PR uses GH bins to install Flannel. I saw support for pkg installs [1]. CoreOS does not have a pkg manager, so pulling GH bins is required. Therefore, I added include: github-release.yml when flannel_source_type == "github-release".

[1] https://github.com/kubernetes/contrib/blob/master/ansible/roles/flannel/tasks/client.yml#L2-L7

ansible/group_vars/all.yml
```diff
@@ -27,6 +27,11 @@ cluster_name: cluster.local
 insecure_registrys:
 #  - "gcr.io"
 
+# Required for CoreOS. CoreOS does not include a Python interpreter. The
+# pre-ansible role installs a python interpreter to /opt/bin/. For more
+# information see https://coreos.com/blog/managing-coreos-with-ansible.html
+#ansible_python_interpreter: "PATH=/opt/bin:$PATH python"
```
eparis: Can this be automatically set in the pre-ansible role using the 'raw' module to verify things?

ansible/roles/common/defaults/main.yaml
```diff
@@ -0,0 +1,8 @@
+---
+# The directory where binaries are stored on Ansible
+# managed systems.
+bin_dir: /usr/local/bin
+
+# The directory used by Ansible to temporarily store
+# files on Ansible managed systems.
+ansible_temp_dir: /tmp/.ansible/files
```
eparis: I think I'd like to see a tmp_dir created in pre-ansible (via mktemp) rather than do this. But that's probably just a matter of taste...

ansible/roles/common/tasks/main.yml
```diff
@@ -23,6 +23,22 @@
     docker_config_dir: "/etc/default"
   when: ansible_distribution == "Debian" or ansible_distribution == "Ubuntu"
 
+- name: Create config file directory
+  file: path={{ docker_config_dir }} state=directory
+  when: ansible_lsb.id == "CoreOS"
```
eparis: shouldn't this always run? Why is CoreOS 'special' ?

ansible/roles/common/tasks/main.yml
```diff
@@ -23,6 +23,22 @@
...
+
+- name: Set the bin directory path for CoreOS
+  set_fact:
+    bin_dir: "/opt/bin"
+  when: ansible_lsb.id == "CoreOS"
+
+- name: Create bin directory for CoreOS
+  file: path={{ bin_dir }} state=directory
+  when: ansible_lsb.id == "CoreOS"
```
eparis: Always run. Simple playbook is always better.

ansible/roles/kubernetes/tasks/main.yml
```diff
@@ -7,17 +7,28 @@
...
+  when: ansible_lsb.id == "CoreOS" and kube_script_dir == "/usr/libexec/kubernetes"
+
 - name: Create kubernetes config directory
   file: path={{ kube_config_dir }} state=directory
 
 - name: Create kubernetes script directory
   file: path={{ kube_script_dir }} state=directory
 
+- name: Create bin dir
+  file: path={{ bin_dir }} state=directory
```
eparis: Would like a space between this and the next action.

ansible/roles/node/tasks/main.yml
```diff
@@ -41,6 +44,9 @@
     kubelet_token: "{{ tokens.results[0].content|b64decode }}"
     proxy_token: "{{ tokens.results[1].content|b64decode }}"
 
+- name: Create the /var/lib/kubelet working directory
+  file: path=/var/lib/kubelet state=directory
```
eparis: This should probably be a variable rather than hard coded path.

ansible/roles/etcd/tasks/main.yml
```diff
@@ -17,10 +17,14 @@
   include: github-release.yml
   when: etcd_source_type == "github-release"
 
+- name: Create config directory
+  file: path="/etc/etcd" state=directory
```
eparis: /etc/etcd should probably be a variable not a hard coded path (not your fault)

ansible/roles/pre-ansible/tasks/main.yml
```diff
@@ -9,6 +9,14 @@
   register: distro
   always_run: yes
 
+- name: Set the is_coreos fact
+  set_fact:
+    is_coreos: true
+  when: "'CoreOS' in distro.stdout"
+
+- include: coreos.yml
+  when: is_coreos
```
eparis: oh man! All those comments and I was going to just merge.  Until the VERY LAST changeset! I believe you must initialize the fact. Else this `when: is_coreos` will crash on non-coreos systems.

You can fix only this, and I will merge hoping you'll update based on the other comments sometime in the future, or you can do as much or as little rework based on my review as you think reasonable.

ansible/group_vars/all.yml
danehans: According to the following, this parameter can only be set in 3 places and group_vars seems to me the most applicable for my use case:

http://stackoverflow.com/questions/22769568/system-specific-variables-in-ansible

ansible/roles/common/tasks/main.yml
danehans: I added ansible_lsb.id == "CoreOS" b/c the dir creation task did not exist prior to my patch. I assumed that it already existed for other distro's and wanted to play it safe. I will remove the when: statement.

ansible/roles/common/tasks/main.yml
danehans: I will remove the when: statement.

ansible/roles/etcd/tasks/main.yml
danehans: Agreed. I'll update.

ansible/roles/kubernetes/tasks/main.yml
danehans: Agreed. I'll update.

ansible/roles/node/tasks/main.yml
danehans: Agreed. I'll update.

ansible/roles/pre-ansible/tasks/main.yml
danehans: OK. I didn't realize that. I'll update.

ingvagabund:
This PR breaks playbook on RHEL
```vim
05:37:33 TASK: [common | Set the bin directory path for CoreOS] ************************ 
05:37:33 fatal: [10.8.55.250] => error while evaluating conditional: ansible_lsb.id == "CoreOS"
```

PR #549, opened 2016-03-08
Service-loadbalancer allows custom return code on /

bprashanth:
@aledbf


bprashanth:
This is needed because some ingress controllers don't like 404 on /, and I'd like serviceloadbalancer to be usable as a generic backend for all ingress controllers

aledbf:
LGTM

bprashanth:
Thanks, merging

PR #551, opened 2016-03-08
Redirect on x-forwarded-for=http in echoserver.

bprashanth:
Anyone sending x-forwarded-for probably cares about the protocol, @aledbf

### DIFF
--- a/ingress/echoheaders/Makefile
+++ b/ingress/echoheaders/Makefile
@@ -1,7 +1,7 @@
 all: push
 
 # TAG 0.0 shouldn't clobber any release builds
-TAG = 1.2
+TAG = 1.3
 PREFIX = gcr.io/google_containers/echoserver
 
 container:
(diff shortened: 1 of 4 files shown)


aledbf:
LGTM

bprashanth:
Thanks merging

carsonoid:
This commit broke the ingress example. Shouldn't we only redirect to https when `$http_x_forwarded_proto` is `https` **not** when it's `http` as is set by this commit. Currently it always redirects all http traffic to https which makes no sense when testing ingress.

bprashanth:
Can you elaborate? the point is to redirect all http to https

bprashanth:
You can use 1.2 to just get the old non-redirect

carsonoid:
That's what I ended up doing. The ingress example docs should probably be
updated though.

PR #552, opened 2016-03-08
Use ubuntu-slim image

aledbf:
This PR removes the use of `alpine` as base image in:
- keepalived
- haproxy image
- nginx-slim image
- nginx-third-party controller
- service-loadbalancer


aledbf:
ping @bprashanth

bprashanth:
You're using debian-slim not ubuntu? also what's the motivation behind the change? I know alpine has some caveats like dns search path, but is there something more going on?

bprashanth:
I was looking at a stale commit, you're using ubuntu-slim indeed. Question about how ubuntu-slim > alpine:edge still stands.

bprashanth:
LGTM when you fix the merge conflicts

bprashanth:
Ah, turns out the dependency is between this make file whic produces the new nginx-slim based on ubuntu, and the controller Dockerfile which runs apt-get in the other pr. So can't push out nginx-third-party till this one goes in.

PR #553, opened 2016-03-09
SSL-enabled Nginx Ingress Controller integrated Hashicorp Vault

devlinmr:
Based on the nginx-alpha controller, this adds functionality to configure SSL nginx servers for ingresses requiring it, pulling in certificates from Vault.

See README.md.

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/553)


bprashanth:
Hi @devlinmr thanks for the PR. Didn't notice it till now, will take a look by monday. Feel free to tag me or @kubernetes/goog-cluster on Ingress/lb prs in the future.

ingress/controllers/nginx-alpha-ssl/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+Access to Vault is predicated on the following:
+
+Vault is running in the local environment.
+
+A valid access token is available at the following location:
+
+/etc/vault-token/ingress-read-only
+
+An example for creating this as a kubernetes secret is shown below.
```
bprashanth: did you forget to add this documentation, or did I miss it?

ingress/controllers/nginx-alpha-ssl/controller.go
```diff
@@ -0,0 +1,217 @@
...
+	vaultEnabled := true
+	config := vault.DefaultConfig()
+	config.Address = os.Getenv("VAULT_ADDR")
+	vault, err := vault.NewClient(config)
+	if err != nil {
+		fmt.Printf("WARN: VAULT_ADDR is not set\n")
+		vaultEnabled = false
+	}
+
+	token := os.Getenv("VAULT_TOKEN")
```
bprashanth: How about actually adding an RC+Service yaml for vault, and then polling here till the vault service is available (--wait-for-value or something can turn off the polling, but if it's true we just don't proceed until vault is up).

ingress/controllers/nginx-alpha-ssl/controller.go
```diff
@@ -0,0 +1,217 @@
...
+			continue
+		}
+		known = ingresses
+
+		if vaultEnabled {
+			for _, ingress := range ingresses.Items {
+				ingressHost := ingress.Spec.Rules[0].Host
+				vaultPath := "secret/ssl/" + ingressHost
+
+				keySecretData, err := vault.Logical().Read(vaultPath)
```
bprashanth: can you include some docs on how one inserts certs into value? for bonus points you can make a simple script that is similar to `kubectl create -f secret.yaml` but inserts into vault instead of apiserver.

ingress/controllers/nginx-alpha-ssl/run.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+
+export VAULT_SKIP_VERIFY=true
+[ -f /etc/vault-token/ingress-read-only ] && export VAULT_TOKEN=`cat /etc/vault-token/ingress-read-only`
```
bprashanth: Why not just read this directly in the controller (--vault-token) instead of using an env var? 
If you add it as a secret it will be mounted in the specified location (or it's a bug in Kubernetes).

ingress/controllers/nginx-alpha-ssl/README.md
bprashanth: Ah, I see a subsequent commit has it

bprashanth:
Did a quick scan. This is nice! If you could add a section about "why use valut?" (just so others can make an informed decision about weather or not they want to deploy the extra rc+svc), add vault as a svc+rc, and squash the godep commits I'll take a second look.

bprashanth:
I think travis is complaining about missing Godeps

bprashanth:
ping @devlinmr

bprashanth:
@devlinmr thats a crazy list of commits :) guessing it was a rebase error?

test-foxish:
recomputing cla status...

PR #555, opened 2016-03-10
Auto create firewall rule and allow user specified static ips

bprashanth:
As the name indicates. Still need to add a few unittests.


bprashanth:
Closed in favor of https://github.com/kubernetes/contrib/pull/563

PR #559, opened 2016-03-10
Replace the (unused) url-list flag with a paths flag 

roberthbailey:
To allow the loader to hit multiple url paths and/or override the default url path on the host under test.

Fix the Makefile to work properly with godeps.


brendandburns:
LGTM

PR #563, opened 2016-03-12
Allow a user specified global static ip via annotation.

bprashanth:
This pr allows a user to specify an existing global static ip through the Ingress annotation: `kubernetes.io/ingress.global-static-ip-name` and have that assigned to both the HTTP and HTTPS forwarding rules created by the Ingress controller. 

It also refactors the TLS loading logic into an interface to facilitate loading from multiple sources (on file, apiserver, fake).

### DIFF
--- a/ingress/controllers/gce/controller/utils.go
+++ b/ingress/controllers/gce/controller/utils.go
@@ -34,7 +34,10 @@ import (
 	"github.com/golang/glog"
 )
 
-const allowHTTPKey = "kubernetes.io/ingress.allowHTTP"
+const (
+	allowHTTPKey    = "kubernetes.io/ingress.allow-http"
+	staticIPNameKey = "kubernetes.io/ingress.global-static-ip-name"
+)
 
 // ingAnnotations represents Ingress annotations.
 type ingAnnotations map[string]string
@@ -50,6 +53,14 @@ func (ing ingAnnotations) allowHTTP() bool {
 		return true
 	}
 	return v
+}
+
+func (ing ingAnnotations) staticIPName() string {
+	val, ok := ing[staticIPNameKey]
+	if !ok {
+		return ""
+	}
+	return val
 }
 
 // errorNodePortNotFound is an implementation of error.
--- a/ingress/controllers/gce/loadbalancers/loadbalancers.go
+++ b/ingress/controllers/gce/loadbalancers/loadbalancers.go
@@ -22,6 +22,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"reflect"
 	"strings"
 
 	compute "google.golang.org/api/compute/v1"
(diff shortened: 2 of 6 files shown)


ingress/controllers/gce/loadbalancers/loadbalancers.go
```diff
@@ -413,15 +422,50 @@ func (l *L7) checkForwardingRule(name, proxyLink, ip, portRange string) (fw *com
...
+	//  - GCE will throw a 400, and the controller will keep trying to use
+	//    the IP in the hope that the user manually resolves the conflict
+	//    or deletes/modifies the Ingress.
+	// TODO: Handle the last case better.
+
+	if l.runtimeInfo.StaticIPName != "" {
+		// Existing static IPs allocated to forwarding rules will get orphaned
+		// till the Ingress is torn down.
+		if ip, err := l.cloud.GetGlobalStaticIP(l.runtimeInfo.StaticIPName); err != nil || ip == nil {
+			glog.Warningf("The given static IP name %v doesn't translate to an existing global static IP, ignoring it and allocating a new IP: %v",
```
freehan: Log an event here? That would be easier for user to debug

ingress/controllers/gce/loadbalancers/loadbalancers.go
```diff
@@ -448,10 +489,16 @@ func (l *L7) checkHttpsForwardingRule() (err error) {
...
 }
 
+// checkStaticIP reserves a static IP allocated to the Forwarding Rule.
 func (l *L7) checkStaticIP() (err error) {
 	if l.fw == nil || l.fw.IPAddress == "" {
 		return fmt.Errorf("Will not create static IP without a forwarding rule.")
 	}
+	// Don't manage staticIPs if the user has specified an IP.
+	if address, manageStaticIP := l.getEffectiveIP(); !manageStaticIP {
+		glog.V(3).Infof("Not managing user specified static IP %v", address)
```
freehan: Log a event here

ingress/controllers/gce/loadbalancers/loadbalancers.go
bprashanth: Sending events from this module is a TODO. All errors returned by it are sent as events to apiserver, but not warnings like this. Currently it doesn't know anything about Kubernetes apiserver (by design), but it's useful to pass a wrapper type or function closure down just to send events back. Can I do this in a follow up pr (since multiple places can send such events)?

bprashanth: The same applies for any non-controller package in `ingress/controllers/gce/` (so backends, instances, healthchecks etc), none of them need to know about Kubernetes except for events

freehan: Alright, sounds good.

ingress/controllers/gce/controller/utils.go
```diff
@@ -34,7 +34,10 @@ import (
 	"github.com/golang/glog"
 )
 
-const allowHTTPKey = "kubernetes.io/ingress.allowHTTP"
+const (
+	allowHTTPKey    = "kubernetes.io/ingress.allow-http"
```
thockin: For future - these should be `ingress.kubernetes.io/allow-http` - it's documented somewhere that I am sure you didn't see :)

thockin: and if it is GCE specific, we should probably call it `gce.ingress.kubernetes.io` or something

bprashanth: I just used the pattern we use for static pods in the kubelet: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/types/pod_update.go#L26. 

Allow-http is shared but global-static-ip-name is gce specific. Easy enough to rename since this isn't in a relased image yet.

Issue #564, opened 2016-03-12
Perfdash should support more e2e performance test.


Random-Liu:
Now, the implementation of *Perfdash* is specific to Density test and Load test. It is almost impossible to add support for other e2e performance tests without a refactoring.

We should refactor and generalize the code to make it support more e2e performance test.

@gmarek @yujuhong 
/cc @kubernetes/sig-testing

gmarek:
+/infty

cc @hongchaodeng @kubernetes/sig-scalability

Random-Liu:
Close this one since #565 and kubernetes/kubernetes#22912 have been merged.

PR #565, opened 2016-03-13
Refactor Perfdash to let it support more e2e performance test

Random-Liu:
For #564 and kubernetes/kubernetes#15554.
Based on kubernetes/kubernetes#22912.

This PR refactored Perfdash to use the new [performance data type](https://github.com/Random-Liu/kubernetes/blob/kubelet-perf/test/e2e/perftype/perftype.go#L45) added in kubernetes/kubernetes#22912.

With this PR, it is easy to add more e2e performance test support. You just need to:
1) Transform e2e performance test result into [`PerfData`](https://github.com/Random-Liu/kubernetes/blob/kubelet-perf/test/e2e/perftype/perftype.go#L45) in _k8s/kubernetes/test/e2e/perftype_, and print the `PerfData` in e2e test log.
2) Add corresponding jenkins project into `JenkinsProjects` in `config.go`.
3) Add test description and test name into `DescriptionName` in `config.go`.
Then, Perfdash will automatically and periodically fetch data from specified jenkins projects, parse the test log and generate graph.

I've tested the PR locally with a fake `downloader`. However it would be better to merge kubernetes/kubernetes#22912 first, and test this PR with real jenkins project.

@gmarek @yujuhong 
/cc @kubernetes/sig-testing


wojtek-t:
@gmarek - can you please review it?
From the description it looks like something we would really like to have :)

@Random-Liu - thanks a lot for working on it!

gmarek:
@Random-Liu - now that previous PR is in, can you please rebase this one?

Random-Liu:
@gmarek, I'll test this one with the real Jenkins project. :)

Random-Liu:
@gmarek @yujuhong  Rebased and test. Now I have one running happily on my cluster. :)

Random-Liu:
Here's some screenshots :)

screenshot from 2016-04-14 15 00 30
screenshot from 2016-04-14 14 59 37

gmarek:
Screenshots look great! I'll try to do review tomorrow.

perfdash/config.go
```diff
@@ -0,0 +1,40 @@
...
+// To add new e2e test support, you need to:
+//   1) Transform e2e performance test result into *PerfData* in k8s/kubernetes/test/e2e/perftype,
+//   and print the PerfData in e2e test log.
+//   2) Add corresponding jenkins project into *JenkinsProjects*.
+//   3) Add test description and test name into *DescriptionName*.
+
+var (
+	// JenkinsProjects are the projects PerfDash supports. Downloader will download
+	// and analyze build log from the these jenkins projects.
+	JenkinsProjects = []string{"kubernetes-e2e-gce-scalability", "kubernetes-e2e-gce-serial"}
```
gmarek: It's not jenkins projects per se now. It's GCS bucket names. Actually we need a two level structure here to allow easy federated testing: top-level GCS folder URL, and set of buckets under this folder. We need to be able to represent a structure such as:
{<google_bucket_url>/kubernetes-e2e-gce-scalability, <google_bucket_url>/kubernetes-e2e-gce-serial, <redhat_bucket_url>/some_redhat_test,...}

perfdash/google-gcs-downloader.go
```diff
@@ -22,65 +22,49 @@ import (
...
 // constants to use for downloading data.
 const (
-	jobName = "kubernetes-e2e-gce-scalability"
 	logFile = "build-log.txt"
 )
 
 // GoogleGCSDownloader that gets data about Google results from the GCS repository
 type GoogleGCSDownloader struct {
-	startFrom            int
+	Builds               int
```
gmarek: The reason for previous logic was to make it easier to make not backward-compatible changes in e.g. format of data. After doing that one could just set a build number to the one when given change went live.

I liked old behavior, but if you have an argument (except simplicity) why new one is better then I'm OK with this change.

perfdash/perfdash.go
```diff
@@ -17,141 +17,17 @@ limitations under the License.
...
-	for _, datum := range data {
-		if datum.Verb == method {
-			return datum.Latency[item]
-		}
-	}
-	return -1
-}
-
 var (
-	www         = flag.Bool("www", false, "If true, start a web-server to server performance data")
```
gmarek: This flag was useful for debugging.

perfdash/perfdash.go
```diff
@@ -161,29 +37,14 @@ func main() {
...
-		if err != nil {
-			fmt.Printf("Failed to get data: %v\n", err)
-			os.Exit(1)
-		}
-		for _, v := range buildLatency {
-			generateCSV(v, resources, methods, os.Stdout)
-		}
-		return
-	}
+	downloader := NewGoogleGCSDownloader(*builds)
```
gmarek: Validate the number of builds. There's probably no reason to allow more than a 100 (and memory usage grows with this number)

perfdash/perfdash.go
```diff
@@ -193,7 +54,7 @@ func main() {
 		}
 	}()
 
-	http.Handle("/api", &buildLatency)
+	http.Handle("/result", &result)
```
gmarek: Please don't change the endpoint if you don't have a very, very, very good reason. I don't see one here.

perfdash/downloader.go
```diff
@@ -17,37 +17,36 @@ limitations under the License.
...
-// TestToBuildData is a map from test name to BuildLatencyData
-type TestToBuildData map[string]BuildLatencyData
-
-// Downloader is the interface that gets a data from a predefined source.
-type Downloader interface {
-	getData() (TestToBuildData, sets.String, sets.String, error)
+// TestToBuildData is a map from test name to BuildData
+type TestToBuildData map[string]BuildData
+
+func (buildLatency *TestToBuildData) ServeHTTP(res http.ResponseWriter, req *http.Request) {
```
gmarek: Why this function?

gmarek:
@Random-Liu - few comments in go part. @bryk - can you please look at JS changes?

perfdash/google-gcs-downloader.go
Random-Liu: The reason is that different Jenkins projects has different build number. Now we may download data from different Jenkins projects, we have to:
1) Specify startFrom for each Jenkins projects.
2) Just say we need the latest X builds, please download them from each Jenkins projects.

I think the 2) one is more straight forward, right? ;)

perfdash/perfdash.go
Random-Liu: ACK. Will change it back. :)

perfdash/perfdash.go
Random-Liu: ACK.

perfdash/config.go
Random-Liu: ACK

perfdash/downloader.go
Random-Liu: https://golang.org/pkg/net/http/#Handler
This is the function we have to implement for http handler.

perfdash/google-gcs-downloader.go
```diff
@@ -22,65 +22,49 @@ import (
...
-		defer testDataBody.Close()
-		testDataScanner := bufio.NewScanner(testDataBody)
-
-		hist := parseTestOutput(testDataScanner, buildNumber, resources, methods)
-
-		for k, v := range hist {
-			if _, ok := buildLatency[k]; !ok {
-				buildLatency[k] = make(BuildLatencyData)
+		fmt.Printf("Last build no: %v\n", lastBuildNo)
+		for buildNumber := lastBuildNo - g.Builds + 1; buildNumber <= lastBuildNo; buildNumber++ {
```
Random-Liu: We should ensure that the buildNumber should >= 1.

perfdash/perfdash.go
Random-Liu: Eh...I removed the CSV related code. Is that used for debugging?

Random-Liu:
@gmarek BTW, the perf-dash is not running now. I can't access http://perf-dash.k8s.io/ any more.

lavalamp:
I upgraded the cluster the other day-- if it wasn't running under an RC/RS,
it needs to be restarted.

gmarek:
I restarted it. I'll need to write a RC for it.

lavalamp:
(Make a deployment instead! :) )

perfdash/www/script.js
```diff
@@ -19,103 +19,104 @@ var app = angular.module('PerfDashApp', ['ngMaterial', 'chart.js']);
...
-    this.selectedVerb = "GET";
     this.testNames = [];
-    this.series = [ "Perc99", "Perc90", "Perc50" ];
-    this.allData = null;
 };
 
 PerfDashApp.prototype.onClick = function(data) {
     console.log(data);
-    window.location = "http://kubekins.dls.corp.google.com/job/kubernetes-e2e-gce-scalability/" + data[0].label + "/"
+    var job = document.getElementById("job").innerHTML
```
bryk: Use `var job = this.job`. You've got the job property on the controller anyways. See index.html#37

perfdash/www/script.js
```diff
@@ -19,103 +19,104 @@ var app = angular.module('PerfDashApp', ['ngMaterial', 'chart.js']);
...
-    return result;
+// Get all of the builds for the data set (e.g. build numbers)
+PerfDashApp.prototype.getBuilds = function() {
+    return Object.keys(this.data)
 };
 
-// Get the set of all verbs (e.g. 'GET') in the data set
-PerfDashApp.prototype.getVerbs = function() {
+// Get the set of all labels (e.g. 'resources', 'verbs') in the data set
+PerfDashApp.prototype.getLabels = function() {
```
bryk: `getLabels` method that returns nothing? Perhaps change name to something that actually explains what this does? :)

Applies to getData too

bryk:
Javascript looks good.

perfdash/www/script.js
Random-Liu: @bryk I tried that before, `this.job` is undefined here.
It seems that I couldn't access `this` in `onClick`. I'm not quite familiar with JS, do you know what happened? :)

Random-Liu:
@bryk Thanks a lot for reviewing!

perfdash/www/script.js
bryk: Do you have a link to any running dash? I could debug.

Random-Liu: @bryk Sure, sent. :)

bryk: In the constructor type:
```js
this.onClick = this.onClickInternal_.bind(this);
```
And rename this function to `onClickInternal_`. After that, you'll be able toa ccess this.job here

Random-Liu: @bryk Nice, it works!!! Thanks a lot!!!!

perfdash/google-gcs-downloader.go
gmarek: Right. You're 100% right. SGTM.

perfdash/perfdash.go
gmarek: Kind of - I don't care about CSV, I care about 'dry run':)

Random-Liu: ACK.

Random-Liu:
@gmarek Addressed comments and squashed.

perfdash/downloader.go
```diff
@@ -17,37 +17,37 @@ limitations under the License.
...
-	Verb     string    `json:"verb"`
+// Downloader is the interface that gets a data from a predefined source.
+type Downloader interface {
+	getData() (TestToBuildData, error)
 }
 
-// LatencyData represents the latency data for a set of RESTful API calls
-type LatencyData struct {
-	APICalls []APICallLatency `json:"apicalls"`
+// BuildData contains jenkins job name and a map from build number to perf data
```
gmarek: nit: remove "jenkins"

perfdash/perfdash.go
```diff
@@ -17,173 +17,61 @@ limitations under the License.
...
 	flag.Parse()
 
-	var downloader Downloader
-	downloader = NewGoogleGCSDownloader(*startFrom)
+	if *builds > maxBuilds || *builds < 0 {
+		fmt.Printf("Invalid builds number: %d", *builds)
+		*builds = maxBuilds
+	}
+
+	// TODO(random-liu): Add a top layer downloader to download build log from different buckets if we support
```
gmarek: s/if/when/ :)

gmarek:
LGTM, sans nit. Please apply the label when you fix it.

Random-Liu:
Addressed comments and self apply LGTM based on https://github.com/kubernetes/contrib/pull/565#issuecomment-211541979

Random-Liu:
@gmarek Could you help me rebuild the image and restart the perfdash? It seems that I have no permission to push image to google_containers. :)

gmarek:
oh yeah.. that thing.

gmarek:
@Random-Liu it's done. The way graphs are display is wrong though. Please fix.

Random-Liu:
@gmarek Really? Let me see. :)

PR #570, opened 2016-03-16
Update nginx controller to use endpoints. Remove lua to configure upstreams

aledbf:
This PR:
- removes lua code to configure upstreams
- adds SSL support for ingress rules
- expose TCP services using a ConfigMap
- nginx.conf customization using a ConfigMap
- improves the verbose logs showing the diff in the configuration file
- dump configmap example

Using the flag `--v=2` it is possible to see in the logs the changes in the configuration when an endpoint is added/removed
```
I0316 12:24:37.581267       1 utils.go:148] NGINX configuration diff a//etc/nginx/nginx.conf b//etc/nginx/nginx.conf
I0316 12:24:37.581356       1 utils.go:149] --- /tmp/922554809	2016-03-16 12:24:37.000000000 +0000
+++ /tmp/079811012	2016-03-16 12:24:37.000000000 +0000
@@ -235,7 +235,6 @@

     upstream default-echoheadersx {
         least_conn;
-        server 10.2.112.124:5000;
         server 10.2.208.50:5000;

     }
I0316 12:24:37.610073       1 command.go:69] change in configuration detected. Reloading...
```

Using the same flag but in level 5 `--v=5` it will configure nginx to run in debug mode

```
$ ./nginx-third-party-lb --dump-nginx—configuration
Example of ConfigMap to customize NGINX configuration:
data:
  body-size: 1m
  error-log-level: info
  gzip-types: application/atom+xml application/javascript application/json application/rss+xml
    application/vnd.ms-fontobject application/x-font-ttf application/x-web-app-manifest+json
    application/xhtml+xml application/xml font/opentype image/svg+xml image/x-icon
    text/css text/plain text/x-component
  hts-include-subdomains: "true"
  hts-max-age: "15724800"
  keep-alive: "75"
  max-worker-connections: "16384"
  proxy-connect-timeout: "30"
  proxy-read-timeout: "30"
  proxy-real-ip-cidr: 0.0.0.0/0
  proxy-send-timeout: "30"
  server-name-hash-bucket-size: "64"
  server-name-hash-max-size: "512"
  ssl-buffer-size: 4k
  ssl-ciphers: ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES256-GCM-SHA384:DHE-RSA-AES128-GCM-SHA256:DHE-DSS-AES128-GCM-SHA256:kEDH+AESGCM:ECDHE-RSA-AES128-SHA256:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA:ECDHE-ECDSA-AES128-SHA:ECDHE-RSA-AES256-SHA384:ECDHE-ECDSA-AES256-SHA384:ECDHE-RSA-AES256-SHA:ECDHE-ECDSA-AES256-SHA:DHE-RSA-AES128-SHA256:DHE-RSA-AES128-SHA:DHE-DSS-AES128-SHA256:DHE-RSA-AES256-SHA256:DHE-DSS-AES256-SHA:DHE-RSA-AES256-SHA:AES128-GCM-SHA256:AES256-GCM-SHA384:AES128-SHA256:AES256-SHA256:AES128-SHA:AES256-SHA:AES:CAMELLIA:DES-CBC3-SHA:!aNULL:!eNULL:!EXPORT:!DES:!RC4:!MD5:!PSK:!aECDH:!EDH-DSS-DES-CBC3-SHA:!EDH-RSA-DES-CBC3-SHA:!KRB5-DES-CBC3-SHA
  ssl-protocols: TLSv1 TLSv1.1 TLSv1.2
  ssl-session-cache: "true"
  ssl-session-cache-size: 10m
  ssl-session-tickets: "true"
  ssl-session-timeout: 10m
  use-gzip: "true"
  use-hts: "true"
  worker-processes: "8"
metadata:
  name: custom-name
  namespace: a-valid-namespace
```

This enables customizations like just updating the timeouts:
````
$ kubectl get configmap nginx-load-balancer-conf -o yaml
apiVersion: v1
data:
  proxy-connect-timeout: "10"
  proxy-read-timeout: "120"
  proxy-send-imeout: "120"
kind: ConfigMap
metadata:
  name: nginx-load-balancer-conf
```


aledbf:
This also removes the dns requirement

ingress/controllers/nginx-third-party/controller.go
```diff
@@ -350,6 +350,41 @@ func (lbc *loadBalancerController) getUpstreamServers(data []interface{}) ([]ngi
...
+				glog.Warningf("Secret %v has no private key", secretName)
+				continue
+			}
+			key, ok := secret.Data[api.TLSPrivateKeyKey]
+			if !ok {
+				glog.Warningf("Secret %v has no cert", secretName)
+				continue
+			}
+
+			pemFileName := lbc.nginx.AddOrUpdateCertAndKey(secretName, string(cert), string(key))
```
bprashanth: It would be really nice if you actually validated the cert using https://golang.org/pkg/crypto/x509/, a TODO is fine for now. 

Also why fatal if a single cert fails? you can still loadbalance existing services right? ideally this would just return an error and we'd requeu and retry on next run.

ingress/controllers/nginx-third-party/nginx/main.go
```diff
@@ -220,9 +220,9 @@ type NginxManager struct {
 	// path to the configuration file to be used by nginx
 	ConfigFile string
 
-	sslCertificates []ssl.Certificate
-	sslDHParam      string
-	servicesL4      []Service
+	sslDHParam string
```
bprashanth: Please add a comment describing high level function of NginxManager and a oneline about sslDHParam.

ingress/controllers/nginx-third-party/controller.go
```diff
@@ -450,9 +450,7 @@ func (lbc *loadBalancerController) Run() {
 	go lbc.svcController.Run(lbc.stopCh)
 
 	// periodic check for changes in configuration
-	go wait.Until(lbc.sync, 5*time.Second, wait.NeverStop)
-
-	time.Sleep(5 * time.Second)
+	go wait.Until(lbc.sync, 10*time.Second, wait.NeverStop)
```
bprashanth: hmm this is confusing, why are you calling lbc.sync here? shouldn't you just startup the controller and let it sync on each watch/relist?

ingress/controllers/nginx-third-party/controller.go
```diff
@@ -291,58 +287,24 @@ func (lbc *loadBalancerController) getUpstreamServers(data []interface{}) ([]ngi
...
 			server.Locations = append(server.Locations, locations...)
-			servers[rule.Host] = server
 		}
 	}
 
-	aUpstreams := make([]nginx.Upstream, 0, len(upstreams))
+	// TODO: find a way to make this more readable
+	// The structs must be ordered to always generate the same file
+	// if the content does not change.
+	aUpstreams := make([]*nginx.Upstream, 0, len(upstreams))
```
bprashanth: please add a unittest for this if you don't already have one (i mean different order of services and endpoints produce the same config file hash).

ingress/controllers/nginx-third-party/controller.go
```diff
@@ -329,6 +220,113 @@ func (lbc *loadBalancerController) registerHandlers() {
...
+				}
+			}
+			locations = append(locations, loc)
+		}
+
+		server.Locations = locations
+		servers = append(servers, server)
+	}
+
+	// lbc.nginx.AddOrUpdateIngress(name, nginx.IngressNGINXConfig{Upstreams: upstreams, Servers: servers})
```
bprashanth: are these todos, or something you plan to / have removed in a subsequent commit?

bprashanth:
Let me know when you're done experimenting with reloads and I'll resume reviewing

ingress/controllers/nginx-third-party/nginx/nginx.go
```diff
@@ -20,6 +36,7 @@ type Upstream struct {
 	Backends []UpstreamServer
 }
 
+// UpstreamByNameServers Upstream sorter by name
```
bprashanth: nit: please reword to 'sorts upstreams by name'

ingress/controllers/nginx-third-party/nginx/utils.go
```diff
@@ -116,3 +119,63 @@ func toMap(iface interface{}) (map[string]interface{}, bool) {
 
 	return map[string]interface{}{}, false
 }
+
+func checkChanges(filename string, data *bytes.Buffer) (bool, error) {
```
bprashanth: it's kind of add that this function takes a file path and a buffer, only to convert the filepath into a buffer. Maybe just make it a method on the nginx manager called needsReload so it has access to self.config?

ingress/controllers/nginx-third-party/nginx/utils.go
```diff
@@ -116,3 +119,63 @@ func toMap(iface interface{}) (map[string]interface{}, bool) {
...
+	res := data.Bytes()
+	if !bytes.Equal(src, res) {
+		err = ioutil.WriteFile(filename, res, 0644)
+		if err != nil {
+			return false, err
+		}
+
+		dData, err := diff(src, res)
+		if err != nil {
+			return false, fmt.Errorf("computing diff: %s", err)
```
bprashanth: does this error bubble up to a requeue? if so i don't think error in computing the diff is news worthy. Can you just log it and return true?

aledbf:
This also also adds support for TCP services using ConfigMaps using the flag
`--tcp-services-configmap=default/tcp-configmap-example`
example
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: tcp-configmap-example
data:
  9000: "default/example-go:8080"
```

aledbf:
ping @bprashanth

bprashanth:
LGTM please squash into fewer commits, and add some sections to the README about how to use configmaps to loadbalance tcp etc

bprashanth:
Also please add a section about lua reload vs -s and that you investigated the performance to be roughly equivalent

ingress/controllers/nginx-third-party/README.md
```diff
@@ -307,6 +308,93 @@ The route `/error` expects two arguments: code and format
...
+## Debug
+
+Using the flag `--v=XX` it is possible to increase the level of logging.
+In particular:
+- `--v=2` shows details using `diff` about the changes in the configuration in nginx
+
+```
+I0316 12:24:37.581267       1 utils.go:148] NGINX configuration diff a//etc/nginx/nginx.conf b//etc/nginx/nginx.conf
+I0316 12:24:37.581356       1 utils.go:149] --- /tmp/922554809  2016-03-16 12:24:37.000000000 +0000
++++ /tmp/079811012  2016-03-16 12:24:37.000000000 +0000
```
bprashanth: please include an example of a simple tls ingress (that's supported right?):
https://github.com/kubernetes/contrib/tree/master/ingress/controllers/gce#tls

Just so we have a consistent experience. I'm ok doing this in another pr if you prefer.

ingress/controllers/nginx-third-party/README.md
```diff
@@ -307,6 +308,93 @@ The route `/error` expects two arguments: code and format
 
 Using a volume pointing to `/var/www/html` directory is possible to use a custom error
 
+
```
bprashanth: this doc indicates that we use lua, please update.

ingress/controllers/nginx-third-party/controller.go
```diff
@@ -350,6 +362,41 @@ func (lbc *loadBalancerController) getUpstreamServers(data []interface{}) ([]ngi
...
+		ing := ingIf.(*extensions.Ingress)
+
+		for _, tls := range ing.Spec.TLS {
+			secretName := tls.SecretName
+			secret, err := lbc.client.Secrets(ing.Namespace).Get(secretName)
+			if err != nil {
+				glog.Warningf("Error retriveing secret %v for ing %v: %v", secretName, ing.Name, err)
+				continue
+			}
+			cert, ok := secret.Data[api.TLSCertKey]
```
bprashanth: Please document that this is the key/name that needs to be in the secret. I think the README isn't specific and I suspect the script in the REDME will create the wrong secret.

bprashanth:
Just a few nits. It would be really nice if we got the README examples to be more similar to the GCE one so people get a cross platform experience. Ideally we'd just have one subdir of examples and link from both the nginx and gce controller to that.

Please also remove wip when you re-up, and I can lgtm.

PR #571, opened 2016-03-16
Remove kubemark-100 from merge blockers.

gmarek:
cc @eparis


eparis:
I've not seen this test causing problems so I don't know why it should be removed.  Thus I'm not going to click the merge button   :)

gmarek:
Oh sorry - we're removing the suite and will substitute it with kubemark-500 after few changes to merge queue.

PR #580, opened 2016-03-18
Introduced a simple example of a CNI plugin

### LINKED ISSUE #576: Need a simple example CNI network plugin
There should be a simple example CNI plugin.

MikeSpreitzer:
Resolves #576

### DIFF
--- a/cni-plugins/to_docker/README.md
+++ b/cni-plugins/to_docker/README.md
@@ -23,12 +23,46 @@ Docker must be configured with a cluster store, and you must have a
 multi-host Docker network.  The Docker name of that multi-host network
 must appear as the value of the `name` field in `c2d.conf`.
 
+For example, you might use the Kuryr libnetwork driver to make a
+multi-host Docker network with the following command.
+
+```
+docker network create -d kuryr --ipam-driver=kuryr --subnet=172.19.0.0/16 --gateway=172.19.0.1 mynet
+```
+
+For an almost-example, you might try Docker's "overlay" driver to make
+a multi-host Docker network with the following command.
+
+```
+docker network create -d overlay --subnet=172.19.0.0/16 --gateway=172.19.0.1 mynet
+```
+
+However, with Docker's "overlay" driver there is no way to add the
+required additional connectivity from hosts to containers.  So do not
+do this.
+
+For another almost-example, you might try Calico's libnetwork driver,
+with commands like the following.
+
+```
+calicoctl pool add 172.19.0.0/16 --nat-outgoing
+calicoctl pool remove 192.168.0.0/16
+docker network create --driver calico --ipam-driver calico mynet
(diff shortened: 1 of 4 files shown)


eparis:
@dcbw can you review please?

MikeSpreitzer:
@eparis @dcbw: what is needed to make progress here?

The travis failure looks like a flake (see https://travis-ci.org/kubernetes/contrib/jobs/116916070#L174).

MikeSpreitzer:
@k8s-bot test this

eparis:
k8s-bot is jenkins. this is travis. I will kick travis.

MikeSpreitzer:
@eparis thanks.  But nothing has happened, as far as I can see.  Is there a way I can kick Travis myself?  I see https://github.com/kubernetes/kubernetes/blob/master/docs/devel/automation.md mentions Travis but omits instructions for how to kick it.

eparis:
kicking travis requires commit privs to the repo. Click on the travis link and if you have privs there is a 'rerun' button near the top right.  But it appears you added a new commit and now travis is happy?

MikeSpreitzer:
Yes, I found a fixed a bug.  Luckily there was no flake this time.

dcbw:
Sorry, I shall review ASAP

cni-plugins/to_docker/README.md
```diff
@@ -0,0 +1,112 @@
...
+```
+--network-plugin=cni --network-plugin-dir=/etc/cni/net.d
+```
+
+If the config file has a field named `debug` with the value `true`
+then each invocation of the plugin will produce some debugging output
+in `/tmp/`.
+
+A multi-host docker network does not necessarily meet the requirement
+(seen in http://kubernetes.io/docs/admin/networking/#kubernetes-model)
```
dcbw: Which specific bit do you mean here, one of these:

- all containers can communicate with all other containers without NAT
- all nodes can communicate with all containers (and vice-versa) without NAT
- the IP that a container sees itself as is the same IP that others see it as

or something else from that doc?

dcbw:
In general it looks fine to me.  Just some questions on the documentation.

cni-plugins/to_docker/README.md
MikeSpreitzer: I mean part of "all nodes can communicate with all containers (and vice-versa) without NAT".  I say *part of* because, as far as I know, the vice-versa is not really required (except as logically required by the use of NodePort).

MikeSpreitzer:
@dcbw: I revised the documentation, does it answer your questions?

MikeSpreitzer:
What do we need to make progress here?

dcbw:
@MikeSpreitzer looks fine to me.  @eparis ping on what needs to happen to proceed?

PR #581, opened 2016-03-18
Try official docker client in docker micro benchmark

Random-Liu:
I changed the docker micro benchmark to using new docker client https://github.com/docker/engine-api, so as to:
1) Test whether there is performance regression (There shouldn't be, just by the way)
2) Have a try of new docker client.
It turns out that there is no performance difference.

@yujuhong @dchen1107


docker-micro-benchmark/helpers/docker_helpers.go
```diff
@@ -244,9 +285,9 @@ func DoParallelContainerStopBenchmark(client *docker.Client, qps float64, routin
 
 // GetContainerIDs returns all the container ids in the system
 func GetContainerIDs(client *docker.Client) (containerIDs []string) {
-	containers, err := client.ListContainers(docker.ListContainersOptions{All: true})
+	containers, err := client.ContainerList(getContext(), types.ContainerListOptions{All: true})
 	if err != nil {
-		panic(fmt.Sprintf("Error list containers: %v", err))
+		LogError("Error list containers: %v", err)
```
yujuhong: I see you change panic to LogError. Are there cases where the test should exit, instead of continuing after getting the error?

yujuhong: A more general question is does the tests retry the failed requests, or just continue, and why.

docker-micro-benchmark/helpers/docker_helpers.go
```diff
@@ -17,40 +17,75 @@ limitations under the License.
...
+	"github.com/docker/engine-api/types/events"
+	"github.com/docker/engine-api/types/network"
+	"github.com/docker/engine-api/types/strslice"
 	"github.com/juju/ratelimit"
+	"golang.org/x/net/context"
+)
+
+const (
+	TestImage   = "ubuntu"
+	TestCommand = "ubuntu"
```
yujuhong: Why is the command "ubuntu"?

docker-micro-benchmark/helpers/docker_helpers.go
```diff
@@ -17,40 +17,75 @@ limitations under the License.
...
+	return context.Background()
+}
+
+func PullTestImage(client *docker.Client) {
+	rc, err := client.ImagePull(getContext(), types.ImagePullOptions{
+		ImageID: TestImage,
+		Tag:     "latest",
+	}, nil)
+	if err != nil {
+		fmt.Errorf("%s", err)
```
yujuhong: Why calling fmt.Errorf("%s", err), and why not using the out put at all?

docker-micro-benchmark/helpers/docker_helpers.go
```diff
@@ -17,40 +17,75 @@ limitations under the License.
...
+		return
+	}
+	defer rc.Close()
+	decoder := json.NewDecoder(rc)
+	var msg events.Message
+	for {
```
yujuhong: hm....why using a for loop?

docker-micro-benchmark/helpers/docker_helpers.go
Random-Liu: I've changed it before. However I met some problem with Godep, and maybe mass up the version a little during trying to fix the Godep. :)

docker-micro-benchmark/helpers/docker_helpers.go
Random-Liu: ? It's used in the following line.

docker-micro-benchmark/helpers/docker_helpers.go
Random-Liu: ACK.

docker-micro-benchmark/helpers/docker_helpers.go
Random-Liu: Changed back to panic for now. May add some error statistics in the future. :)

docker-micro-benchmark/helpers/docker_helpers.go
yujuhong: It doesn't seem like "events.Message" is the type made for image pulls. See https://github.com/docker/engine-api/issues/89 Hopefully they'll settle on something.

Random-Liu: Yeah, I saw that, and just tried `events.Message` here and found it partially works. :)
In fact I don't think there api is stabilized now, also see:
https://github.com/docker/engine-api/pull/140

yujuhong: I saw that too, but I think it's already completed, no?

Random-Liu: I an planning to file an issue to tracking all these stuffs, just have try myself first.

Random-Liu: @yujuhong Yeah, just don't know whether there will be other changes. I'll comment and ask them on docker/engine-api#89.

docker-micro-benchmark/helpers/docker_helpers.go
yujuhong: I see you added a lot of panics  in the PR. I'm concerned that the benchmark will never complete given that it panics whenever there is any error. How about we do something simpler:
 - For operations that are necessary for the test to start, panic or error out and stop the tests (and retry if possible).
 - For others, record the error (e.g., in a counter) and move on. Report the overall error at the end for the human beings to judge whether this is acceptable.  

This should be simple enough and does not require too much work. Let me know if you disagree.

Random-Liu: I've run some test yesterday night and docker works well surprisingly, hahaha.
But what you suggested is really reasonable. I'll add better error handling in this PR.

Random-Liu:
@yujuhong Now, after each sub benchmark, the top 5 errors and total error rate will for each docker operation will be printed. :)

docker-micro-benchmark/helpers/errors.go
```diff
@@ -0,0 +1,82 @@
...
+		return str
+	}
+	for l, errList := range e.errs {
+		str += fmt.Sprintf("%s error: \n", l)
+		for i, err := range errList {
+			if i >= errorTopNum {
+				str += fmt.Sprintf("... %d more errors\n", len(errList)-errorTopNum)
+				break
+			}
+			str += err.Error() + "\n"
```
yujuhong: nit: using strings.Join() may be cleaner.

docker-micro-benchmark/helpers/errors.go
```diff
@@ -0,0 +1,82 @@
...
+	e.RLock()
+	defer e.RUnlock()
+	var str string
+	if len(e.errs) == 0 {
+		return str
+	}
+	for l, errList := range e.errs {
+		str += fmt.Sprintf("%s error: \n", l)
+		for i, err := range errList {
+			if i >= errorTopNum {
```
yujuhong: `i ==  errorTopNum`

docker-micro-benchmark/helpers/errors.go
```diff
@@ -0,0 +1,82 @@
...
+	defer e.RUnlock()
+	return len(e.errs) <= 0
+}
+
+func (e *Error) add(l string, err error) {
+	e.Lock()
+	defer e.Unlock()
+	e.total[l] = e.total[l] + 1
+	if err != nil {
+		e.errs[l] = append(e.errs[l], err)
```
yujuhong: I am slightly wary of keeping all the error messages in memory. Maybe we need a flag t turn off this. Let's add a TODO for this.

docker-micro-benchmark/helpers/errors.go
```diff
@@ -0,0 +1,82 @@
...
+	return &Error{
+		errs:  make(map[string][]error),
+		total: make(map[string]int),
+	}
+}
+
+func (e *Error) isNil() bool {
+	e.RLock()
+	defer e.RUnlock()
+	return len(e.errs) <= 0
```
yujuhong: `== 0` should be enough.

docker-micro-benchmark/helpers/errors.go
```diff
@@ -0,0 +1,82 @@
...
+
+import (
+	"fmt"
+	"sync"
+)
+
+const errorTopNum = 5
+
+// Error is the error returned by functions in docker_helpers
+type Error struct {
```
yujuhong: nit: since you record more than just errors (e.g., you also record the total operations), maybe you can use a more general name (e.g., `Stats`?)

docker-micro-benchmark/helpers/errors.go
Random-Liu: ACK.

docker-micro-benchmark/helpers/errors.go
Random-Liu: ACK.

docker-micro-benchmark/helpers/errors.go
Random-Liu: Just to make it absolutely right, Hahaha. Bad habit.

docker-micro-benchmark/helpers/errors.go
Random-Liu: Agreed. Although errors are really rare, we should avoid this to make the benchmark result more accurate. I'll just keep the top errors and just count the error number.

docker-micro-benchmark/helpers/errors.go
Random-Liu: ACK.

Random-Liu:
@yujuhong Addressed comments.

yujuhong:
@Random-Liu I forgot about this PR completely....Could you check the failure and then we can try getting this in? Thanks!

Random-Liu:
@yujuhong Will do! :)

Random-Liu:
@yujuhong Fixed, PTAL.

PR #582, opened 2016-03-18
Skip golint on golang 1.4

eparis:
apparently lint dropped 1.4 support.


eparis:
@gmarek

gmarek:
Thanks for fixing this.

PR #588, opened 2016-03-18
Ansible: Add flannel support for Ubuntu 14.04

unicell:


eparis:
how much is the same/can be copied reused from https://github.com/kubernetes/contrib/pull/548 ?

unicell:
Just had a quick check, seems the only overlapping part is adding github-release.yml playbook for flannel. And since Ubuntu 14.04 uses upstart instead of systemd, still need service script installation besides flannel github release download and tarball extract.

So for you question, probably 3-4 steps in flannel/tasks/github-release.yml can be copied/reused. @eparis what do you suggest for me to proceed? Do I have to wait until CoreOS change landed in tree or we can do it in parallel?

unicell:
rebased to latest CoreOS commit

ansible/roles/flannel/tasks/client.yml
```diff
@@ -4,15 +4,33 @@
...
 - name: Install from github
   include: github-release.yml
-  when: is_coreos
+  when: flannel_source_type == "github-release"
+
+- set_fact:
+    flannel_use_upstart: ansible_distribution == 'Ubuntu' and ansible_distribution_major_version|int < 15
 
 - name: Install Flannel config file
   template: src=flanneld.j2 dest=/etc/sysconfig/flanneld
```
eparis: lets set a fact called 'flannel_config_dir'.  Default it to /etc/sysconfig. Overwrite it to /etc/flannel/ if you must. Then the follow on plays will need less duplication.

ansible/roles/flannel/tasks/config.yml
```diff
@@ -13,7 +13,7 @@
   delegate_to: "{{ groups['etcd'][0] }}"
 
 - name: Load the flannel config file into etcd
-  shell: "/usr/bin/etcdctl --no-sync --peers={{ peers }} set {{ conf_loc }} < {{ conf_file }}"
+  shell: "etcdctl --no-sync --peers={{ peers }} set {{ conf_loc }} < {{ conf_file }}"
```
eparis: where is etcdctl on your system? (just curious)

ansible/roles/flannel/tasks/client.yml
```diff
@@ -4,15 +4,33 @@
   args:
         name: flannel
         state: latest
-  when: not is_atomic and not is_coreos
+  when: not is_atomic and not is_coreos and ansible_distribution != "Ubuntu"
+
+- name: Force flannel_source_type to github when packages are not available
+  set_fact:
+    flannel_source_type: "github-release"
```
eparis: This is an unacceptable ansible pattern. flannel_source_type will not be fined on anything other than coreos/ubuntu and thus when you try to use it below this will cause the playbook to fail.

Instead, the very first thing you should do in this playbook is unconditionally set flannel_source_type to "package_manager" and then second overwirte it to "github-release" on the correct 2 systems.  Only run the {{ ansible_pkg_mgr }} if it is "package_manager" and run the github-release plays only if it is "github-release"

ansible/roles/flannel/tasks/github-release.yml
```diff
@@ -3,6 +3,9 @@
   set_fact:
     flannel_version: 0.5.5
 
+- stat: path=/usr/local/flannel-{{ flannel_version }}
+  register: st
```
eparis: This is all pretty cool!

ansible/roles/flannel/tasks/ubuntu.yml
```diff
@@ -0,0 +1,16 @@
+---
+- name: Ubuntu | Configure Docker to use Flannel network
+  lineinfile: dest="{{ docker_config_dir }}/docker" state=present
+              line="{{ item.line }}"
+              regexp="{{ item.regexp }}"
+  with_items:
+    - { line: '. /run/flannel/subnet.env', regexp: '. /run/flannel/subnet.env' }
+    - { line: 'DOCKER_OPTS=\"--bip ${FLANNEL_SUBNET} --mtu ${FLANNEL_MTU}\"',
+        regexp: '^DOCKER_OPTS=' }
+  when: ansible_distribution == 'Ubuntu' and ansible_distribution_major_version|int < 15
```
eparis: do you need the `when: ansible_distribution == 'Ubuntu'` inside the `ubuntu.yml` ?  (and below)

eparis:
minor comments, one major comment. You cannot set a fact with a `when` line and then use the fact without first checking that it is defined. The common pattern is to unconditionally set it to a default, then overwrite it to something custom.

ansible/roles/flannel/tasks/client.yml
unicell: Default "flannel_source_type" is defined in roles/flannel/defaults/main.yaml, so here it just overrides to "github-release" for coreos and ubuntu.

ansible/roles/flannel/tasks/client.yml
unicell: done

ansible/roles/flannel/tasks/config.yml
unicell: hmm, it is in /usr/bin, since I used etcd playbook in contrib to deploy etcd. Not really remember the initial reason I removed the path here, will revert this line.

ansible/roles/flannel/tasks/ubuntu.yml
unicell: done

unicell:
Pull request updated to address @eparis 's comments

ansible/roles/flannel/tasks/client.yml
```diff
@@ -1,16 +1,36 @@
...
+  when: is_coreos or ansible_distribution == "Ubuntu"
+
+- name: Install Flannel with package manager
   action: "{{ ansible_pkg_mgr }}"
   args:
         name: flannel
         state: latest
-  when: not is_atomic and not is_coreos
+  when: not is_atomic and flannel_source_type != "github-release"
```
eparis: Not going to block merge, but this really should be source_type == package, no?

ansible/roles/flannel/tasks/client.yml
```diff
@@ -1,16 +1,36 @@
...
   include: github-release.yml
-  when: is_coreos
+  when: flannel_source_type == "github-release"
+
+- set_fact:
+    flannel_use_upstart: ansible_distribution == 'Ubuntu' and ansible_distribution_major_version|int < 15
+
+- name: Set Flannel config file directory
+  set_fact:
+    flannel_config_dir: "/etc/sysconfig"
```
eparis: Not going to block merge, but should this be set in roles/flannel/defaults/main.yaml ?  (yes, I know I told you to do it like this)

eparis:
I'm going to merge.  If you think my two comments are good/valid please send another PR!

rutsky:
This breaks Flannel deployment on Core OS: on Core OS `/usr` is read-only

/cc: @danehans @stephenrlouie

PR #593, opened 2016-03-19
submit-queue: gzip all responses

eparis:
PR list goes from 89.93k to 16.21k


eparis:
@justinsb this is the same as your work in #468 but uses github.com/NYTimes/gziphandler instead of implementing it ourselves. Your thoughts?

PR #595, opened 2016-03-21
WIP: First commit of a tool that reconfigures tests in response to test re…

gmarek:
…sults.

It's not done and extremely hacky, but if someone want to take a look I'm happy to hear any feedback.

cc @wojtek-t @eparis @kubernetes/sig-testing

### DIFF
--- a/test-adjuster/Godeps/Godeps.json
+++ b/test-adjuster/Godeps/Godeps.json
@@ -1,0 +1,39 @@
+{
+	"ImportPath": "k8s.io/contrib/test-adjuster",
+	"GoVersion": "go1.6",
+	"GodepVersion": "v60",
+	"Packages": [
+		"./..."
+	],
+	"Deps": [
+		{
+			"ImportPath": "github.com/golang/glog",
+			"Rev": "44145f04b68cf362d9c4df2182967c2275eaefed"
+		},
+		{
+			"ImportPath": "github.com/google/go-github/github",
+			"Rev": "58e27d2951882f0b3101ba9d6e0c34a0a49350c6"
+		},
+		{
+			"ImportPath": "github.com/google/go-querystring/query",
+			"Rev": "6bb77fe6f42b85397288d4f6f67ac72f8f400ee7"
+		},
+		{
+			"ImportPath": "golang.org/x/net/context",
+			"Rev": "c2528b2dd8352441850638a8bb678c2ad056fd3e"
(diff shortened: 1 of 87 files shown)


test-adjuster/Godeps/Godeps.json
```diff
@@ -0,0 +1,9 @@
+{
```
wojtek-t: @eparis for feedback on this in general

test-adjuster/adjuster.go
```diff
@@ -0,0 +1,68 @@
+/*
```
wojtek-t: I'm not sure if adjuster is the best name, because we don't know what it is adjusting.
Maybe frequency_adjuster or sth like that?

test-adjuster/adjuster.go
```diff
@@ -0,0 +1,68 @@
...
+
+func lockAndGetRepo() string {
+	repoLock.Lock()
+	return *repoPath
+}
+
+func unlockRepo() {
+	repoLock.Unlock()
+}
+
```
wojtek-t: Can you put this into a class instead of random individual functions?
I would like to avoid things like global locks, and stuff like that.

test-adjuster/test-action.go
```diff
@@ -0,0 +1,188 @@
...
+	"os/exec"
+	"strings"
+	"time"
+
+	"github.com/golang/glog"
+	"github.com/google/go-github/github"
+	"golang.org/x/oauth2"
+)
+
+// Action is an interface for actions that should be taken by the Checker if Trigger returns false.
```
wojtek-t: This is pretty misleading - we should perform actions if trigger returns true :-)

test-adjuster/adjuster.go
gmarek: Done.

test-adjuster/adjuster.go
gmarek: Done.

test-adjuster/test-action.go
gmarek: It was a mistake in the comment:)

gmarek:
@eparis @wojtek-t PTAL

gmarek:
All code is there, it just needs to be uncommented and wrapped in appropriate pod. I already tested it on K8s and at least read path works from container. I tested write path locally.

PR #600, opened 2016-03-22
CoreOS Support: fix detection of CoreOS related tasks

### LINKED ISSUE #599: Ansible: CoreOS Support breaks RHEL Support
By introducing support for CoreOS, I can no longer run playbook on RHEL. 

```vim
05:37:33 TASK: [common | Set the bin directory path for CoreOS] ************************ 
05:37:33 fatal: [10.8.55.250] => error while evaluating conditional: ansible_lsb.id == "CoreOS"
```

When running

```
ansible --inventory-file=INVENTORY_FILE masters -m setup
```

ansible_lsb is not defined.

What about to use ``ansible_distribution`` instead of ``ansible_lsb``?

[1] https://github.com/kubernetes/contrib/pull/548

ingvagabund:
Currently, the playbook is broken on RHEL due to undefined facts:

* change ansible_lsb.id to ansible_distribution
* install flannel from github only on CoreOS

Fixes #599

### DIFF
--- a/ansible/roles/common/tasks/main.yml
+++ b/ansible/roles/common/tasks/main.yml
@@ -29,7 +29,7 @@
 - name: Set the bin directory path for CoreOS
   set_fact:
     bin_dir: "/opt/bin"
-  when: ansible_lsb.id == "CoreOS"
+  when: ansible_distribution == "CoreOS"
 
 - name: Create the directory used to store binaries
   file: path={{ bin_dir }} state=directory
--- a/ansible/roles/docker/tasks/main.yml
+++ b/ansible/roles/docker/tasks/main.yml
@@ -6,10 +6,10 @@
   when: ansible_distribution == "Ubuntu"
 
 - include: generic-install.yml
-  when: ansible_distribution != "Debian" and ansible_distribution != "Ubuntu" and ansible_lsb.id != "CoreOS"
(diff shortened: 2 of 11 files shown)


ingvagabund:
Fixes https://github.com/kubernetes/contrib/issues/599

ingvagabund:
@eparis can you review? With this patch the playbook is running on RHEL again.

ncdc:
@ingvagabund FYI you need to put the "fixes" text in the description, not in the 1st comment if you want it to close the issue automatically when this is merged.

ingvagabund:
@ncdc did not know that, thanks.

danehans:
@ingvagabund @eparis Sorry for the breakage of non-CoreOS systems. I will connect with @eparis to figure out a plan to avoid future patches from breaking non-CoreOS systems. Unfortunatly, your patch broke CoreOS :-)

Ansible is unable to obtain the distro information using the ansible_distribution fact. I have created the following issue with Ansible:

https://github.com/ansible/ansible/issues/15096

Until the above issue is resolved, we must use the ansible_lsb fact. I believe the root of the breakage was b/c I was not initializing the is_coreos = false fact. I will submit a new PR that reverts ansible_distribution to ansible_lsb for CoreOS conditionals.

danehans:
Here is an example facts output that details the issue:

https://gist.github.com/danehans/dc7081b4ac5dfe857cfe#file-coreos_ansible_facts-L99-L102

PR #608, opened 2016-03-23
add missing list separator

rutsky:
/cc @ingvagabund

### DIFF
--- a/ansible/roles/kubernetes-addons/tasks/main.yml
+++ b/ansible/roles/kubernetes-addons/tasks/main.yml
@@ -4,7 +4,8 @@
     python_bin: "/opt/bin/python"
   when: ansible_distribution == "CoreOS"
 
-- set_fact:
+- name: Set python_bin for other OS
+  set_fact:
     python_bin: "python"
   when: python_bin is not defined
 
--- a/ansible/roles/kubernetes-addons/tasks/main.yml
+++ b/ansible/roles/kubernetes-addons/tasks/main.yml
@@ -1,13 +1,12 @@
 ---
+- name: Set default python_bin
+  set_fact:
+    python_bin: "python"
+
 - name: Set python_bin fact for CoreOS
   set_fact:
     python_bin: "/opt/bin/python"
   when: ansible_distribution == "CoreOS"
-
-- name: Set python_bin for other OS
-  set_fact:
-    python_bin: "python"
-  when: python_bin is not defined
 
 - name: Assures addons dir exists
   file: path={{ kube_addons_dir }} state=directory
--- a/ansible/roles/kubernetes-addons/tasks/main.yml
+++ b/ansible/roles/kubernetes-addons/tasks/main.yml
@@ -3,7 +3,8 @@
   set_fact:
     python_bin: "/opt/bin/python"
   when: ansible_distribution == "CoreOS"
-  set_fact:
+
+- set_fact:
     python_bin: "python"
   when: python_bin is not defined
 


ingvagabund:
@rutsky is the separator needed?

rutsky:
@ingvagabund yes, these are two different tasks in YAML list of tasks.

In your code:

```
- name: Set python_bin fact for CoreOS
  set_fact:
    python_bin: "/opt/bin/python"
  when: ansible_distribution == "CoreOS"
  set_fact:
    python_bin: "python"
  when: python_bin is not defined
```

its a list of one JS object with duplicated keys `set_fact` and `when`. Ansible even warns about this:

```
 [WARNING]: While constructing a mapping from /home/bob/stuff/kubernetes/contrib/ansible/roles/kubernetes-addons/tasks/main.yml, line 2, column 3, found a duplicate dict key (set_fact).  Using last defined
value only.

 [WARNING]: While constructing a mapping from /home/bob/stuff/kubernetes/contrib/ansible/roles/kubernetes-addons/tasks/main.yml, line 2, column 3, found a duplicate dict key (when).  Using last defined value
only.
```

ingvagabund:
Will the separator help? I could not find any docs describing how to define two facts under the same action. Not sure if it will work this way. Don't you have to put the definition to another action? E.g.

```
- name: Set python_bin fact for CoreOS
  set_fact:
    python_bin: "/opt/bin/python"
  when: ansible_distribution == "CoreOS"

- name: Set python_bin for other OS
  set_fact:
    python_bin: "python"
  when: python_bin is not defined
```

rutsky:
@ingvagabund I don't follow you. What do you want to achieve?

`main.yml` defines list of tasks. `set_fact` is a task. If you want to do multiple `set_fact` you need to write multiple tasks (or set multiple facts in single `set_fact` command).
`name` is just annotation of task.
`when` is condition --- is current task should be ran?

I added `name` annotation for second task so it's now:

```
# If this is CoreOS, set python_bin to `/opt/bin/python`
- name: Set python_bin fact for CoreOS
  set_fact:
    python_bin: "/opt/bin/python"
  when: ansible_distribution == "CoreOS"

# If `python_bin` is not yet set (this implies that this is NOT CoreOS), set python_bin to `python`
- name: Set python_bin for other OS
  set_fact:
    python_bin: "python"
  when: python_bin is not defined
```

eparis:
Thus far the playbook has used the idiom

- set fact to default
- set fact to something other than default when something is true

Instead of the suggestion in https://github.com/kubernetes/contrib/pull/608#issuecomment-200457921 which does

- set fact when something is true
- set fact to default when fact is unset.

But either seem good to fix this issue, I'd just prefer it my way for consistency.

rutsky:
@eparis I reordered fact setting in this part of code according to your style.

ingvagabund:
@rutsky my question was what is the correct/preferred way for defining two or more facts. If define one fact per tasks or if define more facts under the same task. I don't have any preference for either way.

rutsky:
> @rutsky my question was what is the correct/preferred way for defining two or more facts. If define one fact per task or if define more facts under the same task. I don't have any preference for either way.

Code that I fixed doesn't have setting more than one fact: it tries to set single fact (`python_bin`) value based on some condition (`ansible_distribution == "CoreOS"`).
Ansible doesn't have proper `if ... then ... else ...` condition, so we need to do this in two tasks.

According to [documentation](https://docs.ansible.com/ansible/set_fact_module.html#examples) you can set multiple facts in single task:

```
- set_fact:
     one_fact: something
     other_fact: "{{ local_var * 2 }}"
     another_fact: "{{ some_registered_var.results | map(attribute='ansible_facts.some_fact') | list }}"
```

Technically we can use proper `if ... then ... else ...` clause of *Jinja*, like so:

```
- name: Set python_bin
  set_fact:
    python_bin: |-
      {% if ansible_distribution == "CoreOS" -%}
        "/opt/bin/python"
      {%- else %}
        "python"
      {%- endif %}
```

but @eparis style is better IMO.

ingvagabund:
@rutsky thanks.

LGTM.

eparis:
Please no jinja. I like for the ansible to look like it was written by a 13 year old child. That way I can understand it.

PR #610, opened 2016-03-23
add Vagrant working directory to gitignore

rutsky:
Vagrant writes in `.vagrant` directory cache information.

### DIFF
--- a/ansible/vagrant/.gitignore
+++ b/ansible/vagrant/.gitignore
@@ -1,1 +1,2 @@
+/.vagrant/
 openstack_config.yml


ansible/vagrant/.gitignore
```diff
@@ -1 +1,2 @@
+/.vagrant/
```
eparis: You don't need the leading slash, but I don't care enough to ask you to change it.

rutsky: I use leading slash to explicitly state that I'm ignoring `.vagrant` sub-directory in *this* directory.
If I'd use `.vagrant/` git will ignore `.vagrant` sub-directory in *this* directory and *all sub-directories*, which is not what I want, since we don't expect to see `.vagrant` in subdirectories.

@eparis does it make sense or I missing something?

PR #616, opened 2016-03-24
munger: update location of WordSepNormalizeFunc

justinsb:
It moved from k8s.io/kubernetes/pkg/util to
k8s.io/kubernetes/pkg/util/flag


justinsb:
Never mind ... I should be using godeps

eparis:
although feel free to update the godeps.  `make help` has all sorts of useful information so you don't have to remember things...

PR #619, opened 2016-03-24
fix ansible+vagrant+virtualbox+centos7 configuration

rutsky:
Changes in Ansible + Vagrant configuration:

* By default on VirtualBox public interface is not `eth0`, but `eth1`: I automatically use proper interface for flannel and when generating `/etc/hosts`.

  This should fix #586 .

* Fix `bin_dir` on CentOS with K8S installed from packages.

These changes are *almost* enough to start K8S cluster in Vagrant on VirtualBox + CentOS 7.
Unfortunately CentOS 7 yet have only [1.2.0alpha1](https://git.centos.org/summary/rpms!kubernetes.git) which doesn't include [commit with `successThreshold` and `failureThreshold` liveness probe parameter](https://github.com/kubernetes/kubernetes/commit/1e88a682da871a4a7f811dc5efcfe3890be8bbf8). This leads to errors in `kube-addons` like following one:

```
мар 24 09:43:38 kube-master kube-addons.sh[30042]: error validating "/etc/kubernetes/addons/dns/skydns-rc.yaml": error validating data: [found invalid field successThreshold for v1.Probe, found invalid field failureThreshold for v1.Probe]; if you choose to ignore these errors, turn validation off with --validate=false
```

To bypass this issue I locally patched kube-addons scripts to include `--validate=false` in `kubectl` invocations:

```
diff --git a/ansible/roles/kubernetes-addons/files/kube-addon-update.sh b/ansible/roles/kubernetes-addons/files/kube-addon-update.sh
index ede2832..ab0e023 100755
--- a/ansible/roles/kubernetes-addons/files/kube-addon-update.sh
+++ b/ansible/roles/kubernetes-addons/files/kube-addon-update.sh
@@ -259,7 +259,7 @@ function create-object() {
     log INFO "Creating new ${obj_type} from file ${file_path} in namespace ${namespace}, name: ${obj_name}"
     # this will keep on failing if the ${file_path} disappeared in the meantime.
     # Do not use too many retries.
-    run-until-success "${KUBECTL} create --namespace=${namespace} -f ${file_path}" ${NUM_TRIES} ${DELAY_AFTER_ERROR_SEC}
+    run-until-success "${KUBECTL} create --validate=false --namespace=${namespace} -f ${file_path}" ${NUM_TRIES} ${DELAY_AFTER_ERROR_SEC}
 }
 
 function update-object() {
[...]

Issue #586, opened 2016-03-18
ansible Flanneld on centos doesnt start


MansM:
TASK [flannel : Launch Flannel] ************************************************
changed: [kube-master]
fatal: [kube-node-1]: FAILED! => {"changed": false, "failed": true, "msg": "Job for flanneld.service failed because a timeout was exceeded. See \"systemctl status flanneld.service\" and \"journalctl -xe\" for details.\n"}
fatal: [kube-node-2]: FAILED! => {"changed": false, "failed": true, "msg": "Job for flanneld.service failed because a timeout was exceeded. See \"systemctl status flanneld.service\" and \"journalctl -xe\" for details.\n"}

journalctl -xe on node-1:
Mar 18 15:44:52 kube-node-1 flanneld[13777]: E0318 15:44:52.157605 13777 network.go:53] Failed to retrieve network config: client: etcd cluster is unavailable or misconfigured
Mar 18 15:44:53 kube-node-1 flanneld[13777]: E0318 15:44:53.158587 13777 network.go:53] Failed to retrieve network config: client: etcd cluster is unavailable or misconfigured
Mar 18 15:44:54 kube-node-1 flanneld[13777]: E0318 15:44:54.160032 13777 network.go:53] Failed to retrieve network config: client: etcd cluster is unavailable or misconfigured

[root@kube-master ~]# flanneld --version
0.5.3

[root@kube-master ~]# etcd --version
etcd Version: 2.2.2
Git SHA: 6335fdc
Go Version: go1.4.2
Go OS/Arch: linux/amd64

didnt make any changes to any vars
vagrant 1.8.1 (virtualbox)
ansible 2.0.1.0
OSX

rutsky:
@MansM take a look at my changes in #619. They fix flannel configuration issues on Vagrant + VirtualBox.

danehans:
@MansM can you pull the latest master and retry. I believe https://github.com/kubernetes/contrib/commit/33c1671bb42b13a7712771c68118a092565668c0 fixes your issue.

rutsky:
@danehans I don't think 33c1671 fixes @MansM issue: he runs on Ansible on Vagrant+Virtualbox, in this configuration non-default Ethernet device should be used as I described in #619.

danehans:
@rutsky thanks for the clarification. @MansM can you verify if https://github.com/kubernetes/contrib/pull/619 fixes your issue.

MansM:
[...]

Issue #508, opened 2016-02-22
ansible: ansible_fqdn fact causes failed deployments


danehans:
Per @eparis in [1] ansible_fqdn does not work in all situations. It is preferred to use a system fact IP address instead of ansible_fqdn.

[1] https://github.com/kubernetes/contrib/pull/244

Issue #428, opened 2016-02-01
Ansible: Flannel Install Fails on Ubuntu


rsmitty:
While attempting to deploy a k8s cluster on Ubuntu hosts, I've encountered an issue where flannel does not install. This is because of the [client.yml](https://github.com/kubernetes/contrib/blob/master/ansible/roles/flannel/tasks/client.yml) file and the way that the installation is handled. It wrongly assumes that flannel is in the package repository, regardless of OS (as long as it's not AtomicOS). This will need to be fixed to install from github releases for Ubuntu and I presume Debian. I'll take this on when I have a free moment, if someone will go ahead and assign it to me.

rsmitty:
Digging in further, it seems that none of the flannel playbooks support anything but RHEL/CentOS. Looks like this may be a bigger issue than I initially thought. Working around the initial issue lead to a failure in the next step when trying to put the sysconfig file down.

eparis:
At one point in time i made the playbook run on debian. I it's very hazy...  I remember being frustrated because of https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=792704

danehans:
@eparis @rsmitty I am working on adding CoreOS support to the Ansible playbooks and I am sseing issues with flanneld starting due to incorrect ENV key names in flanneld.j2. Please refer to [1] for the list of flannel environment variable names that are supported. For example,  flanneld.j2 uses FLANNEL_ETCD_KEY when it should be FLANNELD_ETCD_PREFIX.

[1] https://github.com/coreos/flannel#environment-variables

eparis:
ok, we might need to 'disto-ize' that. RHEL/Fedora use those as command line args, not as ENV vars. They are launching flannel with something like:
```
/usr/bin/flanneld -etcd-endpoints=${FLANNEL_ETCD} -etcd-prefix=${FLANNEL_ETCD_KEY} $FLANNEL_OPTIONS
```
I don't know how easy/possible it would be to get the distro to switch, so we might have to put that config file only on RHEL/Fedora and use a new config file with the ENV vars on Debian...

danehans:
@eparis I was wondering if that was the case, so I distro-ized the flanneld.j2. [2] is the PR that I'm ready to submit for adding CoreOS support... just waiting on the Corp CLA stuff to complete. It would be much appreciated if you could check out [2] and let me know if any changes are required for a LGTM.

[...]

PR #620, opened 2016-03-24
deploy to ubuntu 15.10 [WIP

feliksik:
I have recently been working to deploy to Ubuntu 15.10 (preparing for systemd in the 16.04 LTS release). I would like to contribute, but am new to the k8s community. I like to have some alignment that assures me we will push the same direction. This PR is intended to gather some insight on how to make the ansible module more portable, as I think that is nontrivial. 

I see multiple people are using the contrib/ansible repo ( #428 #619 #586 #508 #406, a simple search: https://github.com/kubernetes/contrib/search?utf8=%E2%9C%93&q=ansible&type=Issues ). Some nice people like @eparis @rsmitty @danehans @rutsky have been working on improving it. 

I see that we attempt to make a number of things flexible: 
- k8s version, and binary/source distribution (i see the source_type variable is used). 
- OS distribution (the `when: ansible_distribution != "CoreOS"` is used in roles/tasks/flannel/client). I am afraid this when: thingy is tricky to use for all kinds of different OS's. 
- the ip addresses of the peers, and the way in which to define this (`ansible_hostname` or `ansible_fqdn`, but what about if we want to use an internal network?)
- the users that are assumed to be present on the system (e.g. 'etcd' was a user not present on ubuntu; maybe no assumptions should be made at all)
- the location where `etcdctl` turns out to be installed (`usr/bin/etcdctl` didn't hold after installing on ubuntu)

Some choice to make things more portable: 
- standardize on certain choices (e.g. no matter the OS, the implementation of the etcdctl installer _must_ create a binary in /usr/bin ; but I think a better standard would be: no matter the OS, the binaries must end up in $PATH and we don't use a full path in calling the binary);
- be configurable (e.g. the IP address of the peers). You can configure in group_vars/all, but you can also split this into separate groups. Currently i see ansible_fqdn and ansible_hostnames used, but this is pretty big assumption (imagine eth1 is your internal net you which to use -- not external eth0)
- be modular (e.g. the tool that configures the network for ubuntu). 

Tools that create flexibility: 
- use of `group_vars`
- use of `host_vars`
- indirect variables (you can set `i_am_kubelet_with_name: {{ansible_hostname}}` in group_vars/nodes and it will be defined for all hosts in `nodes` group, but have a different value)
- modularity 1: dynamic filename of  `include: install_flannel_on_{{ansible_distribution}}.yml`
- modularity 2: running things conditionally with `when: ansible_distribution == "CoreOS"`

I happen to have quite some experience with Ansible and have made significant effort with group_vars, host_vars, and levels of indirection to make things highly configurable. Ansible is super-liberal in how to approach such issues, making things very much a matter of taste.
It is certainly possible do do it more robust/portable, but I also figured it is not trivial (especially since you don't easily test on all environments). I think it requires a well chosen set of best practices. Is there a philosophy on how to these portability issues? How feels/is in the lead?

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/620)

### DIFF
--- a/ansible/.gitignore
+++ b/ansible/.gitignore
@@ -2,3 +2,5 @@ myinventory
 inventory
 *.swp
 *~
+binaries
+host_vars
--- a/ansible/cluster.yml
+++ b/ansible/cluster.yml
@@ -3,6 +3,15 @@
 # with the default addons.
 
 - hosts: all
+  sudo: yes
+  gather_facts: true
+  tasks: 
+
+    - lineinfile: dest=/etc/hosts line="{{hostvars[groups['masters'][0]].static_lan_ip}}    {{groups['masters'][0]}}"
+
+    - apt: name=util-linux state=installed
+
+- hosts: all
   gather_facts: false
   sudo: yes
   roles:
--- a/ansible/download-release.sh
+++ b/ansible/download-release.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+set -euo pipefail
+IFS=$'\n\t'
+
+mkdir -p binaries
+
+SCRIPT_URL=https://raw.githubusercontent.com/kubernetes/kubernetes/master/cluster/ubuntu/download-release.sh
+
+curl $SCRIPT_URL | bash
--- a/ansible/group_vars/all.yml
+++ b/ansible/group_vars/all.yml
@@ -35,7 +35,7 @@ insecure_registrys:
 # Kubernetes services will get fake IP addresses from this range.
 # This range must not conflict with anything in your infrastructure. These
 # addresses do not need to be routable and must just be an unused block of space.
-kube_service_addresses: 10.254.0.0/16
+kube_service_addresses: 172.30.0.0/16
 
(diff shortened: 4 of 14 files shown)


feliksik:
I signed it!

ingvagabund:
Once https://github.com/kubernetes/contrib/pull/1698 and https://github.com/kubernetes/contrib/pull/1700 get merged, one can deploy the cluster on Ubuntu-16.04. With https://github.com/kubernetes/contrib/pull/1701 merged it can be done via Vagrant as well.

@feliksik will this be enough? Most likely Ubuntu-15.* should be working as well (after slight modification). Though, I could not find any Ubuntu-15.* vagrant box to test it.

test-foxish:
recomputing cla status...

feliksik:
@ingvagabund : I think this can be closed then, 15.04 is obsolete now anyway.

ingvagabund:
@feliksik thanks for letting me know.

PR #618, opened 2016-03-24
Fixed wrong KUBECTL_BIN

deromka:
Reverted the KUBECTL_BIN to the correct value: /usr/bin/kubectl, since the new one {{ bin_dir }} was referencing the common.yaml default which is /usr/local/bin
And there is not kubectl overthere, thus the addon services were not running

Here is the kube-addons service logs with the {{ bin_dir }} value:

`[root@kube-master system]# systemctl status -l kube-addons
● kube-addons.service - Kubernetes Addon Object Manager
   Loaded: loaded (/etc/systemd/system/kube-addons.service; enabled; vendor preset: disabled)
   Active: active (running) since Wed 2016-03-23 21:01:15 GMT; 12h ago
     Docs: https://github.com/GoogleCloudPlatform/kubernetes
 Main PID: 27455 (kube-addons.sh)
   CGroup: /system.slice/kube-addons.service
           ├─27455 /bin/bash /usr/libexec/kubernetes/kube-addons.sh
           └─93200 sleep .5

Mar 24 09:17:43 kube-master kube-addons.sh[27455]: /usr/libexec/kubernetes/kube-addons.sh: line 155: /usr/local/bin/kubectl: No such file or directory
Mar 24 09:17:43 kube-master kube-addons.sh[27455]: /usr/libexec/kubernetes/kube-addons.sh: line 155: /usr/local/bin/kubectl: No such file or directory
Mar 24 09:17:44 kube-master kube-addons.sh[27455]: /usr/libexec/kubernetes/kube-addons.sh: line 155: /usr/local/bin/kubectl: No such file or directory
Mar 24 09:17:44 kube-master kube-addons.sh[27455]: /usr/libexec/kubernetes/kube-addons.sh: line 155: /usr/local/bin/kubectl: No such file or directory
Mar 24 09:17:45 kube-master kube-addons.sh[27455]: /usr/libexec/kubernetes/kube-addons.sh: line 155: /usr/local/bin/kubectl: No such file or directory
Mar 24 09:17:45 kube-master kube-addons.sh[27455]: /usr/libexec/kubernetes/kube-addons.sh: line 155: /usr/local/bin/kubectl: No such file or directory
Mar 24 09:17:46 kube-master kube-addons.sh[27455]: /usr/libexec/kubernetes/kube-addons.sh: line 155: /usr/local/bin/kubectl: No such file or directory
[...]

PR #622, opened 2016-03-24
Updates the Default bin_dir for Ansible

danehans:
Previously, the default bin_dir did not work for some platforms.
The new default follows the contrib/init project settings.

Fixes Issue: https://github.com/kubernetes/contrib/pull/618


eparis:
I'm going to merge. If things break someone will yell.

Issue #623, opened 2016-03-24
Kube Addons python_bin fact is not being properly updated for CoreOS


danehans:
```
$ systemctl status kube-addons
● kube-addons.service - Kubernetes Addon Object Manager
   Loaded: loaded (/etc/systemd/system/kube-addons.service; enabled; vendor preset: disabled)
   Active: active (running) since Thu 2016-03-24 16:45:20 UTC; 1h 23min ago
     Docs: https://github.com/GoogleCloudPlatform/kubernetes
 Main PID: 5160 (kube-addons.sh)
   Memory: 980.0K
      CPU: 1min 839ms
   CGroup: /system.slice/kube-addons.service
           ├─ 5160 /bin/bash /opt/bin/kubernetes/kube-addons.sh
           └─26184 sleep 600

Mar 24 18:05:20 core-master kube-addons.sh[5160]: cat: write error: Broken pipe
Mar 24 18:05:20 core-master kube-addons.sh[5160]: /opt/bin/kubernetes/kube-addon-update.sh: line 102: python: command not found
Mar 24 18:05:20 core-master kube-addons.sh[5160]: /opt/bin/kubernetes/kube-addon-update.sh: line 102: python: command not found
Mar 24 18:05:20 core-master kube-addons.sh[5160]: cat: write error: Broken pipe
```

PYTHON_BIN env var from kube-addon-update.sh:
```
PYTHON_BIN=${PYTHON_BIN:-python}
```
^ should show PYTHON_BIN=${PYTHON_BIN:-/opt/bin/python} for CoreOS deployments.

danehans:
Closing since ^ fixed the issue.

PR #624, opened 2016-03-24
Fixes python_bin path for Kube Addons in Ansible

danehans:
Previously, the python_path was not being properly updated for
CoreOS systems. This patch moves python_bin fact management
from the kubernetes-addons role to the kubernetes role.

Fixes Issue: https://github.com/kubernetes/contrib/issues/623


danehans:
@ingvagabund @atosatto I think this commit should work with your systems. I would appreciate if you could take a look to make sure.

atosatto:
I confirm you that it works (@eparis has been faster than me!) :)

danehans:
yay!!! thx @atosatto

Issue #577, opened 2016-03-17
Support Contiv Networking with Kubernetes


jojimt:
Contiv (https://github.com/contiv) provides policy based container networking with kubernetes integration via CNI plugin as documented in https://github.com/contiv/netplugin/tree/master/mgmtfn/k8splugin. I would like to upstream the corresponding ansible changes to the contrib/ansible repo.

jojimt:
Addressed by https://github.com/kubernetes/contrib/pull/783

PR #625, opened 2016-03-24
Add contiv networking support (https://github.com/kubernetes/contrib/issues/577)

jojimt:
This PR brings in ansible changes needed to setup kubernetes with Contiv networking as outlined in https://github.com/kubernetes/contrib/issues/577.

Summary of changes:
1) Added a new role (contiv) that holds all contiv specific config code
2) Minor tweaks to etcd setup so contiv can access etcd from any node in proxy-mode
3) Exposed a few options via variables, with default values set so the default behavior will be unchanged.
4) Added missing "validate_certs: False" for a couple of addon file fetches.

### DIFF
--- a/ansible/cluster.yml
+++ b/ansible/cluster.yml
@@ -45,6 +45,12 @@
   tags:
     - network-service-install
 
+# install contiv netmaster
+- hosts: masters
+  sudo: yes
+  roles:
+    - { role: contiv, contiv_role: netmaster, when: networking == 'contiv' }
+
 # install kube master services
 - hosts: masters
   sudo: yes
@@ -79,3 +85,9 @@
     - { role: opencontrail-provision, when: networking == 'opencontrail' }
   tags:
     - network-service-config
+
+# install contiv netplugin
+- hosts: nodes
+  sudo: yes
+  roles:
+    - { role: contiv, contiv_role: netplugin, when: networking == 'contiv' }
--- a/ansible/group_vars/all.yml
+++ b/ansible/group_vars/all.yml
@@ -96,3 +96,21 @@ dns_replicas: 1
 
 # See kube documentation for apiserver runtime config options.  Example below enables HPA, deployments features.
 # apiserver_extra_args: "--runtime-config=extensions/v1beta1/deployments=true"
+
+# The port that the Kubernetes apiserver component listens on.
+kube_master_api_port: 443
+
+# This directory is where all the additional config stuff goes
+# the kubernetes normally puts in /srv/kubernets.
+# This puts them in a sane location.
+# Editting this value will almost surely break something. Don't
+# change it. Things like the systemd scripts are hard coded to
+# look in here. Don't do it.
+kube_config_dir: /etc/kubernetes
+
+# This is where all the cert scripts and certs will be located
+kube_cert_dir: "{{ kube_config_dir }}/certs"
+
+# This is the IP where the contiv netmaster can be accessed.
+# Usually, comes via environment or command line
+# contiv_service_vip: 10.6.3.101
--- a/ansible/roles/contiv/defaults/main.yaml
+++ b/ansible/roles/contiv/defaults/main.yaml
@@ -1,0 +1,40 @@
+# This is the ip address of the interface used for control communication within the cluster. It
+# needs to be reachable from all nodes.
+# Defaults to the address used by ansible to reach the node.
+# If using vagrant, this will need to be overriden or edited here.
+contiv_control_ip: "{{ hostvars[inventory_hostname]['ansible_default_ipv4']['address'] }}"
+
+# This is the host interface that contiv uses for all inter-host tenant traffic. Typically set
+# in the inventory file.
+#contiv_network_if: ethX
+
+# Generic way to specify the bin path
+bin_dir: ~/go/bin
+
+# This is the path where contiv build output is kept
+contiv_bin_path: {{ bin_dir }}
+
+# This is the path where contiv binaries are placed on the target
+contiv_bin_dest: /usr/bin
+
+# This is where kubelet looks for plugin files
+kube_plugin_dir: /usr/libexec/kubernetes/kubelet-plugins/net/exec
+
+# Specifies routed mode vs bridged mode for networking (bridge | routing)
+# if you are using an external router for all routing, you should select bridge here
+contiv_fwd_mode: routing
+
+# contiv_def_subnet is the subnet pods will be assigned by default (i.e. pod spec does not
+# specify a network)
+contiv_def_subnet: "20.1.1.0/24"
+contiv_defnet_gw: 20.1.1.254
+
+# The following are aci specific parameters. These must be set only if aci integration
+# is desired. Otherwise, you can ignore these.
+apic_url: ""
+apic_username: ""
+apic_password: ""
+apic_leaf_nodes: ""
+apic_phys_dom: ""
+apic_contracts_unrestricted_mode: no
+apic_epg_bridge_domain: not_specified
--- a/ansible/roles/contiv/files/contiv_cni.conf
+++ b/ansible/roles/contiv/files/contiv_cni.conf
@@ -1,0 +1,5 @@
+{
+  "cniVersion": "0.1.0",
+  "name": "contiv-poc",
+  "type": "contivk8s.bin"
+}
--- a/ansible/roles/contiv/tasks/main.yml
+++ b/ansible/roles/contiv/tasks/main.yml
@@ -1,0 +1,15 @@
+# initial set up of netmaster
+- include: netmaster.yml
+  when: contiv_role == "netmaster"
+
+# clean restart of netmaster
+- include: restart_netmaster.yml tags=contiv_restart
+  when: contiv_role == "netmaster"
+
+# initial set up of netplugin
+- include: netplugin.yml
+  when: contiv_role == "netplugin"
+
+# clean restart of netplugin
+- include: restart_netplugin.yml tags=contiv_restart
+  when: contiv_role == "netplugin"
--- a/ansible/roles/contiv/tasks/netmaster.yml
+++ b/ansible/roles/contiv/tasks/netmaster.yml
@@ -1,0 +1,31 @@
+- name: ensure contiv_bin_dest exists
+  file: path={{ contiv_bin_dest }} recurse=yes state=directory
+
+- name: copy netmaster exec file
+  copy: src={{ contiv_bin_path }}/netmaster dest={{ contiv_bin_dest }}/netmaster mode=0755
+
+- name: copy netctl exec file
+  copy: src={{ contiv_bin_path }}/netctl dest={{ contiv_bin_dest }}/netctl mode=0755
+
+- name: copy netmaster setup script
+  template: src=nw_setup.j2 dest={{ contiv_bin_dest }}/nw_setup.sh mode=0755
+
+- name: copy environment file for netmaster
+  copy: src=netmaster.env dest=/etc/default/netmaster
+
+- name: copy systemd units for netmaster
+  copy: src=netmaster.service dest=/etc/systemd/system/netmaster.service
+
+- name: setup netmaster host alias
+  lineinfile: dest=/etc/hosts line="{{ contiv_service_vip }} netmaster"
+
+- name: restart netmaster
+  service: name=netmaster state=restarted
+
+- name: wait for netmaster
+  pause: seconds=60
+
+- include: aci.yml
+
+- name: configure network
+  shell: {{ contiv_bin_dest }}/nw_setup.sh
--- a/ansible/roles/contiv/tasks/netplugin.yml
+++ b/ansible/roles/contiv/tasks/netplugin.yml
@@ -1,0 +1,86 @@
+- name: ensure pluginDir exists
+  file: path={{ kube_plugin_dir }} recurse=yes state=directory
+
+- name: copy contiv cni conf file
+  copy: src=contiv_cni.conf dest={{ kube_plugin_dir }}/contiv_cni.conf
+
+- name: ensure cni bin dir exists
+  file: path=/opt/cni/bin recurse=yes state=directory
+
+- name: copy contiv cni exec file
+  copy: src={{ contiv_bin_path }}/contivk8s dest=/opt/cni/bin/contivk8s.bin mode=0755
+
+- name: ensure contiv config directory exists
+  file: path=/opt/contiv/config recurse=yes state=directory
+
+- name: setup config for the contiv cni plugin
+  template: src=contiv.cfg.j2 dest=/opt/contiv/config/contiv.json
+
+- name: ensure contiv_bin_dest exists
+  file: path={{ contiv_bin_dest }} recurse=yes state=directory
+
+- name: copy netplugin exec file
+  copy: src={{ contiv_bin_path }}/netplugin dest={{ contiv_bin_dest }}/netplugin mode=0755
+
+- name: get openstack kilo rpm
+  get_url:
+    url: https://repos.fedorapeople.org/repos/openstack/openstack-kilo/rdo-release-kilo-1.noarch.rpm
+    dest: /tmp/rdo-release-kilo-1.noarch.rpm
+    validate_certs: False
+  environment:
+    http_proxy: "{{ http_proxy|default('') }}"
+    https_proxy: "{{ https_proxy|default('') }}"
+    no_proxy: "{{ no_proxy|default('') }}"
+  when: ansible_distribution == "CentOS"
+  tags:
+       - ovs_install
+
+- name: install openstack kilo
+  yum: name=/tmp/rdo-release-kilo-1.noarch.rpm state=present
+  when: ansible_distribution == "CentOS"
+  tags:
+       - ovs_install
+
+- name: install ovs
+  environment:
+    http_proxy: "{{ http_proxy|default('') }}"
+    https_proxy: "{{ https_proxy|default('') }}"
+    no_proxy: "{{ no_proxy|default('') }}"
+  yum: name=openvswitch state=present
+  when: ansible_distribution == "CentOS"
+  tags:
+       - ovs_install
+
+- name: install ntp
+  yum: name=ntp state=present
+  when: ansible_distribution == "CentOS"
+
+- name: start ntp
+  shell: systemctl enable ntpd
+  when: ansible_distribution == "CentOS"
+
+- name: start ovs service
+  service: "name=openvswitch enabled=yes state=started"
+  when: ansible_distribution == "CentOS"
+
+- name: setup ovs
+  shell: "ovs-vsctl set-manager {{ item }}"
+  with_items:
+    - "tcp:127.0.0.1:6640"
+    - "ptcp:6640"
+
+- name: copy environment file for netplugin
+  template: src=netplugin.j2 dest=/etc/default/netplugin mode=0644
+
+- name: copy systemd units for netplugin
+  copy: src=netplugin.service dest=/etc/systemd/system/netplugin.service
+
+- name: setup netmaster host alias
+  lineinfile: dest=/etc/hosts line="{{ contiv_service_vip }} netmaster"
+
+- name: restart netplugin
+  service: name=netplugin state=restarted
+
+- name: restart kubelet
+  service: name=kubelet state=restarted
+
--- a/ansible/roles/contiv/tasks/restart_netmaster.yml
+++ b/ansible/roles/contiv/tasks/restart_netmaster.yml
@@ -1,0 +1,35 @@
+- name: stop netmaster
+  service: name=netmaster state=stopped
+
+- name: stop aci-gw
+  service: name=aci-gw state=stopped
+  ignore_errors: yes
+
+- name: clean netmaster state
+  shell: etcdctl rm --recursive /contiv.io || true
+
+- name: copy netmaster exec file
+  copy: src={{ contiv_bin_path }}/netmaster dest={{ contiv_bin_dest }}/netmaster mode=0755
+
+- name: copy netctl exec file
+  copy: src={{ contiv_bin_path }}/netctl dest={{ contiv_bin_dest }}/netctl mode=0755
+
+- name: copy n/w setup script
+  template: src=nw_setup.j2 dest={{ contiv_bin_dest }}/nw_setup.sh mode=0755
+
+- name: copy environment file for netmaster
+  copy: src=netmaster.env dest=/etc/default/netmaster
+
+- name: copy systemd units for netmaster
+  copy: src=netmaster.service dest=/etc/systemd/system/netmaster.service
+
+- name: start netmaster
+  service: name=netmaster state=started
+
+- name: wait for netmaster
+  pause: seconds=60
+
+- include: aci.yml
+
+- name: setup network
+  shell: {{ contiv_bin_dest }}/nw_setup.sh
--- a/ansible/roles/contiv/tasks/restart_netplugin.yml
+++ b/ansible/roles/contiv/tasks/restart_netplugin.yml
@@ -1,0 +1,71 @@
+# stop services
+- name: stop kubelet
+  service: name=kubelet state=stopped
+
+- name: stop docker
+  service: name=docker state=stopped
+
+- name: stop netplugin
+  service: name=netplugin state=stopped
+
+# perform clean up
+- name: delete contivk8s log file
+  file: path=/var/log/contivk8s.log state=absent
+
+- name: copy clean up script
+  copy: src=netplugin_clean.sh dest={{ contiv_bin_dest }}/netplugin_clean.sh mode=0755
+
+- name: clean up
+  shell: {{ contiv_bin_dest }}/netplugin_clean.sh || true
+
+# set up services
+- name: ensure pluginDir exists
+  file: path={{ kube_plugin_dir }} recurse=yes state=directory
+
+- name: copy contiv cni conf file
+  copy: src=contiv_cni.conf dest={{ kube_plugin_dir }}/contiv_cni.conf
+
+- name: ensure cni bin dir exists
+  file: path=/opt/cni/bin recurse=yes state=directory
+
+- name: copy contiv cni exec file
+  copy: src={{ contiv_bin_path }}/contivk8s dest=/opt/cni/bin/contivk8s.bin mode=0755
+
+- name: ensure contiv config directory exists
+  file: path=/opt/contiv/config recurse=yes state=directory
+
+- name: setup config for the contiv cni plugin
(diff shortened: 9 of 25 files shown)


ansible/cluster.yml
```diff
@@ -65,6 +71,7 @@
 # install kubernetes on the nodes
 - hosts: nodes
   sudo: yes
+  serial: 5
```
danehans: Is this required for adding contiv support? If not, can you please create a separate pr for the patch?

ansible/group_vars/all.yml
```diff
@@ -96,3 +96,21 @@ dns_replicas: 1
 
 # See kube documentation for apiserver runtime config options.  Example below enables HPA, deployments features.
 # apiserver_extra_args: "--runtime-config=extensions/v1beta1/deployments=true"
+
```
danehans: lines 99-101 already exist and should not be in this pr: https://github.com/kubernetes/contrib/blob/master/ansible/group_vars/all.yml#L40-L41

ansible/group_vars/all.yml
```diff
@@ -96,3 +96,21 @@ dns_replicas: 1
...
+# The port that the Kubernetes apiserver component listens on.
+kube_master_api_port: 443
+
+# This directory is where all the additional config stuff goes
+# the kubernetes normally puts in /srv/kubernets.
+# This puts them in a sane location.
+# Editting this value will almost surely break something. Don't
+# change it. Things like the systemd scripts are hard coded to
+# look in here. Don't do it.
+kube_config_dir: /etc/kubernetes
```
danehans: This already exists: roles/kubernetes/defaults/main.yml
kube_config_dir: /etc/kubernetes

We do not want to expose this in group_vars.

ansible/cluster.yml
jojimt: Good point. This is only required when the cluster has many nodes (I tested with 20). If it will be helpful, I can take it out of this PR.

ansible/group_vars/all.yml
```diff
@@ -96,3 +96,21 @@ dns_replicas: 1
...
+
+# This is where all the cert scripts and certs will be located
+kube_cert_dir: "{{ kube_config_dir }}/certs"
```
danehans: Same feedback as above. From what I gather, nothing should change in this file except adding contiv to this line: https://github.com/kubernetes/contrib/blob/master/ansible/group_vars/all.yml#L49

ansible/roles/contiv/defaults/main.yaml
```diff
@@ -0,0 +1,31 @@
...
+# Defaults to the address used by ansible to reach the node.
+# If using vagrant, this will need to be overriden or edited here.
+contiv_control_ip: "{{ hostvars[inventory_hostname]['ansible_default_ipv4']['address'] }}"
+
+# This is the host interface that contiv uses for all inter-host tenant traffic. Typically set
+# in the inventory file.
+#contiv_network_if: ethX
+
+# This is the path where contiv binaries are kept
+contiv_bin_path: ~/go/bin
```
danehans: Can centavo follow the same bin_dir pattern found here: roles//common/defaults/main.yaml:
bin_dir: /usr/bin

bin_dir fact is used for supporting multiple distros

ansible/roles/contiv/defaults/main.yaml
```diff
@@ -0,0 +1,31 @@
...
+# if you are using an external router for all routing, you should select bridge here
+contiv_fwd_mode: routing
+
+# contiv_def_subnet is the subnet pods will be assigned by default (i.e. pod spec does not
+# specify a network)
+contiv_def_subnet: "20.1.1.0/24"
+contiv_defnet_gw: 20.1.1.254
+
+# The following are aci specific parameters. These must be set only if aci integration
+# is desired. Otherwise, you can ignore these.
```
danehans: Do you have a link that explains aci integration in more detail? If so, pls add it as a remark.

ansible/roles/contiv/files/contiv_cni.conf
```diff
@@ -0,0 +1,5 @@
+{
+  "cniVersion": "0.1.0",
```
danehans: can u parameterize the chi version instead of hard coding. Throw the param/value in defaults/main

ansible/group_vars/all.yml
jojimt: Actually, this is needed for other roles. That's why it is moved to group_vars now.

ansible/roles/contiv/tasks/main.yml
```diff
@@ -0,0 +1,19 @@
...
+
+# initial set up of netplugin
+- include: netplugin.yml
+  when: contiv_role == "netplugin"
+
+# clean restart of netplugin
+- include: restart_netplugin.yml tags=contiv_restart
+  when: contiv_role == "netplugin"
+
+# copy/load images used for demo
```
danehans: I would like to see the demo stuff removed. If the demo stuff ids used for testing, then I'm all for that. Lets just change the name accordingly and make sure their is a test clean-up after a successful test.

ansible/roles/contiv/tasks/netmaster.yml
```diff
@@ -0,0 +1,29 @@
+
+- name: copy netmaster exec file
+  copy: src={{ contiv_bin_path }}/netmaster dest=/usr/bin/netmaster mode=0755
```
danehans: We can't have /usr/bin/ hardcoded b/c /usr/bin/ is RO on coreos. Can you follow the existing design pattern for copying bins?

ansible/group_vars/all.yml
jojimt: I am installing the certificates on the nodes so the contiv client can access the api server. That's why this is moved to group_vars.

ansible/roles/contiv/tasks/netmaster.yml
```diff
@@ -0,0 +1,29 @@
...
+- name: copy systemd units for netmaster
+  copy: src=netmaster.service dest=/etc/systemd/system/netmaster.service
+
+- name: setup netmaster host alias
+  lineinfile: dest=/etc/hosts line="{{ contiv_service_vip }} netmaster"
+
+- name: restart netmaster
+  service: name=netmaster state=restarted
+
+- name: wait for netmaster
```
danehans: Is this needed? I believe service: name=netmaster state=restarted will fail if the restart is not successful. Unusable runs are long, so we want to everything we can to make things quick.

ansible/group_vars/all.yml
jojimt: Not sure why it appears after the rebase. No conflicts are marked either.

ansible/roles/contiv/defaults/main.yaml
jojimt: Yes, will do.

ansible/roles/contiv/tasks/main.yml
jojimt: Yes, it is used for testing. I will update the names.

ansible/roles/contiv/tasks/netmaster.yml
jojimt: Sure.

ansible/roles/contiv/tasks/netmaster.yml
jojimt: This is needed for now. I will optimize it to the exact required delay later.

ansible/roles/contiv/tasks/netplugin.yml
```diff
@@ -0,0 +1,81 @@
+- name: ensure pluginDir exists
+  file: path=/usr/libexec/kubernetes/kubelet-plugins/net/exec recurse=yes state=directory
```
danehans: We will need to work around this for cores support b/c /usr/libexec is RO.

ansible/roles/contiv/tasks/netplugin.yml
```diff
@@ -0,0 +1,81 @@
...
+
+- name: copy contiv cni conf file
+  copy: src=contiv_cni.conf dest=/usr/libexec/kubernetes/kubelet-plugins/net/exec/contiv_cni.conf
+
+- name: ensure cni bin dir exists
```
danehans: can the chi bin live with the rest of the bins using the bin_dir variable? I We want to have a single location where we manage bins.

ansible/roles/contiv/tasks/netplugin.yml
jojimt: Sure, I can do that. However, if I remember right, this path is hard-coded in kubelet.

ansible/roles/contiv/tasks/netplugin.yml
jojimt: As of now, no.

ansible/roles/contiv/tasks/netplugin.yml
```diff
@@ -0,0 +1,81 @@
...
+
+- name: get openstack kilo rpm
+  get_url:
+    url: https://repos.fedorapeople.org/repos/openstack/openstack-kilo/rdo-release-kilo-1.noarch.rpm
+    dest: /tmp/rdo-release-kilo-1.noarch.rpm
+    validate_certs: False
+  environment:
+    http_proxy: "{{ http_proxy|default('') }}"
+    https_proxy: "{{ https_proxy|default('') }}"
+  when: ansible_os_family == "RedHat"
```
danehans: Will ^ work using ansible_distribution instead? The rest of the project uses ansible_distribution, so it would be nice to continue using that fact.

ansible/roles/contiv/tasks/netplugin.yml
```diff
@@ -0,0 +1,81 @@
...
+  copy: src={{ contiv_bin_path }}/netplugin dest=/usr/bin/netplugin mode=0755
+
+- name: get openstack kilo rpm
+  get_url:
+    url: https://repos.fedorapeople.org/repos/openstack/openstack-kilo/rdo-release-kilo-1.noarch.rpm
+    dest: /tmp/rdo-release-kilo-1.noarch.rpm
+    validate_certs: False
+  environment:
+    http_proxy: "{{ http_proxy|default('') }}"
+    https_proxy: "{{ https_proxy|default('') }}"
```
danehans: I think this also needs no_proxy. See the following for an example:
roles//etcd/tasks/github-release.yml

ansible/roles/contiv/tasks/netplugin.yml
```diff
@@ -0,0 +1,81 @@
...
+- name: ensure contiv config directory exists
+  file: path=/opt/contiv/config recurse=yes state=directory
+
+- name: setup config for the contiv cni plugin
+  template: src=contiv.cfg.j2 dest=/opt/contiv/config/contiv.json
+
+- name: copy netplugin exec file
+  copy: src={{ contiv_bin_path }}/netplugin dest=/usr/bin/netplugin mode=0755
+
+- name: get openstack kilo rpm
```
danehans: This will be another problem for supporting coreOS. Do you mind opening GH Issues for these problem ares so we don't loose track of them?

ansible/roles/contiv/tasks/netplugin.yml
jojimt: Sorry, did not understand this comment. Can you explain what you mean by "Will ^ work using ansible_distribution instead?"

ansible/roles/contiv/tasks/netplugin.yml
```diff
@@ -0,0 +1,81 @@
...
+  environment:
+    http_proxy: "{{ http_proxy|default('') }}"
+    https_proxy: "{{ https_proxy|default('') }}"
+  when: ansible_os_family == "RedHat"
+  tags:
+       - ovs_install
+
+- name: install openstack kilo
+  yum: name=/tmp/rdo-release-kilo-1.noarch.rpm state=present
+  when: ansible_os_family == "RedHat"
```
danehans: s/ansible_os_family == "RedHat"/ansible_distribution == "CentOS"/g

ansible/roles/contiv/tasks/netplugin.yml
```diff
@@ -0,0 +1,81 @@
...
+  tags:
+       - ovs_install
+
+- name: install ovs
+  environment:
+    http_proxy: "{{ http_proxy|default('') }}"
+    https_proxy: "{{ https_proxy|default('') }}"
```
danehans: add no_proxy

ansible/roles/contiv/tasks/netplugin.yml
```diff
@@ -0,0 +1,81 @@
...
+  yum: name=openvswitch state=present
+  when: ansible_os_family == "RedHat"
```
danehans: s/ansible_os_family == "RedHat"/ansible_distribution == "CentOS"/g

ansible/roles/contiv/tasks/netplugin.yml
```diff
@@ -0,0 +1,81 @@
...
+  tags:
+       - ovs_install
+
+- name: install ntp
```
danehans: Is ntp required for contiv? If not, pls remove. I think it's a good idea to add nt. support to the common role, but that should be done in a separate PR if centavo does not require ntp.

ansible/roles/contiv/tasks/netplugin.yml
jojimt: Normally, it won't require it. However, I guess, it makes sense to have no_proxy also exported when we export the proxy. I will add it.

ansible/roles/contiv/tasks/netplugin.yml
jojimt: I don't really know the details of all CoreOS issues. If you're aware of the issue, why don't you open it?

ansible/roles/contiv/tasks/netplugin.yml
jojimt: Yes, it is required.

ansible/roles/contiv/tasks/restart_netmaster.yml
```diff
@@ -0,0 +1,35 @@
+- name: stop netmaster
```
danehans: restart_netmaster.yml performs the same tasks as net master.yml? Is this really needed? Would a handler and notify help? Take a look at the following for add'l info on the handler/notify idea:

roles/docker/handlers/main.yml

ansible/roles/contiv/tasks/restart_netplugin.yml
```diff
@@ -0,0 +1,71 @@
+# stop services
```
danehans: same with restart_netmaster.yml. It would be really nice if we can get away from having to duplicate all the code in the start netplugin/master just to restart the services.

ansible/roles/contiv/tasks/restart_netmaster.yml
jojimt: Yes, this is needed because sometimes we want to explicitly restart the networking.

ansible/roles/etcd/templates/etcd.conf.j2
```diff
@@ -1,40 +1,48 @@
 {% macro initial_cluster() -%}
 {% for host in groups[etcd_peers_group] -%}
 {% if loop.last -%}
-{{ hostvars[host]['ansible_hostname'] }}={{ etcd_peer_url_scheme }}://{{ hostvars[host].ansible_default_ipv4.address }}:{{ etcd_peer_port }}
+{{ hostvars[host]['ansible_hostname'] }}={{ etcd_peer_url_scheme }}://{{ hostvars[host]['ansible_fqdn'] }}:{{ etcd_peer_port }}
```
danehans: you need to use ansible_default_ipv4.address and not ansible_fqdn

ansible/roles/etcd/templates/etcd.conf.j2
```diff
@@ -1,40 +1,48 @@
...
 {%- else -%}
-{{ hostvars[host]['ansible_hostname'] }}={{ etcd_peer_url_scheme }}://{{ hostvars[host].ansible_default_ipv4.address }}:{{ etcd_peer_port }},
+{{ hostvars[host]['ansible_hostname'] }}={{ etcd_peer_url_scheme }}://{{ hostvars[host]['ansible_fqdn'] }}:{{ etcd_peer_port }},
```
danehans: you need to use ansible_default_ipv4.address and not ansible_fqdn

ansible/roles/etcd/templates/etcd.conf.j2
```diff
@@ -1,40 +1,48 @@
...
 {%- endif -%}
 {% endfor -%}
 {% endmacro -%}
 
-{% if groups[etcd_peers_group] and groups[etcd_peers_group] | length > 1 %}
+{% if groups[etcd_peers_group] and groups[etcd_peers_group] | length > 0 %}
```
danehans: Is length > 0 required for contiv support. If so, please explain.

ansible/roles/etcd/templates/etcd.conf.j2
```diff
@@ -1,40 +1,48 @@
...
 #ETCD_SNAPSHOT_COUNTER="10000"
 #ETCD_HEARTBEAT_INTERVAL="100"
 #ETCD_ELECTION_TIMEOUT="1000"
-ETCD_LISTEN_CLIENT_URLS="{{ etcd_listen_client_urls }}"
 #ETCD_MAX_SNAPSHOTS="5"
 #ETCD_MAX_WALS="5"
 #ETCD_CORS=""
 
-{% if groups[etcd_peers_group] and groups[etcd_peers_group] | length > 1 %}
+
```
danehans: Are you adding support for etcd ha below? If so, @adamschaub is working on that in a separate pr. Please remove. I'm a bit confused on why all these etc changes are required for Contiv.

ansible/roles/etcd/templates/etcd.conf.j2
jojimt: Why is that? In my set up, ansible_default_ipv4.address actually gives a syntax error. It can be ['ansible_default_ipv4'].address. But ansible_fqdn works just fine. Can you explain what is the issue here?

ansible/roles/kubernetes-addons/tasks/cluster-monitoring.yml
```diff
@@ -13,6 +13,7 @@
     url=https://raw.githubusercontent.com/GoogleCloudPlatform/kubernetes/master/cluster/addons/cluster-monitoring/influxdb/{{ item }}
     dest="{{ local_temp_addon_dir }}/cluster-monitoring/{{ item }}.j2"
     force=yes
+    validate_certs=False
```
danehans: pls remove and do an overall scrub on changes that are outside the scope of what's needed to add Contiv support.

ansible/roles/etcd/templates/etcd.conf.j2
jojimt: Yes, because we run etcd in proxy mode on all the nodes and non-proxy mode on a single master.

ansible/roles/kubernetes-addons/tasks/dns.yml
```diff
@@ -13,6 +13,7 @@
     url=https://raw.githubusercontent.com/GoogleCloudPlatform/kubernetes/master/cluster/addons/dns/skydns-rc.yaml.in
     dest="{{ local_temp_addon_dir }}/dns/skydns-rc.yaml.j2"
     force=yes
+    validate_certs=False
```
danehans: pls remove and do an overall scrub on changes that are outside the scope of what's needed to add Contiv support. If their is an issue with cert validation, pls open a GH issue and provide an explanation or a a PR that explains and fixes the issue.

ansible/roles/kubernetes-addons/tasks/dns.yml
```diff
@@ -38,6 +39,7 @@
     url=https://raw.githubusercontent.com/GoogleCloudPlatform/kubernetes/master/cluster/addons/dns/skydns-svc.yaml.in
     dest="{{ local_temp_addon_dir }}/dns/skydns-svc.yaml.j2"
     force=yes
+    validate_certs=False
```
danehans: pls remove and do an overall scrub on changes that are outside the scope of what's needed to add Contiv support. If their is an issue with cert validation, pls open a GH issue and provide an explanation or a a PR that explains and fixes the issue.

ansible/roles/etcd/templates/etcd.conf.j2
jojimt: No, I am only adding proxy mode support as contiv requires it.

ansible/roles/kubernetes/tasks/secrets.yml
```diff
@@ -39,10 +39,42 @@
   set_fact:
```
danehans: All these changes seems outside the scope of Contiv support. If I'm wrong, please explain.

ansible/roles/node/defaults/main.yml
```diff
@@ -1,2 +1,4 @@
 kubelet_working_dir: /var/lib/kubelet
 localBuildOutput: ../../_output/local/go/bin
+# Specifies the number of pods allowed on a node.
+max_pods: "40"
```
danehans: This should be separate of Contiv support, right?

ansible/roles/kubernetes-addons/tasks/cluster-monitoring.yml
jojimt: The problem is that the run fails without this. When I see that other similar code has this set, it seems reasonable.

ansible/roles/node/templates/kubelet.j2
```diff
@@ -14,12 +14,16 @@ KUBELET_HOSTNAME="--hostname-override={{ inventory_hostname }}"
 KUBELET_API_SERVER="--api-servers=https://{{ groups['masters'][0] }}:{{ kube_master_api_port }}"
 
 {% set kubelet_options = [] -%}
+{{ kubelet_options.append('--max-pods=' + max_pods|string)|default('', true) -}}
```
danehans: This should be separate of Contiv support, right?

ansible/roles/kubernetes-addons/tasks/dns.yml
jojimt: Same response as above.

ansible/roles/kubernetes/tasks/secrets.yml
jojimt: No, this is required so we can access the api server using certificate auth.

ansible/roles/node/defaults/main.yml
jojimt: Sure, I can remove this.

ansible/cluster.yml
danehans: We need the PR to only contain changes that are required for adding Contiv. Any additional changes that contrib will benefit from, pls create separate pr's.

ansible/group_vars/all.yml
danehans: Weird, take them out then. We need to make sure this PR does not change these params.

ansible/group_vars/all.yml
danehans: Please explain, if the centavo roles needs this, then the centavo role should have a dependency on the cabernets role. Checkout how open contrail does this:
```
$ cat roles/opencontrail/meta/main.yml 
---
dependencies:
  - { role: common }
  - { role: kubernetes }
```

ansible/group_vars/all.yml
danehans: If the Contiv roles needs this, then the Contiv role should have a dependency on the Kubernetes role. Checkout how opencontrail does this:
```
$ cat roles/opencontrail/meta/main.yml 
---
dependencies:
  - { role: common }
  - { role: kubernetes }
```

We def do not want to expose these dir's in group_vars.

ansible/roles/contiv/tasks/main.yml
danehans: OK, I think it's also important to include a task that cleans-up after the test.

ansible/group_vars/all.yml
jojimt: I need to template the location of certificates so contiv client can access the kube api server. Please see ansible/roles/contiv/templates/contiv.cfg.j2. That's why these variables need to be exposed. What would be the other way to do that?

ansible/roles/contiv/tasks/netplugin.yml
danehans: Right, but we will need to use an Ansible role outside of contrib that creates the dir using cloud-init to support coreOS. In general, we want to eliminate as much hard-coding in tasks as possible.

ansible/roles/contiv/tasks/netplugin.yml
danehans: OK, I will.

ansible/roles/contiv/tasks/netplugin.yml
danehans: The rest of the contrib project uses the ansible_distribution fact for OS detection. For example: ansible_distribution == "CentOS"

Can you use ansible_distribution == "CentOS" instead of ansible_os_family == "RedHat" so the Contiv role follows the same design pattern as the rest of the project?

jojimt: OK, will change that.

ansible/roles/kubernetes-addons/tasks/cluster-monitoring.yml
danehans: It looks like you found a bug, where some add ons are missing this. I agree it's needed. For the purposes of keeping the PR clean, add notes to the PR description for changes above and beyond Contiv support. This helps if we need to revert, etc.. for PR's to match what changes they make.

danehans:
@jojimt my vagrant install fails when using Contiv:

```
TASK [contiv : copy netmaster exec file] ***************************************
fatal: [kube-master]: FAILED! => {"changed": false, "failed": true, "msg": "could not find src=/Users/daneyonhansen/go/bin/netmaster"}
```

Seems like you need to create contiv_bin_path before copy netmaster exec file. Are you using contrib vagrant to test or are you using a different test environment?

jojimt:
Yes, you need to create that directory on a local machine. I use a shell script to do that before starting vagrant.

danehans:
I think the ansible contiv role should be doing that. The other roles provide similar functionality.

danehans:
@jojimt can you send me the shell script so I can replicate and possibly add it to vagrant. As I mentioned above though, I believe the ansible role should be creating that bin dir.

jojimt:
Here is an example: https://github.com/contiv/netplugin/blob/master/vagrant/k8s/setup_cluster.sh

From what I understand, even in k8s case, the ansible does not create the localBinOutput directory. It needs to be created with appropriate binaries before invoking ansible. I just followed the same pattern.

danehans:
@eparis you can close this since https://github.com/kubernetes/contrib/pull/783 merged.

PR #631, opened 2016-03-25
Add retries to reading data from GCS as it's not reliable enough.

gmarek:
@spiffxp @wojtek-t


wojtek-t:
LGTM

eparis:
We don't need any sleep or backoff or anything?  We can give it a try.

gmarek:
I don't know - we can add sleep to handle the race case, but for 503's it should be OK.

PR #642, opened 2016-03-26
Improve nginx-ingress-controller documentation.

aledbf:
Add flag to enable vts status module


aledbf:
ping @bprashanth

ingress/controllers/nginx-third-party/README.md
```diff
@@ -426,6 +426,16 @@ Please check the example `rc-custom-configuration.yaml`
 
 If the Configmap it is updated, NGINX will be reloaded with the new configuration
 
+### NGINX status page
+
```
bprashanth: please add a section with how to deploy the controller, i.e choose one rc.yaml as the main one, move it to the top level directory, and ask people to create that one before starting any of the examples.

bprashanth: Also make this yaml contain everything needed to start the ingress controller (i.e includs default svc+rc etc)

bprashanth: please also rename the rc so it doesn't say something like 3rdparty etc, i.e can we just call this one nginx-ingress-controller? 

Actually now that we're not dependent on third party (right?) I think we can rename the whole directory to just controllers/nginx. WDYT?

aledbf: >  to just controllers/nginx. WDYT?

good idea. 
done

ingress/controllers/nginx/README.md
```diff
@@ -91,242 +94,54 @@ kubectl create -f examples/default-backend.yaml
...
-* Connection #0 to host 172.17.4.99 left intact
--no body in request-
-```
+## Troubleshooting
 
+Problems encountered during [1.2.0-alpha7 deployment](https://github.com/kubernetes/kubernetes/blob/master/docs/getting-started-guides/docker.md):
+* make setup-files.sh file in hypercube does not provide 10.0.0.1 IP to make-ca-certs, resulting in CA certs that are issued to the external cluster IP address rather then 10.0.0.1 -> this results in nginx-third-party-lb appearing to get stuck at "Utils.go:177 - Waiting for default/default-http-backend" in the docker logs.  Kubernetes will eventually kill the container before nginx-third-party-lb times out with a message indicating that the CA certificate issuer is invalid (wrong ip), to verify this add zeros to the end of initialDelaySeconds and timeoutSeconds and reload the RC, and docker will log this error before kubernetes kills the container.
+  * To fix the above, setup-files.sh must be patched before the cluster is inited (refer to https://github.com/kubernetes/kubernetes/pull/21504)
 
 ## Custom errors
```
bprashanth: these make more sense as nested subheadings of the Troubleshooting section (eg: ### Custom, Debug,  errors)

bprashanth:
I assume you're still working on this since the main README doesn't have a deploy/create controller section with some top level all-in-one yaml, let me know if that's not the case and I'll review.

aledbf:
ping @bprashanth

ingress/controllers/nginx/README.md
```diff
@@ -0,0 +1,184 @@
...
+
+
+
+## Create a loadbalancer
+
+Loadbalancers are created via a ReplicationController or Daemonset
+
+First we need to deploy some application to publish. To keep this simple we will use the [echoheaders app](https://github.com/kubernetes/contrib/blob/master/ingress/echoheaders/echo-app.yaml) that just returns information about the http request as output
+```
+kubectl run echoheaders --image=gcr.io/google_containers/echoserver:1.1 --replicas=1 --port=8080
```
bprashanth: We have echoserver 1.2 and 1.3, use one of those (see README of echoserver for differences)?

ingress/controllers/nginx/README.md
```diff
@@ -0,0 +1,184 @@
...
+NAME      RULE          BACKEND   ADDRESS
+echomap   -
+          foo.bar.com
+          /foo          echoheaders-x:80
+          bar.baz.com
+          /bar          echoheaders-y:80
+          /foo          echoheaders-x:80
+```
+
+Before the deploy of the Ingress controller we need a default backend [404-server](https://github.com/kubernetes/contrib/tree/master/404-server)
```
bprashanth: Why do we actually need a default backend RC? can we just run the default-http-server as a container in the same pod as the nginx controller and redirect to that? (I think the gce controller does this)

ingress/controllers/nginx/README.md
```diff
@@ -0,0 +1,184 @@
...
+```
+kubectl create -f examples/default-backend.yaml
```
bprashanth: Lets keep this section *just* about creating the loadblancer. like so:

## Deploy the Ingress controller

$ kubectl create -f all-in-one.yaml

## HTTP 

kubctl create echoserver...

## TLS

## TCP

## Troubleshooting

### Custom Errors 

### Debugging

## Limitations

ingress/controllers/nginx/README.md
```diff
@@ -0,0 +1,184 @@
...
+kubectl expose rc default-http-backend --port=80 --target-port=8080 --name=default-http-backend
+```
+
+The final step:
+```
+kubectl create -f examples/default/rc-default.yaml
+```
```
bprashanth: We should show them how they can verify things are working, something like:
```
$ kubectl get ing
.. see some ip...
$ curl ip 
.. see some output
```

ingress/controllers/nginx/README.md
```diff
@@ -0,0 +1,184 @@
...
+  9000: "default/example-go:8080"
+```
+
+
+Please check the [tcp services](examples/tcp/README.md) example
+
+
+## Custom NGINX configuration
+
+Using a ConfigMap it is possible to customize the defaults in nginx.
```
bprashanth: If we have an example for this please link to it like you did for tcp

ingress/controllers/nginx/README.md
aledbf: I prefer to keep this as a parameter to allow me to customize the default backend without changing the controller descriptor (and allow to scale the backend independently)

bprashanth: SG

bprashanth: Hmm, can we add a fallback backend then? (not necessarily in this pr, just spin up a go server and redirect there if the given svc is missing). That way we divorce ourselves of this dependence.

ingress/controllers/nginx/README.md
aledbf: > .. see some ip...

I will that this functionality in another PR (add/update IP where the controller it's running)

ingress/controllers/nginx/README.md
aledbf: done

ingress/controllers/nginx/README.md
aledbf: > Hmm, can we add a fallback default backend then

I thought the idea was to not start the controller if the default backend is missing or does not exists
(I can change that in the next PR)

ingress/controllers/nginx/README.md
aledbf: done

ingress/controllers/nginx/README.md
bprashanth: Sure, i can push it as part of the new nginx controller image if it's O(days), otherwise we'll just cut 2. Eitherways please add some text to this doc about how people can find the IP of their loadbalancer. You can just use:
```
$ NODE_IP=$(kubectl get node `kubectl get po -l name=nginx-ingress-lb --template '{{range .items}}{{.spec.nodeName}}{{end}}'` --template '{{range $i, $n := .status.addresses}}{{if eq $n.type "ExternalIP"}}{{$n.address}}{{end}}{{end}}')
$ curl $NODE_IP/foo -H 'Host: foo.bar.com'
```

ingress/controllers/nginx/README.md
bprashanth: Yeah, why not start it? I think that'll lead to confusion.

bprashanth:
Thanks, I'll LGTM if you just add https://github.com/kubernetes/contrib/pull/642#discussion_r57808996

ingress/controllers/nginx/README.md
aledbf: done

bprashanth:
LGTM thanks

Issue #643, opened 2016-03-27
Use `source_type: packageManager` by default in Ansible


rutsky:
I think this is more reasonable default for users that want to use these scripts to deploy Kubernetes.
This will make deployment process one step easier and is good workaround for bugs like #636 for users that just want to get working installation of Kubernetes. 

`localBuild` is a good default for developers of Kubernetes, so if they are target audience of these script, then `localBuild` default should be kept.

/cc @eparis, @danehans, @stephenrlouie

eparis:
I say we default to packageManager

danehans:
+1 on packageManager default. @eparis @rutsky PR is here: https://github.com/kubernetes/contrib/pull/645

rutsky:
@danehans, thanks! PR is merged.

ingvagabund:
This breaks any OS that is not CentOS or Fedora as did_install fact is never set to False.

Fix is here [1]. Not yet tested as I have to leave now. This will fall to generic installator. Need to check if it is needed for RHEL. If not, I will create the missing piece.

[1] https://github.com/kubernetes/contrib/pull/658

ingvagabund:
@eparis ^^

PR #645, opened 2016-03-27
Changes source_type default to packageManager

danehans:
Fixes Issue: https://github.com/kubernetes/contrib/issues/643


danehans:
cc @eparis

PR #646, opened 2016-03-27
Add cross-compiling support for exechealthz

luxas:
Compile `exechealthz` for different architectures
Please push the generated images
Tracking issue: kubernetes/kubernetes#17981

@bprashanth @thockin @brendandburns @aledbf @eparis

### DIFF
--- a/exec-healthz/Makefile
+++ b/exec-healthz/Makefile
@@ -1,17 +1,71 @@
-all: push
+# Copyright 2016 The Kubernetes Authors All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
 
-# 0.0 shouldn't clobber any released builds
-TAG = 0.0
-PREFIX = gcr.io/google_containers/exechealthz
+# Build the exechealthz binary or image for any GOARCH
+#
+# Usage:
+# 	[TAG=1.0] [PREFIX=gcr.io/google_containers/exechealthz] [ARCH=amd64] make (server|container|push)
+
+all: container
+
+TAG?=1.0
+PREFIX?=gcr.io/google_containers/exechealthz
+ARCH?=amd64
+GOLANG_VERSION=1.6
+TEMP_DIR:=$(shell mktemp -d)
+
+# Set default base image dynamically for each arch
+ifeq ($(ARCH),amd64)
+	BASEIMAGE?=busybox
+endif
+ifeq ($(ARCH),arm)
+	BASEIMAGE?=armel/busybox
+endif
+ifeq ($(ARCH),arm64)
+	BASEIMAGE?=aarch64/busybox
+endif
+ifeq ($(ARCH),ppc64le)
+	BASEIMAGE?=ppc64le/busybox
+endif
 
 server: exechealthz.go
-	CGO_ENABLED=0 GOOS=linux godep go build -a -installsuffix cgo -ldflags '-w' -o exechealthz ./exechealthz.go
+	CGO_ENABLED=0 GOOS=linux GOARCH=$(ARCH) GOARM=6 go build -a -installsuffix cgo -ldflags '-w' -o exechealthz ./exechealthz.go
 
-container: server
-	docker build -t $(PREFIX):$(TAG) .
+container:
+	# Copy the whole directory to a temporary dir and set the base image
+	cp -r ./* $(TEMP_DIR)
+	cd $(TEMP_DIR) && sed -i "s|BASEIMAGE|$(BASEIMAGE)|g" Dockerfile
+
+	# Compile the binary inside a container for reliable builds
+	docker run --rm -it -v $(TEMP_DIR):/build golang:$(GOLANG_VERSION) /bin/bash -c "make -C /build server ARCH=$(ARCH)"
+
+	docker build -t $(PREFIX)-$(ARCH):$(TAG) $(TEMP_DIR)
+
+ifeq ($(ARCH),amd64)
+	# Backward compatibility. TODO: deprecate this image tag
+	docker tag -f $(PREFIX)-$(ARCH):$(TAG) $(PREFIX):$(TAG)
+endif
+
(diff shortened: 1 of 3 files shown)


exec-healthz/Makefile
```diff
@@ -1,17 +1,70 @@
...
+ifeq ($(ARCH),ppc64le)
+	BASEIMAGE?=ppc64le/busybox
+endif
 
 server: exechealthz.go
-	CGO_ENABLED=0 GOOS=linux godep go build -a -installsuffix cgo -ldflags '-w' -o exechealthz ./exechealthz.go
+	CGO_ENABLED=0 GOOS=linux GOARCH=$(ARCH) GOARM=6 go build -a -installsuffix cgo -ldflags '-w' -o exechealthz ./exechealthz.go
+
+container:
+	# Copy the whole directory to a temporary dir and set the base image
```
eparis: Why are you doing this? Just so you can run sed on the Dockerfile?

luxas: Yes, so it won't mess with the source directory. 
That's a very convenient way of doing it, makes it possible to run `make` over and over again for multiple arches.

eparis: Should we switch to `go install` so that we can cache past build results? And instead of cp-ing the entire directory create a `Dockerfile-arch` (which would be in .gitignore) and build the container from there?

luxas: I would like to have the dockerized build, cause I don't know which setup you have (`go1.6`, things already cached in `GOPATH` etc), so I think it's more reliable to put it inside docker.
I would say `go1.6` is pretty required, at least for `ppc64le`.

I don't think the caching is such a big deal, you are only supposed to do this once or twice.
If you don't have too hard feelings against this approach (that cross-Makefiles in main Kubernetes also has), could we please leave as-is?

exec-healthz/Makefile
```diff
@@ -1,17 +1,69 @@
-all: push
+# Copyright 2016 The Kubernetes Authors All rights reserved.
```
bprashanth: I don't think we add copyrights to any of our makefiles

exec-healthz/Makefile
```diff
@@ -1,17 +1,69 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
 
-# 0.0 shouldn't clobber any released builds
-TAG = 0.0
-PREFIX = gcr.io/google_containers/exechealthz
+# Build the exechealthz binary or image for any GOARCH
+#
+# Usage:
+# 	[TAG=1.0] [REGISTRY=gcr.io/google_containers] [ARCH=amd64] make [server|container|push]
```
bprashanth: why change this to registry? it's now different from the others, which or good or bad use prefix (in contrib and in kubernetes/test/images).

exec-healthz/Makefile
```diff
@@ -1,17 +1,69 @@
...
+ARCH?=amd64
+GOLANG_VERSION=1.6
+TEMP_DIR:=$(shell mktemp -d)
+
+# Set default base image dynamically for each arch
+ifeq ($(ARCH),amd64)
+	BASEIMAGE?=busybox
+endif
+ifeq ($(ARCH),arm)
+	BASEIMAGE?=armel/busybox
```
bprashanth: do you actually need all, or just speculating that we will need them? I'd rather leave them out if we don't care, since it places a burden on the release manager.

exec-healthz/Makefile
```diff
@@ -1,17 +1,69 @@
...
+endif
 
 server: exechealthz.go
-	CGO_ENABLED=0 GOOS=linux godep go build -a -installsuffix cgo -ldflags '-w' -o exechealthz ./exechealthz.go
+	CGO_ENABLED=0 GOOS=linux GOARCH=$(ARCH) GOARM=6 go build -a -installsuffix cgo -ldflags '-w' -o exechealthz ./exechealthz.go
+
+container:
+	# Copy the whole directory to a temporary dir and set the base image
+	cp -r ./* $(TEMP_DIR)
+	cd $(TEMP_DIR) && sed -i "s|BASEIMAGE|$(BASEIMAGE)|g" Dockerfile
```
bprashanth: where are you cleaning the TEMP_DIR?

bprashanth:
So how does one actually test an image for the other architectures before release?

exec-healthz/Makefile
luxas: See https://github.com/kubernetes/kubernetes/commit/05bd1073012afd33ceaf59e8f3b9eafc73f1e1ef

exec-healthz/Makefile
luxas: I used registry as we do in main k8s Makefiles (not test/images), but may change to prefix if you like and feel it's more in line with contrib

exec-healthz/Makefile
luxas: nowhere, it's in `/tmp` so it'll be cleaned on reboot.
also it might be useful to check the files/binaries there

luxas:
It's possible to test via qemu, but I think we just trust the cross-compilation forces that go have
`arm`, `arm64` and `ppc64le` are considered experimental anyway (I'm adding compilation scripts for them now)
exechealthz for arm works fine

My side-project is here https://github.com/luxas/kubernetes-on-arm if you like to check it up and the tracking issue is here: kubernetes/kubernetes#17981

exec-healthz/Makefile
bprashanth: Hmm, well ok we need to do the same for contrib as well. I suppose this could be the first.

exec-healthz/Makefile
bprashanth: Like any CVE will require us to ping you saying hey test this, and us to repush

exec-healthz/Makefile
bprashanth: Eh. Don't leak dirs even in /tmp :)
If anyone wants to debug they can open the Makefile and comment out the line.

exec-healthz/Makefile
bprashanth: Yeah please change to PREFIX, because it doesn't have to be a registry and if I wrote a build script that rebuilds all images in contrib it would be nice to not have to special case this one.

luxas:
@bprashanth Updated. PTAL

bprashanth:
please add a "Releases" section to the README that explicitly states that on needs to release multi arch, and mention how to do so (i.e make TAG=1.0 PREFIX=arm..). Can you also confirm that you tested this manually on the other architectures?

luxas:
@bprashanth Added section to README and tested manually.

bprashanth:
LGTM

luxas:
Thank you so much @bprashanth! :smile:
Do you want me to send a follow-up that forces the releaser to set a tag and remove the default or is it fine as-is?

bprashanth:
Yes please send a follow up when you have time

PR #647, opened 2016-03-28
Clarify firewall rule doc.

bprashanth:


bprashanth:
Oneline doc change, no need to bug people about review. self-merging when all green.

PR #648, opened 2016-03-28
use same /etc/sysconfig/flanneld env file as on other distributions

rutsky:
In CentOS 7 flanneld.service contains:

```
ExecStart=/usr/bin/flanneld -etcd-endpoints=${FLANNEL_ETCD} -etcd-prefix=${FLANNEL_ETCD_KEY} $FLANNEL_OPTIONS
```

Which allows to use variables like `FLANNEL_ETCD` in
`/etc/sysconfig/flanneld`.

Currently used approach on CoreOS uses flannel-specific environment
variables `FLANNELD_*`. However there is no way to pass arbitrary options
to flanneld daemon using `FLANNELD_*` environment variables (and currently
used `FLANNELD_OPTIONS` doesn't work actually).

Lets start flanneld in the same way on CoreOS to unify using of
environment files.

This fixes settings public ethernet interface in Ansible+Vagrant+VirtualBox+CoreOS configuration, noted in #644.

/cc @eparis @danehans


eparis:
Is the .service file getting installed on all hosts, even pkg-mgr ?

rutsky:
> Is the .service file getting installed on all hosts, even pkg-mgr ?

Flannel is installed from packages on all OSes except Atomic (where flannel is preinstalled) and Core OS.
So `flanneld.service` is written only on Core OS.
I assume that all other used systems have compatible `flanneld.service` (I checked only Cent OS 7 that it's true), and make Core OS' `flanneld.service` "compatible".

PR #652, opened 2016-03-28
Resilience fixes for the Ingress controller

bprashanth:
Reasoning for each commit:
1. UrlMap updating logic was too complicated, because it accounted for "joining" which we don't care about right now.
2. Errors at the beginning of the sync() function would prevent cloud urlmaps from getting cleaned up
3. The controller would silently forget about resources if it was restarted, and while down someone modified an Ingress
4. Retry creating GCE client since it can fail if we don't resolve metadata server

### DIFF
--- a/ingress/controllers/gce/backends/backends.go
+++ b/ingress/controllers/gce/backends/backends.go
@@ -59,7 +59,11 @@ func portKey(port int64) string {
 func NewBackendPool(
 	cloud BackendServices,
 	healthChecker healthchecks.HealthChecker,
-	nodePool instances.NodePool, namer utils.Namer, ignorePorts []int64, resyncWithCloud bool) *Backends {
+	nodePool instances.NodePool,
+	namer utils.Namer,
+	ignorePorts []int64,
+	resyncWithCloud bool) *Backends {
+
 	ignored := []string{}
 	for _, p := range ignorePorts {
 		ignored = append(ignored, portKey(p))
--- a/ingress/controllers/gce/backends/backends.go
+++ b/ingress/controllers/gce/backends/backends.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"net/http"
 	"strconv"
+	"time"
 
 	"k8s.io/kubernetes/pkg/util/sets"
 
@@ -38,6 +39,10 @@ type Backends struct {
 	healthChecker healthchecks.HealthChecker
 	snapshotter   storage.Snapshotter
 	namer         utils.Namer
+	// ignoredPorts are a set of ports excluded from GC, even
+	// after the Ingress has been deleted. Note that invoking
+	// a Delete() on these ports will still delete the backend.
+	ignoredPorts sets.String
 }
 
 func portKey(port int64) string {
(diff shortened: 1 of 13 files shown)


ingress/controllers/gce/controller/cluster_manager.go
```diff
@@ -138,6 +153,32 @@ func defaultInstanceGroupName(clusterName string) string {
...
+	// we know we're on GCE.
+	for {
+		cloudInterface, err := cloudprovider.GetCloudProvider("gce", nil)
+		if err == nil {
+			cloud := cloudInterface.(*gce.GCECloud)
+
+			// If this controller is scheduled on a node without compute/rw
+			// it won't be allowed to list backends. We can assume that the
+			// user has no need for Ingress in this case. If they grant
+			// permissions to the node they will have to restart the controller
```
freehan: FMOK, does user have to restart controller? why?

bprashanth: If I loop around here continuously re-creating the client till ListBackendServices succeedes, the liveness probe will fail at some point and cause a crashloop. I want to *not* crash if scheduled on a r/o node. 

Another way to handle this is to keep re-creating the client object in the liveness probe, and if it suddenly starts succeeding ListbackendServices calls, somehow re-notify everyone I've given the old client to refresh their clients. Didn't think this level of effort was worth it, since we can just doc that the controller needs permissions and leave it at that.

ingress/controllers/gce/backends/backends.go
```diff
@@ -46,18 +51,46 @@ func portKey(port int64) string {
...
 // - nodePool: implements NodePool, used to create/delete new instance groups.
+// - namer: procudes names for backends.
+// - ignorePorts: is a set of ports to avoid syncing/GCing.
+// - resyncWithCloud: if true, periodically syncs with cloud resources.
 func NewBackendPool(
 	cloud BackendServices,
 	healthChecker healthchecks.HealthChecker,
-	nodePool instances.NodePool, namer utils.Namer) *Backends {
-	return &Backends{
+	nodePool instances.NodePool, namer utils.Namer, ignorePorts []int64, resyncWithCloud bool) *Backends {
```
freehan: nit: reformat return list here

ingress/controllers/gce/storage/pools.go
```diff
@@ -51,3 +56,78 @@ func NewInMemoryPool() *InMemoryPool {
...
+		glog.Warningf("Failed to list: %v", err)
+		return
+	}
+	for i := range items {
+		key, err := c.keyGetter(items[i])
+		if err != nil {
+			glog.V(4).Infof("CloudListingPool: %v", err)
+			continue
+		}
+		c.InMemoryPool.Add(key, items[i])
```
freehan: Only Add. No delete?

bprashanth: Delete happens in sync. The point of this is so we don't leak, deleting from this pool would defeat the purpose, eg:
1. Create ingress foo
2. Create fw rule for foo
3. Container stops
4. While container is down, someone deletes ingress foo
5. Container starts
6. Container syncs existing ingress with cloud, ignores rule foo because there is no corresponding ingress

This ReplishPool will add the foo fw rule into the pool so when we sync we take a diff and note that there's an extra foo fw rule without a corresponding ingress that needs deletion. 

Everything in the pool is reconfirmed periodically, so eg: the controller adds fw rule foo to the pool, a user goes to cloud console and deletes foo (this would be a legit case where we need to delete here), the periodic sync of Ingress will reconfirm that foo doesn't exist, create it, and update the pool.

ingress/controllers/gce/storage/pools.go
```diff
@@ -51,3 +56,78 @@ func NewInMemoryPool() *InMemoryPool {
...
+	// An interface that lists objects from the cloud.
+	lister cloudLister
+	// A function capable of producing a key for a given object.
+	// This key must match the key used to store the same object in the user of
+	// this cache.
+	keyGetter keyFunc
+}
+
+// ReplinishPool lists through the cloudLister and inserts into the pool.
+func (c *CloudListingPool) ReplinishPool() {
```
freehan: Typo: Replenish

ingress/controllers/gce/backends/backends.go
bprashanth: Confused by the comment, what do you mean, put each arg on a newline?

bprashanth:
@freehan PTAL

freehan:
LGTM

ingress/controllers/gce/backends/backends.go
freehan: I meant the last line of args are too long

bprashanth:
Fixed nit, inheriting lgtm and merging on green

PR #653, opened 2016-03-28
Updated adhoc/uninstall.yml to support CoreOS

adamschaub:
Adds support for CoreOS distributions to the uninstall playbook.

### DIFF
--- a/ansible/playbooks/adhoc/uninstall.yml
+++ b/ansible/playbooks/adhoc/uninstall.yml
@@ -18,6 +18,20 @@
     - set_fact:
         is_containerized: "{{ is_atomic or containerized | default(false) | bool }}"
 
+    - name: Determine if CoreOS
+      raw: "grep '^NAME=' /etc/os-release | sed s'/NAME=//'"
+      register: distro
+      always_run: yes
+
+    - name: Init the is_coreos fact
+      set_fact:
+        is_coreos: false
+
+    - name: Set the is_coreos fact
+      set_fact:
+        is_coreos: true
+      when: "'CoreOS' in distro.stdout"
+
     - name: Remove br0 interface
       shell: ovs-vsctl del-br br0
       changed_when: False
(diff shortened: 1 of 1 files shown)


adamschaub:
@danehans Could use your help to call on appropriate reviewers

eparis:
If it doesn't work, we can fix it then.

ansible/playbooks/adhoc/uninstall.yml
```diff
@@ -33,21 +47,46 @@
...
     - name: Remove packages
       action: "{{ ansible_pkg_mgr }} name={{ item }} state=absent"
-      when: not is_atomic | bool
+      when: not is_coreos and not is_atomic | bool
       with_items:
         - flannel
         - kubernetes-node
         - kubernetes-master
         - etcd
+        - etcd2
```
danehans: @adamschaub I don;t think we need to remove etcd2 since this is not is_coreos. Feel free to send a follow-on PR that removes if the change works in your lab.

ansible/playbooks/adhoc/uninstall.yml
```diff
@@ -33,21 +47,46 @@
...
         - kube-addons
         - atomic-enterprise
         - atomic-enterprise-master
         - atomic-enterprise-node
 
+    - name: CoreOS | Remove packages
```
danehans: small nit ^ to change the name from packages to bins and systemd files. I just want to make sure we don;t confuse anyone. This change can be combined with the other one in the same future PR.

danehans:
fyi @eparis @adamschaub I successfully tested this patch in my test lab.

ansible/playbooks/adhoc/uninstall.yml
adamschaub: We need to disable the etcd2 service at some point. I figured it didn't hurt to put it in the list of existing services to stop, as opposed to making a separate 'CoreOS' stop service task. Thoughts?

danehans: I think it might throw an error for non-coreos deployments. If it doesn't then it's no biggie. I'm submitting a follow-on PR that makes a few changes b/c I had an issue with this one. I'll cc you on the PR so you can review and test.

ansible/playbooks/adhoc/uninstall.yml
eparis: agreed, would love a followup PR

PR #656, opened 2016-03-29
Fixes Ansible Vagrant for CoreOS

danehans:
Previously, Vagrant would not properly set the box download
url:

kube-node-1: Downloading: https://storage.googleapis.com/\
.release.core-os.net/amd64-usr//coreos_production_vagrant.json

An error occurred while downloading the remote file. The error
message, if any, is reproduced below. Please fix this error and
try again.

The requested URL returned error: 400 Bad Request

This commit properly sets the coreos update_channel
and image_version variables.


danehans:
cc @rutsky @stephenrlouie

rutsky:
Looks like images were cached on my machine.
LGTM, thanks @danehans!

PR #658, opened 2016-03-29
Set implicit value for did install fact for node role

ingvagabund:
Init the did_install fact in packageManagerInstall.yml for node role.
Otherwise it is undefined for other distros than CoreOS or Fedora


eparis:
Can you please squash into a single commit?

ingvagabund:
Right, forget to rebase, on it.

PR #661, opened 2016-03-29
download Flannel releases to /opt/

rutsky:
/usr/ is readonly on Core OS, and /opt/ should work on Ubuntu too.

Intallation failure on Core OS was introduces in #588 

/cc @unicell @danehans @stephenrlouie

### DIFF
--- a/ansible/roles/flannel/tasks/github-release.yml
+++ b/ansible/roles/flannel/tasks/github-release.yml
@@ -3,7 +3,11 @@
   set_fact:
     flannel_version: 0.5.5
 
-- stat: path=/usr/local/flannel-{{ flannel_version }}
+- name: Set Flannel releases directory
+  set_fact:
+    flannel_releases_dir: /opt/
+
+- stat: path={{ flannel_releases_dir }}flannel-{{ flannel_version }}
   register: st
 
 - name: Download tar file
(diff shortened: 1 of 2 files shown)


ansible/roles/flannel/tasks/github-release.yml
```diff
@@ -3,7 +3,11 @@
   set_fact:
     flannel_version: 0.5.5
 
-- stat: path=/usr/local/flannel-{{ flannel_version }}
+- name: Set Flannel releases directory
+  set_fact:
+    flannel_releases_dir: /opt/
```
eparis: should we move it to roles/flannel/defaults/ ?

danehans: and drop the trailing /

ansible/roles/flannel/tasks/github-release.yml
```diff
@@ -20,13 +24,13 @@
...
     src: "{{ ansible_temp_dir }}/flannel-{{ flannel_version }}-linux-amd64.tar.gz"
-    dest: /usr/local
+    dest: "{{ flannel_releases_dir }}"
     copy: no
   when: st.stat.isdir is not defined
 
 - name: Create symlinks
   file:
-    src: /usr/local/flannel-{{ flannel_version }}/{{ item }}
+    src: "{{ flannel_releases_dir }}flannel-{{ flannel_version }}/{{ item }}"
```
danehans: Add the trailing slash from 8 after {{ flannel_releases_dir }}

eparis: agreed

danehans:
@rutsky @eparis fyi this patch works in my coreos testbed. Before getting merged, I would like the inline comments addressed. Otherwise, thanks a bunch @rutsky for fixing the breakage.

rutsky:
@eparis, @danehans I fixed your comments, please take a look.

PR #663, opened 2016-03-29
Bump K8s version to 1.2.0 on Core OS + download k8s binaries one by one

rutsky:
Kubernetes GitHub release tarball has size of about 400+ MB.
Previously it's being downloaded, extracted, and inner tarball extracted
one more time (on each host).
This took roughly 400+ MB * NUM_NODES traffic and about 1 GB of space in
/tmp/ (on each host).

Starting from 1.2.0 it takes more than 1 GB of space in /tmp by Ansible,
which fails for example on Core OS, where /tmp is 1 GB ramdisk.

With these changes only required binaries are downloaded (not
compressed, but it's still significantly less than whole release tarball size).

/cc: @danehans @stephenrlouie @eparis

### DIFF
--- a/ansible/roles/kubernetes/defaults/main.yml
+++ b/ansible/roles/kubernetes/defaults/main.yml
@@ -18,7 +18,7 @@ kube_script_dir: /usr/libexec/kubernetes
 kube_config_dir: /etc/kubernetes
 
 # The URL to download Kubernetes binaries from.
-kube_download_url: https://github.com/kubernetes/kubernetes/releases/download/v{{ kube_version }}/kubernetes.tar.gz
+kube_download_url_base: https://storage.googleapis.com/kubernetes-release/release/v{{ kube_version }}/bin/linux/amd64/
 
 # This is where all the cert scripts and certs will be located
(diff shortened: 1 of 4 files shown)


danehans:
@eparis @rutsky this commit works for me in my coreos testbed. @rutsky thanks for optimizing this piece of the code and bumping to 1.2.0

eparis:
My concern with this approach is updates. Changing the version to 1.2.1 will not download the bins again. Can we download them into a subdir with the version as part of the path and keep the symlinks?

rutsky:
@eparis I changed role to download into versioned directory and symlink files from it to `{{ bin_dir }}`.

@eparis @danehans any idea why binaries are downloaded on CoreOS in `kubernetes` role, not in `masters`/`nodes` roles as done in other distributions?

eparis:
@rutsky I think it was done that way for simplicity. We can change it later if we want (although there was some thought about just always making the master an (unschedulable) node)

danehans:
@rutsky @eparis this func was part of the kube role b/c it was the same tar (containing all kube bins) being downloaded for master/node roles.

PR #666, opened 2016-03-29
add requirement for minimal Vagrant version to use for Core OS images

rutsky:
Vagrant prior version 1.8.0 doesn't write group variables into inventory file:
<https://github.com/mitchellh/vagrant/commit/dd4ae1a51cfb246d561bced89d3b34ee90a0a38f>

@stephenrlouie this is a check for issue that you observed with wrong `ansible_python_path`.


stephenrlouie:
@rutsky Good find on the vagrant commit. I was about to start looking for a legitimate source/fix for this error.

rutsky:
@eparis can you take a look at this PR?

PR #669, opened 2016-03-29
CoreOS pre-ansible: un-hardcode bin_dir

adamschaub:
Updates CoreOS pre-ansible tasks so that bin_dir isn't hardcoded in coreos.yml. 

@eparis @danehans

### DIFF
--- a/ansible/roles/pre-ansible/defaults/main.yml
+++ b/ansible/roles/pre-ansible/defaults/main.yml
@@ -1,3 +1,7 @@
+# The directory where binaries are stored on Ansible
+# # managed systems.
+bin_dir: /usr/bin
+
 # The directory where scripts used for bootstrapping CoreOS
 # reside.
 bootstrap_script_dir: /opt/bin
--- a/ansible/roles/pre-ansible/tasks/coreos.yml
+++ b/ansible/roles/pre-ansible/tasks/coreos.yml
@@ -34,7 +34,7 @@
   raw: >
           printf "%s\n%s"
           "#! /usr/bin/bash"
-          "PATH=\$PATH:/opt/bin"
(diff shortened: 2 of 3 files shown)


ansible/roles/pre-ansible/defaults/main.yml
```diff
@@ -1,3 +1,7 @@
+# The directory where binaries are stored on Ansible
+# # managed systems.
```
rutsky: Double comment here: `# #`.

rutsky:
@adamreese, @eparis, @danehans why `common` role don't depend on `pre-ansible`, and other roles don't depend on `common` role?

I see that `bin_dir` is redefined for Core OS in [`common` role](https://github.com/kubernetes/contrib/blob/6aaa2f148a5b1348422e9c6653b7c984273e93f1/ansible/roles/common/tasks/main.yml#L45), [`kubernetes` role](https://github.com/kubernetes/contrib/blob/0ba7c0a9076fd571f02214b4cde7ae92b65459d2/ansible/roles/kubernetes/tasks/main.yml#L16) and now in `pre-ansible` role too, which worries me.

eparis:
@rutsky uggh, yes...  the problem was always me not wanting the run `pre-ansible` or `common` (many) more times than needed.... It's definitely a problem, and if you think of a way to fix it, awesome...

rutsky:
@adamreese can you add proper dependencies and and move`bin_dir` definition to a single place, since @eparis agreed with this solution?

eparis:
I only 'agree' if we don't start running common or pre-ansible multiple times....  So i'd rather see this merge and then try again on a different PR.

rutsky:
@eparis I opened separate issue for adding dependencies: https://github.com/kubernetes/contrib/issues/671

PR #676, opened 2016-03-30
Updated kube-ui to use new kubernetes dashboard ui

deromka:
Updated kube-ui to use new kubernetes dashboard ui from [here](https://github.com/kubernetes/kubernetes/tree/master/cluster/addons/dashboard)

### DIFF
--- a/ansible/roles/kubernetes-addons/tasks/kube-ui.yml
+++ b/ansible/roles/kubernetes-addons/tasks/kube-ui.yml
@@ -4,7 +4,7 @@
 
 - name: KUBE-UI | Download kube-ui files from Kubernetes repo
   get_url:
-    url=https://raw.githubusercontent.com/GoogleCloudPlatform/kubernetes/master/cluster/addons/kube-ui/{{ item }}
+    url=https://raw.githubusercontent.com/kubernetes/kubernetes/master/cluster/addons/dashboard/{{ item }}
     dest="{{ kube_addons_dir }}/kube-ui/"
     force=yes
     validate_certs=False
@@ -13,5 +13,5 @@
     https_proxy: "{{ https_proxy|default('') }}"
     no_proxy: "{{ no_proxy|default('') }}"
   with_items:
-    - kube-ui-rc.yaml
-    - kube-ui-svc.yaml
+    - dashboard-controller.yaml
+    - dashboard-service.yaml


rutsky:
There are separate task for installing Dashboard: https://github.com/deromka/contrib/blob/master/ansible/roles/kubernetes-addons/tasks/kube-dash.yml
We should either fix kube-ui, or completely remove it.

deromka:
[kube-dash](https://github.com/kubernetes/kubedash) is a cluster performance analytics UI, rather than management UI for
The previously known [kube-ui](https://github.com/kubernetes/kube-ui) is now forked under a new name [kubernetes-dashboad](https://github.com/kubernetes/dashboard) which is also UI.

So Ansible scripts will now deploy the kubernetes-dashboard instead of kube-ui
And keep deploying the kubedash

rutsky:
@deromka you right, I though that `kube-dash` is `kubernetes-dashboard`, LGTM.

Can you squash commits?

ansible/roles/kubernetes-addons/templates/kube-addons.upstart.j2
```diff
@@ -4,7 +4,7 @@ start on started kube-apiserver
 stop on runlevel [!2345]
 
 env TOKEN_DIR={{ kube_token_dir }}
-env KUBECTL_BIN={{ bin_dir }}/kubectl
+env KUBECTL_BIN=/usr/bin/kubectl
```
danehans: We can't have this hard coded. Please set bin_dir in the associated playbook task if you need to set bin_dir: /usr/bin/  However, I thought /usr/bin/ was already the default for bin_dir.

ansible/roles/kubernetes-addons/templates/kube-addons.service.j2
```diff
@@ -4,7 +4,7 @@ Documentation=https://github.com/GoogleCloudPlatform/kubernetes
 
 [Service]
 Environment="TOKEN_DIR={{ kube_token_dir }}"
-Environment="KUBECTL_BIN={{ bin_dir }}/kubectl"
+Environment="KUBECTL_BIN=/usr/bin/kubectl"
```
danehans: Same here.

danehans:
@deromka see my inline comments and as @rutsky mentions, the commits need to be squashed into 1: https://ariejan.net/2011/07/05/git-squash-your-latests-commits-into-one/

deromka:
@danehans @rutsky done.

PR #680, opened 2016-03-31
Allocate and remember cluster uid 

bprashanth:
Remembers existing cluster uids in a config map so even if it changes we don't delete/re-create working loadbalancers. This means once a cluster is up and running with a uid, it doesn't change no matter how many time the controller is restarted with a different uid, unless someone deletes the uid config map in kube-system. 

This is needed so we can rollout a change that auto-assigns uids to controllers, which in turn is required before we start running the controller on the master.


freehan:
LGTM

bprashanth:
Thanks!

PR #686, opened 2016-03-31
cherrypick: auto-approve cherrypicks if the pick to master was approved

eparis:


eparis:
@bgrant0607   tested, and this would have automatically approved #23685 so it would not have gotten the `do-not-merge` since you already approved https://github.com/kubernetes/kubernetes/pull/23646

PR #687, opened 2016-03-31
Clarify default backend creation.

bprashanth:
Clarify how users create a default backend so they don't see a crashlooping nginx controller.
@aledbf


bprashanth:
Just docs, so merging

Issue #602, opened 2016-03-22
Switch munger to use GCS as a source of truth instead of Jenkins.


eparis:
```
I0322 17:13:04.205714       1 e2e.go:119] Got a non-success response 404 while reading data for kubernetes-e2e-gke/3126/finished.json
E0322 17:13:04.206332       1 submit-queue.go:242] GCS stable check returned different value than Jenkins.
```

gmarek:
Fascinating...

gmarek:
There's something clearly wrong there:
```
I0323 11:30:43.180208       1 e2e.go:119] Got a non-success response 404 while reading data for kubernetes-test-go/0/finished.json
```

If not for the fact that it happened few times today I'd say that it's some kind of race. Are we sure that we update 'last-build.txt' as a last thing when uploading results? It seems that the code https://github.com/kubernetes/contrib/blob/master/test-utils/utils/utils.go#L46 returned 0 for some reason. @ixdy @spxtr

ixdy:
Yeah, there is a race right now. Jenkins calls `hack/jenkins/upload-to-gcs.sh` which uploads the build logs and updates `latest-build.txt`, and then it calls `hack/jenkins/upload-finished.sh`, which uploads the `finished.json` file.

The eventual plan is to combine these two scripts, since there's no reason to have both of them, but the PR Jenkins jobs don't use `upload-finished.sh` yet. As an easier workaround we could probably switch the order they're called.

ixdy:
The build number of 0 is probably an artifact from Jenkins restarting.

gmarek:
@ixdy - can you take care of it?

ixdy:
Quick fix merged. Proper fix in process.

gmarek:
Thanks @ixdy - it seems to be working now.
@eparis, @spxtr - I'd like to switch to GCS as a main source of truth on Monday, if we won't see any differences until then. WDYT?

@justinsb - this should make it quicker for your PRs to go through, as we're testing GCS source to see if it's reliable enough.

cc @kubernetes/sig-testing

justinsb:
@gmarek that's great - I was looking at reintroducing my patches, but I think you've done a lot of the work in them for me!  I was going to start with the configuration-in-files PR (allowing more settings for each test runner), but it was sort of complicated by the fast that we are still using Jenkins.  So if we switch to GCS as the source of truth I should be able to redo my PRs much more easily!  (And there might not be a lot left in them!)

[...]

PR #688, opened 2016-03-31
Add more logging to GCS stable check

gmarek:
Additional log in e2e.go will probably be quite verbose, so we need to fix this problem after changing to GCS as source of truth. #602

cc @wojtek-t

### DIFF
--- a/test-utils/utils/utils.go
+++ b/test-utils/utils/utils.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"time"
 
 	"github.com/golang/glog"
 )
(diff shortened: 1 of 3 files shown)


test-utils/utils/utils.go
```diff
@@ -43,6 +44,7 @@ func getResponseWithRetry(url string) (*http.Response, error) {
 		if response.StatusCode == http.StatusOK {
 			return response, nil
 		}
+		time.Sleep(100 * time.Millisecond)
```
wojtek-t: please make this a constant

gmarek: Done.

wojtek-t:
LGTM

PR #697, opened 2016-04-01
Enable security context in clusters by default

ingvagabund:
Enable security context in clusters by default to have testing environment match the production one


pweil-:
LGTM

ansible/roles/master/templates/apiserver.j2
```diff
@@ -20,7 +20,7 @@ KUBE_SERVICE_ADDRESSES="--service-cluster-ip-range={{ kube_service_addresses }}"
 KUBE_ETCD_SERVERS="--etcd-servers={% for node in groups['etcd'] %}http://{{ node }}:2379{% if not loop.last %},{% endif %}{% endfor %}"
 
 # default admission control policies
-KUBE_ADMISSION_CONTROL="--admission-control=NamespaceLifecycle,NamespaceExists,LimitRanger,SecurityContextDeny,ServiceAccount,ResourceQuota"
+KUBE_ADMISSION_CONTROL="--admission-control=NamespaceLifecycle,NamespaceExists,LimitRanger,ServiceAccount,ResourceQuota"
```
eparis: Should we put these in an ansible variable so users can overwrite out defaults?

PR #701, opened 2016-04-01
Avoid unnecessary 'changed' for kubernetes-addons playbooks

unicell:
Currently several addon spec file are always downloaded from Github,
each time Ansible run. Then installed to target configuration file after
inline changes.

This creates too many 'changed' noise and breaks the expectation that
Ansible run should be idempotent and only report 'changed' when really
does.

This change mute 'changed' message for downloading and inline replacing
part, and let the final install to /etc part report 'changed' or not.

### DIFF
--- a/ansible/roles/kubernetes-addons/tasks/cluster-monitoring.yml
+++ b/ansible/roles/kubernetes-addons/tasks/cluster-monitoring.yml
@@ -20,6 +20,7 @@
     - heapster-service.yaml
     - influxdb-grafana-controller.yaml
     - influxdb-service.yaml
+  changed_when: false
 
 - name: MONITORING | Remove instances of getting pillar dicts
   local_action: lineinfile
@@ -33,6 +34,7 @@
     - heapster-service.yaml
     - influxdb-grafana-controller.yaml
     - influxdb-service.yaml
+  changed_when: false
 
 - name: MONITORING | Add tests for pillar vars converted to ansible vars
   local_action: lineinfile
@@ -47,6 +49,7 @@
     - heapster-service.yaml
     - influxdb-grafana-controller.yaml
     - influxdb-service.yaml
+  changed_when: false
 
 - name: MONITORING | Install template from converted saltfile
   template:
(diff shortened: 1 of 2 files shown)


rutsky:
IMO better approach would be to use specific version of file and use [`checksum`](https://docs.ansible.com/ansible/get_url_module.html#options) parameter of `get_url`, e.g.:

```
 - name: DNS | Download skydns-rc.yaml file from Kubernetes repo
   local_action: get_url
     url=https://raw.githubusercontent.com/kubernetes/kubernetes/v1.2.1/cluster/addons/dns/skydns-rc.yaml.in
     dest="{{ local_temp_addon_dir }}/dns/skydns-rc.yaml.j2"
     force=yes
     checksum=sha256:2dae13ae58656b2f20919740caa953d632de1f017de8df1807f80f03168f7169
   sudo: no
```

version (`1.2.1`) and checksum (`sha256:2dae13a...`) can be moved to `defaults.yml` for easier further upgrading.

Pros of this approach:

1. We will use fixed versions that are guaranteed to work. Currently version in master branch can be changed in incompatible way or even removed.

2. We will verify that downloaded has expected contents.

3. Ansible will check, if file already downloaded and have expected checksum, then no redownload will happen and no change event will be generated.

Cons: `checksum` was introduced in Ansible 2.0 release (January 12, 2016). There is deprecated [`sha256sum`](https://docs.ansible.com/ansible/get_url_module.html#options) parameter, but I'm unsure is [it's behaviour is good enough](https://github.com/ansible/ansible-modules-core/issues/625).

AFAIKT, [minimum supported version of Ansible in this project is currently 1.9](https://github.com/kubernetes/contrib/blame/master/ansible/README.md#L11).

unicell:
I agree it is better to have it versioned. 

But on the other hand, these spec file are downloaded and processed in temp directory ( local_temp_addon_dir) before copying over to /etc/kubernetes/addons. That means next time I run playbooks, contents in temp is not guaranteed to be kept. Then Ansible will again treated as "changed".

So, to implement your way, we need to:
- making those file versioned
- some refactoring to move away from temp directories
- backward compatibility for Ansible 1.9

That's why I proposed the change this way. Simply treat download / inline replacing part as intermediate steps, as it doesn't really touch real config, and let final step report "changed" or not when copying to /etc/kubernetes/addons. Sure we can do more refactoring work after bumping minimum Ansible version to 2.0.

What do you think?

rutsky:
I haven't noticed that you ignore changed state for all steps except last one, which is only important.
I like your idea, changes LGTM!

PR #706, opened 2016-04-04
Updated with correction for Ingress TLS spec

joonathan:
Documentation fix for `error validating data: expected type array, for field spec.tls, got map; if you choose to ignore these errors, turn validation off with --validate=false`


joonathan:
I signed it!

bprashanth:
Thanks, LGTM

PR #707, opened 2016-04-04
Allow traffic to default server _ in nginx Ingress controller

aledbf:


bprashanth:
Still wip?

aledbf:
@bprashanth ready

ingress/controllers/nginx/utils.go
```diff
@@ -109,6 +109,15 @@ func getLBDetails(kubeClient *unversioned.Client) (*lbInfo, error) {
 		return nil, fmt.Errorf("Unable to get POD information")
 	}
 
+	if pod.Status.Phase != api.PodRunning {
+		// we wait up to 30 seconds until the pod is running and
+		// it is possible to get the IP and name of the node
+		err := waitForPodRunning(kubeClient, podNs, podName, time.Millisecond*200, time.Second*30)
```
bprashanth: Why do this? we're always going to call this function from a running pod to get its own lb details, right?

bprashanth: I think you should either not do it for the mentioned reason, or move it into a generic: getPodDetails function that takes a pod name and returns details only if the pod is running.

ingress/controllers/nginx/utils.go
```diff
@@ -128,6 +137,8 @@ func getLBDetails(kubeClient *unversioned.Client) (*lbInfo, error) {
 		}
 	}
 
+	podIP := os.Getenv("POD_IP")
```
bprashanth: what're we using pod ip for?

ingress/controllers/nginx/Makefile
```diff
@@ -4,8 +4,16 @@ all: push
 TAG = 0.5
 PREFIX = gcr.io/google_containers/nginx-ingress-controller
 
+REPO_INFO=$(shell git config --get remote.origin.url)
```
bprashanth: Please update the readme with instructions on how to build since it's now a little more than just running make all. Or write a RELEASES.md file like we have in kube-dns: https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dns/kube2sky/RELEASES.md

ingress/controllers/nginx/utils.go
aledbf: @bprashanth because of this https://github.com/kubernetes/contrib/issues/752. 
For some reason `kubeClient.Nodes().Get(pod.Spec.NodeName)` returns an error unless I add this wait. All this is required just to get the IP of the node (to update the ingress rules).

ingress/controllers/nginx/Makefile
aledbf: @bprashanth why? that code is just add additional information about the build from git. No user intervention required. This is the first line in the logs:
```
I0410 21:36:01.457316       1 main.go:89] Using build: https://github.com/aledbf/contrib - git-b580f13
```

bprashanth: Oh, i thought you were building with a specific commit hash of godep. I guess you'er just piping the current commit hash down into the controller.

bprashanth:
LGTM will merge on green

ingress/controllers/nginx/main.go
```diff
@@ -41,6 +41,10 @@ const (
 )
 
 var (
+	// value overwritten during build. This can be used to resolve issues.
+	version = "0.0"
```
bprashanth: don't you want to keep this up to date with the latest? i.e 0.5 in this case

bprashanth: It's helpful when people give you logs and don't know if they changed the image on the rc/restart pods etc, or even if they're using latest

aledbf: 👍

PR #708, opened 2016-04-05
Munger uses GCS as a source of truth

gmarek:
Let's do this.

cc @kubernetes/sig-testing @spiffxp @spxtr @ixdy @justinsb


eparis:
this breaks `godep go test ./...`

eparis:
the reason it breaks is because the submit-queue tests look for/emulate e2e pass/failure, so we need to make that work.

gmarek:
This should be fixed now.

PR #709, opened 2016-04-05
Use -o name on kubectl run

thockin:
Also arrange tmux panes for easier keybaord input.

Also rename scripts to have meaning on splits.


micro-demos/daemon_sets/demo.sh
```diff
@@ -33,6 +33,6 @@ run "kubectl --namespace=demos create -f $(relative daemon.yaml) --validate=fals
 run "kubectl --namespace=demos describe ds daemons-demo"
 
 tmux new -d -s my-session \
-    "$(dirname ${BASH_SOURCE})/split1_lhs.sh" \; \
-    split-window -h -d "sleep 10; $(dirname $BASH_SOURCE)/split1_rhs.sh" \; \
+    "sleep 10; $(dirname ${BASH_SOURCE})/split1_color_nodes.sh" \; \
```
caesarxuchao: In case you don't already know, you can use  `tmux select-pane -t` to select the pane to focus on (e.g., https://github.com/kubernetes/contrib/blob/master/micro-demos/deployments/demo.sh#L35)

micro-demos/pods/demo.sh
```diff
@@ -28,7 +28,6 @@ run "kubectl --namespace=demos get pods"
 desc "Get the pod's IP"
 run "kubectl --namespace=demos get pod pods-demo -o yaml | grep podIP"
 
-trap "" SIGINT
```
caesarxuchao: For my own education, when will such a trap be useful?

caesarxuchao:
LGTM. I asked a question for my own education. It needs rebase.

micro-demos/daemon_sets/demo.sh
thockin: you're correct, I did.  But I actually like it better this way :)

micro-demos/pods/demo.sh
thockin: I don't honestly recall why I added it, but it was annoying me when running tests..

thockin:
rebased and deployment demo fixed (grr)

micro-demos/deployments/deployment.yaml
```diff
@@ -14,14 +17,12 @@ spec:
     metadata:
       labels:
         demo: deployment
+        version: v1
```
caesarxuchao: This label makes the script much cleaner. Thanks.

Issue #700, opened 2016-04-01
[Ansible+Vagrant+VirtualBox] K8s doesn't work due to incorrect endpoint IP of `kubernetes` service


rutsky:
On Vagrant+VirtualBox public IP is on `eth1`, not on `eth0` (which is used for NAT-ed Internet access).
K8s apiserver when started by default uses default ethernet device to get public IP address, which leads to publishing unroutable IP:

```
kube-master ~ # kubectl get endpoints --namespace=default
NAME         ENDPOINTS       AGE
kubernetes   10.0.2.15:443   2d
kube-master ~ # ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default 
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000
    link/ether 08:00:27:96:9e:8a brd ff:ff:ff:ff:ff:ff
    inet 10.0.2.15/24 brd 10.0.2.255 scope global dynamic eth0
       valid_lft 80637sec preferred_lft 80637sec
    inet6 fe80::a00:27ff:fe96:9e8a/64 scope link 
       valid_lft forever preferred_lft forever
3: eth1: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000
    link/ether 08:00:27:48:ea:40 brd ff:ff:ff:ff:ff:ff
    inet 172.28.128.5/24 brd 172.28.128.255 scope global dynamic eth1
       valid_lft 679sec preferred_lft 679sec
    inet6 fe80::a00:27ff:fe48:ea40/64 scope link 
       valid_lft forever preferred_lft forever
4: flannel.1: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1450 qdisc noqueue state UNKNOWN group default 
    link/ether be:98:8d:1e:3e:6e brd ff:ff:ff:ff:ff:ff
    inet 172.16.11.0/12 scope global flannel.1
       valid_lft forever preferred_lft forever
    inet6 fe80::bc98:8dff:fe1e:3e6e/64 scope link 
[...]

PR #712, opened 2016-04-05
[Vagrant+VBox+Ansible] use static ip + explicitly set API Server bind address

### LINKED ISSUE #696: [Ansible+Vagrant+VirtualBox] use static IP instead of DHCP
Currently on Vagrant we rewrite `/etc/hosts` with public IPs of nodes that were received at the time of provision in [vagrant-ansible.yml](https://github.com/kubernetes/contrib/blob/master/ansible/vagrant/vagrant-ansible.yml#L17), e.g.:

```
$ cat /etc/hosts 
172.28.128.5 kube-master
172.28.128.3 kube-node-1
172.28.128.4 kube-node-2
```

this is OK for quick testing, but after some time IP addresses of nodes may change (for example if you shutdown host for weekend), which will lead to communication failures between different Kubernetes parts.

I suggest to use fixed IP addresses at least on VirtualBox, where VirtualBox software DHCP server is used.

The question is how to properly select IP addresses range to use?

/cc: @eparis @danehans

rutsky:
Fixes #696 and #700 .

/cc @danehans @eparis @adamschaub

### DIFF
--- a/ansible/roles/master/defaults/main.yml
+++ b/ansible/roles/master/defaults/main.yml
@@ -3,3 +3,5 @@ kube_master_insecure_port: 8080
 localBuildOutput: ../../_output/local/go/bin
 
 admission_controllers: NamespaceLifecycle,NamespaceExists,LimitRanger,ServiceAccount,ResourceQuota
+
+kube_apiserver_bind_address: "0.0.0.0"
--- a/ansible/roles/master/templates/apiserver.j2
+++ b/ansible/roles/master/templates/apiserver.j2
@@ -23,4 +23,4 @@ KUBE_ETCD_SERVERS="--etcd-servers={% for node in groups['etcd'] %}http://{{ node
 KUBE_ADMISSION_CONTROL="--admission-control={{ admission_controllers }}"
 
 # Add your own!
-KUBE_API_ARGS="{{ apiserver_extra_args | default('') }} --tls-cert-file={{ kube_cert_dir }}/server.crt --tls-private-key-file={{ kube_cert_dir }}/server.key --client-ca-file={{ kube_cert_dir }}/ca.crt --token-auth-file={{ kube_token_dir }}/known_tokens.csv --service-account-key-file={{ kube_cert_dir }}/server.crt"
+KUBE_API_ARGS="{{ apiserver_extra_args | default('') }} --tls-cert-file={{ kube_cert_dir }}/server.crt --tls-private-key-file={{ kube_cert_dir }}/server.key --client-ca-file={{ kube_cert_dir }}/ca.crt --token-auth-file={{ kube_token_dir }}/known_tokens.csv --service-account-key-file={{ kube_cert_dir }}/server.crt --bind-address={{ kube_apiserver_bind_address }}"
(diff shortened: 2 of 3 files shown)


danehans:
@rutsky LGTM

Tested successfully on my vagrant/coreos setup

/cc @eparis

rutsky:
@eparis can you please take a look at this PR, since @danehans tested these changes?

I'll rebase #724 after you merge this.

eparis:
i broke it merging something else  :-(

eparis:
I also just generally don't like it because we are doing vbox custom things. Can you please look for any reasonable way to solve this in ansible instead of with hard coding and vbox specifics.  If you tell me no, there is nothing better I'll take it, but I'm not in love.

rutsky:
Rebased.

@eparis this PR includes two things:

1. It adds Ansible variable `kube_apiserver_bind_address` that allows to specify bind address of API server: https://github.com/kubernetes/contrib/pull/712/commits/a3aacd1b39d7c82361d72db98c90f6dbb8e961b3

2. Changes Vagrant configuration to use static IP addresses for VMs on VirtualBox (https://github.com/kubernetes/contrib/pull/712/commits/cecbf160e880d03bd0954deeb50902151d68264c) and explicitly uses master IP in `kube_apiserver_bind_address` (https://github.com/kubernetes/contrib/pull/712/commits/7648fbc3fd4de8a346434f8e3ab7109c8554b439).

@eparis 
> I also just generally don't like it because we are doing vbox custom things. Can you please look for any reasonable way to solve this in ansible instead of with hard coding and vbox specifics.

I'm not really following you here. Vagrant+VirtualBox is specific environment that in one or another way requires specific configuration.
It's not possible to treat all Vagrant+VirtualBox quirks only in Ansible, because some information should be explicitly specified for Ansible, either from `group_vars/all`, or from Vagrantfile.
I chose to write proper Ansible environment in Vagrantfile because in this way it simplifies K8s deployment on Vagrant (just `vagrant up`, not `edit group_vars/all; vagrant up`).

By switching to static IPs instead of DHCP in VirtualBox I make VirtualBox configuration closer to other configurations (i.e. bare metal).
Currently hostnames are managed in Vagrant-configuration by explicitly writing `/etc/hosts` file in [vagrant-ansible.yml](https://github.com/kubernetes/contrib/blob/master/ansible/vagrant/vagrant-ansible.yml), so we can't use DHCP.

In ideal world Vagrant+VirtualBox configuration should be identical to bare metal configuration, but due to issues like https://github.com/mitchellh/vagrant/issues/2779 it's far from being identical to real world configurations.
There is no need of additional "Vagrant+VirtualBox" configuration in ideal world, because Vagrant+VirtualBox is trying to emulate real world configuration, i.e. bare metal.

rutsky:
Note that I changed VBox IPs range from `172.30.0.0/24` to `172.32.0.0/24` in the last commit: https://github.com/kubernetes/contrib/pull/712/commits/740d6af5c3e52f804c26250daff11dbda6e486e1

`172.30.0.0` intersects with Flannel default IPs range: `172.16.0.0/12`.
Note, that I previously used not just random range, but range that VirtualyBox is actually assigned on my local setup.
So technically this PR is also fixes potential collision between VBox VMs IPs and Flanneld virtual IPs.

PR #714, opened 2016-04-05
Created variable for flannel download url

stephenrlouie:


rutsky:
How about moving `"{{ flannel_download_url_base }}v{{ flannel_version }}/flannel-{{ flannel_version }}-linux-amd64.tar.gz"` to defaults too?

```
# defaults/main.yaml

#The default url to download the flannel tar from
flannel_download_url_base: "https://github.com/coreos/flannel/releases/download/"
flannel_download_url: "{{ flannel_download_url_base }}v{{ flannel_version }}/flannel-{{ flannel_version }}-linux-amd64.tar.gz"
```

In this way all information about download URL will be in one place.

stephenrlouie:
I signed it!

rutsky:
Oh, and I think developers here prefer slash in expression, rather then in variable value (someone commented about this one of my previous commits):

```
flannel_download_url_base: "https://github.com/coreos/flannel/releases/download"
flannel_download_url: "{{ flannel_download_url_base }}/v{{ flannel_version }}/flannel-{{ flannel_version }}-linux-amd64.tar.gz"
```

stephenrlouie:
Thanks @rutsky, made those changes

stephenrlouie:
- Would it be possible to include `flannel_download_url_base`, `kube_download_url_base`, and `pypy_base_url` commented out in `group_vars/all.yml` It would enable anyone to host these files and fetch them from a local server. I saw cluster set up time go from ~45 minutes to less than 10 minutes with this change.

- If not, this should at least be a variable instead of a hard coded download url to enable the option

cc @rutsky @danehans @eparis

Issue #713, opened 2016-04-05
It should be possible to run a local submit queue with fake GitHub/Jenkins.


spxtr:
Right now we have the ability to do a `--dry-run`, which doesn't mutate GitHub but still tries to read from it and Jenkins. It would be useful to have `--fake`, which just generates fake PRs and the like for local testing.

I haven't quite looked through everything so this might be possible already, but if not I'd be happy to implement it.

@kubernetes/sig-testing what do you think?

eparis:
there is probably a lot of that code already written in the submit-queue_test.go.  Sounds like a good idea to me...

spxtr:
The combination of `--dry-run` and `--fake-e2e` is enough to do most of what I need.

PR #715, opened 2016-04-06
Add a fake e2e tester.

spxtr:
Pass in `--fake-e2e` and all jobs will be stable. I'm really not sure if this is the idiomatic way of doing this in go, but it seems reasonable.

It's probably a good idea to switch the submit-queue_test to use this, but that can wait.

ref #713


spxtr:
@eparis mind taking a look?

PR #723, opened 2016-04-06
Matching the url naming convention to the kube and flannel names

stephenrlouie:
- Changed naming convention to match kube and flannel
- Added the slash in the path, given @rutsky's comment: https://github.com/kubernetes/contrib/pull/714


eparis:
needs a rebase.

PR #729, opened 2016-04-06
Add job health to the internal e2e page.

spxtr:
Every munge loop, the submit queue now tracks which jobs are stable and which aren't, and keeps a running total. The internal e2e page now displays these percents next to the job like so:

ss

I'd like to have a chart showing health as a function of time, and I would like to have it save the health data across restarts. It also won't be tough to make the color scale from green to red as the job fails more and more. Those can happen in future PRs.

cc @kubernetes/sig-testing

### DIFF
--- a/mungegithub/submit-queue/www/index.html
+++ b/mungegithub/submit-queue/www/index.html
@@ -147,9 +147,12 @@
               <md-chips ng-model="cntl.builds" readonly="true">
                 <md-chip-template title="{{$chip.msg}}">
                   <span style="color: {{$chip.color}}">{{$chip.state}}</span> <a ng-href="https://console.cloud.google.com/storage/kubernetes-jenkins/logs/{{$chip.name}}/{{$chip.id}}/">{{$chip.name}}</a>
-                  <span style="color: {{$chip.color}}">{{$chip.msg}}</span>
+                  <span style="color: {{$chip.color}}">{{$chip.msg}}</span> <span title="Job stability">{{$chip.stability}}</span>
                 </md-chip-template>
               </md-chips>
+              <br>
+              <h2 class="md-title" ng-show="cntl.OverallHealth.length > 0">Overall Health: {{ cntl.OverallHealth }}</h2>
+              <p>Health percents are the fraction of the time that a given job is stable since the submit queue restarted<span ng-show="cntl.StartTime.length > 0"> ({{ cntl.StartTime }})</span>.</p>
             </md-tab-body>
           </md-tab>
 
--- a/mungegithub/submit-queue/www/script.js
+++ b/mungegithub/submit-queue/www/script.js
@@ -8,12 +8,15 @@ function SQCntl(dataService, $interval, $location) {
   self.prs = {};
   self.users = {};
   self.builds = {};
+  self.health = {};
   self.prQuerySearch = prQuerySearch;
   self.historyQuerySearch = historyQuerySearch;
   self.goToPerson = goToPerson;
   self.selectTab = selectTab;
   self.queryNum = 0;
   self.selected = 0;
+  self.OverallHealth = "";
+  self.StartTime = "";
   self.StatOrder = "-Count";
   self.location = $location;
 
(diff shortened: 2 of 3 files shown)


mungegithub/submit-queue/www/index.html
```diff
@@ -148,9 +148,12 @@ <h2 class="md-toolbar-tools">Google End-to-End Results</h2>
                 <md-chips ng-model="cntl.builds" readonly="true">
                   <md-chip-template title="{{$chip.msg}}">
                     <span style="color: {{$chip.color}}">{{$chip.state}}</span> <a ng-href="https://storage.cloud.google.com/kubernetes-jenkins/logs/{{$chip.name}}/{{$chip.id}}/">{{$chip.name}}</a>
-                    <span style="color: {{$chip.color}}">{{$chip.msg}}</span>
+                    <span style="color: {{$chip.color}}">{{$chip.msg}}</span> <span title="Job stability">{{$chip.stability}}</span>
                   </md-chip-template>
                 </md-chips>
+                <br>
+                <h2 class="md-title">Total Health: <span id="e2e-health"></span></h2>
+                <p>Health percents are the fraction of the time that a given job is stable since the submit queue restarted.</p>
```
ixdy: can we print out when that was?

spxtr:
@eparis do you mind taking a look at this? I'm not sure how often the submit queue gets restarted, and it might be worth implementing some more features before merging this PR. Plus, my JS isn't particularly strong so I probably did something silly.

spiffxp:
@spxtr re: more features before merging, was wondering if it made sense to tackle #22580

eparis:
submit queue gets updated/restarted on average about once a week.

I kinda feel like I'd like to see the chip % be supplied with the chip information rather than in the separate "overall health" but that's up to you....

mungegithub/submit-queue/www/index.html
```diff
@@ -148,9 +148,12 @@ <h2 class="md-toolbar-tools">Google End-to-End Results</h2>
                 <md-chips ng-model="cntl.builds" readonly="true">
                   <md-chip-template title="{{$chip.msg}}">
                     <span style="color: {{$chip.color}}">{{$chip.state}}</span> <a ng-href="https://storage.cloud.google.com/kubernetes-jenkins/logs/{{$chip.name}}/{{$chip.id}}/">{{$chip.name}}</a>
-                    <span style="color: {{$chip.color}}">{{$chip.msg}}</span>
+                    <span style="color: {{$chip.color}}">{{$chip.msg}}</span> <span title="Job stability">{{$chip.stability}}</span>
                   </md-chip-template>
                 </md-chips>
+                <br>
+                <h2 class="md-title">Overall Health: <span id="e2e-health"></span></h2>
```
eparis: I'm not certain but I think you can replace the `<span></span>` with just `{{ cntl.OverAllHealth }}` and then set `self.OverAllHealth` everytime you update it...

mungegithub/submit-queue/www/script.js
```diff
@@ -153,6 +170,14 @@ function SQCntl(dataService, $interval, $location) {
         obj.msg = job.Status;
         failedBuild = true;
       }
+      if (self.health.TotalLoops !== 0 &&
```
eparis: I wish that the response here came with health information...

spxtr:
The reason I supply the chip % separately is that in the near future I would like to supply data to plot % as a function of time in the `/health` request. It's certainly not too big of a deal to do both or one or the other, so I'd be fine with switching.

eparis:
I won't argue if this is going to leave you in a better place. I personally would rather have simpler easier javascrip/html and more complex golang, but that just because I know golang better. Stitching 2 data sources together in the javescript is fine if it's going to help down the road.

spxtr:
I think it might be worth putting it into the google-internal-ci request. I'll make that switch tomorrow.

spxtr:
@eparis PTAL

The reason I'm keeping them separate is that eventually I'll want to send lots more data in this request, but it only updates every 10 minutes server-side and it might get pretty hefty. It's probably a premature optimization, but it also keeps the go code much simpler. The JS wouldn't really look much different either way.

Issue #716, opened 2016-04-06
[Vagrant+VBox+Ansible+CoreOS] DNS is not working in containers


rutsky:
With #712 I managed to get almost working Kubernetes cluster (yes, again "almost" :) ).

```
kube-master ~ # kubectl cluster-info
Kubernetes master is running at http://localhost:8080
Elasticsearch is running at http://localhost:8080/api/v1/proxy/namespaces/kube-system/services/elasticsearch-logging
Heapster is running at http://localhost:8080/api/v1/proxy/namespaces/kube-system/services/heapster
Kibana is running at http://localhost:8080/api/v1/proxy/namespaces/kube-system/services/kibana-logging
KubeDNS is running at http://localhost:8080/api/v1/proxy/namespaces/kube-system/services/kube-dns
Grafana is running at http://localhost:8080/api/v1/proxy/namespaces/kube-system/services/monitoring-grafana
InfluxDB is running at http://localhost:8080/api/v1/proxy/namespaces/kube-system/services/monitoring-influxdb
kube-master ~ # kubectl get pods --all-namespaces
NAMESPACE     NAME                                    READY     STATUS    RESTARTS   AGE
default       wip-2165332859-lc4zn                    1/1       Running   0          4h
kube-system   elasticsearch-logging-v1-790pr          1/1       Running   0          8h
kube-system   elasticsearch-logging-v1-sqtin          1/1       Running   1          8h
kube-system   fluentd-elasticsearch-kube-node-1       1/1       Running   0          5h
kube-system   fluentd-elasticsearch-kube-node-2       1/1       Running   0          5h
kube-system   heapster-v1.1.0.beta1-345332335-vhp32   4/4       Running   0          8h
kube-system   kibana-logging-v1-x0w4y                 1/1       Running   0          8h
kube-system   kube-dns-v11-wcwna                      4/4       Running   0          3h
kube-system   monitoring-influxdb-grafana-v3-h9bxv    2/2       Running   0          8h
```

DNS addon is started and passes it's own health checks, but DNS in other containers doesn't work:

```
kube-master ~ # kubectl run wip --rm --tty -i --image=busybox -- /bin/sh
Waiting for pod default/wip-2165332859-lc4zn to be running, status is Pending, pod ready: false

Hit enter for command prompt

/ # nslookup kubernetes.default.svc.cluster.local
Server:    10.254.0.10
Address 1: 10.254.0.10

[...]

PR #731, opened 2016-04-07
add Docker dropin to start it after Flannel on CoreOS

rutsky:
Without this Docker can start before Flannel which will lead to
incorrect configuration of docker0 device (since configuration for it
prepares Flannel).

This should fix issue #716.

As this issue was not reported on other configuration, I believe on
other OSes Docker already depends on Flannel.

/cc: @danehans @eparis

### DIFF
--- a/ansible/roles/flannel/files/40-docker-depends-on-flannel.conf
+++ b/ansible/roles/flannel/files/40-docker-depends-on-flannel.conf
@@ -1,0 +1,12 @@
+# Dropin for docker.service
+
+[Unit]
+# Flannel must be started before Docker.
+# Flannel acquires subnet for the host and writes appropriate
+# configuration in /run/flannel_docker_opts.env that is being read
+# by Docker service.
+# See https://coreos.com/kubernetes/docs/latest/deploy-master.html
+# and https://github.com/coreos/flannel/issues/112#issuecomment-72708347
+# for details.
+Requires=flanneld.service
+After=flanneld.service
--- a/ansible/roles/flannel/tasks/client.yml
+++ b/ansible/roles/flannel/tasks/client.yml
@@ -14,7 +14,7 @@
   when: not is_atomic and flannel_source_type != "github-release"
 
 - set_fact:
-    flannel_use_upstart: ansible_distribution == 'Ubuntu' and ansible_distribution_major_version|int < 15
+    flannel_use_upstart: "{{ ansible_distribution == 'Ubuntu' and ansible_distribution_major_version|int < 15 }}"
 
 - name: Install Flannel with github release
   include: github-release.yml
--- a/ansible/roles/flannel/tasks/client.yml
+++ b/ansible/roles/flannel/tasks/client.yml
@@ -1,4 +1,6 @@
 ---
+# Install Flannel.
+
 - name: Force to use github-release when packages are not available
   set_fact:
     flannel_source_type: "github-release"
@@ -11,12 +13,12 @@
         state: latest
   when: not is_atomic and flannel_source_type != "github-release"
 
+- set_fact:
+    flannel_use_upstart: ansible_distribution == 'Ubuntu' and ansible_distribution_major_version|int < 15
+
 - name: Install Flannel with github release
   include: github-release.yml
   when: flannel_source_type == "github-release"
-
-- set_fact:
-    flannel_use_upstart: ansible_distribution == 'Ubuntu' and ansible_distribution_major_version|int < 15
 
 - name: Set Flannel config file directory
   set_fact:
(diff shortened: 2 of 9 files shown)


danehans:
A couple thoughts:

A) I think this fix should live in the docker role and not the flannel role.
B) This patch will break non-flannel networking. Contrib also support Opencontrail and soon Contiv.
C) I would like to avoid implementing fixes that are OS specific, unless absolutely required. This will help us a great deal maintaining multi-OS's. Can we try to figure out a way to create a dep that is OS and network provider agnostic?

Maybe a flannel handler should be created to restart flannel. As part of the flannel restart handler, docker is restart. You can checkout out the docker restart handler to see what I'm getting at. Or maybe we implement the same logic as a PostExec systemd unit/drop-in for all OS types. I believe this logic is needed to support this use case:

I deploy k8s using contrib. I make a mistake on the flannel subnet I use, so I update group_vars with a different subnet and run setup.sh. In this scenario, I believe flanneld gets the new network info, but docker daemon is not restarted to have it read the new BIP from the flannel env file.

Let me know your thoughts.

rutsky:
Regarding A) and B) notes. I placed Docker dropin in Flannel role intentionally: this way it will not be applied for other networking types.
We already assume in Flannel role that Docker is used, so this doesn't add any new dependencies between roles.

rutsky:
Regarding C) I don't quite follow you. We have specific rules for different OSes (i.e. `is_coreos`, `is_atomic` usage) and I don't understand why adding additional dep on CoreOS is different?
Step with adding dependency dropin is documented in CoreOS installation guide.
And if this issue is not reproduced on other OSes, I hope it fixed in them in upstream, so we shouldn't change anything for them.

rutsky:
Regarding handlers: having properly propagated changes during repetitive runs of Ansible is a must.
We have issues with this in lots of places and probably with flannel configuration too (I'm on a phone now and can't easily check atm), as you noted @danehans.
But it's not related to this specific issue with current implementation, let's open separate bug report for this (I planned to address it, because I want to use these Ansible scripts not for one-time deployment, but rather for continuous support of a single deployment).

ansible/roles/flannel/files/40-docker-depends-on-flannel.conf
```diff
@@ -0,0 +1,9 @@
+# Dropin for docker.service
+
+[Unit]
+# Flannel must be started before Docker.
```
danehans: @rutsky do you mind adding a link to the coreos url you mentioned in your feedback as a comment? I think this may be it: https://github.com/coreos/flannel/issues/112

danehans:
Thanks for the additional information. I think I understand the situation better now. Something still irks me about a role managing a drop-in for a different role. If @eparis is fine with it, then LGTM.

eparis:
I think I am not ok with it as a CoreOS only thing. It feels to me it should be more of a 'what is your package source'. If you install docker/flannel from the RH/Fedora package manager (something like) this will be done automatically. But if you were to get sources from elsewhere (is that possible for docker or flannel, I can't remember any more) then we need to lay it down for you...

ansible/roles/flannel/files/40-docker-depends-on-flannel.conf
rutsky: @danehans I added links on bug-report you found and CoreOS documentation from which I learned this.

rutsky:
@eparis I agree with you that this shouldn't be CoreOS-specific.

Unfortunately this requires refactoring of flannel role, which I've done.

I moved init-files writing to `github-release.yml` --- custom upstart or systemd service file should be used only if non-packaged flannel is used.
In `github-release.yml` I switch by systemd/upstart and include either `systemd-service.yml`, or `upstart-service.yml`.

I tested changes in configuration with CoreOS with incremental Ansible invocation, @danehans I'll appreciate if you test these changes in your configuration.

I haven't tested Ubuntu/upstart case, since there is no Vagrant configuration for it and I don't have appropriate bare metal setup.

eparis:
I'm happy. thank you!

rutsky:
I tested this PR with from-scratch deploying of K8s on Vagrant+VirtualBox+Ansible+CoreOS - it works.

ingvagabund:
This breaks ansible. When running setup.sh I get:
```
ERROR: that is not a legal parameter in an Ansible task or handler
```

It is caused by missing indent inside of ``roles/flannel/tasks/upstart-service.yml``.
Instead of this:

```
- assert:
  that:
    - flannel_source_type == "github-release"
```

it has to be
```
- assert:
    that:
      - flannel_source_type == "github-release"
```

ingvagabund:
Already fixed by https://github.com/kubernetes/contrib/pull/767

PR #736, opened 2016-04-07
Do not require (or set) release-note-* labels on non-master

eparis:


eparis:
@roberthbailey @david-mcmahon

eparis:
(And I slid some other things in there too, like updating to the latest angular/material and a file rename)

david-mcmahon:
During some testing of release notes today I noticed this - https://github.com/kubernetes/kubernetes/pull/23755.  So there are cases when changes land directly on the non-master branches.  
This is either wrong and needs to be stopped from happening or we probably need to account for it since it will have no release-note-* labels and no ref back to a master PR.

Does anyone know if this is common?  I guess the worst thing is that it simply won't be release noted.

@bgrant0607

spiffxp:
@david-mcmahon kubernetes/kubernetes#23650 is another example, couldn't directly cherry-pick from master; some of the other v1.1 conformance tests vs. v1.2 cluster issues will probably require something similar

david-mcmahon:
A couple of options:
1. Leave the bot as is and a release-note-* label is optional on non-master PRs.  I can adjust tooling and doc to account for that.
1. Change the bot again to require release-note-* labels on non-master PRs when it wasn't cherry-picked.  This will still require tooling and (different) doc changes.

roberthbailey:
There are cases where the release branch has diverged too far from master to do a cherry pick, in which case we submit a change directly onto the release branch. This is less common than doing a cherry pick, but isn't incredibly rare either.

david-mcmahon:
I've updated the PR that is clarifying the docs - https://github.com/kubernetes/kubernetes/pull/23939/files#diff-6984223442de65b99c008b3a6a5a6721R87 - with info about this exception.

PR #737, opened 2016-04-07
Make utils URL configurable

gmarek:
cc @wojtek-t


gmarek:
@eparis PTAL - it's needed to write tests of the GCS logic.

PR #738, opened 2016-04-07
Require labels on non-master PRs whose 'parents' do not have release-note-*

eparis:


eparis:
https://github.com/kubernetes/contrib/pull/738/files#diff-7294a858e5f70293980043d7fe211686R64 is the interesting part of the PR. The rest is refactoring to make that possible.

eparis:
@david-mcmahon @roberthbailey 

Right now I have completely disabled the release-note label stuff on non-master. this brings it back for some PRs

david-mcmahon:
How do you bang these out so fast?  LGTM.  I'll update the doc to state that we now, in fact, do enforce this!

PR #748, opened 2016-04-08
conditionally create /etc/hosts file

rutsky:
Now /etc/hosts file is not being touched (i.e. modification time is not
changed) if it already exists.

Removed unnecessary "changed" state in Ansible run summary.

/cc @danehans

### DIFF
--- a/ansible/vagrant/vagrant-ansible.yml
+++ b/ansible/vagrant/vagrant-ansible.yml
@@ -11,8 +11,13 @@
     - public_iface: "{{ ansible_default_ipv4.interface }}"
   tasks:
     # On stock Core OS there is no /etc/hosts.
+    - name: "Check is /etc/hosts file exists"
+      stat: path=/etc/hosts
+      register: hosts
+
     - name: "Create hosts file if it is not present"
       file: path=/etc/hosts state=touch
+      when: not hosts.stat.exists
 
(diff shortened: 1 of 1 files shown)


danehans:
LGTM

ansible/vagrant/vagrant-ansible.yml
```diff
@@ -11,8 +11,12 @@
     - public_iface: "{{ ansible_default_ipv4.interface }}"
   tasks:
     # On stock Core OS there is no /etc/hosts.
+    - name: "Check is /etc/hosts file exists"
+      stat: path=/etc/hosts
+      register: hosts
+
     - name: "Create hosts file if it is not present"
-      file: path=/etc/hosts state=touch
+      file: path=/etc/hosts state={{ "file" if hosts.stat.exists else "touch" }}
```
eparis: should we just use 'copy' with 'force=false' ?  and some empty src file? Maybe this is better, but I'm not in love with the conditional in state= ......

rutsky:
@eparis 
> should we just use 'copy' with 'force=false' ? and some empty src file? Maybe this is better, but I'm not in love with the conditional in state= ......

Good point! Fixed by conditionally touching file, please take a look.

eparis:
I like that better.  Great.

PR #749, opened 2016-04-08
update K8s version used in github-release to 1.2.2

rutsky:
/cc @danehans @eparis


eparis:
wish we had an automatic was to get the latest stable release by default.

PR #751, opened 2016-04-10
Update munger tests so they will work for GCS based checker

gmarek:
Needed for moving to GCS. @kubernetes/sig-testing

### DIFF
--- a/mungegithub/mungers/e2e/e2e.go
+++ b/mungegithub/mungers/e2e/e2e.go
@@ -19,6 +19,7 @@ package e2e
 import (
 	"bufio"
 	"fmt"
+	"strconv"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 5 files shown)


mungegithub/mungers/e2e/e2e.go
```diff
@@ -110,68 +111,83 @@ const (
...
+		response, err := e.GoogleGCSBucketUtils.GetFileFromJenkinsGoogleBucket(job, lastBuildNumber, path)
+		if err != nil {
+			glog.Errorf("Error while getting data for %v/%v/%v: %v", job, lastBuildNumber, path, err)
+			e.setBuildStatus(job, "Not Stable", strconv.Itoa(lastBuildNumber))
+			continue
+		}
+		defer response.Body.Close()
+
+		thisStable := true
+		for response.StatusCode != 200 {
```
eparis: this is actually the introduction of the 200.

gmarek: Actually I should change this to http.StatusOK

PR #753, opened 2016-04-10
Update keepalived

aledbf:
## 0.4
- Update keepalived to [1.2.20](https://github.com/acassen/keepalived/blob/master/ChangeLog)
- Use iptables parameter to not respond on addresses that it does not own
- Replace annotations with ConfigMap to specify services and IP addresses
- Avoid unnecessary reloads if the configuration did not change
- The parameter `--password` was removed because is not supported in vrrp v3


aledbf:
ping @bprashanth

keepalived-vip/main.go
```diff
@@ -54,6 +50,8 @@ var (
 
 func main() {
 	clientConfig := kubectl_util.DefaultClientConfig(flags)
+
+	flags.AddGoFlagSet(flag.CommandLine)
```
bprashanth: do you need this AND Parse(os.Args) ?

keepalived-vip/vip-rc.yaml
```diff
@@ -16,24 +16,19 @@ spec:
...
-              fieldRef:
-                fieldPath: metadata.namespace
         volumeMounts:
         - mountPath: /lib/modules
           name: modules
           readOnly: true
+        # to use unicast
+        args:
+        - --services-configmap=default/vip-configmap
+        #- --use-unicast=true
```
bprashanth: please either add a comment about when someone would want to enable this.

bprashanth:
LGTM but for the 2 nits

keepalived-vip/main.go
aledbf: @bprashanth without this is not possible to use flags like `--v` to increase the verbosity of the logs

bprashanth: Ah, does that show up in --help? (how to increase verbosity i mean), if not might be worth adding it, or noting it in some doc.

aledbf: @bprashanth yes (and I don't ned to add those flags in main.go)
```
$ ./kube-keepalived-vip --help
Usage of :
      --alsologtostderr[=false]: log to standard error as well as files
      --certificate-authority="": Path to a cert. file for the certificate authority.
      --client-certificate="": Path to a client certificate file for TLS.
      --client-key="": Path to a client key file for TLS.
      --cluster="": The name of the kubeconfig cluster to use
      --context="": The name of the kubeconfig context to use
      --insecure-skip-tls-verify[=false]: If true, the server's certificate will not be checked for validity. This will make your HTTPS connections insecure.
      --kubeconfig="": Path to the kubeconfig file to use for CLI requests.
      --log_backtrace_at=:0: when logging hits line file:N, emit a stack trace
      --log_dir="": If non-empty, write log files in this directory
      --logtostderr[=true]: log to standard error instead of files
      --namespace="": If present, the namespace scope for this CLI request.
      --password="": Password for basic authentication to the API server.
  -s, --server="": The address and port of the Kubernetes API server
      --services-configmap="": Name of the ConfigMap that contains the definition of the services to expose.
		The key in the map indicates the external IP to use. The value is the name of the
		service with the format namespace/serviceName and the port of the service could be a number or the
		name of the port.
      --stderrthreshold=2: logs at or above this threshold go to stderr
      --token="": Bearer token for authentication to the API server.
      --use-kubernetes-cluster-service[=true]: If true, use the built in kubernetes
        cluster for creating the client
      --use-unicast[=false]: use unicast instead of multicast for communication
		with other keepalived instances
      --user="": The name of the kubeconfig user to use
      --username="": Username for basic authentication to the API server.
      --v=0: log level for V logs
      --vmodule=: comma-separated list of pattern=N settings for file-filtered logging
F0411 22:19:35.756241   91584 main.go:67] Please specify --services-configmap
```

bprashanth:
LGTM will merge on green

aledbf:
> could you explain to me why the migration to configmaps was required and what advantages are expected?

The main reason for this change is to avoid collisions of IP and ports (is easy to visualize the conf and fix conflicts)

> There is only one configmap for all services. When using multiple namespaces for different customers that are allowed to manage their own services this is a great problem.

Yes this is one of the trade-offs. Open to any suggestion that avoids the mentioned issue earlier.

> You can't have multiple services using different ports on the same external ip.

Right. In the current state this is not possible. I think it's easy to change this and allow the use of same IP different services if there's no port conflicts.

PR #754, opened 2016-04-11
Fix WeakStable in munger and add tests

gmarek:
Depends on #751

cc @wojtek-t

### DIFF
--- a/mungegithub/mungers/e2e/e2e.go
+++ b/mungegithub/mungers/e2e/e2e.go
@@ -19,6 +19,7 @@ package e2e
 import (
 	"bufio"
 	"fmt"
+	"net/http"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


mungegithub/mungers/e2e/e2e.go
```diff
@@ -168,9 +168,8 @@ func (e *RealE2ETester) GCSWeakStable() bool {
 			continue
 		}
 		defer response.Body.Close()
-
 		thisStable := true
-		for response.StatusCode != 200 {
+		for response.StatusCode == 200 {
```
eparis: bast you 200!  http.StatusOK for the ignorant among us (aka me)

eparis:
you happy?  I'm happy to merge both.

gmarek:
Well - it'd be great if you could take a look first, if I didn't do anything terribly stupid. But other then that - yes, I'm happy with them. We'll be able to move to GCS tomorrow.

PR #755, opened 2016-04-11
Run nginx controller on a non-minion node

bprashanth:
Add a `--running-in-cluster=false` mode. This is useful for running the controller on any non-minion node (eg: edge router, kubernetes master, local desktop for development etc). 

Also adds an all-in-one rc example.
@aledbf

### DIFF
--- a/ingress/controllers/nginx/main.go
+++ b/ingress/controllers/nginx/main.go
@@ -17,7 +17,6 @@ limitations under the License.
 package main
 
 import (
-	"flag"
 	"fmt"
 	"net/http"
 	"net/http/pprof"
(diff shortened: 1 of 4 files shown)


aledbf:
LGTM

ingress/controllers/nginx/main.go
```diff
@@ -94,25 +99,36 @@ func main() {
...
+	}
 	if err != nil {
 		glog.Fatalf("failed to create client: %v", err)
 	}
 
-	podInfo, err := getPodDetails(kubeClient)
-	if err != nil {
-		glog.Fatalf("unexpected error getting runtime information: %v", err)
+	runtimePodInfo := &podInfo{NodeIP: "127.0.0.1"}
+	if *inCluster {
```
bprashanth: Actually this doesn't make sense if it's running in a static pod on the kubernetes master. Will think of something better.

bprashanth:
I'll send a follow-up pr for the netlink ip address fix, merging this as is.

PR #758, opened 2016-04-11
submit-queue: Give information about how long the queue has been blocked

### LINKED ISSUE #721: submit queue dashboard should display time since last merge
To make it easier to detect when it's been blocked.

eparis:
sq

And how fast it is running...

These number ignore PRs with e2e-not-required and those merged by hand
(since both require almost nothing from the submit queue)

resolves #721

### DIFF
--- a/mungegithub/mungers/submit-queue.go
+++ b/mungegithub/mungers/submit-queue.go
@@ -27,6 +27,7 @@ import (
 	"sync"
 	"time"
 
+	"k8s.io/kubernetes/pkg/util"
 	"k8s.io/kubernetes/pkg/util/sets"
 
 	"k8s.io/contrib/mungegithub/features"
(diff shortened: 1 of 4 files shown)


eparis:
@gmarek do you think I should have tacked onto the side of `submitQueueHealth`

bgrant0607:
Thanks!

cc @ixdy @fejta

fejta:
Please sync up with @spxtr who is also making some similar changes

spxtr:
An alternative is to keep a list of PR merge times in the `SubmitQueue` struct. When we call `updateHealth`, remove entries older than a day and calculate the merge rate for the last day as a simple (num in merge times array) / (1 day). I think that would be conceptually simpler, but it might be less interesting than this smoothed statistic.

eparis:
@spxtr that is so much less interesting!  (but admittedly probably more reasonable)  sq.statusHistory keeps the last 128 items. That could likely be changed somehow to include 128 items or the last 24 hours, whichever is greater. Although statusHistory includes both `e2e-not-required` and things which got onto the queue and were subsequently merged by a human in its merged list. (It obviously could be debated if those should count)

I'm trying to decide what I think about it. Lets assume the queue was running at top speed, 2 PRs per hour for quite. So it shows 48 per day.
If the queue stops completely for:
6 hours yours shows 36/day mine shows 15/day
12 hours yours shows 24/day mine shows 10/day
18 hours yours shows 12/day mine shows 7/day
24 hours yours shows 0/day mine shows ~0/day

Which one would we rather show people? In some ways yours is obviously 'more correct' but I sorta think mine is more useful at casual glance. (same thing happens in the other direction, my numbers respond faster)

spxtr:
One other option is to show both merges in the last day as well as merges in the last 6 hours. It's certainly not ideal, but it's easy to understand. Yours is definitely more interesting at a glance. I like it, but I think it should say something like "Estimated Merging X Per Day". Anyway that's mostly bikeshedding :)

I think I intended the `submitQueueHealth` struct to contain this sort of data. Is there a particular reason why it shouldn't?

eparis:
No I think `submitQueueHealth` makes sense. I started to put it there when I rebased but didn't know if you wanted it to be e2e health instead of sq health.

I'll take a yay/nay. This method (+ bikshedding) or should I do averages?

spxtr:
I think this method (+ "estimated") is good. I like that it's responsive.

PR #759, opened 2016-04-11
Add tests for WeakStable and enable WeakStable logic

gmarek:
@kubernetes/sig-testing @spxtr

### DIFF
--- a/mungegithub/mungers/submit-queue.go
+++ b/mungegithub/mungers/submit-queue.go
@@ -259,9 +259,8 @@ func (sq *SubmitQueue) AddFlags(cmd *cobra.Command, config *github.Config) {
 		"kubernetes-e2e-gce-scalability",
 		"kubernetes-kubemark-5-gce",
 	}, "Comma separated list of jobs in Jenkins to use for stability testing")
-	cmd.Flags().StringSliceVar(&sq.WeakStableJobNames, "weak-stable-jobs", []string{
-		"kubernetes-kubemark-500-gce",
-	}, "Comma separated list of jobs in Jenkins to use for stability testing that needs only weak success")
+	cmd.Flags().StringSliceVar(&sq.WeakStableJobNames, "weak-stable-jobs", []string{},
+		"Comma separated list of jobs in Jenkins to use for stability testing that needs only weak success")
 	cmd.Flags().StringVar(&sq.JenkinsHost, "jenkins-host", "http://jenkins-master:8080", "The URL for the jenkins job to watch")
 	cmd.Flags().StringSliceVar(&sq.RequiredStatusContexts, "required-contexts", []string{}, "Comma separate list of status contexts required for a PR to be considered ok to merge")
 	cmd.Flags().StringVar(&sq.E2EStatusContext, "e2e-status-context", jenkinsE2EContext, "The name of the github status context for the e2e PR Builder")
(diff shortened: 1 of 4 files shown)


mungegithub/mungers/submit-queue_test.go
```diff
@@ -409,8 +416,10 @@ func TestSubmitQueue(t *testing.T) {
 		events           []github.IssueEvent
 		ciStatus         *github.CombinedStatus
 		jenkinsJob       jenkins.Job
-		lastBuildNumber  string
+		lastBuildNumber  int
```
spiffxp: just curious, what prompted the move from `string` to `int`?

gmarek: Necessity of computing `value - 1` for weakStable check.

mungegithub/mungers/submit-queue.go
```diff
@@ -259,9 +259,8 @@ func (sq *SubmitQueue) AddFlags(cmd *cobra.Command, config *github.Config) {
 		"kubernetes-e2e-gce-scalability",
 		"kubernetes-kubemark-5-gce",
 	}, "Comma separated list of jobs in Jenkins to use for stability testing")
-	cmd.Flags().StringSliceVar(&sq.WeakStableJobNames, "weak-stable-jobs", []string{
-		"kubernetes-kubemark-500-gce",
```
eparis: This test shouldn't be here any more?

mungegithub/mungers/submit-queue_test.go
```diff
@@ -164,13 +165,19 @@ func FailJenkins() jenkins.Job {
 	}
 }
 
+func getJUnit(testsNo int, failuresNo int) []byte {
```
eparis: Does this belong here or somehow over in the utils? Also does it matter that you are no longer hard coding `time=1234` ?

mungegithub/mungers/submit-queue.go
gmarek: I don't want to add the code that enables it, together with adding the test in a single PR.

mungegithub/mungers/submit-queue_test.go
gmarek: No - I just found it useful for debugging to have different timestamps.

mungegithub/mungers/submit-queue_test.go
```diff
@@ -746,6 +853,42 @@ func TestSubmitQueue(t *testing.T) {
...
+				data, err := json.Marshal(test.weakResults[buildNumberCopy])
+				if err != nil {
+					t.Errorf("Unexpected error: %v", err)
+				}
+				w.Write(data)
+			})
+		}
+		for junitFile, xml := range test.gcsJunit {
+			path = fmt.Sprintf("/bar/%v/artifacts/%v", test.lastBuildNumber, junitFile)
+			// workaround go for loop semantics
```
eparis: Isn't xml already a copy? What is the semantic issue here? (just for me knowledge)

mungegithub/mungers/submit-queue_test.go
spiffxp: thanks

mungegithub/mungers/submit-queue_test.go
gmarek: It's not really a copy. It's one of mine biggest issue with go. When you iterate over something, i.e. write `for x := range bla` compiler allocates a space for a single copy of element of `bla` and reuses it in all iterations. This is OK, for 'normal' processing, but it's kind of painful when you write a go routine inside a for loop or, like here, create a function to be called later.

PR #765, opened 2016-04-13
mungegithub: delete bot comments when they are no longer useful

eparis:
Almost all of the bot's comments serve a purpose that can be resolved.
Once they are resolved the comments no longer need to be present and
just make review more difficult.

This PR removes the comments created by this bot. It does NOT remove
comments created by the cherrypick or the k8s-bot.

Resolves: https://github.com/kubernetes/kubernetes/issues/24007

### DIFF
--- a/mungegithub/mungers/block_paths.go
+++ b/mungegithub/mungers/block_paths.go
@@ -26,11 +26,18 @@ import (
 	"k8s.io/kubernetes/pkg/util/yaml"
 
 	"github.com/golang/glog"
+	githubapi "github.com/google/go-github/github"
 	"github.com/spf13/cobra"
 )
 
+const (
+	blockPathFormat = `Adding label:%s because PR changes docs prohibited to auto merge
+See http://kubernetes.io/editdocs/ for information about editing docs`
+)
+
 var (
-	_ = fmt.Print
+	_             = fmt.Print
+	blockPathBody = fmt.Sprintf(blockPathFormat, doNotMergeLabel)
 )
 
 type configBlockPath struct {
@@ -46,7 +53,9 @@ type BlockPath struct {
 }
 
 func init() {
-	RegisterMungerOrDie(&BlockPath{})
+	b := &BlockPath{}
+	RegisterMungerOrDie(b)
+	registerShouldDeleteCommentFunc(b.isStaleComment)
 }
 
 // Name is the name usable in --pr-mungers
@@ -129,12 +138,21 @@ func (b *BlockPath) Munge(obj *github.MungeObject) {
 				if matchesAny(*f.Filename, b.doNotBlockRegexp) {
 					continue
 				}
-				body := fmt.Sprintf(`Adding label:%s because PR changes docs prohibited to auto merge
-See http://kubernetes.io/editdocs/ for information about editing docs`, doNotMergeLabel)
-				obj.WriteComment(body)
+				obj.WriteComment(blockPathBody)
 				obj.AddLabels([]string{doNotMergeLabel})
 				return
 			}
 		}
 	}
+}
+
+func (b *BlockPath) isStaleComment(obj *github.MungeObject, comment *githubapi.IssueComment) bool {
+	if *comment.Body != blockPathBody {
+		return false
+	}
+	stale := !obj.HasLabel(doNotMergeLabel)
+	if stale {
+		glog.V(6).Infof("Found stale BlockPath comment")
+	}
+	return stale
 }
--- a/mungegithub/mungers/comment-deleter.go
+++ b/mungegithub/mungers/comment-deleter.go
@@ -1,0 +1,94 @@
+/*
+Copyright 2016 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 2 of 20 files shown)


eparis:
@gmarek @spxtr you two have been looking at the queue recently.

mungegithub/mungers/submit-queue_test.go
```diff
@@ -182,7 +182,6 @@ func getTestSQ(startThreads bool, config *github_util.Config, server *httptest.S
 	}
 	sq.JobNames = []string{"foo"}
 	sq.WeakStableJobNames = []string{"bar"}
-	sq.WhitelistOverride = "ok-to-merge"
```
thockin: what happened to this?  Not in the PR comments?

thockin: NM, need to read whole PR before commenting.  sigh

mungegithub/mungers/rebuild.go
```diff
@@ -19,12 +19,13 @@ package mungers
 import (
 	"fmt"
 	"regexp"
+	"strings"
 
 	"k8s.io/contrib/mungegithub/features"
 	"k8s.io/contrib/mungegithub/github"
 
 	"github.com/golang/glog"
-	githubapi "github.com/google/go-github/github"
```
thockin: FWIW the runonname is more go-style than nicely_formatted_name, but I don't care

mungegithub/mungers/lgtm_after_commit.go
```diff
@@ -69,10 +74,28 @@ func (LGTMAfterCommitMunger) Munge(obj *github.MungeObject) {
...
+func isLGTMAfterCommitComment(comment *github_api.IssueComment) bool {
+	body := *comment.Body
+	return body == lgtmRemovedBody
+}
+
+func lgtmAfterComment(obj *github.MungeObject, comment *github_api.IssueComment) bool {
+	if comment.CreatedAt == nil {
+		return false
+	}
+	if !obj.HasLabel("lgtm") {
```
thockin: constant for "lgtm" ?

mungegithub/mungers/comment-deleter.go
```diff
@@ -0,0 +1,146 @@
...
+		if comment.User == nil || comment.User.Login == nil {
+			continue
+		}
+		if *comment.User.Login != botName {
+			continue
+		}
+		if comment.Body == nil {
+			continue
+		}
+		if isRebaseComment(comment) {
```
thockin: The SW engineer in me wants to complain that this is an open-coded list, rather than a set of registered-in-on-place plugins that have an optional method.  It's not too egregious given the scope, but for evolution...

thockin:
LGTM overall.  nits which you may choose to ignore and then self-LGTM

mungegithub/mungers/block_paths.go
```diff
@@ -26,11 +26,18 @@ import (
 	"k8s.io/kubernetes/pkg/util/yaml"
 
 	"github.com/golang/glog"
+	github_api "github.com/google/go-github/github"
 	"github.com/spf13/cobra"
 )
 
+const (
+	blockPathFormat = `Adding label:%s because PR changes docs prohibited to auto merge
```
gmarek: I'm not a native speaker, but I don't really understand this comment. Is something along the lines "You need to update docs"?

eparis: I'll rewrite it in this PR.

mungegithub/mungers/comment-deleter.go
```diff
@@ -0,0 +1,146 @@
...
+// CommentDeleter looks for comments which are no longer useful
+// and deletes them
+type CommentDeleter struct{}
+
+func init() {
+	RegisterMungerOrDie(CommentDeleter{})
+}
+
+// Name is the name usable in --pr-mungers
+func (CommentDeleter) Name() string { return commentDeleterName }
```
gmarek: Out of curiosity - is there a reason it works on value instead of a pointer?

mungegithub/mungers/comment-deleter.go
gmarek: I agree with Tim - we should add an interface for comment operations, make each munger register one for itself and here have simple for loop.

On top of that - I'm probably missing something, but how do we check if e.g. RebaseComment is no longer needed?

gmarek:
I like the idea!

mungegithub/mungers/comment-deleter.go
eparis: Already mostly finished doing this. Although I have a debate in my head. Should it only remove stale comments for 'active/selected' mungers, or for all mungers if they are enabled or not...

Today the 'cherrypick-must-have-milestone' mungers runs as part of the cherrypick-queue pocess and the others mungers run as part of the submit-queue.  If I only register for active mungers I need to run the comment deleter in both processes. If I register no matter what then I only need to run the comment-deleter in the submit-queue...


I could make the 'isRebaseComment' a little smarter. I assumed that one was only useful to send the e-mail and after that the 'needs-rebase' label would be the indicator. If you want we can leave the comment until the needs-rebase label has been removed....

gmarek: I'm fine with registering everything. It should be easy enough to change in future if we need to.

I guess Rebase should go away only if PR is rebased, that's my intuition. Generally comments can go away if someone acted on them.

PR #766, opened 2016-04-13
Add support for named port, better docs for TLS nginx Ingress

### LINKED ISSUE #743: NGINX ingress controller does not work properly with a named targetPort
When using a named targetPort on the service it does not resolve which endpoint port to use, as far as I can tell.

This is a bit of a hard problem to solve because it's not immediately apparent from the endpoints list what the port name is.

### LINKED ISSUE #781: Nginx Ingress Controller problem with named targetPort in Service
service.yaml:
```
apiVersion: v1
kind: Service
metadata:
  name: kibana-logging
  namespace: kube-system
  labels:
    k8s-app: kibana-logging
    kubernetes.io/cluster-service: "true"
    kubernetes.io/name: "Kibana"
spec:
  ports:
  - port: 5601
    protocol: TCP
    targetPort: 5601
  selector:
    k8s-app: kibana-logging
```

with ingress.yaml:
```
kind: Ingress
metadata:
  name: kibana-logging
  namespace: kube-system
spec:
  rules:
  - host: phive-logs.pmi.ag
    http:
      paths:
      - backend:
          serviceName: kibana-logging
          servicePort: 5601
        path: /
```

works flawlessly :)

but if i set `targetPort: ui` (as it is named in the pod definition) nginx ingress controller logs: `service kube-system/kibana-logging does no have any active endpoints` ... but:
```
$ kubectl get ep --namespace=kube-system kibana-logging
NAME             ENDPOINTS          AGE
kibana-logging   10.103.11.3:5601   4h
```

### LINKED ISSUE #858: Ingress servers duplicated when multiple ingresses
When I test against the ingress controller in https://github.com/kubernetes/contrib/tree/master/ingress/controllers/nginx, using the ingresses in https://github.com/kubernetes/contrib/blob/master/ingress/controllers/nginx/examples/ingress.yaml, multiple of each IP are listed the server config of nginx:
```
    upstream default-echoheaders-x-80 {
        least_conn;
        server 10.138.27.3:8080;
        server 10.138.27.3:8080;
        
    }
    
    upstream default-echoheaders-y-80 {
        least_conn;
        server 10.138.27.3:8080;
        
    }
```

From what I can tell, this causes issues when combined with the `proxy_next_upstream` flag in nginx for bigger server configs like:
nginx vhost traffic status monitor 2016-04-27 16-21-41

Is it possible to de-dup this server list or is this for some reason intentional?

### LINKED ISSUE #871: Ingress - SSL should not redirect by default
The nginx.tmpl redirects to https by default, I'm not sure this should be the standard behaviour.

```
        {{ if $server.SSL }}
        if ($scheme = http) {
            return 301 https://$host$request_uri;
        }
        {{ end }}
```

All the examples will break for example...

```
$ curl -v --resolve foo.bar.com:80:54.171.67.92 foo.bar.com/foo
* Added foo.bar.com:80:54.171.67.92 to DNS cache
* Hostname foo.bar.com was found in DNS cache
*   Trying 54.171.67.92...
* Connected to foo.bar.com (54.171.67.92) port 80 (#0)
> GET /foo HTTP/1.1
> Host: foo.bar.com
> User-Agent: curl/7.47.0
> Accept: */*
> 
< HTTP/1.1 301 Moved Permanently
< Server: nginx/1.9.13
< Date: Sat, 30 Apr 2016 06:24:53 GMT
< Content-Type: text/html
< Content-Length: 185
< Connection: keep-alive
< Location: https://foo.bar.com/foo
< 
<html>
<head><title>301 Moved Permanently</title></head>
<body bgcolor="white">
<center><h1>301 Moved Permanently</h1></center>
<hr><center>nginx/1.9.11</center>
</body>
</html>
* Connection #0 to host foo.bar.com left intact
```

### LINKED ISSUE #872: Ingress - nginx - option to specify path to nginx.tmpl
The nginx.tmpl is stored in the root of the docker image. It would be ideal to place within a directory and/or specify the path to the nginx.tmpl

An example use case is to override the nginx.tmpl using a configmap.

```
apiVersion: v1
kind: ReplicationController
metadata:
  name: nginx-ingress-controller
  labels:
    k8s-app: nginx-ingress-lb
spec:
  replicas: 1
  selector:
    k8s-app: nginx-ingress-lb
  template:
    metadata:
      labels:
        k8s-app: nginx-ingress-lb
        name: nginx-ingress-lb
    spec:
      terminationGracePeriodSeconds: 60      
      containers:
      - image: gcr.io/google_containers/nginx-ingress-controller:0.5
        name: nginx-ingress-lb
        imagePullPolicy: Always
        livenessProbe:
          httpGet:
            path: /healthz
            port: 10249
            scheme: HTTP
          initialDelaySeconds: 30
          timeoutSeconds: 5
        # use downward API
        env:
          - name: POD_NAME
            valueFrom:
              fieldRef:
                fieldPath: metadata.name
          - name: POD_NAMESPACE
            valueFrom:
              fieldRef:
                fieldPath: metadata.namespace
        ports:
        - containerPort: 80
          hostPort: 80
        - containerPort: 443
          hostPort: 4444
        args:
        - /nginx-ingress-controller
        - --default-backend-service=default/default-http-backend
        volumeMounts:
          - mountPath: /etc/nginx-template-tmpl
            name: nginx-template-volume
            readOnly: true
      volumes:
        - name: nginx-template-volume
          configMap:
            name: nginx
            items:
            - key: nginx.tmpl
              path: nginx.tmpl

```
Hard coded here.
https://github.com/kubernetes/contrib/blob/master/ingress/controllers/nginx/nginx/template.go#L46

### LINKED ISSUE #876: ingress - nginx - default http path rule creates an additional server.
The default http path rule creates an additional server, with an empty server_name and does not map to "_". An Example config.

```
# An Ingress with 2 hosts and 3 endpoints
kind: Ingress
metadata:
  name: echomap
spec:
  rules:
  - http:
      paths:
      - path: /testpath
        backend:
          serviceName: echoheaders-x
          servicePort: 80
  - host: foo.bar.com
    http:
      paths:
      - path: /foo
        backend:
          serviceName: echoheaders-x
          servicePort: 80
  - host: bar.baz.com
    http:
      paths:
      - path: /bar
        backend:
          serviceName: echoheaders-y
          servicePort: 80
      - path: /foo
        backend:
          serviceName: echoheaders-x
          servicePort: 80
```
nginx config
```
    server {
        listen 80;
        ...
        server_name ;
    }
    server {
        listen 80;
        ...
        server_name _;
        ...
    }
    server {
        listen 80;
        ...
        server_name bar.baz.com;
        ...
    }
```

full example.

https://gist.github.com/ajohnstone/4b75473fb2e8937d5fc7144d46aad4c4#file-02-nginx-conf-L2-L296

aledbf:
fixes #743
fixes #781
fixes #858
fixes #871
closes #872
fixes #876

### DIFF
--- a/ingress/controllers/nginx/README.md
+++ b/ingress/controllers/nginx/README.md
@@ -11,7 +11,6 @@ This is a nginx Ingress controller that uses [ConfigMap](https://github.com/kube
 - custom ssl_dhparam (optional). Just mount a secret with a file named `dhparam.pem`.
 - support for TCP services (flag `--tcp-services-configmap`)
 - custom nginx configuration using [ConfigMap](https://github.com/kubernetes/kubernetes/blob/master/docs/proposals/configmap.md)
-- custom error pages. Using the flag `--custom-error-service` is possible to use a custom compatible [404-server](https://github.com/kubernetes/contrib/tree/master/404-server) image
 
 
 ## Requirements
@@ -120,10 +119,13 @@ Please follow [test.sh](https://github.com/bprashanth/Ingress/blob/master/exampl
 
 Check the [example](examples/tls/README.md)
 
-### Force HTTPS
+### HTTP Strict Transport Security
 
-By default the controller redirects (301) to HTTPS if there is a TLS Ingress rule. To disable this behavior use `use-hts=false` in the NGINX ConfigMap.
+HTTP Strict Transport Security (HSTS) is an opt-in security enhancement specified through the use of a special response header. Once a supported browser receives this header that browser will prevent any communications from being sent over HTTP to the specified domain and will instead send all communications over HTTPS.
 
+By default the controller redirects (301) to HTTPS if there is a TLS Ingress rule. 
+
+To disable this behavior use `hsts=false` in the NGINX ConfigMap.
 
 #### Optimizing TLS Time To First Byte (TTTFB)
 
@@ -190,24 +192,21 @@ Please check the example `example/rc-default.yaml`
 To extract the information in JSON format the module provides a custom URL: `/nginx_status/format/json`
 
 
+
+### Custom errors
+
+In case of an error in a request the body of the response is obtained from the `default backend`. Each request to the default backend includes two headers: 
+- `X-Code` indicates the HTTP code
+- `X-Format` the value of the `Accept` header
+
+Using this two headers is possible to use a custom backend service like [this one](https://github.com/aledbf/contrib/tree/nginx-debug-server/Ingress/images/nginx-error-server) that inspect each request and returns a custom error page with the format expected by the client. This images handles `html` and `json` responses.
+
+
 ## Troubleshooting
 
 Problems encountered during [1.2.0-alpha7 deployment](https://github.com/kubernetes/kubernetes/blob/master/docs/getting-started-guides/docker.md):
 * make setup-files.sh file in hypercube does not provide 10.0.0.1 IP to make-ca-certs, resulting in CA certs that are issued to the external cluster IP address rather then 10.0.0.1 -> this results in nginx-third-party-lb appearing to get stuck at "Utils.go:177 - Waiting for default/default-http-backend" in the docker logs.  Kubernetes will eventually kill the container before nginx-third-party-lb times out with a message indicating that the CA certificate issuer is invalid (wrong ip), to verify this add zeros to the end of initialDelaySeconds and timeoutSeconds and reload the RC, and docker will log this error before kubernetes kills the container.
   * To fix the above, setup-files.sh must be patched before the cluster is inited (refer to https://github.com/kubernetes/kubernetes/pull/21504)
-
-### Custom errors
-
-The default backend provides a way to customize the default 404 page. This helps but sometimes is not enough.
-Using the flag `--custom-error-service` is possible to use an image that must be 404 compatible and provide the route /error
-[Here](https://github.com/aledbf/contrib/tree/nginx-debug-server/Ingress/images/nginx-error-server) there is an example of the the image
-
-The route `/error` expects two arguments: code and format
-* code defines the wich error code is expected to be returned (502,503,etc.)
-* format the format that should be returned For instance /error?code=504&format=json or /error?code=502&format=html
-
-Using a volume pointing to `/var/www/html` directory is possible to use a custom error
-
 
 ### Debug
 
--- a/ingress/controllers/nginx/README.md
+++ b/ingress/controllers/nginx/README.md
@@ -180,6 +180,15 @@ Using a ConfigMap it is possible to customize the defaults in nginx.
 Please check the [tcp services](examples/custom-configuration/README.md) example
 
 
+## Custom NGINX template
+
+The NGINX template is located in the file `/etc/nginx/template/nginx.tmpl`. Mounting a volume is possible to use a custom version.
+Use the [custom-template](examples/custom-template/README.md) example as a guide
+
+**Please note the template is tied to the go code. Be sure to no change names in the variable `$cfg`**
+
+
+
 ### NGINX status page
 
 The ngx_http_stub_status_module module provides access to basic status information. This is the default module active in the url `/nginx_status`.
--- a/ingress/controllers/nginx/README.md
+++ b/ingress/controllers/nginx/README.md
@@ -120,6 +120,9 @@ Please follow [test.sh](https://github.com/bprashanth/Ingress/blob/master/exampl
 
 Check the [example](examples/tls/README.md)
 
+### Force HTTPS
+
+By default the controller redirects (301) to HTTPS if there is a TLS Ingress rule. To disable this behavior use `use-hts=false` in the NGINX ConfigMap.
 
 
 #### Optimizing TLS Time To First Byte (TTTFB)
--- a/ingress/controllers/nginx/README.md
+++ b/ingress/controllers/nginx/README.md
@@ -241,3 +241,5 @@ The previous behavior can be restored using `retry-non-idempotent=true` in the c
 ## Limitations
 
 - Ingress rules for TLS require the definition of the field `host`
+- The IP address in the status of loadBalancer could contain old values
+
--- a/ingress/controllers/nginx/controller.go
+++ b/ingress/controllers/nginx/controller.go
@@ -618,15 +618,18 @@ func (lbc *loadBalancerController) getUpstreamServers(data []interface{}) ([]*ng
 	upstreams[defUpstreamName] = lbc.getDefaultUpstream()
 
 	servers := lbc.createServers(data)
-	// default server - no servername.
-	servers[defServerName] = &nginx.Server{
-		Name: defServerName,
-		Locations: []*nginx.Location{{
-			Path:         rootLocation,
-			IsDefBackend: true,
-			Upstream:     *lbc.getDefaultUpstream(),
-		},
-		},
+	if _, ok := servers[defServerName]; !ok {
+		// default server - no servername.
+		// there is no rule with default backend
+		servers[defServerName] = &nginx.Server{
+			Name: defServerName,
+			Locations: []*nginx.Location{{
+				Path:         rootLocation,
+				IsDefBackend: true,
+				Upstream:     *lbc.getDefaultUpstream(),
+			},
+			},
+		}
 	}
 
 	for _, ingIf := range data {
@@ -637,7 +640,14 @@ func (lbc *loadBalancerController) getUpstreamServers(data []interface{}) ([]*ng
 				continue
 			}
 
-			server := servers[rule.Host]
+			host := rule.Host
+			if host == "" {
+				host = defServerName
+			}
+			server := servers[host]
+			if server == nil {
+				server = servers["_"]
+			}
 
 			for _, path := range rule.HTTP.Paths {
 				upsName := fmt.Sprintf("%v-%v-%v", ing.GetNamespace(), path.Backend.ServiceName, path.Backend.ServicePort.String())
@@ -768,18 +778,23 @@ func (lbc *loadBalancerController) createServers(data []interface{}) map[string]
 		ing := ingIf.(*extensions.Ingress)
 
 		for _, rule := range ing.Spec.Rules {
-			if _, ok := servers[rule.Host]; !ok {
+			host := rule.Host
+			if host == "" {
+				host = defServerName
+			}
+
+			if _, ok := servers[host]; !ok {
 				locs := []*nginx.Location{}
 				locs = append(locs, &nginx.Location{
 					Path:         rootLocation,
 					IsDefBackend: true,
 					Upstream:     *lbc.getDefaultUpstream(),
 				})
-				servers[rule.Host] = &nginx.Server{Name: rule.Host, Locations: locs}
+				servers[host] = &nginx.Server{Name: host, Locations: locs}
 			}
 
-			if pemFile, ok := pems[rule.Host]; ok {
-				server := servers[rule.Host]
+			if pemFile, ok := pems[host]; ok {
+				server := servers[host]
 				server.SSL = true
 				server.SSLCertificate = pemFile
 				server.SSLCertificateKey = pemFile
--- a/ingress/controllers/nginx/controller.go
+++ b/ingress/controllers/nginx/controller.go
@@ -291,42 +291,52 @@ func (lbc *loadBalancerController) updateEpNamedPorts(key string) {
 				return
 			}
 
-			lbc.checkSvcForUpdate(svc)
+			err = lbc.checkSvcForUpdate(svc)
+			if err != nil {
+				lbc.svcEpQueue.requeue(key, err)
+				return
+			}
 		}
 	}
 }
 
-func (lbc *loadBalancerController) checkSvcForUpdate(svc *api.Service) {
+// checkSvcForUpdate verifies if one of the running pods for a service contains
+// named port. If the annotation in the service does not exists or is not equals
+// to the port mapping obtained from the pod the service must be updated to reflect
+// the current state
+func (lbc *loadBalancerController) checkSvcForUpdate(svc *api.Service) error {
+	// get the pods associated with the service
 	pods, err := lbc.client.Pods(svc.Namespace).List(api.ListOptions{
 		LabelSelector: labels.Set(svc.Spec.Selector).AsSelector(),
 	})
 	if err != nil {
-		glog.Errorf("error searching service pods %v/%v: %v", svc.Namespace, svc.Name, err)
-		return
+		return fmt.Errorf("error searching service pods %v/%v: %v", svc.Namespace, svc.Name, err)
+	}
+
+	if len(pods.Items) == 0 {
+		return nil
 	}
 
 	namedPorts := map[string]string{}
+	// we need to check only one pod searching for named ports
+	pod := &pods.Items[0]
+	glog.V(4).Infof("checking pod %v/%v for named port information", pod.Namespace, pod.Name)
+	for i := range svc.Spec.Ports {
+		servicePort := &svc.Spec.Ports[i]
 
-	for i := range pods.Items {
-		pod := &pods.Items[i]
-		glog.V(4).Infof("checking pod %v/%v for named port information", pod.Namespace, pod.Name)
-		for i := range svc.Spec.Ports {
-			servicePort := &svc.Spec.Ports[i]
-
-			_, err := strconv.Atoi(servicePort.TargetPort.StrVal)
+		_, err := strconv.Atoi(servicePort.TargetPort.StrVal)
+		if err != nil {
+			portNum, err := podutil.FindPort(pod, servicePort)
 			if err != nil {
-				portNum, err := podutil.FindPort(pod, servicePort)
-				if err != nil {
-					glog.V(4).Infof("failed to find port for service %s/%s: %v", svc.Namespace, svc.Name, err)
-					continue
-				}
-
-				if servicePort.TargetPort.StrVal == "" {
-					continue
-				}
-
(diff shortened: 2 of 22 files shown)


aledbf:
ping @bprashanth

aledbf:
In case of services with namedPorts the Ingress controller add an annotation to the service
````
$ kubectl get ing
NAME        RULE           BACKEND   ADDRESS
demo-port   -                        172.17.4.99
            foo.bar.com
            /              echoheaders-xtp:echo-port
```

```
kubectl get svc echoheaders-multiple -o yaml
apiVersion: v1
kind: Service
metadata:
  annotations:
    kubernetes.io/ingress-named-ports: '{"echo-port":"8080"}'
  labels:
    app: echoheaders
  name: echoheaders-multiple
  namespace: default
spec:
  clusterIP: 10.3.0.214
  ports:
  - name: demo
    port: 80
    protocol: TCP
    targetPort: echo-port
  - name: demo2
    port: 81
    protocol: TCP
    targetPort: 8080
  selector:
    app: echoheaders
```

ingress/controllers/nginx/controller.go
```diff
@@ -217,6 +257,75 @@ func (lbc *loadBalancerController) getUDPConfigMap(ns, name string) (*api.Config
...
+	}
+
+	if !reflect.DeepEqual(svc.ObjectMeta.Annotations, namedPorts) {
+		data, _ := json.Marshal(namedPorts)
+		if svc.ObjectMeta.Annotations == nil {
+			svc.ObjectMeta.Annotations = map[string]string{}
+		}
+		svc.ObjectMeta.Annotations[namedPortAnnotation] = string(data)
+		glog.Infof("updating service %v with new named port mappings", svc.Name)
+		_, err := lbc.client.Services(svc.Namespace).Update(svc)
```
bprashanth: What happens if this update fails? (eg: resource version conflict). And if it passes, will we actuallly just come back in this function and relist pods again?

ingress/controllers/nginx/controller.go
```diff
@@ -217,6 +257,75 @@ func (lbc *loadBalancerController) getUDPConfigMap(ns, name string) (*api.Config
...
+		glog.Warningf("service %v not found", key)
+		return
+	}
+
+	svc := svcObj.(*api.Service)
+	if svc.Spec.Selector == nil {
+		return
+	}
+
+	pods, err := lbc.client.Pods(svc.Namespace).List(api.ListOptions{
```
bprashanth: I'd suggest you don't actually list all pods. This is a really heavy operation which if done too often can have all sorts of side effects (oom on this container in large clusters -- this happened with kube2sky in 1000 nodes, load on apiserver, request throttling etc). The only sane way to retrieve all pods from apiserver (with or without a label selector) is via watch.

I'm fine with a todo for now. 
I dislike the growing dependency graph. 

Can you do this less often? (currently it happens every svc enqueue right?)

bprashanth:
I'm ok with the pod list if we can restrict it to just when the user has the named targetport.Other stuff looks fine.

ingress/controllers/nginx/controller.go
aledbf: > I'd suggest you don't actually list all pods

this just get the pods associated with one service, not all the pods in the cluster

> currently it happens every svc enqueue right?

I changed that to Ingress add and update events.

ingress/controllers/nginx/controller.go
aledbf: requeue added in case of error during update

ingress/controllers/nginx/controller.go
bprashanth: It's still a list of all pods between etcd and the apiserver. Also you're only decreasing the chances of oom, because whose to say you're not running a single svc with 30 pods per node on a 1000 node cluster. I'd vote for:
1. List in a very rare case, if you can detect exactly when it's required
2. If you're listing on "every X update/add/delete", then use a watch

If you want to put that in a todo it's fine. Please move the listing out into a funtion called ListAllPods so we can easily swap it out when we refactor.

aledbf: @bprashanth how feasible is to propose the addition of the namedPort name in the endpoint?
That removes all this "unnecessary" code.

alanhartless:
It seems something with this has broken generating a correct nginx.conf for me. My nginx.conf ends up with something like:

```
    upstream default-web- {
        least_conn;
        server 127.0.0.1:8181;

    }

    upstream upstream-default-backend {
        least_conn;
        server 10.0.0.3:8080;

    }
```

It's like it's not finding the ports for the `web` service. 

```
apiVersion: v1
kind: Service
metadata:
  name: web
  labels:
    name: web
spec:
  ports:
    - port: 80
      targetPort: 8080
      name: http
  selector:
    name: web
```

```
kind: Ingress
metadata:
  name: foo-bar-net
spec:
  tls:
  - hosts:
    - foo.bar.net
    secretName: foo-bar-tls
  rules:
  - host: foo.bar.net
    http:
      paths:
      -
        path: /
        backend:
          serviceName: web
          servicePort: 80
```

I'm double checking to make sure my local patched ok and to see if I can figure something out in the mean time but wanted to mention it in case the cause was something obvious. I don't see any errors in the log.

alanhartless:
Just an update - if I change the `servicePort` to a named port, then my conf added the name correct:

```
    upstream default-web-http {
        least_conn;
        server 127.0.0.1:8181;

    }```

Maybe that part is just cosmetic. Still not finding my web services though so continuing to dig into that.

**Comment by nottix** [comment]
@bprashanth any plan to integrate this pull request?

**Comment by bprashanth** [comment]
Yeah I'll make another pass sometime today since it looks like Manual updated the commits after my last comment.

File: ingress/controllers/nginx/controller.go [hunk 3]
```diff
@@ -217,6 +248,126 @@ func (lbc *loadBalancerController) getUDPConfigMap(ns, name string) (*api.Config
...
+	}
+
+	namedPorts := map[string]string{}
+	// we need to check only one pod searching for named ports
+	pod := &pods.Items[0]
+	glog.V(4).Infof("checking pod %v/%v for named port information", pod.Namespace, pod.Name)
+	for i := range svc.Spec.Ports {
+		servicePort := &svc.Spec.Ports[i]
+
+		_, err := strconv.Atoi(servicePort.TargetPort.StrVal)
```
bprashanth: can you bail early (before pod list) if none of these ports is a string?

If we can't selectively list only when there's a named port, I'd suggest setting up a pod watch, eg: 
```go
	podStore.Store, podController = framework.NewInformer(
 		&cache.ListWatch{
 			ListFunc: func(options api.ListOptions) (runtime.Object, error) {
 				return e.client.Core().Pods(api.NamespaceAll).List(options)
 			},
 			WatchFunc: func(options api.ListOptions) (watch.Interface, error) {
 				return e.client.Core().Pods(api.NamespaceAll).Watch(options)
 			},
 		},
 		&api.Pod{},
 		resyncPeriod(),
 		framework.ResourceEventHandlerFuncs{
 			AddFunc:    addPod,
 			UpdateFunc: updatePod,
 			DeleteFunc: deletePod,
 		},
 	)
```
And then you have a local cache you can list for pods:
https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/replication/replication_controller.go#L542

ingress/controllers/nginx/README.md
```diff
@@ -193,19 +198,6 @@ Problems encountered during [1.2.0-alpha7 deployment](https://github.com/kuberne
 * make setup-files.sh file in hypercube does not provide 10.0.0.1 IP to make-ca-certs, resulting in CA certs that are issued to the external cluster IP address rather then 10.0.0.1 -> this results in nginx-third-party-lb appearing to get stuck at "Utils.go:177 - Waiting for default/default-http-backend" in the docker logs.  Kubernetes will eventually kill the container before nginx-third-party-lb times out with a message indicating that the CA certificate issuer is invalid (wrong ip), to verify this add zeros to the end of initialDelaySeconds and timeoutSeconds and reload the RC, and docker will log this error before kubernetes kills the container.
   * To fix the above, setup-files.sh must be patched before the cluster is inited (refer to https://github.com/kubernetes/kubernetes/pull/21504)
 
-### Custom errors
-
```
bprashanth: why are you removing this?

ingress/controllers/nginx/README.md
```diff
@@ -120,7 +119,13 @@ Please follow [test.sh](https://github.com/bprashanth/Ingress/blob/master/exampl
 
 Check the [example](examples/tls/README.md)
 
+### HTTP Strict Transport Security
 
+HTTP Strict Transport Security (HSTS) is an opt-in security enhancement specified through the use of a special response header. Once a supported browser receives this header that browser will prevent any communications from being sent over HTTP to the specified domain and will instead send all communications over HTTPS.
+
+By default the controller redirects (301) to HTTPS if there is a TLS Ingress rule. 
+
+To disable this behavior use `force-hsts=false` in the NGINX ConfigMap.
```
bprashanth: can we call this just hsts=true/false? since hsts is supposed to force https it feels redundant to say force hsts.

ingress/controllers/nginx/controller.go
```diff
@@ -217,6 +248,126 @@ func (lbc *loadBalancerController) getUDPConfigMap(ns, name string) (*api.Config
...
+	}
+}
+
+// checkSvcForUpdate verifies if one of the running pods for a service contains
+// named port. If the annotation in the service does not exists or is not equals
+// to the port mapping obtained from the pod the service must be updated to reflect
+// the current state
+func (lbc *loadBalancerController) checkSvcForUpdate(svc *api.Service) error {
+	// get the pods associated with the service
+	pods, err := lbc.client.Pods(svc.Namespace).List(api.ListOptions{
```
bprashanth: please add a todo to switch this with a watch.

ingress/controllers/nginx/controller.go
```diff
@@ -217,6 +248,126 @@ func (lbc *loadBalancerController) getUDPConfigMap(ns, name string) (*api.Config
...
+			return fmt.Errorf("error getting service %v/%v: %v", svc.Namespace, svc.Name, err)
+		}
+
+		if svc.ObjectMeta.Annotations == nil {
+			svc.ObjectMeta.Annotations = map[string]string{}
+		}
+
+		newSvc.ObjectMeta.Annotations[namedPortAnnotation] = string(data)
+		glog.Infof("updating service %v with new named port mappings", svc.Name)
+		_, err = lbc.client.Services(svc.Namespace).Update(svc)
```
bprashanth: So we're updating the Service and waiting for that update to come down the watch to update the nginx config? or are we using the cached value up the stack? I feel like the first is a really round about way, and the second is not reliable since an intermediate update can cause spurious reloads. Or have I misunderstood how this gets into the config?

ingress/controllers/nginx/controller.go
```diff
@@ -373,7 +524,12 @@ func (lbc *loadBalancerController) getServices(data map[string]string, proto api
 		var endps []nginx.UpstreamServer
 		targetPort, err := strconv.Atoi(svcPort[1])
 		if err != nil {
-			endps = lbc.getEndpoints(svc, intstr.FromString(svcPort[1]), proto)
+			for _, sp := range svc.Spec.Ports {
```
bprashanth: nit: can you cache svcPort[0,1] in appropriately named local vars instead of re-indexing?

ingress/controllers/nginx/controller.go
```diff
@@ -747,8 +921,11 @@ func (lbc *loadBalancerController) Run() {
 	go lbc.endpController.Run(lbc.stopCh)
 	go lbc.svcController.Run(lbc.stopCh)
 
+	time.Sleep(1 * time.Second)
```
bprashanth: why?

ingress/controllers/nginx/nginx/main.go
```diff
@@ -105,18 +105,18 @@ type nginxConfiguration struct {
...
 
-	// Enables or disables the header HTS in servers running SSL
-	UseHTS bool `structs:"use-hts,omitempty"`
+	// Enables or disables the header HSTS in servers running SSL
+	ForceHSTS bool `structs:"force-hsts,omitempty"`
 
-	// Enables or disables the use of HTS in all the subdomains of the servername
-	HTSIncludeSubdomains bool `structs:"hts-include-subdomains,omitempty"`
+	// Enables or disables the use of HSTS in all the subdomains of the servername
+	HSTSIncludeSubdomains bool `structs:"hsts-include-subdomains,omitempty"`
```
bprashanth: can you include a comment specifying the defaults of these fields?

ingress/controllers/nginx/nginx/ssl.go
```diff
@@ -53,8 +53,10 @@ func (nginx *Manager) CheckSSLCertificate(pemFileName string) ([]string, error)
 		return []string{}, err
 	}
 
-	var block *pem.Block
-	block, _ = pem.Decode(pemCerts)
+	block, _ := pem.Decode(pemCerts)
+	if block == nil {
+		return []string{}, fmt.Errorf("No valid PEM formatted block found")
```
bprashanth: can we surface this as an event if it isn't already?

ingress/controllers/nginx/README.md
aledbf: because this can be done with the default backend. I will add an example

ingress/controllers/nginx/README.md
aledbf: ok

ingress/controllers/nginx/controller.go
aledbf: > So we're updating the Service and waiting for that update to come down the watch to update the nginx config

that's correct

ingress/controllers/nginx/controller.go
aledbf: to avoid N "deferring sync till endpoints control" errors before the initial sync

ingress/controllers/nginx/controller.go
bprashanth: That's icky, can we make it more straightforward? 
1. Only trigger searching through pods if there is a named port
2. Return the matching endpoint instead of waiting for out of band watch to fire

ingress/controllers/nginx/controller.go
bprashanth: put that in the defferring if statement instead of here? i.e something like https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/replication/replication_controller.go#L514

ingress/controllers/nginx/nginx/ssl.go
aledbf: and use `glog.Error` in the caller?

bprashanth:
@aledbf i see you're still listing, should i wait for something along the lines of https://github.com/kubernetes/contrib/pull/766/files#r60281158 or is that a follow up?

aledbf:
>  or is that a follow up?

Yes, I prefer to do that.

aledbf:
@bprashanth ready for another review

ingress/controllers/nginx/controller.go
```diff
@@ -541,9 +717,38 @@ func (lbc *loadBalancerController) createUpstreams(data []interface{}) map[strin
 			}
 
 			for _, path := range rule.HTTP.Paths {
-				name := fmt.Sprintf("%v-%v-%v", ing.GetNamespace(), path.Backend.ServiceName, path.Backend.ServicePort.IntValue())
+				name := fmt.Sprintf("%v-%v-%v", ing.GetNamespace(), path.Backend.ServiceName, path.Backend.ServicePort.String())
 				if _, ok := upstreams[name]; !ok {
```
bprashanth: can we ditch the extra indent and continue when ok?

ingress/controllers/nginx/controller.go
bprashanth: Hmm, Ok so this is the part I'm not convinced about. Why not just do in getEndpoints? enqueue the parent ingress for any service when it changes, and invoke the same sync routine we do. In getEndpoints if the service port has a string, do the whole relist and lookup business. Am I missing some reason that won't work?

bprashanth:
Sorry for the delay, my only reservation at this point is that checkSvcForUpdate is icky, I'd vastly prefer if it didn't update services. If you can either do as suggested or explain why we can't just handle this through the normal ingress sync routine, I'll be ok with it.

aledbf:
@bprashanth testing doing the update in getEndpoints

aledbf:
ping @bprashanth please review

ingress/controllers/nginx/controller.go
```diff
@@ -41,14 +44,37 @@ import (
...
 )
 
+type namedPortMapping map[string]string
+
+func (npm namedPortMapping) getPort(name string) (string, bool) {
+	val, ok := npm.getMappings()[name]
+	return val, ok
+}
+
+func (npm namedPortMapping) getMappings() map[string]string {
```
bprashanth: please add a comment about what mappings, since the name is very general and it returns a map. Or rename the method.

bprashanth:
So much better, thanks!
LGTM but for the one naming nit

aledbf:
@bprashanth done

bprashanth:
LGTM

sekka1:
Just seen all of this new stuff today.  Was trying to upgrade our old ingress (container version 0.2).  When will a new container be build from this?  It seems `gcr.io/google_containers/nginx-ingress-controller:0.6` does not have the `hsts` fix in it.  I set the flag in the config map to `false` and it has no effect.

bprashanth:
@sekka1 please try with 0.61, it contains https://github.com/kubernetes/contrib/pull/849

sekka1:
Thanks @bprashanth it is working now.

bprashanth:
Thank @simonswine and @aledbf, I just pushed the image :)

PR #767, opened 2016-04-13
flannel/tasks/upstart-service.yml: fix indentation

jlebon:
Otherwise ansible crashes with the singular cryptic message:

> ERROR: that is not a legal parameter in an Ansible task or handler


ingvagabund:
You are faster :) Thanks for the pack

PR #768, opened 2016-04-13
fix indentation

rutsky:
Previously on some versions of Ansible this was reported as following error:

```
ERROR: that is not a legal parameter in an Ansible task or handler
```

(however my Ansible 2.0.0.2 was silent).

/cc @stephenrlouie


rutsky:
Oh, there is already fix for this: https://github.com/kubernetes/contrib/pull/767

PR #771, opened 2016-04-13
Test-enable weak stable check for kubemark-500

gmarek:
cc @wojtek-t @kubernetes/sig-testing

### DIFF
--- a/mungegithub/mungers/submit-queue_test.go
+++ b/mungegithub/mungers/submit-queue_test.go
@@ -713,86 +713,86 @@ func TestSubmitQueue(t *testing.T) {
 			reason:          noMerge,
 			state:           "pending",
 		},
-		// Should pass even though last 'weakStable' build failed, as it wasn't "strong" failure
-		// and because previous two builds succeeded.
-		{
-			name:            "Test20",
-			pr:              ValidPR(),
-			issue:           NoOKToMergeIssue(),
-			events:          NewLGTMEvents(),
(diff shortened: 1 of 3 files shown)


eparis:
Any interest or need to show a weak failure as a differently colored chip in the UI?  might not be useful information....

gmarek:
I'm not really into UI:). It may be useful, as green is not really green, so people might be surprised. Surprises are bad.

This is a test run for now - I want to see how it behaves. It we won't have any failures until tomorrow I'll cause some, to make sure it works properly.

gmarek:
Ping @eparis

mungegithub/mungers/submit-queue_test.go
```diff
@@ -713,86 +713,86 @@ func TestSubmitQueue(t *testing.T) {
...
-				"junit_03.xml": getJUnit(7, 0),
-			},
-			e2ePass:  true,
-			unitPass: true,
-			reason:   e2eFailure,
-			state:    "success",
-		},
+		// // Should pass even though last 'weakStable' build failed, as it wasn't "strong" failure
+		// // and because previous two builds succeeded.
+		// {
```
ixdy: why did this test case get commented out?

gmarek: Because the logic was temporary disabled - it's currently only logging. I'll reenable it when enabling the logic.

PR #779, opened 2016-04-13
Uninstall: Removed service configuration files, cleaned up flannel files

adamschaub:
Includes miscellaneous fixes for coreos uninstall.

/cc @danehans @eparis

### DIFF
--- a/ansible/playbooks/adhoc/uninstall.yml
+++ b/ansible/playbooks/adhoc/uninstall.yml
@@ -41,6 +41,14 @@
       changed_when: False
       failed_when: False
 
+    - name: Clear etcd configuration data
+      command: etcdctl rm --recursive {{ item }}
+      with_items:
+        - "/registry"
+        - "/cluster.local" #cluster name. default in group_vars/all.yml
+      ignore_errors: True
+      when: inventory_hostname == groups['etcd'][0]
+
     - name: Stop services
       service: name={{ item }} state=stopped
       with_items:
@@ -61,7 +69,8 @@
         - flannel
         - kubernetes-node
         - kubernetes-master
-        - etcd
+        - kubernetes-client
+        - "{{ etcd_svc_name }}"
         - kube-addons
         - atomic-enterprise
         - atomic-enterprise-master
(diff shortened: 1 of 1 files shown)


ansible/playbooks/adhoc/uninstall.yml
```diff
@@ -41,6 +41,14 @@
       changed_when: False
       failed_when: False
 
+    - name: Clear etcd configuration data
+      command: etcdctl rm --recursive {{ item }}
+      with_items:
+        - "/registry"
+        - "/cluster.local" #cluster name. default in group_vars/all.yml
```
eparis: so why not use the cluster name instead of hard coding it here?

danehans: @adamschaub I agree with @eparis 

It would be much appreciated if you could test and send a follow-on PR that incorporates: https://github.com/kubernetes/contrib/blob/master/ansible/group_vars/all.yml#L9-L12

ansible/playbooks/adhoc/uninstall.yml
```diff
@@ -113,26 +122,34 @@
         - "{{ bin_dir }}/kubelet"
         - "{{ bin_dir }}/kube-proxy"
         - "{{ bin_dir }}/kube-scheduler"
+        - "{{ bin_dir }}/flanneld"
+        - /run/flannel_docker_opts.env
+        - /run/flannel
+        - /tmp/flannel-conf.json
+        - /opt/flannel-0.5.5 #flannel version can change
```
danehans: Same with: https://github.com/kubernetes/contrib/blob/master/ansible/roles/flannel/tasks/github-release.yml#L4

ansible/playbooks/adhoc/uninstall.yml
```diff
@@ -113,26 +122,34 @@
...
         - /etc/systemd/system/kube-controller-manager.service
         - /etc/systemd/system/kube-proxy.service
+        - /etc/systemd/system/kube-proxy.service.wants
         - /etc/systemd/system/kube-apiserver.service
+        - /etc/systemd/system/kube-apiserver.service.wants
         - /etc/systemd/system/kubelet.service
         - /etc/systemd/system/kube-scheduler.service
+        - /etc/systemd/system/flanneld.service
+        - /etc/systemd/system/flanneld.service.wants
+        - /etc/systemd/system/multi-user.target.wants
```
danehans: Since ansible is not managing the following:

kube-proxy.service.wants
kube-apiserver.service.wants
flanneld.service.wants
multi-user.target.wants
delay-master-services.target
delay-node-services.target

Can you try removing and see if uninstall works as expected? I prefer not removing things we don't config.

PR #783, opened 2016-04-15
[ansible] Adds Support for Contiv

danehans:
Supercedes https://github.com/kubernetes/contrib/pull/625

This PR brings in ansible changes needed to setup Kubernetes
with Contiv networking as outlined in:

https://github.com/kubernetes/contrib/issues/577

Change Summary:

1) Added the contiv role that holds all contiv specific config code.
2) Expose minimal contiv configuration options in group_vars/all
3) Updates ansible/cluster.yml to expose contiv role.
4) Updates etcd to support legacy (:4001/:7001) ports and adds
   proxy support.
5) Added missing "validate_certs: False" for a couple of addon
   file fetches.

### DIFF
--- a/ansible/cluster.yml
+++ b/ansible/cluster.yml
@@ -44,6 +44,12 @@
   tags:
     - network-service-install
 
+# install contiv netmaster
+- hosts: masters
+  sudo: yes
+  roles:
+    - { role: contiv, contiv_role: netmaster, when: networking == 'contiv' }
+
 # install kube master services
 - hosts: masters
   sudo: yes
@@ -78,3 +84,9 @@
     - { role: opencontrail-provision, when: networking == 'opencontrail' }
   tags:
     - network-service-config
+
+# install contiv netplugin
+- hosts: nodes
+  sudo: yes
+  roles:
+    - { role: contiv, contiv_role: netplugin, when: networking == 'contiv' }
--- a/ansible/roles/etcd/defaults/main.yaml
+++ b/ansible/roles/etcd/defaults/main.yaml
@@ -4,6 +4,10 @@ etcd_source_type: package
 
 etcd_client_port: 2379
 etcd_peer_port: 2380
+# Remove etcd_client_legacy_port & etcd_peer_legacy_port
+# when https://github.com/contiv/netplugin/issues/323 is fixed
+etcd_client_legacy_port: 4001
+etcd_peer_legacy_port: 7001
 etcd_peers_group: etcd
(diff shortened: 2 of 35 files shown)


danehans:
@eparis @jojimt started this PR and I refactored it while he is on leave. I know it's big, but we would really appreciate your review.

/cc @stephenrlouie @adamschaub @rutsky @jainvipin

danehans:
@jojimt and I both have signed CLA's. I think it's because I cherry-picked his original patch. Do you have any recommendations on passing the CLA in this scenario?

danehans:
@eparis fyi... I fixed the CLA issue.

ansible/roles/etcd/defaults/main.yaml
```diff
@@ -4,6 +4,10 @@ etcd_source_type: package
 
 etcd_client_port: 2379
 etcd_peer_port: 2380
+# Remove etcd_client_legacy_port & etcd_peer_legacy_port
+# when https://github.com/contiv/netplugin/issues/323 is fixed
+etcd_client_legacy_port: 4001
```
eparis: why do I want this legacy stuff? can we please get rid of it?

danehans: Take a look at line 8. I know how you feel about this legacy stuff. When https://github.com/contiv/netplugin/pull/325 gets merged and a release cut, I will go back and remove the etcd_legacy. We're in a bit of a pinch to get this in, otherwise I'd hold off for 325 to get merged.

ansible/cluster.yml
```diff
@@ -44,6 +44,12 @@
   tags:
     - network-service-install
 
+# install contiv netmaster
```
eparis: Is there any way to move flannel, opencontrail and contiv into their own playbook(s)/file(s)? declutter the top level?

danehans: I believe we can remove the redundant hosts entries for each networking provider in cluster.yml, and use ```when: networking == flannel | contrail | contiv``` under the appropriate hosts entries to select the proper networking role. I was just trying to follow existing patterns with the contiv PR. Could this PR be accepted as-is? I can create an issue such as "improve support for multiple networking providers" and submit a follow-on PR that addresses the issue.

eparis: As long as you agree to come back and clean up. Yes. I will merge.

danehans: Count on it.

PR #787, opened 2016-04-17
SSL termination URL rewritting changes

jdavis7257:
These changes will allow ssl termination to work without an AclMatch and it will act just like the http services. If the user wants they can specify an AclMatch which will make the load balancer act the same as it does currently.

### DIFF
--- a/service-loadbalancer/README.md
+++ b/service-loadbalancer/README.md
@@ -116,8 +116,7 @@ A couple of points to note:
 - You need to take care of ensuring there is no collision between these service ports on the node.
 
 #### SSL Termination
-
-- Annotate a service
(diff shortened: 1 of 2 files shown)


service-loadbalancer/README.md
```diff
@@ -124,7 +124,6 @@ metadata:
   name: myservice
   annotations:
     serviceloadbalancer/lb.sslTerm: "true"
-    serviceloadbalancer/lb.aclMatch: "-i /t1 /t2"
```
bprashanth: can i trouble you to give a better example (I know the readme didn't have a good example to begin with, so feel free to defer)? Perhaps just add a line saying: here's how you get the /foo, and here's how you keep it as /.

bprashanth:
LGTM but for the nit
@aledbf fyi

bprashanth:
Can you also please squash your first and third commit?

bprashanth:
@jdavis7257 great thanks! LGTM

PR #791, opened 2016-04-18
Create firewall rule, update certs when secrets are updated

bprashanth:
Fixes #1 on https://github.com/kubernetes/contrib/blob/master/ingress/controllers/gce/BETA_LIMITATIONS.md.

Also allows certificate update when the Kubernetes secret changes, and contains changes in preperation for running the controller on the master.

Assigning to Minhan, @kubernetes/goog-cluster fyi


bprashanth:
@freehan tests passed

ingress/controllers/gce/firewalls/firewalls.go
```diff
@@ -0,0 +1,78 @@
...
+
+import (
+	"github.com/golang/glog"
+
+	compute "google.golang.org/api/compute/v1"
+	"k8s.io/contrib/ingress/controllers/gce/utils"
+	netset "k8s.io/kubernetes/pkg/util/net/sets"
+)
+
+// Src range from which the GCE L7 performs health checks.
```
freehan: nit: add link to doc here `https://cloud.google.com/compute/docs/load-balancing/health-checks#health_check_source_ips_and_firewall_rules`

I have no idea where is it come from in my first glance.

ingress/controllers/gce/firewalls/firewalls.go
```diff
@@ -0,0 +1,78 @@
...
+const l7SrcRange = "130.211.0.0/22"
+
+// FirewallRules manages firewall rules.
+type FirewallRules struct {
```
freehan: This is actually only one rule right?

ingress/controllers/gce/firewalls/interfaces.go
```diff
@@ -0,0 +1,39 @@
...
+type SingleFirewallPool interface {
+	// TODO: Take a list of node ports for the firewall.
+	Sync(nodePorts []int64, nodeNames []string) error
+	Shutdown() error
+}
+
+// Firewall interfaces with the GCE firewall api.
+// This interface is a little different from the rest because it dovetails into
+// the same firewall methods used by the TCPLoadBalancer.
+type Firewall interface {
```
freehan: This is the cloud client right?

ingress/controllers/gce/firewalls/firewalls.go
```diff
@@ -0,0 +1,78 @@
...
+	cloud    Firewall
+	namer    utils.Namer
+	srcRange netset.IPNet
+}
+
+// NewFirewallPool creates a new firewall rule manager.
+// cloud: the cloud object implementing Firewall.
+// namer: cluster namer.
+func NewFirewallPool(cloud Firewall, namer utils.Namer) SingleFirewallPool {
```
freehan: NewFirewallRule?

ingress/controllers/gce/main.go
```diff
@@ -83,6 +75,13 @@ var (
 		`Optional, if this controller is running in a kubernetes cluster, use the
 		 pod secrets for creating a Kubernetes client.`)
 
+	// TODO: Consolidate this flag and running-in-cluster. People already use
+	// the first one to mean "running in dev", unfortunately.
+	useRealCloud = flags.Bool("use-real-cloud", false,
```
freehan: I do not get the justification for adding this flag. This is for dev purpose right?  e.g. not running in cluster but use real cloud?

ingress/controllers/gce/firewalls/firewalls.go
bprashanth: It doesn't need to be, that's just a detail. We could well create a rule per nodeport through the Sync function since it has all the information required. It's a "pool" in that it manages a group of cloud resources in a generic way (syncs periodically, shutdown etc). A pool can contain 1 item, but the other way around doesn't hold. In an ideal world all of : backends, instance groups, firewalls etc would satisfy a pool interface, but today that's not really possible unless I use `interface{}` everywhere which is gross, so I decided to just convey the intent through naming and similar methods.

ingress/controllers/gce/main.go
bprashanth:
```
--running-in-cluster=false --use-real-cloud=false(default): create localhost apiserver client + fake cloud client
--running-in-cluster=false --use-real-cloud=true: create localhost apiserver client + real cloud client
```
the first is for dev, the second is for running this pod on the master (so it needs to use localhost apiserver client but create resources in gce)

bprashanth: Meaning "master" is not "in-cluster", because master doesn't run kube-proxy and has a kubectl config and all master pods run with host networking, so in this aspect it's more like a desktop machine than a node.

freehan:
LGTM. Sorry for keep you waiting.

bprashanth:
Thanks!

aknuds1:
Eagerly waiting on this to make its way into GKE. AFAICT TLS secret changes aren't currently applied.

zhaytee:
Did this ever make it into GKE? It's a critical requirement for any production service using Ingress objects on GKE and LetsEncrypt TLS certs (which need to be rotated out every 90 days).

gkop:
I just tried a simple apply on an edited secret, and yes it worked. The web console and `gcloud compute ssl-certificates list` reported the change right away, and the load balancer started serving it up in about 10 minutes. It would be nice to have this officially documented!

gkop:
I just tried a simple apply on an edited secret, and yes it worked. The web console and `gcloud compute ssl-certificates list` reported the change right away, and the load balancer started serving it up in about 10 minutes. It would be nice to have this officially documented! _Especially_ because there are other corners of k8s where changes to secrets aren't automatically picked up, like deployments, so we don't take it for granted.

evanj:
I can verify the same thing, although in my case gcloud compute ssl-certificates list did not reflect the change immediately. It appeared that nothing happened, but after about ~10 minutes or so the change was applied in all the UIs and on the load balancer itself. Seamless!

PR #796, opened 2016-04-18
Library to operate on MIGs in GCE for Cluster Autoscaler

piosz:
ref https://github.com/kubernetes/kubernetes/issues/24404


cluster-autoscaler/config/migconfig.go
```diff
@@ -18,16 +18,37 @@ package config
...
+		Name: splitted[2],
+	}, nil
+}
+
 // MigConfig contains managed instance group configuration details.
 type MigConfig struct {
 	MinSize int
 	MaxSize int
-	Url     url.URL
+	Zone string
```
mwielgus: Please add Project.

mwielgus:
LGTM

PR #803, opened 2016-04-19
Move task to copy kube-gen-token script

Ladicle:
Task to copy kube-gen-token script is in secrets.yml.
However, secrets.yml is called only the first node of master.

The `kube-gen-token.sh` is also used in main.yml of kubernetes-addons.
Fail to deploy HA Cluster, in the second or later master node that does not have a script.

### DIFF
--- a/ansible/roles/kubernetes/tasks/gen_tokens.yml
+++ b/ansible/roles/kubernetes/tasks/gen_tokens.yml
@@ -1,10 +1,4 @@
 ---
-- name: Copy the token gen script
-  copy:
-    src=kube-gen-token.sh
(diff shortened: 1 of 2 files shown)


eparis:
we shouldn't be running kubernets-addons.sh on anything other than the first master...  they would fight each other and bring the system down...

Ladicle:
Thanks comments. I'm misled into thinking action of kubernets addons.

In order to apply only to first master, is better that was changed to `master[0]` from `master` the following line?
https://github.com/kubernetes/contrib/blob/master/ansible/cluster.yml#L56

PR #815, opened 2016-04-20
Perfdash: Add version support in perfdash

Random-Liu:
For https://github.com/kubernetes/kubernetes/pull/24542

This PR makes the perfdash aware of the metrics version.
Now, perfdash will only show the result with the newest version.

@gmarek Could you help me take a look? The change is minor. :)

/cc @yujuhong @vishh

### DIFF
--- a/perfdash/parser.go
+++ b/perfdash/parser.go
@@ -49,24 +49,29 @@ func parseTestOutput(scanner *bufio.Scanner, job string, tests Tests, buildNumbe
 				continue
 			}
 			state = inTest
-			if _, found := result[testName]; !found {
-				result[testName] = BuildData{Job: job, Builds: map[string][]perftype.DataItem{}}
-			} else if result[testName].Job != job {
+			if _, found := result[testName]; found && result[testName].Job != job {
 				// If the job is different, we'll skip the test and keep the old result
 				state = scanning
 			}
 		}
 		if state == processing {
-			// TODO(random-liu): This is brittle, we should emit a tail delimiter too
-			if strings.Contains(line, "INFO") || strings.Contains(line, "STEP") || strings.Contains(line, "Failure") || strings.Contains(line, "[AfterEach]") {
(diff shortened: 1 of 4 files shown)


perfdash/parser.go
```diff
@@ -49,24 +49,28 @@ func parseTestOutput(scanner *bufio.Scanner, job string, tests Tests, buildNumbe
...
-			} else if result[testName].Job != job {
+			if _, found := result[testName]; found && result[testName].Job != job {
 				// If the job is different, we'll skip the test and keep the old result
 				state = scanning
 			}
 		}
 		if state == processing {
-			// TODO(random-liu): This is brittle, we should emit a tail delimiter too
-			if strings.Contains(line, "INFO") || strings.Contains(line, "STEP") || strings.Contains(line, "Failure") || strings.Contains(line, "[AfterEach]") {
+			// The perf data is end with an empty line.
```
gmarek: I'm a bit scared of this change: empty lines happen. Can you add some end of data marker for printing perf data?

gmarek:
I'm worried about the empty line as an ending marker. LGTM otherwise.

perfdash/parser.go
Random-Liu: @gmarek There won't be empty line in json format. And even if there is some empty line accidentally, the loop will keep trying. :)

gmarek: What I did previously was adding a FINISHED (or something like that) at the end of the data. We can write printer that adds it at the end of JSON.

Random-Liu: @gmarek All right. :)

Random-Liu: @gmarek Done~

gmarek:
LTGM. Add label after kubernetes/kubernetes#24542 is merged

Random-Liu:
kubernetes/kubernetes#24542 has been merged, self apply LGTM according to https://github.com/kubernetes/contrib/pull/815#issuecomment-212916207

PR #816, opened 2016-04-20
Remove single inner quotes for insecure registry option

williamcaban:
Remove single inner quotes which prevented docker from starting in CentOS when using _insecure registry_.

When the following entry is generated in CentOS (/etc/sysconfig/docker), the Docker service won't start and complain about misconfiguration. When the single inner quotes are removed everything works fine.

Docker do not start --> INSECURE_REGISTRY="'--insecure-registry=gcr.io '"

Docker start Okay --> INSECURE_REGISTRY="--insecure-registry=gcr.io "

### DIFF
--- a/ansible/roles/docker/tasks/main.yml
+++ b/ansible/roles/docker/tasks/main.yml
@@ -60,7 +60,7 @@
     - restart docker
 
 - name: Add any insecure registrys to docker config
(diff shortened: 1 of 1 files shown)


williamcaban:
I signed it!

PR #820, opened 2016-04-21
Cluster-autoscaler: Utils + more scale up flow

mwielgus:
kubernetes/kubernetes#24404

cc: @piosz @fgrzadkowski

### DIFF
--- a/cluster-autoscaler/cluster_autoscaler.go
+++ b/cluster-autoscaler/cluster_autoscaler.go
@@ -22,11 +22,7 @@ import (
 	"time"
 
 	"k8s.io/contrib/cluster-autoscaler/config"
-	kube_api "k8s.io/kubernetes/pkg/api"
-	"k8s.io/kubernetes/pkg/client/cache"
 	kube_client "k8s.io/kubernetes/pkg/client/unversioned"
-	"k8s.io/kubernetes/pkg/fields"
-	"k8s.io/kubernetes/pkg/labels"
 
 	"github.com/golang/glog"
 )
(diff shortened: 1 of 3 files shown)


cluster-autoscaler/cluster_autoscaler.go
```diff
@@ -53,27 +49,55 @@ func main() {
...
+
+				// TODO: Checking if all nodes are present.
+
+				// Checks if scheduler tried to schedule the pods after thew newest node was added.
+				newestNode := GetNewestNode(nodes.Items)
+				if newestNode == nil {
+					glog.Errorf("No newest node")
+					continue
+				}
+				oldesSchedulingTrial := GetOldestFailedSchedulingTrail(pods)
```
fgrzadkowski: oldes**t**SchedulingTrial

cluster-autoscaler/cluster_autoscaler.go
```diff
@@ -53,27 +49,55 @@ func main() {
...
+				if oldesSchedulingTrial == nil {
+					glog.Errorf("No oldest unschedueled trial: %v", err)
+					continue
+				}
+
+				if newestNode.CreationTimestamp.After(oldesSchedulingTrial.Add(-1 * time.Minute)) {
```
fgrzadkowski: Let's add a TODO that we should find a better way to verify that scheduler took into account newest node.

cluster-autoscaler/cluster_autoscaler.go
```diff
@@ -53,27 +49,55 @@ func main() {
...
+					// Lets give scheduler another chance.
+					glog.V(1).Infof("One of the pods have not been tried after adding %s", newestNode.Name)
+					continue
 				}
```
fgrzadkowski: Maybe an info log at the end that we'd try to increase the size of the cluster?

cluster-autoscaler/utils.go
```diff
@@ -0,0 +1,82 @@
...
+// List returns all unscheduled pods.
+func (unscheduledPodLister *UnscheduledPodLister) List() ([]*kube_api.Pod, error) {
+	//TODO: Extra filter based on pod condition.
+	return unscheduledPodLister.podLister.List(labels.Everything())
+}
+
+// NewUnscheduledPodLister returns a lister providing pods that failed to be scheduled.
+func NewUnscheduledPodLister(kubeClient *kube_client.Client) *UnscheduledPodLister {
+	// watch unscheduled pods
+	selector := fields.ParseSelectorOrDie("spec.nodeName==" + "" + ",status.phase!=" +
```
fgrzadkowski: I think it's enough to list pods in phase `Pending`

cluster-autoscaler/utils.go
```diff
@@ -0,0 +1,82 @@
...
+	podReflector.Run()
+
+	return &UnscheduledPodLister{
+		podLister: podLister,
+	}
+}
+
+// NewNodeLister builds a node lister.
+func NewNodeLister(kubeClient *kube_client.Client) *cache.StoreToNodeLister {
+	listWatcher := cache.NewListWatchFromClient(kubeClient, "nodes", kube_api.NamespaceAll, fields.Everything())
```
fgrzadkowski: I think should take into account only nodes in state `Ready`. Unhealthy nodes are not interesting to us.

fgrzadkowski:
Mostly minor things.

cluster-autoscaler/cluster_autoscaler.go

cluster-autoscaler/cluster_autoscaler.go

cluster-autoscaler/cluster_autoscaler.go
mwielgus: We will increase the size of the cluster once @piosz finishes his PR. Then the log will be added.

cluster-autoscaler/utils.go
mwielgus: This is how k8s scheduler does it.

cluster-autoscaler/cluster_autoscaler.go
fgrzadkowski: sgtm

cluster-autoscaler/utils.go
mwielgus: Done.

cluster-autoscaler/utils.go
fgrzadkowski: As discussed offline I believe we don't need to have the same options here, but I won't insist.

cluster-autoscaler/utils.go
```diff
@@ -0,0 +1,107 @@
...
+		podLister: podLister,
+	}
+}
+
+// ReadyNodeLister lists ready nodes.
+type ReadyNodeLister struct {
+	nodeLister *cache.StoreToNodeLister
+}
+
+// List retuns ready nodes.
```
fgrzadkowski: retu**r**ns

fgrzadkowski:
One misspelling. Otherwise LGTM.

PR #822, opened 2016-04-21
Cluster-autoscaler: New node count estimator

mwielgus:
kubernetes/kubernetes#24404

cc: @piosz @fgrzadkowski

### DIFF
--- a/cluster-autoscaler/estimator/estimator.go
+++ b/cluster-autoscaler/estimator/estimator.go
@@ -1,0 +1,100 @@
+/*
+Copyright 2016 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package estimator
+
+import (
+	"math"
+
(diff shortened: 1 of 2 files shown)


cluster-autoscaler/estimator/estimator.go
```diff
@@ -0,0 +1,98 @@
...
+package estimator
+
+import (
+	"math"
+
+	kube_api "k8s.io/kubernetes/pkg/api"
+	"k8s.io/kubernetes/pkg/api/resource"
+)
+
+// BasicNodeEstimator estimates the number of needed nodes to handle the given amount of pods.
```
fgrzadkowski: Should we mention that it will never overestimate but may underestimate?

cluster-autoscaler/estimator/estimator_test.go
```diff
@@ -0,0 +1,105 @@
...
+func TestEstimate(t *testing.T) {
+
+	pod := &kube_api.Pod{
+		Spec: kube_api.PodSpec{
+			Containers: []kube_api.Container{
+				{
+					Resources: kube_api.ResourceRequirements{
+						Requests: kube_api.ResourceList{
+							kube_api.ResourceCPU:    *resource.NewMilliQuantity(500, resource.DecimalSI),
+							kube_api.ResourceMemory: *resource.NewQuantity(1000*1024*1024, resource.DecimalSI),
```
fgrzadkowski: How about we add a const for mem and cpu and use it here and below in assertions?

fgrzadkowski:
Small comments. In general LGTM.

cluster-autoscaler/estimator/estimator.go
mwielgus: Done.

cluster-autoscaler/estimator/estimator_test.go
mwielgus: Done.

PR #824, opened 2016-04-21
[mungegithub] Only request 40px avatars.

rmmh:


eparis:
would we rather do it on the server side, since it would be smaller easier?  maybe not...

eparis:
by updating https://github.com/kubernetes/contrib/blob/master/mungegithub/mungers/submit-queue.go#L467 I think?

ixdy:
I have no strong opinion. The image size is specified in CSS, which is pretty far away from the place we reference the URL already.

rmmh:
Since the width:40px is in the CSS, I thought doing this in the JS keeps the modifications for display as localized to the presentation layer as possible.

Issue #826, opened 2016-04-21
Missing proxy environment settings on some url tasks


jojimt:
Some of the url tasks in ansible/roles/kubernetes-addons/tasks/main.yml are missing the env setting for proxy. As a result, these tasks fail when executed behind a proxy.

PR #827, opened 2016-04-21
Fix https://github.com/kubernetes/contrib/issues/826

jojimt:
Added the missing environment setting and validate_certs, consistent with other addon tasks.


jojimt:
cc @danehans

PR #828, opened 2016-04-22
Cluster-autoscaler: Implemented waiting for operation in GCE Manager

piosz:
ref https://github.com/kubernetes/kubernetes/issues/24404


mwielgus:
LGTM

PR #830, opened 2016-04-22
Cluster-autoscaler: check node/mig state

mwielgus:
cc: @piosz @fgrzadkowski

### DIFF
--- a/cluster-autoscaler/utils.go
+++ b/cluster-autoscaler/utils.go
@@ -17,7 +17,11 @@ limitations under the License.
 package main
 
 import (
+	"fmt"
 	"time"
+
+	"k8s.io/contrib/cluster-autoscaler/config"
+	"k8s.io/contrib/cluster-autoscaler/utils/gce"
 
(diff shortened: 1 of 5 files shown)


cluster-autoscaler/utils.go
```diff
@@ -105,3 +109,35 @@ func GetOldestFailedSchedulingTrail(pods []*kube_api.Pod) *time.Time {
...
+		if err != nil {
+			return err
+		}
+
+		migConfig, err := gceManager.GetMigForInstance(instanceConfig)
+		if err != nil {
+			return err
+		}
+		url := migConfig.Url()
+		count, _ := migCount[url]
```
piosz: nit: you don't need the second variable here

piosz:
LGTM, just a nit

piosz:
:facepunch: :facepunch: :facepunch: 
```
Verifying hack/../hack/verify-golint.sh
FAILED
```

cluster-autoscaler/utils.go
mwielgus: Done.

Issue #819, opened 2016-04-20
The submit queue page takes forever to load


spxtr:
Looking at the timeline, we've loaded all of the JSON within about a second, but we spend the next 5-7 seconds laying out the tabs/lists and then loading a billion github avatars.

http://submit-queue.k8s.io/

spxtr:
It looks like we spend 3s handling`/prs`, 1.5s handling `/history`, and 1s handling `/usrs`, and we're doing a whole lot of DOMSubtreeModified events.

ixdy:
We can probably save some time/bandwidth by adding a size parameter to the avatar URLs, e.g. `&s=40`.

eparis:
Things might be made a lot faster if we use https://material.angularjs.org/1.0.7/demo/virtualRepeat on the PRs (and maybe history/queue) pages. I'm not sure how that plays with the search boxes and things...

spxtr:
It's much better for now thanks to #831.

PR #831, opened 2016-04-22
Optimize submit-queue frontend

rmmh:
The Queue tab is fully painted in .7s vs 5s before.

- Default to showing the Queue tab, since that's the most relevant one.
- Load tab data lazily, to avoid the delay of rendering everything first.
- Update AngularJS dependency from 1.5.3 to 1.5.5 (seems harmless).
- Vendor a patched version of angular-material.js with faster rendering.
  angular/material#5681


spxtr:
#819

PR #839, opened 2016-04-25
Cluster-autoscaler: scale up cluster if needed.

mwielgus:
kubernetes/kubernetes#24404

cc: @piosz @fgrzadkowski

### DIFF
--- a/cluster-autoscaler/cluster_autoscaler.go
+++ b/cluster-autoscaler/cluster_autoscaler.go
@@ -22,6 +22,7 @@ import (
 	"time"
 
 	"k8s.io/contrib/cluster-autoscaler/config"
+	"k8s.io/contrib/cluster-autoscaler/estimator"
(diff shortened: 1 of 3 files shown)


cluster-autoscaler/cluster_autoscaler.go
```diff
@@ -33,6 +34,12 @@ var (
 	kubernetes    = flag.String("kubernetes", "", "Kuberentes master location. Leave blank for default")
 )
 
+// ExpansionOption describes an option to expand the cluster.
+type ExpansionOption struct {
```
piosz: optional: better naming of this struct

piosz:
LGTM

cluster-autoscaler/cluster_autoscaler.go
mwielgus: Ack

PR #761, opened 2016-04-12
Implement kube-master HA for multiple masters

gitschaub:
We need to cover HA for kube-master components. Kubernetes HA cluster guide [1]. #725 

User specifies multiple nodes under the master role [2]. Ansible should seamlessly create a cluster with a collection of masters, with api services appropriately load balanced, and other kube-master components leader-elected.
- [x] Enable static etcd clustering (working with #711)
- [x] Deploy kube-apiserver on multiple nodes
- [x] Load balance kube-apiserver instances (e.g. add haproxy/keepalived roles)
- [x] Create controller-manager & scheduler services for each apiserver (leader-elected)

(optional, good to have)
- [x] add in support to run master components in containers (related, #673)

[1] http://kubernetes.io/docs/admin/high-availability/
[2] https://github.com/kubernetes/contrib/blob/master/ansible/inventory.example.ha

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/761)

PR #843, opened 2016-04-25
Loadbalancer role (keepalived/haproxy)

gitschaub:
Added load-balancer role for HA master configuration of cluster. 

When there is more than one master node, an haproxy and leader-elected keepalived pod (via unicast) is run on each master node. A VIP is specified in group_vars for the apiserver. Haproxy uses round-robin load balancing with SSL pass-through.

This PR depends on #761. I'll make changes here as that evolves.

/cc: @danehans @rutsky @eparis

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/843)

### DIFF
--- a/ansible/group_vars/all.yml
+++ b/ansible/group_vars/all.yml
@@ -40,6 +40,10 @@ insecure_registrys:
 # The port that the Kubernetes apiserver component listens on.
 kube_master_api_port: 443
 
+# Routeable VIP for load balancing api servers.
+# This VIP must be unassigned and routeable from your cluster nodes
+#kube_apiserver_vip: 10.10.10.10
+
 # Kubernetes internal network for services.
 # Kubernetes services will get fake IP addresses from this range.
 # This range must not conflict with anything in your infrastructure. These
--- a/ansible/roles/kubernetes/defaults/main.yml
+++ b/ansible/roles/kubernetes/defaults/main.yml
@@ -37,7 +37,7 @@ kube_current_release_directory: "{{ kube_releases_directory }}/{{ kube_version }
 kube_cert_dir: "{{ kube_config_dir }}/certs"
 
 # The IP(s) for which the certificate will be valid
-kube_cert_ip: "{{ hostvars[inventory_hostname]['ansible_all_ipv4_addresses'] | join (',IP:') }}"
+kube_cert_ip: "{{ (hostvars[inventory_hostname]['ansible_all_ipv4_addresses'] | join (',IP:'))}}{% if kube_apiserver_vip %},IP:{{ kube_apiserver_vip }}{% endif %}"
 
 # This is where all of the bearer tokens will be stored
 kube_token_dir: "{{ kube_config_dir }}/tokens"
--- a/ansible/roles/load-balancer/defaults/main.yml
+++ b/ansible/roles/load-balancer/defaults/main.yml
@@ -1,0 +1,7 @@
+keepalived_config_dir: /etc/keepalived
+kube_keepalived_password: kube-pass
+
+haproxy_config_dir: /etc/haproxy
+haproxy_stats_port: 9999
+
+kube_apiserver_interface: "{{ kube_apiserver_interface|default(ansible_default_ipv4.interface) }}"
--- a/ansible/roles/load-balancer/templates/haproxy.cfg.j2
+++ b/ansible/roles/load-balancer/templates/haproxy.cfg.j2
@@ -1,0 +1,40 @@
+#
+# {{ ansible_managed }}
+#
+global
+    log /dev/log    local0
+    log /dev/log    local1 notice
+    chroot /var/lib/haproxy
+    user haproxy
+    group haproxy
+    daemon
+    stats socket /var/lib/haproxy/stats level admin
+
+defaults
+    log global
+    option  dontlognull
+    timeout connect 5000
+    timeout client 50000
+    timeout server 50000
+
+frontend apiserver-incoming
+    bind {{ kube_apiserver_vip }}:{{ kube_master_api_port }}
+    mode tcp
+    option tcplog
+    default_backend apiserver
+
+backend apiserver
+    mode tcp
+    balance roundrobin
+    option ssl-hello-chk
+{% for host in groups['masters'] %}
+    server {{ hostvars[host].ansible_hostname }} {{ hostvars[host]['ansible_' + kube_apiserver_interface].ipv4.address }}:{{ kube_master_api_port }} check
+{% endfor %}
+
+listen stats
+    bind *:{{ haproxy_stats_port }}
+    mode http
+    stats enable
+    stats hide-version
+    stats realm Haproxy\ Statistics
+    stats uri /
--- a/ansible/roles/load-balancer/templates/keepalived.conf.j2
+++ b/ansible/roles/load-balancer/templates/keepalived.conf.j2
@@ -1,0 +1,33 @@
+#check that haproxy is runnning on this machine
+vrrp_script chk_haproxy {
+  script "pidof haproxy"
+  interval 2
+  weight 2
+}
+
+vrrp_instance VI_1 {
+  interface {{ kube_apiserver_interface }}
+  state MASTER
+  virtual_router_id 51
+  priority {% if inventory_hostname == groups['masters'][0] %} 101 {% else %} 100 {% endif %} # 101 on master, 100 on backups
+
+  #use unicast so we don't have to tool with multicast
+  unicast_peer {
+    {% for master in groups['masters'] %}
+    {{ hostvars[master]['ansible_' + kube_apiserver_interface].ipv4.address }}
+    {% endfor %}
+  }
+
+  virtual_ipaddress {
+    {{ kube_apiserver_vip }}
+  }
+
+  track_script {
+    chk_haproxy
+  }
+
+  authentication {
+    auth_type PASS
+    auth_pass {{ kube_keepalived_password }}
+  }
+}
--- a/ansible/roles/master/templates/controller-manager.kubeconfig.j2
+++ b/ansible/roles/master/templates/controller-manager.kubeconfig.j2
@@ -5,7 +5,7 @@ preferences: {}
 clusters:
 - cluster:
     certificate-authority: {{ kube_cert_dir }}/ca.crt
-    server: https://{{ groups['masters'][0] }}:{{ kube_master_api_port }}
+    server: https://{% if groups['masters']|length > 1 %}{{ kube_apiserver_vip }}{% else %}{{ groups['masters'][0] }}{% endif %}:{{ kube_master_api_port }}
   name: {{ cluster_name }}
 contexts:
 - context:
--- a/ansible/roles/master/templates/kubectl.kubeconfig.j2
+++ b/ansible/roles/master/templates/kubectl.kubeconfig.j2
@@ -5,7 +5,7 @@ preferences: {}
 clusters:
 - cluster:
     certificate-authority-data: {{ kube_ca_cert|b64encode }}
-    server: https://{{ groups['masters'][0] }}:{{ kube_master_api_port }}
+    server: https://{% if groups['masters']|length > 1 %}{{ kube_apiserver_vip }}{% else %}{{ groups['masters'][0] }}{% endif %}:{{ kube_master_api_port }}
   name: {{ cluster_name }}
 contexts:
 - context:
--- a/ansible/roles/contiv/templates/contiv.cfg.j2
+++ b/ansible/roles/contiv/templates/contiv.cfg.j2
@@ -1,5 +1,5 @@
 {
-  "K8S_API_SERVER": "https://{{ groups['masters'][0] }}:{{ kube_master_api_port }}",
+  "K8S_API_SERVER": "https://{% if groups['masters']|length > 1 %}{{ kube_apiserver_vip }}{% else %}{{ groups['masters'][0] }}{% endif %}:{{ kube_master_api_port }}",
   "K8S_CA": "{{ kube_cert_dir }}/ca.crt",
   "K8S_KEY": "{{ kube_cert_dir }}/kubecfg.key",
   "K8S_CERT": "{{ kube_cert_dir }}/kubecfg.crt"
--- a/ansible/roles/kubernetes/templates/config.j2
+++ b/ansible/roles/kubernetes/templates/config.j2
@@ -20,4 +20,4 @@ KUBE_LOG_LEVEL="--v=0"
 KUBE_ALLOW_PRIV="--allow-privileged=true"
 
 # How the replication controller, scheduler, and proxy
-KUBE_MASTER="--master=https://{{ groups['masters'][0] }}:{{ kube_master_api_port }}"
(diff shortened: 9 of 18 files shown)


ansible/group_vars/all.yml
```diff
@@ -40,6 +40,18 @@ insecure_registrys:
 # The port that the Kubernetes apiserver component listens on.
 kube_master_api_port: 443
 
+# Required for HA master configuration
+# Kubernetes master apiserver interface
+#kube_apiserver_interface: eth1
```
danehans: Is it possible to follow the etcd design pattern:

```
ansible//roles/etcd/defaults/main.yaml:
etcd_interface: "{{ ansible_default_ipv4.interface }}"

ansible//roles/etcd/defaults/main.yaml:etcd_machine_address:
"{{ hostvars[inventory_hostname]['ansible_' + etcd_interface].ipv4.address }}"

ansible//roles/etcd/templates/etcd.conf.j2: 
{{ hostvars[host]['ansible_hostname'] }}={{ etcd_peer_url_scheme }}://{{ hostvars[host]['ansible_' + etcd_interface].ipv4.address }}:{{ etcd_peer_port }}

ansible//vagrant/Vagrantfile:
etcd_interface: "eth1",
```

adamschaub: Yeah, that makes better sense. I'll move it to the appropriate role's defaults.

ansible/group_vars/all.yml
```diff
@@ -40,6 +40,18 @@ insecure_registrys:
...
+
+# Required for HA master configuration
+# A floating ip address for multiple kube-apiservers
```
danehans: I suggest changing floating to a virtual. Make note that this address should not be assigned to any hosts. The address should come from the same network used to assign real IP addresses to Kubernetes Masters.

ansible/group_vars/all.yml
```diff
@@ -40,6 +40,18 @@ insecure_registrys:
...
+# Set to an address that is routeable from all nodes in the cluster
+#kube_apiserver_vip: 172.32.128.100
+
+# The hostname or IP address of the primary apiserver, or VIP for in an HA configuration.
+kube_apiserver_endpoint: "{% if groups['masters']|length > 1 %}{{ kube_apiserver_vip }}{% else %}{{ groups['masters'][0] }}{% endif %}"
```
danehans: I don't believe this param should be exposed to a user. Can this be implemented in the defaults/main of the masters role?

ansible/roles/kubernetes/defaults/main.yml
```diff
@@ -37,14 +37,15 @@ kube_current_release_directory: "{{ kube_releases_directory }}/{{ kube_version }
...
-kube_cert_ip: "{{ hostvars[inventory_hostname]['ansible_all_ipv4_addresses'] | join (',IP:') }}"
+kube_cert_ip: "{{ (hostvars[inventory_hostname]['ansible_all_ipv4_addresses'] | join (',IP:'))}}{% if kube_apiserver_vip %},IP:{{ kube_apiserver_vip }}{% endif %}"
 
 # This is where all of the bearer tokens will be stored
 kube_token_dir: "{{ kube_config_dir }}/tokens"
 
 # This is where you can drop yaml/json files and the kubelet will run those
 # pods on startup
 kube_manifest_dir: "{{ kube_config_dir }}/manifests"
+kube_standby_manifest_dir: "{{ kube_config_dir }}/standby_manifests"
```
danehans: Can you add a space and an inline comment that describes the param? Can kube_manifest_dir simply be used or is a separate standby_manifests directory required?

ansible/roles/load-balancer/defaults/main.yml
```diff
@@ -0,0 +1,6 @@
+keepalived_config_dir: /etc/keepalived/
```
danehans: Pls drop the trailing / for lines 1 and 2

ansible/roles/load-balancer/defaults/main.yml
```diff
@@ -0,0 +1,6 @@
...
+haproxy_config_dir: /etc/haproxy/
+
+keepalived_password: foobar
```
danehans: This should also be exposed in group_vars/all so users can change. By default, keepalived_password in group vars would be commented-out.

ansible/roles/load-balancer/meta/main.yml
```diff
@@ -0,0 +1,4 @@
+---
+dependencies:
+    - { role: kubernetes }
```
danehans: master role includes a dep for the kubernetes role, so line 3 should be unneeded.

ansible/roles/master/defaults/main.yml
```diff
@@ -1,7 +1,13 @@
-kube_master_insecure_port: 8080
+kube_api_insecure_port: 8080
+
+kube_master_backend_api_port: 4443
```
danehans: When using keepalived, we should be able to preserve default ports. For example, the kube-api front-end and backend can listen 443. This is because the kube-api front-end will listen on a different IP (the keepalived vip) than the backend, which listens to 443 of the actual ip.

In this scenario, we can simply use the existing port vars:

```ansible//roles/kubernetes/defaults/main.yml:kube_master_api_port: 443
ansible//roles/master/defaults/main.yml:kube_master_insecure_port: 8080``

File: ansible/roles/load-balancer/templates/haproxy.cfg.j2 [hunk 9]
```diff
@@ -0,0 +1,40 @@
...
+
+defaults
+    log global
+    option  dontlognull
+    timeout connect 5000
+    timeout client 50000
+    timeout server 50000
+
+frontend localhost
+    bind *:{{ kube_master_frontend_api_port }}
```
danehans: Instead of using * in line 21, I believe you need to specify the vip:

```{{ kube_apiserver_vip }}```

I believe if you do not specify kube_apiserver_vip, then haproxy will listen on all addresses.

File: ansible/roles/load-balancer/templates/haproxy.cfg.j2 [hunk 10]
```diff
@@ -0,0 +1,40 @@
...
+    mode tcp
+    option tcplog
+    default_backend apiserver
+
+backend apiserver
+    mode tcp
+    balance roundrobin
+    option ssl-hello-chk
```
danehans: Is option ssl-hello-chk req'd for ssl pass-thru?

ansible/group_vars/all.yml
danehans: As I understand how the kube_apiserver_endpoint var is used, I question whether it should be introduced. I would prefer that we use the existing kube_apiserver_bind_address var:

ansible//roles/master/defaults/main.yml:kube_apiserver_bind_address: "0.0.0.0"


Update kube_apiserver_bind_address to something like: "{% if groups['masters']|length > 1 %}{{ kube_apiserver_vip }}{% else %}"0.0.0.0"{% endif %}"

Let me know your thoughts.

ansible/roles/load-balancer/templates/keepalived.conf.j2
```diff
@@ -0,0 +1,22 @@
...
+
+  #use unicast so we don't have to tool with multicast
+  unicast_peer {
+    {% for master in groups['masters'] %}
+    {{ hostvars[master]['ansible_' + kube_apiserver_interface].ipv4.address }}
+    {% endfor %}
+  }
+
+  virtual_ipaddress {
+    {{ kube_apiserver_endpoint }}
```
danehans: i would expect line 15 to be kube_apiserver_vip.

ansible/roles/load-balancer/templates/keepalived.conf.j2
```diff
@@ -0,0 +1,22 @@
+vrrp_instance VI_1 {
```
danehans: Pls look at the keepalived.conf [here](http://docwiki.cisco.com/wiki/OpenStack_Havana_Release:_High-Availability_Manual_Deployment_Guide#Keepalived_.26_HAProxy)

I think it's a good idea to implement the haproxy tracking script implemented in the above guide. The script will force a vrrp switch over when the local haproxy process dies.

ansible/roles/master/tasks/ha_master.yml
```diff
@@ -0,0 +1,82 @@
+- name: Generic | Install kubernetes node
+  action: "{{ ansible_pkg_mgr }}"
+  args:
+    name: kubernetes-node
+    state: latest
+  notify:
+    - restart daemons
+  when: not is_coreos
+
+- name: CoreOS | Get Systemd Unit Files for kubelet and kube-proxy
```
danehans: Does ha req proxy on the master? I thought it was kubelet. If so, pls remove proxy on master.

ansible/roles/master/tasks/ha_master.yml
```diff
@@ -0,0 +1,82 @@
...
+  get_url:
+    url=https://raw.githubusercontent.com/kubernetes/contrib/master/init/systemd/{{ item }}.service
+    dest=/etc/systemd/system/{{ item }}.service
+    force=yes
+  register: "{{ item }}_service"
+  notify:
+    - reload systemd
+  with_items:
+    - kubelet
+    #- kube-proxy
```
danehans: remove #- kube-proxy if not used.

ansible/roles/master/templates/controller-manager.kubeconfig.j2
```diff
@@ -5,7 +5,7 @@ preferences: {}
 clusters:
 - cluster:
     certificate-authority: {{ kube_cert_dir }}/ca.crt
-    server: https://{{ groups['masters'][0] }}:{{ kube_master_api_port }}
+    server: https://{{ kube_apiserver_endpoint }}:{{ kube_master_api_port }}
```
danehans: I don't think we can remove groups['masters'][0]. I believe you need to set the vip using a conditional. Ex> if masters length > 1, then set the vip. Make sense?

ansible/roles/master/templates/kubectl.kubeconfig.j2
```diff
@@ -5,7 +5,7 @@ preferences: {}
 clusters:
 - cluster:
     certificate-authority-data: {{ kube_ca_cert|b64encode }}
-    server: https://{{ groups['masters'][0] }}:{{ kube_master_api_port }}
+    server: https://{{ kube_apiserver_endpoint }}:{{ kube_master_api_port }}
```
danehans: I don't think we can remove groups['masters'][0]. I believe you need to set the kube_apiserver_vip using a conditional. Ex> if masters length > 1, then set the vip. Make sense?

ansible/roles/master/templates/kubelet.j2
```diff
@@ -4,14 +4,11 @@
 # The address for the info server to serve on (set to 0.0.0.0 or "" for all interfaces)
 KUBELET_ADDRESS="--address=0.0.0.0"
 
-# The port for the info server to serve on
-# KUBELET_PORT="--port=10250"
-
 # You may leave this blank to use the actual hostname
-KUBELET_HOSTNAME="--hostname-override={{ inventory_hostname }}"
+KUBELET_HOSTNAME=""
```
danehans: Is their a reason why we need to remove:

--hostname-override={{ inventory_hostname }}"

ansible/roles/master/templates/kubelet.j2
```diff
@@ -4,14 +4,11 @@
...
 
-# location of the api-server
-KUBELET_API_SERVER="--api-servers=http://localhost:{{ kube_master_insecure_port }}"
+# location of the api-server. Set to none for unregistered master kubelet
+KUBELET_API_SERVER=""
 
 # Add your own!
-KUBELET_ARGS="--register-node=false --config={{ kube_manifest_dir }}"
+KUBELET_ARGS="--register-node=false --register-schedulable=false --config={{ kube_manifest_dir }} --allow-privileged=true"
```
danehans: I don't think we want to allow priv containers on all kubelets, just the kubelets in the masters host group.

ansible/group_vars/all.yml
adamschaub: apiserver_endpoint is intended to be whatever address every node should contact to reach the apiserver. Bind address in this case is what the local apiserver hyperkube container will bind to. In that way endpoint != bind_address.

ansible/roles/master/templates/kubelet.j2
adamschaub: I believe I removed it because it doesn't affect the kubelet as it's currently unregistered/unscheduleable. It's going to have to be added back in to cover the master/node on one host that we discussed earlier.

ansible/roles/master/templates/kubelet.j2
adamschaub: That's true. This kubelet config is just for the master role. Separate config from nodes.

aledbf:
@adamschaub a couple of question:
- are you using haproxy to balance the traffic to the api servers? 
- why not use keepalived for that too using DR?

adamschaub:
@aledbf Great question. As it is currently configured, haproxy+keepalived (forwarding the TCP connection) is functionally similar to keepalived DR. I followed haproxy+keepalived setup mainly as it was the more common setup I found, and partly because we can configure haproxy for ssl-termination which has some benefits if we expose these apis on a public network (perhaps useful for "Ubernetes" [1])

Admittedly I'm no load balancing expert, very open to discussion/opinions. Input on this @danehans?

[1] https://github.com/kubernetes/kubernetes/blob/master/docs/proposals/federation.md

ansible/group_vars/all.yml
danehans: Thanks for the clarification. Even with my new understanding, I still feel that we should not introduce this var. At I high level s/kube_apiserver_endpoint/kube_apiserver_vip

Even if kube-api on master 0 needs to communicate with kube-api, it should go through the lb to hit the api.

Keep in mind that the masters role should have a dep (meta/main.yml) on the lb role when masters is greater than 1.

ansible/roles/master/templates/kubelet.j2
danehans: OK. Unless we need to remove/change existing code to support HA, let's leave it as-is. Any code removal with be scrutinized by maintainers.

ansible/roles/master/templates/kubelet.j2
danehans: OK. We should be able to include the masters > 1 conditional logic to manage the vars.

ansible/roles/load-balancer/templates/haproxy.cfg.j2
adamschaub: Not required, just an additional check for ssl on the apiserver. Removing it is fine.

adamschaub:
@danehans Updated dependencies so master depends on lb role, unified all the port bindings, removed kube-proxy, added kube-master hostname override back in, updated vagrant file to include kube_apiserver_interface, etc.

Still need to implement haproxy status check in keepalived.conf. For now, master/slave weighting should get us by.

test-foxish:
recomputing cla status...

grodrigues3:

- **ansible**

cc @eparis 

PR #846, opened 2016-04-26
Adapt nginx config for supporting proxy_protocol on HTTP/HTTPS customer facing

simonswine:
Hi everyone,

I tried to use this controller on my baremetal cluster. (The real frontend is haproxy based and doesn't terminate HTTP/HTTPS. It just forwards the client IP information with proxy_protocol.

The current master nginx ingress controller doesn't not enable proxy protocol for these customer facing sockets, when use-proxy-protocol is true. 

I changed the template accordingly

Cheers,

Christian

### DIFF
--- a/ingress/controllers/nginx/README.md
+++ b/ingress/controllers/nginx/README.md
@@ -132,6 +132,14 @@ To disable this behavior use `hsts=false` in the NGINX ConfigMap.
 NGINX provides the configuration option [ssl_buffer_size](http://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_buffer_size) to allow the optimization of the TLS record size. This improves the [Time To First Byte](https://www.igvita.com/2013/12/16/optimizing-nginx-tls-time-to-first-byte/) (TTTFB). The default value in the Ingress controller is `4k` (nginx default is `16k`);
 
 
+## Proxy Protocol
+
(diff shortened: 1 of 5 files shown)


simonswine:
Another bug I hit. When enabling use-proxy-protocol the management port 8080 is set into proxy_protocol mode, which makes the health check fail. I fixed that in the latest commit...

ingress/controllers/nginx/nginx.tmpl
```diff
@@ -153,8 +153,8 @@ http {
 
     {{ range $server := .servers }}
     server {
-        listen 80;
-        {{ if $server.SSL }}listen 443 ssl http2;
+        listen 80{{ if $cfg.useProxyProtocol }} proxy_protocol{{ end }};
+        {{ if $server.SSL }}listen 443{{ if $cfg.useProxyProtocol }} proxy_protocol{{ end }} ssl http2;
```
bprashanth: Can i trouble you to add a proxy protocol example to the docs? I've actually heard others asking for the same and explaining it everytime is burdensome. If you have the time, I suggest a section that describes eg: start haproxy by creating this pod, now start the nginx controller behind haproxy, now do this to see src etc ...doesn't need to be very elaborate, just demonstrate a situation where user doesn't get src, and they does get it.

simonswine:
@bprashanth: I have just added documentation, maybe you give it a quick check, as it's still quite early here.
:laughing: 

The fix itself (in the template) is already in the master as part of 6133c84470851e0d3a2746685ce1da0bdf55bcfd

bprashanth:
@simonswine great thanks!

PR #849, opened 2016-04-26
Fix nginx ingress controller bug around config map merging

simonswine:
Hi I just discovered a bug around the nginx config merge (defaults are merged with configmap)

 * a config map bool value of false cannot overwritte a true value from
   defaults
 * implement merging in ReadConfig
 * remove helper function merge
 * adds tests to ensure config is read properly


bprashanth:
@aledbf

aledbf:
Code LGTM

bprashanth:
@simonswine thanks, merging

PR #853, opened 2016-04-26
Perfdash: Fetch build logs from latest to oldest

Random-Liu:
Fetch build logs from latest to oldest so as to make sure metrics with newer version are parsed first.

/cc @gmarek


Random-Liu:
@gmarek Since you said you are on vocation the following days, and this is really a minor change, I think I shouldn't bother you. :)

Self apply LGTM.

PR #859, opened 2016-04-28
On Atomic hosts address the conditional check ''etcd-legacy' not in..

andrew-j-price:
The following error occurs when running against Atomic hosts.  

TASK [etcd : Open etcd client legacy port with iptables] ***********************
fatal: [<host>]: FAILED! => {"failed": true, "msg": "The conditional check ''etcd-legacy' not in iptablesrules.stdout and networking == contiv' failed. The error was: error while evaluating conditional ('etcd-legacy' not in iptablesrules.stdout and networking == contiv): 'contiv' is undefined\n\nThe error appears to have been in...

### DIFF
--- a/ansible/roles/etcd/tasks/iptables.yml
+++ b/ansible/roles/etcd/tasks/iptables.yml
@@ -18,7 +18,7 @@
 
 - name: Open etcd client legacy port with iptables
   command: /sbin/iptables -I INPUT 1 -p tcp --dport {{ item }} -j ACCEPT -m comment --comment "etcd-legacy"
-  when: "'etcd-legacy' not in iptablesrules.stdout and networking == contiv"
+  when: "'etcd-legacy' not in iptablesrules.stdout and networking == 'contiv'"
   notify:
     - Save iptables rules
   with_items:
--- a/ansible/roles/etcd/tasks/firewalld.yml
+++ b/ansible/roles/etcd/tasks/firewalld.yml
@@ -19,7 +19,7 @@
   firewalld: port={{ item }}/tcp permanent=false state=enabled
   # in case this is also a node where firewalld turned off
   ignore_errors: yes
-  when: networking == contiv
+  when: networking == 'contiv'
   with_items:
     - "{{ etcd_client_legacy_port }}"
     - "{{ etcd_peer_legacy_port }}"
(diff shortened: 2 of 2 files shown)


andrew-j-price:
I signed it!

eparis:
I think this is not the right fix...

@danehans   should 'contriv' be in quotes?
@andrew-j-price does putting the 'contrib' in quotes fix your issue?

andrew-j-price:
@eparis you are correct for 'contriv' needing to be in quotes.  Updated the commit.

ansible/roles/etcd/tasks/iptables.yml
```diff
@@ -24,3 +24,4 @@
   with_items:
     - "{{ etcd_client_legacy_port }}"
     - "{{ etcd_peer_legacy_port }}"
+  when: not is_atomic
```
danehans: This should not be added. I think it's a workaround, instead of fixing the core issue. The project supports multiple networking implementations. The legacy ports are required (for the time being) by contiv networking.

ansible/roles/etcd/tasks/iptables.yml
```diff
@@ -18,10 +18,9 @@
 
 - name: Open etcd client legacy port with iptables
   command: /sbin/iptables -I INPUT 1 -p tcp --dport {{ item }} -j ACCEPT -m comment --comment "etcd-legacy"
-  when: "'etcd-legacy' not in iptablesrules.stdout and networking == contiv"
+  when: "'etcd-legacy' not in iptablesrules.stdout and networking == 'contiv'"
```
danehans: this should be the fix to the problem.

Do you mind updating ansible//roles/etcd/tasks/firewalld.yml to quote contiv in:

when: networking == contiv

and test in Atomic? I would like to make sure the firewall/iptables implementation are as similar as possible.

danehans:
@eparis or @andrew-j-price I have created https://github.com/kubernetes/contrib/issues/861 to highlight the root cause of this issue. Can either of you address the above issue or know someone that can?

danehans:
@andrew-j-price can u squash, remove ```when: not is_atomic``` from iptables.yml, test and resubmit?

eparis:
looks good! can you squash to a single commit? https://help.github.com/articles/about-git-rebase/ should talk about how....

andrew-j-price:
@eparis and @danehans , hopefully my limited git skills got the needful accomplished above.  but tested and works.

eparis:
blast. no  :-(
```
git remote update
git rebase -i b95de190654e32ce3dd7452013a2642db63ddb2a
change the first word of the second and later lines to: s
save/exit
update so there is a single commit message
git show (make sure it shows all of the changes and your commit message)
git push -f $REMOTE $LOCAL_BRANCH:master
```

andrew-j-price:
thank you for the hand holding @eparis .  Think I got it this time.

eparis:
looks great! Thank you.

Issue #862, opened 2016-04-28
shame mailer mysteriously stopped working


ixdy:
The daily shame mailer mysteriously stopped working a few days ago.

From the log on Jenkins:
```
Running gcr.io/google_containers/shame-mailer:2016-03-25-d0e88f1
I0426 15:40:53.052672       6 github.go:215] Made 0 API calls since the last Reset 0.000000 calls/sec
Error: Error running shame: exit status 1
Usage:
  mungegithub [flags]
  mungegithub [command]

Available Commands:
  gencommiters Generate the list of people with commit access

Flags:
      --address=":8080": The address to listen on for HTTP Status
      --alsologtostderr[=false]: log to standard error as well as files
      --block-path-config="block-path.yaml": file containing the pathnames to block or not block
      --blunderbuss-reassign[=false]: Assign PRs even if they're already assigned; use with -dry-run to judge changes to the assignment algorithm
      --committers="./committers.txt": File in which the list of authorized committers is stored; only used if this list cannot be gotten at run time.  (Merged with whitelist; separate so that it can be auto-generated)
      --dry-run[=true]: If true, don't actually merge anything
      --e2e-status-context="Jenkins GCE e2e": The name of the github status context for the e2e PR Builder
      --generated-files-config="generated-files.txt": file containing the pathname to label mappings
      --issue-reports=[]: A list of issue reports to run. If set, will run the reports and exit.
      --jenkins-host="http://jenkins-master:8080": The URL for the jenkins job to watch
      --jenkins-jobs=[kubelet-gce-e2e-ci,kubernetes-build,kubernetes-test-go,kubernetes-e2e-gce,kubernetes-e2e-gce-slow,kubernetes-e2e-gke,kubernetes-e2e-gke-slow,kubernetes-e2e-gce-scalability,kubernetes-kubemark-5-gce]: Comma separated list of jobs in Jenkins to use for stability testing
      --kubernetes-dir="./gitrepos/kubernetes": Path to git checkout of kubernetes tree
      --labels=[]: CSV list of label which should be set on processed PRs. Unset is all labels.
      --log-backtrace-at=:0: when logging hits line file:N, emit a stack trace
      --log-dir="": If non-empty, write log files in this directory
      --log-flush-frequency=5s: Maximum number of seconds between log flushes
      --logtostderr[=true]: log to standard error instead of files
      --max-pr-number=9223372036854775807: The maximum PR to start with
      --min-pr-number=0: The minimum PR to start with
[...]

PR #863, opened 2016-04-28
Make shame command print out to stdout/stderr

ixdy:
We're running `mailx` in verbose mode, but the stdout/stderr was going nowhere, which was of no use.

This change helped me root-cause #862.

@eparis @krousey @lavalamp


lavalamp:
merge & push at will

PR #866, opened 2016-04-29
Cluster-autoscaler: build k8s scheduler node info structure based on kubectl drain

mwielgus:
kubernetes/kubernetes#24404

It would be nice if :
https://github.com/kubernetes/kubernetes/pull/24929
https://github.com/kubernetes/kubernetes/pull/24923
were merged before this PR (although it is not a super hard requirement, we can update godeps again once they are merged).

cc: @piosz @fgrzadkowski


piosz:
LGTM

PR #887, opened 2016-05-02
Fix nginx ingress unittests.

bprashanth:
Bad merge between https://github.com/kubernetes/contrib/pull/849 and #766 
@aledbf


bprashanth:
Green, merging

PR #889, opened 2016-05-03
Bump up nginx controller to 0.61

bprashanth:
@aledbf 
0.61 contains https://github.com/kubernetes/contrib/pull/849


bprashanth:
@simonswine fyi, in case you need to version that contains your fix

aledbf:
LGTM

PR #890, opened 2016-05-03
Add a flag to control which domains we send shame emails to

ixdy:
Intent is to make it easy to add permitted domains like redhat.com, as discussed in https://github.com/kubernetes/test-infra/pull/11#discussion_r61953400, all without having to rebuild everything.

cc @kubernetes/sig-testing


eparis:
LGTM merge whenever you have the 'other side' ready to make use of it.

ixdy:
FYI I just updated the body of the email to use the domains list too, instead of saying that only googlers get emails.

PR #894, opened 2016-05-04
[ansible] Make sure kubelet is restarted

gouyang:


gouyang:
close this as the commit message seems dirty, will re-send one later

PR #895, opened 2016-05-04
[ansible] Make sure kubelet service is up.

gouyang:
When kubelet service is stopped, run the playbook does not get the
service up.

### DIFF
--- a/ansible/roles/node/tasks/main.yml
+++ b/ansible/roles/node/tasks/main.yml
@@ -59,7 +59,7 @@
     - restart kubelet
 
 - name: Enable kubelet
-  service: name=kubelet enabled=yes 
+  service: name=kubelet enabled=yes state=started 
   notify:
     - restart kubelet
 


ansible/roles/node/tasks/main.yml
```diff
@@ -59,7 +59,7 @@
     - restart kubelet
 
 - name: Enable kubelet
-  service: name=kubelet enabled=yes 
+  service: name=kubelet enabled=yes state=started
```
eparis: This should happen in the `restart kubelet` handler I thought to prevent numerous wasted bounces

gouyang: `restart kubelet` should only be invoked when changes happen.
Sometimes I stop the kube cluster and not delete the config files, re-run the playbook, only kubelet service is not running.

```
TASK [node : Set token facts] **************************************************
ok: [...]

TASK [node : Create the kubelet working directory] *****************************
ok: [...]

TASK [node : write the config files for kubelet] *******************************
ok: [...]

TASK [node : write the kubecfg (auth) file for kubelet] ************************
ok: [...]

TASK [node : Enable kubelet] ***************************************************
ok: [...]

TASK [node : write the config files for proxy] *********************************
ok: [...]

TASK [node : write the kubecfg (auth) file for kube-proxy] *********************
ok: [...]

TASK [node : Enable proxy] *****************************************************
changed: [...]

TASK [node : write the delay-node-services target] *****************************
ok: [...]

TASK [node : Enable delay-node-services] ***************************************
ok: [...]
```

Could see kubelet notify is not happened because no changes happen, however kube-proxy is changed because the service is stopped and `state=started` make sure it should be running. 

And I seems `state=started` only miss for kubelet.

Issue #852, opened 2016-04-26
[ansible] Support Lifecycle Management


danehans:
Currently, the contrib/ansible project does a good job of performing an initial deployment of a Kubernetes cluster. After the cluster is initially deployed, operators have a multitude of tasks that must be performed to manage the lifecycle of the cluster. For example:
- Add a master or node to the cluster
- Upgrade the version of Kubernetes
- etc..

Ansible playbooks should be created, tested, documented, supported, etc. to support common lifecycle management tasks. Additionally, the overall design of the project should be addressed to support multiple playbooks.

galexrt:
From my perspective there is currently especially one thing that keeps us from having a better lifecycle management for clusters with Ansible:
The certificate generation. This can be added/fixed by altering the scripts to check with `openssl` what names are in the current certificate and generating a new one for newly added services (/servers) (that require a certificate "change".

Another point to mention is that it would be better to use notify instead of the `*_modified` variables to restart the services (something I have been working on a bit already, if it is wanted).

smenon78:
What's the best way to Upgrade kubernetes on CentOS 7?

galexrt:
@smenon78 Have you just trying to re-run the deployment? That's the way I updated my clusters as already mentioned.
See https://github.com/kubernetes/contrib/blob/master/ansible/roles/master/tasks/pkgMgrInstallers/centos-install.yml#L5 as the latest version of Kubernetes master and nodes will be always installed by running the playbook(s).

smenon78:
Latest package available for me via packageinstall is 1.5.2. I would like to upgrade to 1.6.1
yum --showduplicates list kubernetes
Loaded plugins: fastestmirror
Loading mirror speeds from cached hostfile
 * base: ftp.usf.edu
 * epel: mirror.compevo.com
 * extras: repos-va.psychz.net
 * updates: ftp.osuosl.org
Available Packages
kubernetes.x86_64                                                                                                          1.3.0-0.3.git86dc49a.el7                                                                                                          extras
kubernetes.x86_64                                                                                                          1.4.0-0.1.git87d9d8d.el7                                                                                                          extras
kubernetes.x86_64                                                                                                          1.5.2-0.2.gitc55cf2b.el7                                                                                                          extras
kubernetes.x86_64                                                                                                          1.5.2-0.5.gita552679.el7

galexrt:
Then there are three posibilites to do it with the Ansible deployment:
[...]

Issue #838, opened 2016-04-24
[ansible] Refactor Kubernetes Roles Based on Services


danehans:
The current kubernetes ansible roles are organized into masters and nodes. This makes breaking the master/nodes services into individual units of management, deployment, etc.. In the past, I have seen issues with other projects that do not manage services individually. I suggest changing master into individual roles based on master service, such as:
1. kube-apiserver
2. kube-controller-mansger
3. kube-scheduler

The same for the node role:
1. kubelet
2. kube-proxy

and the client role:
1. kubectl

A single master inventory would stay the same.

site manifest (i.e. cluster.yml) would look like:

```
- hosts: all
  gather_facts: false
  sudo: yes
  roles:
    - pre-ansible
  tags:
    - pre-ansible

# Install etcd
- hosts: etcd
  sudo: yes
  roles:
    - etcd
  tags:
    - etcd

- hosts: all
  sudo: yes
  roles:
    - common
    - docker
  tags:
    - docker

# install flannel
- hosts:
    - masters
    - nodes
  sudo: yes
  roles:
    - { role: flannel, when: networking == 'flannel' }
  tags:
    - network-service-install

# install kube master services
- hosts: masters
  sudo: yes
  roles:
    - kube-apiserver
   - kube-controller-manager
   - kube-scheduler
   - kubectl
   - kubernetes-addons
  tags:
    - masters
    - dns

# install kubernetes on the nodes
- hosts: nodes
  sudo: yes
  roles:
    - kubelet
    - kube-proxy
  tags:
    - nodes
```

ingvagabund:
Not sure if you will be able to separate master service without increasing complexity. As both ``master`` and ``node`` roles depend on ``kubernetes`` role, each invocation of ``apiserver``, ``scheduler`` and other playbooks will call ``kubernetes`` and ``common`` role. At the same time, ``kubernetes`` role is responsible for generating of certificates and tokens.

If would be great to make sure the ``kubernetes`` role gets run only once. Besides, the entire ``master`` role gets run one more time when you run ``kubernetes-addons`` playbook. Assuming all tasks are idempotent, it does not matter. Still, you run some tasks even when there is no need for it.

I can think of the following scenarios:

* remove dependency of ``kubernetes`` roles from ``master``/``node`` and run the ``kubernetes`` right before them. E.g.

```yaml
- hosts: masters
  sudo: yes
  roles:
   - kubernetes
   - kube-apiserver
   - kube-controller-manager
[...]

Issue #793, opened 2016-04-18
[ansible] Allow Roles to be loosely coupled


danehans:
For the most part, roles within contrib/ansible are tightly coupled. For example, I can not easily swap out the contrib/ansible etcd role with a different ectd role from a upstream such as Ansible Galaxy [1].

[1] https://galaxy.ansible.com/

danehans:
https://github.com/kubernetes/contrib/issues/792

PR #899, opened 2016-05-04
Etcd refactor

stephenrlouie:
The refactor breaks down the etcd role into smaller component plays that could be reused in larger playbooks, such as cluster wide, deployments, upgrades, etc. If all the roles are more modular it would make additional playbooks much easier to create and this gives us a consistent pattern to follow with the other roles. 

Hit an ansible bug: https://github.com/ansible/ansible/issues/13818 that effects ansible > 2.0, so I rolled back to 1.9.4, and this works. The bug makes ansible unable to find the templates/files in the role when run from Vagrant.

- Related to #793
- Should make #852 easier
- An example of what #838 should be

----------

Would appreciate feedback @danehans @rutsky @adamschaub 

Refactor of etcd role to be modeled after [openstack-ansible](https://github.com/openstack/openstack-ansible/tree/kilo) and [hpe-helion] (https://github.com/hpe-helion-os/helion-ansible). Confirmed on Virtualbox + CoreOS and Virtualbox + Centos

- I think we should always use the binaries from github, regardless of OS, so we can always control the version of etcd present. CoreOS comes with etcd, we would have to uninstall and install the specified version for CoreOS. Where Centos uses the [generic package installer] (http://docs.ansible.com/ansible/package_module.html) as seen [here] (https://github.com/kubernetes/contrib/blob/master/ansible/roles/etcd/tasks/main.yml#L2). It does not support versions, we currently just specify `latest` and end up with the latest copy of etcd that is in yum/apt-get. (A very unlucky person could be running the install script and the version served by yum/apt-get might get changed during their deployment, leaving them with different versions of etcd) Ubuntu is the only one using the binaries right now, seen [here] (https://github.com/kubernetes/contrib/blob/master/ansible/roles/etcd/tasks/main.yml#L11). Using binaries will let anyone regardless of OS to use a specific version of etcd.

Design Question:
- Not sure where `deploy.yml` should live. I put it one level up from `tasks` right now since its wraps a bunch of `etcd/tasks` playbooks together
- How granular do we want bottom level playbooks that live in the task file.
  - `install.yml` has all install instructions regardless of OS.
  - `install.yml` just cycles through each OS supported, and calls plays like `install-coreos.yml` `install-ubuntu.yml`
    - Right now I'm breaking it out if there is more than 2 statements that apply to the same `when`, see `coreos-start.yml` and `coreos-stop.yml`. I see this is being a bit more manageable down the road at the cost of the directory getting a little crowded.


stephenrlouie:
@danehans @adamschaub - could you take a look, no longer a WIP

stephenrlouie:
@eparis & @danehans 

Could you guys provide feedback for this ansible refactor. I think more granular tasks will make our lives easier down the road. I modeled this after hpe-helion & openstack-ansible. Adding functionality for upgrades / uninstall and whatever comes in the future should be more manageable like this as opposed to what was there before.

 
I confirmed this works on Virtualbox + CoreOS and Virtualbox + Centos. I'll be submitting another PR soon to support updates for CoreOS

ansible/vagrant/Vagrantfile
```diff
@@ -159,7 +159,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     set_common_ansible_options(ansible)
 
     # This sets up both flannel and kube.
-    ansible.playbook = "../cluster.yml"
+    ansible.playbook = "../deploy.yml"
```
adamschaub: Should this be ../playbooks/deploy.yml?

stephenrlouie: good catch

stephenrlouie:
The second commit organizes the top level playbooks into a `playbooks` directory. This causes vagrant to fail because ansible cannot find `group_vars`. 

The fix is to move `group_vars` into the directory where the playbook is being called so to `contrib/ansible/playbooks` **or** where the inventory file lives for vagrant `contrib/ansible/vagrant/.vagrant/provisioners/ansible/inventory/vagrant_ansible_inventory`

A different PR: https://github.com/kubernetes/contrib/pull/933 is related and puts a symlink to `group_vars` where the `vagrant_ansible_inventory` file lives.

PR #909, opened 2016-05-05
Retest stale tests after 96 hours, not 48.

spxtr:
@eparis assuming this works I think it should go in ASAP. We currently can't merge because Jenkins is overloaded and it takes too long for a test to get through the Jenkins queue.

### DIFF
--- a/mungegithub/mungers/stale-green-ci.go
+++ b/mungegithub/mungers/stale-green-ci.go
@@ -29,7 +29,7 @@ import (
 )
 
 const (
-	staleGreenCIHours = 48
+	staleGreenCIHours = 96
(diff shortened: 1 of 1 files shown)


eparis:
It looks like things are making progress. Want to split the difference at 3 days? And tell @ixdy to make Jenkins keep up    :)

ixdy:
Going back to the original issue (kubernetes/contrib/issues/197), we added this because we didn't re-run any verify/unit/integration tests before merge. We now do this, so I'm not sure periodically running tests gets us much benefit.

eparis:
@ixdy I think there is some value, as our queue takes a week, for people to know sooner when their PRs are broken. But you are right, the root cause is solved.  So I'll be happy with 4 days.

PR #911, opened 2016-05-06
Make WeakStable blocking - it was running OK for a couple weeks

gmarek:
cc @kubernetes/sig-testing


eparis:
Is the health % accounting correct for 'weak stable' ?  I'm not sure what we want...  Right now i think the % shows failure for a 'weak' failure, but the queue isn't blocked...  Which do you think is correct?

gmarek:
It's not AFAICT, but I wasn't putting to much attention to that.

PR #922, opened 2016-05-07
Multimaster vagrant

andreamelloncelli:
Added an environment variable to choose the number of masters to Vagrantfile.


andreamelloncelli:
I signed it!

PR #924, opened 2016-05-07
GCE Ingress Controller - Added flag for healthcheck interval

ajcurated:
When using GCE ingress and the default 1 second interval you can end up generating a large number of requests to your services. This is made worse if you have a large cluster as it will be receiving the health check to each node for each service every second, it wouldn't be unrealistic for you to end up with 100rps from just health checks in a large cluster.

I believe there are discussions to have more granular ways to configure healthcheck and other params, but for now this added a flag to allow the interval to be increased.

Much like the `--health-check-path` this is simply `--health-check-internval`

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/924)

### DIFF
--- a/ingress/controllers/gce/backends/backends_test.go
+++ b/ingress/controllers/gce/backends/backends_test.go
@@ -31,7 +31,7 @@ func newBackendPool(f BackendServices, fakeIGs instances.InstanceGroups, syncWit
 	namer := utils.Namer{}
 	return NewBackendPool(
 		f,
-		healthchecks.NewHealthChecker(healthchecks.NewFakeHealthChecks(), "/", namer),
+		healthchecks.NewHealthChecker(healthchecks.NewFakeHealthChecks(), 1, "/", namer),
 		instances.NewNodePool(fakeIGs, "default-zone"), namer, []int64{}, syncWithCloud)
 }
 
--- a/ingress/controllers/gce/controller/cluster_manager.go
+++ b/ingress/controllers/gce/controller/cluster_manager.go
@@ -34,8 +34,9 @@ import (
 )
 
 const (
-	defaultPort            = 80
-	defaultHealthCheckPath = "/"
+	defaultPort                = 80
+	defaultHealthCheckInterval = 1
+	defaultHealthCheckPath     = "/"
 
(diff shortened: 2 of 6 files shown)


bprashanth:
@ajcrowe Thanks for the pr!
I'm ok doing this. But if you have the cycles I think reading both these options from Ingress annotations would be even better (https://github.com/kubernetes/contrib/issues/934). Let me know if you want to try that or just get this in.

ajcrowe:
@bprashanth I think that would be great,

Can you point to me more information about annotations, as I would be keen to make this flexible per ingress resource

bprashanth:
@ajcrowe awesome!
This is how you read annotations: https://github.com/kubernetes/contrib/blob/eeb9a5d1cfaf8f71b23774cb03cca846716f3ad9/ingress/controllers/gce/controller/controller.go#L371

To set it through kubectl you can either use `kubectl annotate`, or `kubectl edit` and write the annotation yourself.

And here's how you'd write it: https://github.com/kubernetes/contrib/blob/eeb9a5d1cfaf8f71b23774cb03cca846716f3ad9/ingress/controllers/gce/controller/controller.go#L351

Using annotations for this feels like to overall right thing to do. The sketch in the #934 is just a high leve idea, I'm open to others as well.

bprashanth:
Also fyi, noticed https://ajcrowe.io/2016/04/kubernetes-gcp-http-load-balancing-tips (nice writeup!), the first 2 points should be fixed in HEAD:
Firewall rules creation: https://github.com/kubernetes/contrib/pull/791/commits/d734cd7722c7604cffdbe6e8f4cefbb0cb3c60f2 (was blocked on a bug upstream, that we hit through the vendored Godep)
Cluster uid: https://github.com/kubernetes/kubernetes/pull/24391 (it was blocked on setup time config templating)

we're just testing those changes, should be included with the next image (0.6.2).

test-foxish:
recomputing cla status...

ajcrowe:
This has been solved in other way like it using the healthchecks of the pods in a service for the path and the cluster uid etc

PR #926, opened 2016-05-08
[nginx-ingress-controller] Custom errors should be optional

### LINKED ISSUE #897: Nginx Ingress Controller - Option to disable interception of errors
It seems to me right now there is no way to configure the nginx ingress controller to **not** intercept errors and serve them from the default/error backend.

In our case this is problematic. Our backends serve meaningful error pages for 500 errors which contain exception ids which help us debugging problems when customers create support requests.
We also have a JSON API backend where 404 error contain additional information that help api clients to fail more gracefully.

How about adding an option to (potentially partially) disable the *error interception* ?
I can submit an initial PR of this would be considered useful.

aledbf:
Without this change the errors in the upstream server are redirected to the default backend.

fixes #897

### DIFF
--- a/ingress/controllers/nginx/nginx/main.go
+++ b/ingress/controllers/nginx/nginx/main.go
@@ -124,19 +124,11 @@ type nginxConfiguration struct {
 	// accessed using HTTPS.
 	HSTSMaxAge string `structs:"hsts-max-age,omitempty"`
 
-	// enables or disable if HTTP codes should be passed to a client or be redirected to nginx for
-	// processing with the error_page directive
+	// enables which HTTP codes should be passed for processing with the error_page directive
 	// http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_intercept_errors
+	// http://nginx.org/en/docs/http/ngx_http_core_module.html#error_page
 	// By default this is disabled
-	InterceptHTTP403 bool `structs:"intercept-error-403,omitempty"`
-	InterceptHTTP404 bool `structs:"intercept-error-404,omitempty"`
-	InterceptHTTP405 bool `structs:"intercept-error-405,omitempty"`
-	InterceptHTTP408 bool `structs:"intercept-error-408,omitempty"`
-	InterceptHTTP413 bool `structs:"intercept-error-413,omitempty"`
-	InterceptHTTP501 bool `structs:"intercept-error-502,omitempty"`
-	InterceptHTTP502 bool `structs:"intercept-error-502,omitempty"`
-	InterceptHTTP503 bool `structs:"intercept-error-503,omitempty"`
-	InterceptHTTP504 bool `structs:"intercept-error-504,omitempty"`
+	CustomHTTPErrors []int
 
 	// Time during which a keep-alive client connection will stay open on the server side.
 	// The zero value disables keep-alive client connections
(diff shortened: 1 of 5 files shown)


databus23:
Any progress on this? I'm eagerly waiting for this. :)
/cc @bprashanth

ingress/controllers/nginx/nginx/main.go
```diff
@@ -124,6 +124,20 @@ type nginxConfiguration struct {
 	// accessed using HTTPS.
 	HSTSMaxAge string `structs:"hsts-max-age,omitempty"`
 
+	// enables or disable if HTTP codes should be passed to a client or be redirected to nginx for
```
bprashanth: Hmm, can you just pass a list and parse in the controller instead of each code?

bprashanth: can you also add some text to the docs about this?

aledbf: done

ingress/controllers/nginx/nginx/template.go
```diff
@@ -56,15 +56,9 @@ func (ngx *Manager) writeCfg(cfg nginxConfiguration, ingressCfg IngressConfig) (
...
 	conf["cfg"] = fixKeyNames(structs.Map(cfg))
 
-	buffer := new(bytes.Buffer)
-	err := ngx.template.Execute(buffer, conf)
-	if err != nil {
-		glog.Infof("NGINX error: %v", err)
-		return false, err
-	}
-
 	if glog.V(3) {
```
bprashanth: suggest simply having a global --verbose=true and flipping log level to V(4) everywhere. That's been helpful for me at least, I almost never really want levelled logs. Just everything, or default. If you want to do that (not in this pr) you can copy how https://github.com/kubernetes/contrib/blob/master/ingress/controllers/gce/main.go#L111 is handled

ingress/controllers/nginx/nginx/utils.go
```diff
@@ -91,6 +96,19 @@ func (ngx *Manager) ReadConfig(config *api.ConfigMap) nginxConfiguration {
 		Metadata:         metadata,
 	})
 
+	cErrors := make([]int, 0)
+	if val, ok := config.Data[customHTTPErrors]; ok {
+		delete(config.Data, customHTTPErrors)
+		for _, i := range strings.Split(val, ",") {
+			j, err := strconv.Atoi(i)
+			if err != nil {
+				glog.Warningf("%v is not a valid http code", i)
```
bprashanth: all you can say is that this is not an int, right? can you also log the error?

ingress/controllers/nginx/nginx/utils.go
```diff
@@ -115,9 +133,23 @@ func (ngx *Manager) ReadConfig(config *api.ConfigMap) nginxConfiguration {
...
 	}
 
+	cfgDefault.CustomHTTPErrors = ngx.filterErrors(cErrors)
 	return cfgDefault
 }
 
+func (ngx *Manager) filterErrors(errCodes []int) []int {
+	fa := make([]int, 0)
+	for _, errCode := range errCodes {
+		if errCode > 299 && errCode < 600 {
```
bprashanth: is this a continguous range?

bprashanth:
LGTM sorry for the delay, all I had were a couple of nits that can easily be fixed in a follow up pr

ingress/controllers/nginx/nginx/utils.go
aledbf: yes, any code in that range can be customized

ingress/controllers/nginx/nginx/utils.go
aledbf: ok, duly noted for the next PR

PR #928, opened 2016-05-09
Make WeakStable even weaker

gmarek:
For some reason failures tend to go in pairs - we should ignore one additional failure.


eparis:
```
--- FAIL: TestCheckGCSWeakBuilds (0.61s)
	e2e_test.go:464: expected: false, saw: true
	e2e_test.go:467: expected: map[foo:{Not Stable 42} bar:{Stable 44}], saw: map[foo:{Stable 42} bar:{Stable 44}]
```

gmarek:
Aa, I forgot to update tests. I'll do it tomorrow.

gmarek:
Fixed. @eparis PTAL

PR #940, opened 2016-05-10
Make copy when deleting instead of messing up range operations

eparis:
We would walk the labels and comments lists and could then delete from
them. Since we used the form `a = append(a[:i], a[i+1:...])` we were
actually changing the list as it was being used in a `range a`
statement. This could cause the comment deleter and the LGTM label
checker to go nuts.

### DIFF
--- a/mungegithub/github/github.go
+++ b/mungegithub/github/github.go
@@ -654,7 +654,12 @@ func (obj *MungeObject) RemoveLabel(label string) error {
 		}
 	}
 	if which != -1 {
-		obj.Issue.Labels = append(obj.Issue.Labels[:which], obj.Issue.Labels[which+1:]...)
+		// We do this crazy delete since users might be iterating over `range obj.Issue.Labels`
+		// Make a completely new copy and leave their ranging alone.
+		temp := make([]github.Label, len(obj.Issue.Labels)-1)
+		copy(temp, obj.Issue.Labels[:which])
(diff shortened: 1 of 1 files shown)


mungegithub/github/github.go
```diff
@@ -654,7 +654,12 @@ func (obj *MungeObject) RemoveLabel(label string) error {
 		}
 	}
 	if which != -1 {
-		obj.Issue.Labels = append(obj.Issue.Labels[:which], obj.Issue.Labels[which+1:]...)
+		// We do this crazy delete since users might be iterating over `range obj.Issue.Labels`
+		// Make a completly new copy and leave their ranging alone.
```
knobunc: completely

mungegithub/github/github.go
```diff
@@ -1345,13 +1350,22 @@ func (obj *MungeObject) DeleteComment(comment *github.IssueComment) error {
...
 		if c.ID == nil || *c.ID != *comment.ID {
 			continue
 		}
-		obj.comments = append(obj.comments[:i], obj.comments[i+1:]...)
-		break
+		which = i
+	}
+	if which != -1 {
+		// We do this crazy delete since users might be iterating over `range obj.comments`
+		// Make a completly new copy and leave their ranging alone.
```
knobunc: completely

knobunc:
LGTM

eparis:
Thank you @knobunc @jayunit100

PR #374, opened 2016-01-13
remove global Godeps and create one per subproject

mikedanese:

PR #943, opened 2016-05-10
Move Ingress godeps to vendor/

aledbf:


aledbf:
@bprashanth this change is mandatory right? If that's true then this requires the change of the scripts `hack/verify-*`

eparis:
I'm looking at this for mungegithub (updating verify-*)   give me a bit.

eparis:
https://github.com/kubernetes/contrib/pull/947/files

eparis:
looks like we ended up in about the same place!

hack/verify-golint.sh
```diff
@@ -30,7 +30,12 @@ fi
 cd "${KUBE_ROOT}"
 
 GOLINT=${GOLINT:-"golint"}
-bad_files=$($GOLINT -min_confidence=0.9 ./...)
+
+find_files() {
+  $(go list -f '{{.Dir}}' ./... | grep -v vendor)
+}
+
+bad_files=$(find_files | xargs $GOLINT -min_confidence=0.9)
```
eparis: This doesn't work. Multiple arguments to golint are supposed to be files, not package directories...

aledbf: @eparis thanks for this :)
I will remove the last commit

eparis: if you update this, i'll merge yours and then rebase mine.

aledbf: ok

eparis:
https://github.com/kubernetes/contrib/pull/943/commits/3a3b0f5f71f848a8b64aef83d4c15fc5741deb76#diff-c29febb4cb6d2f120186cb03752938a6R82

should that have `vendor` ?

aledbf:
@eparis no need https://github.com/aledbf/contrib/blob/vendor-ingress/hack/verify-flags-underscore.py#L218

eparis:
what if used with args?  https://github.com/aledbf/contrib/blob/vendor-ingress/hack/verify-flags-underscore.py#L216  like if this was part of a pre-commit git hook?

eparis:
ohhhh, do we need to disable golang < 1.6 on travis as well?

aledbf:
@eparis I was writing that question :)

eparis:
we did it in kube proper, I say do it here.

eparis:
or we can support 1.5, if you sent the ENV var.

eparis:
but 1.4 is dead to me. (and I don't care about 1.5)

aledbf:
@eparis it's ok to remove 1.5? It fails even with `GO15VENDOREXPERIMENT=1`

eparis:
kill it and move along.

aledbf:
ping @bprashanth

.travis.yml
```diff
@@ -2,10 +2,12 @@ sudo: false
 
 language: go
 
+env:
+  global:
+  - GO15VENDOREXPERIMENT=1
```
eparis: now we don't need this part...   doesn't matter.

thockin: Shouldn't we be vendoring at the contrib/ level, not at subdirs?  I
*reallllly* don't want subdirs of contrib/ vendoring other subdirs of
contrib/.

eparis: https://github.com/kubernetes/contrib/pull/374 from @mikedanese moved us to this mechanism. I agree, it's not the best...

Then again, there are other efforts to break this repo apart further which makes this distinction less meaningful...

bprashanth: We moved from global contrib/ godeps to per project to make splitting them out easier, and because people complained that sub-project godep revs broke other projects (https://github.com/kubernetes/contrib/pull/374)

thockin: then we need to find a way to tell tools (godep, glide) to NOT vendor
k8s.io/contrib.  It's very surprising that a change to contrib/util would
not take effect until all the various subdirs re-vendor it.

eparis:
I'm ready to click `merge`. This PR is a good thing.

.travis.yml
eparis: One can not vendor it. golang will try to pick it up from the GOPATH. However golang can then get 'confused' about recursive dependencies...

This question has nothing to do with godep, it is entirely a question of golang and what you put in `vendor/`.

Ingress doesn't use other packages in `contrib/` so its a non-issue here.

PR #945, opened 2016-05-10
Change namespace.yaml path to /opt

### LINKED ISSUE #944: [ansible] namespace kube-system is not created
namespace `kube-system` is not created because kube-addons.sh is expecting to find it at /opt/namespace.yaml.

https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/addon-manager/kube-addons.sh#L68

gouyang:
Fixes #944

### DIFF
--- a/ansible/roles/kubernetes-addons/tasks/main.yml
+++ b/ansible/roles/kubernetes-addons/tasks/main.yml
@@ -14,7 +14,7 @@
 - name: Make sure the system services namespace exists
   get_url:
     url=https://raw.githubusercontent.com/kubernetes/kubernetes/master/{{ kube_addons_download_source_dir }}/namespace.yaml
-    dest="{{ kube_config_dir }}/addons/"
+    dest=/opt
     force=yes
     validate_certs=False
   environment:


eparis:
I think the script looking in `/opt` needs to be fixed to look somewhere else. In atomic you can't even create `/opt`...

gouyang:
There is /opt exists on Atomic host.

```
$ ls -lsh / | grep opt
   0 lrwxrwxrwx.   1 root root    7 May  4 23:44 opt -> var/opt
```

It need to adjust on one side. Actually I don't see why it changes the path.

gouyang:
Open kubernetes/kubernetes#25440 in kubernetes/kubernetes

mikedanese:
/opt is a path inside the kube-addon-manager container.

gouyang:
@mikedanese browsing files inhttps://github.com/kubernetes/kubernetes/blob/master/cluster/saltbase/salt/kube-addons/, /etc/kubernetes and /etc/kubernetes/addons is used there. is it /opt must be used for namespace.yaml?

jasonbrooks:
#797 is how I'm fixing this issue in my fork of the scripts.

PR #947, opened 2016-05-10
Update mungegithub to support vendor/

eparis:


eparis:
@thockin although I had to do so by hand, this does not vendor `k8s.io/contrib/test-utils/`

bgrant0607:
cc @fejta @thockin @david-mcmahon

thockin:
I'm really not keep on by-hand vendoring.  Some poor fool (probably me)
will have to figure out WTF you did.

eparis:
actually, I had to give up on the by-hand non-vendoring. And just vendor `contrib/test-utils`.  This is goLANG issue, not godep.

Back when we were using `Godep/_workspace` golang was resolving deps via GOPATH. First it would look in `Godep/_owrkspace` and then it would look other places in GOPATH.

But using `vendor/` golang will only resolve deps for this package with stuff in `vendor/` for this package. If i do not put `contrib/test-utils` in `vendor/` golang will find it. It then needs `google/glog`. golang will then satisfy glog from GOPATH ($GOPATH/src/github.com/google/glog). The problem comes in that now both `$GOPATH/src/github.com/google/glog` and `vendor/github.com/google/glog` both will create flags (these are different and distinct packages) which cause failure.

If you have `glog` in `contrib/test-utils/vendor` you get the same problem (trying to build both `vendor/github.com/google/glog` and `contrib/test-utils/vendor/github.com/google/glog` into the same binary.

But golang refuses to resolve the `contrib/test-utils/` usage of `glog` with stuff in mungegithub's `vendor`.

Basically this says we must either vendor it all. Or we must move vendor back to the top level.

thockin:
I still like the second option - top-level.

PR #958, opened 2016-05-11
[nginx-slim] update nginx and add module for digest authentication

### LINKED ISSUE #960: Nginx Ingress Controller - Safari 9 craps out with http2
I've literally wasted hours upon hours tracking this down. :-)

Using the nginx controller with SSL, Safari 9 will refuse to load any site. I'm sure this is a bug on Safari's part but I finally tracked it down to http2 in the listen 443 ssl line. I had to create a custom template to remove the http2 lines in order to get Safari (mobile or mac) to load our sites. None of the other browsers including older Safari browsers had an issue. I think http2 support was added in recent Safari version on the latest OS but obviously something is not right. Unfortunately, it wasn't just my local mac safari either as we had customers all over the world complain they couldn't access the sites with safari! 

Maybe this should be a setting? Or if there's a known workaround I couldn't figure out? 

https://github.com/kubernetes/contrib/blob/master/ingress/controllers/nginx/nginx.tmpl#L157

aledbf:
fixes #960

### DIFF
--- a/images/nginx-slim/Makefile
+++ b/images/nginx-slim/Makefile
@@ -1,7 +1,7 @@
 all: push
 
 # 0.0.0 shouldn't clobber any released builds
-TAG = 0.6
+TAG = 0.7
 PREFIX = gcr.io/google_containers/nginx-slim
 
 container:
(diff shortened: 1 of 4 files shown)


bprashanth:
LGTM, but not merging till I'm able to push (afk currently)

PR #966, opened 2016-05-12
Updated the etcd_download_url

stephenrlouie:
- Variablized the download url, only applicable for Ubuntu. Removed the v from the version number.
- Follows the same pattern we use in [flannel](https://github.com/kubernetes/contrib/blob/master/ansible/roles/flannel/defaults/main.yaml#L18-L19) , [kube](https://github.com/kubernetes/contrib/blob/master/ansible/roles/kubernetes/defaults/main.yml#L21), and [pypy] (https://github.com/kubernetes/contrib/blob/master/ansible/roles/pre-ansible/defaults/main.yml#L10-L11)

### DIFF
--- a/ansible/roles/etcd/defaults/main.yaml
+++ b/ansible/roles/etcd/defaults/main.yaml
@@ -1,7 +1,7 @@
 ---
 # etcd_source_type: package | github-release
 etcd_source_type: package
-etcd_version: v2.2.3
+etcd_version: 2.2.3
 
 etcd_client_port: 2379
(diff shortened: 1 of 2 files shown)


stephenrlouie:
Use #965 to resolve conflict since it has the {{ ansible_temp_dir }} instead of the hard coded path

eparis:
I merged 965, you'll want/need to rebase

stephenrlouie:
@eparis should be good now

PR #967, opened 2016-05-12
Created a top level playbooks dir and scripts dir

stephenrlouie:
**Dependency** - #933 because the `group_vars` directory needs to be in the _same directory_ as the `inventory` file or `deploy.yml`
- rename of the top level script from cluster to deploy. 
- Modeled after [openstack-ansible](https://github.com/openstack/openstack-ansible)
- top level play books will all live in the playbooks dir, called by scripts in the scripts dir. This organizes us well for new playbooks, such as uninstall, upgrade, etc.

Confirmed on Virtualbox & CoreOS & CentOS and BM


eparis:
woo hoo, rebase!

stephenrlouie:
@eparis rebased

ansible/scripts/ansible.cfg
```diff
@@ -0,0 +1,2 @@
+[defaults]
+roles_path = ../roles
```
ingvagabund: Does the ``ansible.cfg`` file has to be under ``scripts`` directory? Why not put the file into ``ansible`` directory and set  ``roles_path = roles``?

ansible/scripts/cluster.sh
```diff
@@ -14,6 +14,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-inventory=${INVENTORY:-inventory}
+inventory=${INVENTORY:-../inventory}
 
-ansible-playbook -i ${inventory} cluster.yml $@
+ansible-playbook -i ${inventory} ../playbooks/cluster.yml $@
```
ingvagabund: As there will be more scripts, I would rather create ``init.sh`` what will set all paths. Something like that:

```shell
$ cat init.sh
#!/bin/bash

INVENTORY_DIR="../inventories"
PLAYBOOKS_DIR="../playbooks"
```

Then:

```shell
$ cat deploy-cluster.sh
#!/bin/bash

. ./init.sh

inventory=${INVENTORY:-${INVENTORY_DIR}/inventory}
ansible-playbook -i ${inventory} ${PLAYBOOK_DIR}/deploy-cluster.yml $@
```

ansible/vagrant/Vagrantfile
```diff
@@ -173,7 +173,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     set_common_ansible_options(ansible)
 
     # This sets up both flannel and kube.
-    ansible.playbook = "../cluster.yml"
+    ansible.playbook = "../playbooks/cluster.yml"
```
ingvagabund: s/cluster.yml/deploy-cluster.yml/

ingvagabund:
Before this is merged, I need to update my CI. It it assuming the ``setup.sh`` is under ``ansible`` directory.

ansible/scripts/ansible.cfg
stephenrlouie: From [ansible] (http://docs.ansible.com/ansible/intro_configuration.html)
"Changes can be made and used in a configuration file which will be processed in the following order:"
```
* ANSIBLE_CONFIG (an environment variable)
* ansible.cfg (in the current directory)
* .ansible.cfg (in the home directory)
* /etc/ansible/ansible.cfg
```

I'm not certain what current directory is defined as. I'd guess it would be the `scripts/` directory in our case

ansible/scripts/cluster.sh
stephenrlouie: I'm fine with this, will update the PR

ansible/scripts/ansible.cfg
stephenrlouie: tested. its the scripts dir. Did not with in `ansible/` or `playbooks/`

ansible/scripts/cluster.sh
stephenrlouie: Creating an inventories dir gives us a problem when using group_vars. 

Copied from [ansible's site] (http://docs.ansible.com/ansible/intro_inventory.html)

> Tip: In Ansible 1.2 or later the group_vars/ and host_vars/ directories can exist in the playbook directory OR the inventory directory. If both paths exist, variables in the playbook directory will override variables set in the inventory directory

stephenrlouie:
I'm seeing successful builds in virtualbox.

Deploying on packet.net CentOS works. CoreOS on packet.net I'm seeing `Unable to connect to the server: dial tcp: lookup localhost: no such host`

ansible/scripts/cluster.sh
ingvagabund: I don't think we are ever going to create inventory directory with anything else than inventory examples. So we should be fine here.

ingvagabund:
@redhat-k8s-bot test please

ingvagabund:
```The conditional check 'source_type == "localBuild"' failed. The error was: error while evaluating conditional (source_type == "localBuild"): 'source_type' is undefined

The error appears to have been in '/home/jenkins/workspace/rhel-kubernetes-contrib-github-worker-2-2-runtest/contrib/ansible/roles/common/tasks/main.yml': line 53, column 3, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:

# enable cpu and memory accounting for systemd services by default
- name: Create systemd dropin directory
  ^ here```

Let me check the code

**Comment by ingvagabund** [comment]
Moving ``group_vars`` directory under scripts works. Well, that is just ugly.

**Comment by ingvagabund** [comment]
@stephenrlouie what do you think is better?

```bash
cd scripts; ./deploy-cluster.sh
```

or

```bash
./scripts/deploy-cluster.sh
```

?

ingvagabund:
Interested readings:
* http://stackoverflow.com/questions/25346796/how-can-i-change-the-path-of-group-vars-and-host-vars
* http://rosstuck.com/multistage-environments-with-ansible/
* https://github.com/ansible/ansible/issues/11238

And surprisingly, ``openstack-ansible`` has ``groups_vars`` inside of inventory:
* https://github.com/openstack/openstack-ansible/tree/master/playbooks/inventory

ansible/scripts/cluster.sh
ingvagabund: Taking my argument back. Reading more about variables, inventory directory seems the most appropriate for ``group_vars``. As long as we avoid ``group_vars`` inside playbook directory, we are then fine.

ingvagabund:
For both static and dynamic inventories, let's put the ``group_vars`` directory under /ansible/inventory directory with an example of ``local.ini``:

```
[masters]
localhost ansible_connection=local

[nodes]
localhost ansible_connection=local

[etcd]
localhost ansible_connection=local
```

so users can run the playbook right away. Later, we can provide ``dynamic_hosts.py`` script for dynamic inventory.

ingvagabund:
At the same time user should be able to run the scripts inside ``scripts`` directory from everywhere (providing its own ``ansible.cfg`` of course). So something like:

```
$ cat init.sh
#!/bin/bash

CUR_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"

INVENTORY_DIR="${CUR_DIR}/../"
PLAYBOOKS_DIR="${CUR_DIR}/../playbooks"
```

will not hurt. @stephenrlouie what do you think?

stephenrlouie:
@ingvagabund I think `cd scripts; ./deploy-cluster.sh` is better.

Its unfortunate that group_vars isn't configurable, but I guess under inventory is the best we can do at this point.

Let me update and see how it goes

ansible/playbooks/deploy-cluster.yml
```diff
@@ -91,11 +91,3 @@
   sudo: yes
   roles:
     - { role: contiv, contiv_role: netplugin, when: networking == 'contiv' }
-
-# install runtime dependencies
```
ingvagabund: This is needed for some e2e tests

ansible/README.md
```diff
@@ -14,7 +14,7 @@ can be real hardware, VMs, things in a public cloud, etc. Anything that you can
 
 ### Configure inventory
 
-Add the system information gathered above into the 'inventory' file, or create a new inventory file for the cluster.
+Add the system information gathered above into the 'inventory' file, or create a new inventory file for the cluster. Place the 'inventory' file into the `./inventories` directory.
```
ingvagabund: s/`./inventories`/`./inventory`/

ingvagabund:
@stephenrlouie up to the 2 nits, LGTM. After slight modification seems to work in my environment.

ingvagabund:
@jasonbrooks how do you call the ansible playbook? By running ``setup,sh``? With the PR ``setup.sh`` is about to change location and name. New script is ``scripts/deploy-cluster.sh``. How much will it affect your environment?

ingvagabund:
@redhat-k8s-bot test please

jasonbrooks:
@ingvagabund Sometimes I use setup.sh, more often I use the Vagrant file in the vagrant subdir. I'm sure I can adapt to the new arrangement.

ingvagabund:
@stephenrlouie per this article [1] it makes sense to have only one playbook for each role unless we decide to embed the control flow mechanism via variables. Which is kind of awkward. 

[1] http://stackoverflow.com/questions/25674649/ansible-in-a-playbook-filter-a-role-by-tags-without-passing-at-the-command-l?answertab=votes#tab-top

ansible/playbooks/deploy-cluster.yml
ingvagabund: Let's keep it there in this PR and move the code to a playbook in another PR.

ingvagabund:
``README.md`` has other parts that needs updating:

* all occurrences of ``setup.sh`` need to be changed to ``deploy-cluster.sh``
* right after ``Running the playbook`` headings one needs to mention the deployment script must run from within the ``scripts`` directory

stephenrlouie:
@ingvagabund you are talking about having one main.yml in each role?

I believe we need to leverage tags to get control flow. We would have to use some common language for tags, 'install', 'reconfigure', 'uninstall' and tag all appropriate tasks. The ordering of those tasks will important. I'm not certain if we can call, roles/etcd/install.yml or roles/etcd/uninstall.yml from the top level while getting the benefit of the entire role structure.

ingvagabund:
> @ingvagabund you are talking about having one main.yml in each role?

@stephenrlouie let's move the discussion to #1380

ingvagabund:
ping @stephenrlouie

ansible/README.md
```diff
@@ -14,23 +14,23 @@ can be real hardware, VMs, things in a public cloud, etc. Anything that you can
 
 ### Configure inventory
 
-Add the system information gathered above into the 'inventory' file, or create a new inventory file for the cluster.
+Add the system information gathered above into the 'inventory' file, or create a new inventory file for the cluster. Place the 'inventory' file into the `./inventory` directory.
```
sttts: So the file is called `inventory`?

Better: ... into a file called `inventory` in the `./inventory` directory, or create a new one for the cluster.

Also use back-ticks for verbatim names.

ansible/README.md
```diff
@@ -14,23 +14,23 @@ can be real hardware, VMs, things in a public cloud, etc. Anything that you can
...
 
-Look though all of the options in `group_vars/all.yml` and
+Look though all of the options in `inventory/group_vars/all.yml` and
 set the variables to reflect your needs. The options are described there
 in full detail.
 
 ## Running the playbook
 
-After going through the setup, run the setup script provided:
+After going through the setup, run the deploy-cluster script provided by going into the scripts directory:
```
sttts: In the PR description it's called `deploy`.

ansible/README.md
```diff
@@ -14,23 +14,23 @@ can be real hardware, VMs, things in a public cloud, etc. Anything that you can
...
-`$ ./setup.sh`
+`$ cd scripts/ && ./deploy-cluster.sh`
 
 You may override the inventory file by doing:
 
-`INVENTORY=myinventory ./setup.sh`
+`INVENTORY=myinventory ./deploy-cluster.sh`
 
 
 In general this will work on very recent Fedora, rawhide or F21.  Future work to
```
sttts: recent Fedora + F21? What does it mean? F22, F23 do not work?

ansible/README.md
```diff
@@ -42,20 +42,20 @@ You can just setup certain parts instead of doing it all.
...
 
-`$ ./setup.sh --tags=etcd`
+`$ ./deploy-cluster.sh --tags=etcd`
 
 #### Kubernetes master
 
-`$ ./setup.sh --tags=masters`
+`$ ./deploy-cluster.sh --tags=masters`
 
 #### kubernetes nodes
```
sttts: capital `K`

ansible/scripts/init.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+CUR_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
```
sttts: nit: no spaces after `$(`

ansible/scripts/deploy-cluster.sh
```diff
@@ -14,6 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-inventory=${INVENTORY:-inventory}
+. ./init.sh
 
-ansible-playbook -i ${inventory} deploy-cluster.yml $@
+inventory=${INVENTORY:-${INVENTORY_DIR}/inventory}
```
sttts: Do we really need these scripts? They are just hiding what ansible-playbook offers natively.

sttts:
In general I question the sense of the scripts (as they are now). They don't do anything than passing through ansible-playbook options.

ansible/scripts/deploy-cluster.sh
ingvagabund: Currently, the script does not do anything at all except it points to kubernetes default inventory and roles directories. In future, it can be altered and extended accordingly.

The main point of the scripts directory is to provide common ground for other scripts so there is no need for additional settings. This is a step towards re-usability of playbooks.

E.g. what I do in my CI is to copy dynamic inventory script under ${INVENTORY_DIR} so it is at the same directory as ``group_vars``.

ansible/README.md
ingvagabund: That is quite outdated. The README needs to be extended and polished. All three OS are already supported, including CoreOS. Subject of this PR is not to fix README, rather to establish new top directories structure. I plan to update all playbooks/documents and introduce new ones. Let's postpone the fix until we have improved framework including CI.

ansible/README.md
ingvagabund: One can ``deploy`` etcd cluster only, or just nodes. So we decide to rename so it is aligned with entire cluster deployment use case.

ingvagabund:
> In general I question the sense of the scripts (as they are now). They don't do anything than passing through ansible-playbook options.

It will be visible later. Each script will run a subset of playbooks. This discussion is more suitable for ``etcd``, ``kubernetes`` and other specific roles.

ingvagabund:
Resolved via #1527.

PR #978, opened 2016-05-13
[cluster-autoscaler] Add support for PodScheduled condition.

fgrzadkowski:
The new logic is the following:
1. Take time of the newest node becoming available
2. Reset status for `PodScheduled` condition to `Unknown` for all the pods with status `True` and reason `Unschedulable` with `LastTransitionTime` before timestamp from 1)
3. Proceed only for newer unschedulable pods.

Ref: #25553

### DIFF
--- a/cluster-autoscaler/cluster_autoscaler.go
+++ b/cluster-autoscaler/cluster_autoscaler.go
@@ -56,7 +56,7 @@ func main() {
 	}
 
 	kubeClient := kube_client.NewOrDie(kubeConfig)
-	unscheduledPodLister := NewUnscheduledPodLister(kubeClient)
+	unschedulablePodLister := NewUnschedulablePodLister(kubeClient)
 	nodeLister := NewNodeLister(kubeClient)
 
 	migConfigs := make([]*config.MigConfig, 0, len(migConfigFlag))
@@ -75,20 +75,6 @@ func main() {
 		select {
 		case <-time.After(time.Minute):
 			{
-				pods, err := unscheduledPodLister.List()
-				if err != nil {
-					glog.Errorf("Failed to list pods: %v", err)
-					continue
-				}
-				if len(pods) == 0 {
-					glog.V(1).Info("No unscheduled pods")
-					continue
-				}
-
-				for _, pod := range pods {
-					glog.V(1).Infof("Pod %s/%s is not scheduled", pod.Namespace, pod.Name)
-				}
-
 				nodes, err := nodeLister.List()
 				if err != nil {
 					glog.Errorf("Failed to list nodes: %v", err)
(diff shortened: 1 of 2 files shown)


fgrzadkowski:
FYI I haven't tested this manually e2e.

cluster-autoscaler/utils.go
```diff
@@ -25,26 +25,39 @@ import (
...
 	"k8s.io/kubernetes/pkg/client/cache"
 	kube_client "k8s.io/kubernetes/pkg/client/unversioned"
 	"k8s.io/kubernetes/pkg/fields"
 	"k8s.io/kubernetes/pkg/labels"
 	"k8s.io/kubernetes/plugin/pkg/scheduler/schedulercache"
+
+	"github.com/golang/glog"
 )
 
 // UnscheduledPodLister list unscheduled pods
```
piosz: typo: should be UnschedulablePodLister

cluster-autoscaler/utils.go
```diff
@@ -119,24 +132,52 @@ func NewNodeLister(kubeClient *kube_client.Client) *ReadyNodeLister {
...
+func GetNewestNodeTime(nodes []*kube_api.Node) time.Time {
+	var result time.Time
 	for _, node := range nodes {
-		if result == nil || node.CreationTimestamp.After(result.CreationTimestamp.Time) {
-			result = node
+		if node.CreationTimestamp.After(result) {
+			result = node.CreationTimestamp.Time
+		}
+	}
+	return result.Add(1 * time.Minute)
```
piosz: How about adding 1 minute somewhere else and making it constant? As for now the function GetNewestNodeTime doesn't return what its name declares.

cluster-autoscaler/cluster_autoscaler.go
```diff
@@ -104,25 +90,27 @@ func main() {
...
+					glog.Errorf("Failed to list unscheduled pods: %v", err)
 					continue
 				}
 
-				// TODO: Find better way to check if all pods were checked after the newest node
-				// was added.
-				if newestNode.CreationTimestamp.After(oldestSchedulingTrial.Add(-1 * time.Minute)) {
-					// Lets give scheduler another chance.
-					glog.V(1).Infof("One of the pods have not been tried after adding %s", newestNode.Name)
+				// We need to reset all pods that have been marked as unschedulable before newest
```
piosz: nit: AFAIU also pods marked as unschedulable in the same time the node became available are reset here, so to be super precise the comment should be updated.

cluster-autoscaler/cluster_autoscaler.go
```diff
@@ -104,25 +90,27 @@ func main() {
...
+				// node became available for the scheduler.
+				newestNodeTime := GetNewestNodeTime(nodes)
+				resetOldPods(kubeClient, pods, newestNodeTime);
+
+				// From now on we only care about unschedulable pods that were marked after the newest
+				// node became available for the scheduler.
+				pods = filterOldPods(pods, newestNodeTime)
```
piosz: I think this is not needed - once you updated the old pods there is no more unschedulable pods older than newestNodeTime (assuming the watch is updated).

cluster-autoscaler/cluster_autoscaler.go
```diff
@@ -104,25 +90,27 @@ func main() {
...
 
-				// TODO: Find better way to check if all pods were checked after the newest node
-				// was added.
-				if newestNode.CreationTimestamp.After(oldestSchedulingTrial.Add(-1 * time.Minute)) {
-					// Lets give scheduler another chance.
-					glog.V(1).Infof("One of the pods have not been tried after adding %s", newestNode.Name)
+				// We need to reset all pods that have been marked as unschedulable before newest
+				// node became available for the scheduler.
+				newestNodeTime := GetNewestNodeTime(nodes)
+				resetOldPods(kubeClient, pods, newestNodeTime);
```
piosz: Add comment somewhere that resetOldPods and filterOldPods should operate on disjoint sets of pods and in the future it would be great to refactor it.

cluster-autoscaler/utils.go
fgrzadkowski: As discussed offline I'd like to hide this 1 minute. Renaming the function to better describe what it does.

cluster-autoscaler/cluster_autoscaler.go
fgrzadkowski: I don't know how to make sure that the watch is up to date. I think it's easier to just filter them out.

cluster-autoscaler/cluster_autoscaler.go
fgrzadkowski: Done

cluster-autoscaler/cluster_autoscaler.go
fgrzadkowski: Done.

fgrzadkowski:
All comments addressed. PTAL.

cluster-autoscaler/utils.go
fgrzadkowski: Done.

piosz:
lgtm

PR #982, opened 2016-05-13
etcd refactor, more granular tasks for etcd. renamed top level playbooks

stephenrlouie:
The first of the roles to undergo a refactor to be more like [openstack-ansible] (https://github.com/openstack/openstack-ansible). Granular roles to enable more top level playbooks to be made such as `upgrade.yml` `uninstall.yml`, etc. This breaks down the few task files into more granular files. Renamed the top level playbooks to be more descriptive, once we have more playbooks at the top level we should use something like #967 to organize the top level directory. The current issue is roles are made for a first time deployment and that is it, the role structure did not leave much room for modification / reuse in other scenarios.

The end state is to deploy each role like a 'micro-service' that has granular control so one can spin up, spin down, individual roles without having damaging effects across the cluster. All the roles should be broken down in a similar way, and broken out into their own repos. The ansible folder could have a repo called kube-ansible, or [kube-deploy] (https://github.com/kubernetes/kube-deploy), and each role will have its own repo that will get pulled in by kube-deploy as specified in a [requirements file] (https://github.com/openstack/openstack-ansible/blob/master/ansible-role-requirements.yml).

- Validated on Virtualbox: CoreOS & CentOS

Could you all take a glance and provide feedback? I think contrib/ansible needs to start taking steps in this direction if we want it to be usable for production deployments and be plug-able for future deployment.

CC @danehans @eparis @adamschaub @ingvagabund


stephenrlouie:
CC @danehans

danehans:
@stephenrlouie can you ping the os-ansible community to understand why they decided to repeat the role name under tasks? I prefer dropping the role name within the tasks, similar to how Kolla names their tasks. Check out this [example](https://github.com/openstack/kolla/tree/master/ansible/roles/heat/tasks).

stephenrlouie:
@danehans 

OS-ansible prefixes the plays for readability when plays are passing by and clarity when a single role might have multiple items its handling like this [os-neutron repo] (https://github.com/openstack/openstack-ansible-os_neutron/tree/master/tasks). It has prefixes neutron, nuage, ovs, plumgrid.

I should rename `etcd-configure-firewalld.yml` to `firewalld-configure`  and iptables. That kind of convention makes more sense to me.

stephenrlouie:
@danehans updated plays names / structure with the second commit. Will squash if we see those changes as reasonable.

ansible/roles/etcd/tasks/install.yml
```diff
@@ -0,0 +1,18 @@
+---
+- name: Install etcd
+  action: "{{ ansible_pkg_mgr }}"
+  args:
+        name: etcd
+        state: latest
+  notify:
+    - restart etcd
+  when: not is_atomic and ansible_distribution != "Ubuntu" and not is_coreos
```
eparis: not saying you need to fix it here, but it seems like this test should be somehow just a `when: etcd_source_type == "package-manager"`     If you ever stumble into this code again.

eparis:
rebase needed.

ansible/roles/etcd/tasks/install.yml
stephenrlouie: ^ that broke on CentOS

ansible/deploy.sh
```diff
@@ -16,4 +16,4 @@
 
 inventory=${INVENTORY:-inventory}
 
-ansible-playbook -i ${inventory} cluster.yml $@
+ansible-playbook -i ${inventory} deploy.yml $@
```
ingvagabund: ``deploy-cluster.yml`` would be more appropriate. Although, in future the deployment scripts will get migrated under ``kubernetes/kube-deploy``, ansible scripts can provide additional functionality, e.g. collection of logs from cluster nodes.

ingvagabund:
As ``ansible/setup.sh`` is the main entry point and already in use, we should be careful when renaming it. I would suggest to rename any entry point or global variable once the repository gets migrated into ``kubernetes/kube-deploy`` repository.

ansible/roles/etcd/tasks/install.yml
ingvagabund: what got broken? ``etcd_source_type`` is not defined on CentOS? If so, please open a PR for it and reference it here. This needs to be fixed. I can imagine other values for ``etcd_source_type`` such as installation of the packages from the latest build in distribution builder not yet pushed to stable (e.g. ``etcd_source_type == "distribution_builder"``). On Fedora it makes sense to install etcd from rawhide build. I have already use case for this for docker.

ansible/roles/etcd/tasks/install.yml
```diff
@@ -0,0 +1,18 @@
...
+
+- name: Force etcd_source_type to github when packages are not available
+  set_fact:
+    etcd_source_type: "github-release"
```
ingvagabund: What about to generalize the concept and introduce ``package_source_type``? It could be set in ``pre-ansible`` role. Of course, ``etcd_source_type`` could be still defined to provide higher refinement.

ansible/roles/etcd/tasks/main.yml
```diff
@@ -1,52 +1,4 @@
...
-
-- include: coreos.yml
-  when: is_coreos
-
-- include: firewalld.yml
-  when: has_firewalld
-
-- include: iptables.yml
-  when: not has_firewalld and has_iptables
+- include: install.yml
```
ingvagabund: NL after install.yml

ansible/roles/etcd/tasks/main.yml
```diff
@@ -1,52 +1,4 @@
...
+- include: configure.yml
```
ingvagabund: NL after configure.yml

ansible/roles/etcd/tasks/start.yml
```diff
@@ -0,0 +1,5 @@
+---
+- include: etcd-start.yml
+  when: not is_coreos
```
ingvagabund: NL after ``is_coreos``

ansible/roles/etcd/tasks/upgrade.yml
```diff
@@ -0,0 +1,3 @@
+---
+
+- name: Test Update
```
ingvagabund: What is a purpose of this task?

ansible/roles/etcd/tasks/etcd2-start.yml
```diff
@@ -0,0 +1,7 @@
+---
+- name: CoreOS | Enable etcd2
+  service: name=etcd2 enabled=yes
```
ingvagabund: Make this conditional? Introduce ``enable_services: true|false`` in ``group_vars/all.yml``? When I run the ansible in CI environment, the cluster is destroyed right after it is used. Thus, no need to enable/disable any service. This will not save much time but it could decrease dependency on systemd.

ansible/roles/etcd/tasks/etcd2-configure.yml
```diff
@@ -11,10 +11,3 @@
 - name: CoreOS | Write etcd config file
   template: src=etcd.conf.j2 dest=/etc/etcd/etcd.conf
   notify: restart etcd2
```
ingvagabund: Based on this [1], restart etcd2 should not work as etcd2_started is not defined any more in this task block. From ``ansible/roles/etcd/handlers/main.yml +11``:

```
- name: restart etcd2
  service: name=etcd2 state=restarted
  when: etcd2_started.changed == false
```

[1] http://docs.ansible.com/ansible/playbooks_intro.html#handlers-running-operations-on-change

ingvagabund: From [2] citing: "Registered variables are valid on the host the remainder of the playbook run, which is the same as the lifetime of “facts” in Ansible."

So once defined they live forever. However, ``etcd2_started`` is defined in ``start.yml`` after it is used.

[2] http://docs.ansible.com/ansible/playbooks_variables.html#registered-variables

ansible/roles/etcd/tasks/etcd-configure.yml
```diff
@@ -0,0 +1,12 @@
+---
+- name: Write etcd config file
+  template: src=etcd.conf.j2 dest=/etc/etcd/etcd.conf
+  notify:
+    - restart etcd
```
ingvagabund: The same here. ``restart etcd`` handler has ``when`` condition with ``etcd_started.changed``. Again, the register variable is not defined at the time of execution of the handler.

ansible/roles/etcd/tasks/upgrade.yml
stephenrlouie: did not mean to commit. Was just there to toy around with potentially updating.

ansible/roles/etcd/tasks/etcd2-configure.yml
stephenrlouie: @ingvagabund I see your point. I guess the fix would be just to have `start.yml` called before `configure.yml` [here] (https://github.com/stephenrlouie/contrib/blob/etcd-ref/ansible/roles/etcd/tasks/main.yml#L4-L6)

I believe this would resolve the `etcd2.started` and `etcd.started` handler issue. But feels sloppy to me.

This should work as is since the configuration is put in place before we start it, so it shouldn't require a restart. A more pressing concern would be when you want to run a reconfigure play or an upgrade play. You need to make sure that handler gets called properly. I'm not sure what would be optimal, keep that condition on the handler and make sure its called on a reconfigure or upgrade, or remove that condition and always restart when that handler is called?

Advice on the best choice would be appreciated @danehans @eparis @ingvagabund

ansible/roles/etcd/tasks/install.yml
stephenrlouie: We use `source_type` [numerous times] (https://github.com/kubernetes/contrib/search?utf8=%E2%9C%93&q=source_type) in the project. I'll use that for now. If we need more granularity down the road each instance of `source_type` can be prepended by its service or package name.

ansible/roles/etcd/tasks/etcd2-start.yml
stephenrlouie: Added it to the `etcd/defaults/main.yml`, set to true. I don't think it should belong in the top level group_vars. But very easy to overwrite if its useful to optionally skip it.

ansible/roles/etcd/tasks/start.yml
stephenrlouie: addressed it

ansible/roles/etcd/tasks/etcd2-configure.yml
stephenrlouie: [this] (https://github.com/stephenrlouie/contrib/blob/etcd-ref/ansible/roles/etcd/tasks/install.yml#L7-L8) is also in the same place, etcd.started is not set when this is called

ingvagabund: ``restart etcd`` handler exists to make the deployment script idempotent.

* When running the role on vanilla cluster, etcd is not installed (thus not running). Running ``Start etcd`` task sets ``etcd_started.changed`` to ``true``. Before the ``etcd`` role ends, handlers are checked. As ``etcd_started.changed`` is not ``false``, ``restart etcd`` handler is not called.
* When running the role on a cluster with etcd installed and running, ``Start etcd`` task sets ``etcd_started.changed`` to ``false``. Thus, the ``restart etcd`` handler is called.

To run the ``restart etcd`` handler at all, ``Start etcd`` has to be run at the first place. So if you run ``install`` and/or ``etcd-configure`` without ``etcd-start``, the handler is not usable. At the same time you need to restart the etcd service if you change configuration file or update to newer version of etcd. So the ``restart etcd`` must be independent of ``Start etcd`` task.

Thus, instead of keeping ``restart etcd`` conditional, I suggest:

* make the ``restart etcd`` handler unconditional
* introduce new ``etcd_modified`` variable in ``defaults/main.yaml`` set to ``false`` by default
* if any of ``install-etcd`` and ``configure-etcd`` (and its equivalents) gets invoked, ``etcd_modified`` variable gets set to ``true``
* at the end of ``tasks/main.yml`` we will test if ``etcd_modified`` is ``true`` and ``etcd_started.changed`` is ``false``. If so, ``restart etcd`` gets notified and called.

ansible/roles/etcd/tasks/etcd2-start.yml
ingvagabund: Fair enough

ansible/roles/etcd/tasks/etcd2-configure.yml
stephenrlouie: Ended up making a new play file for `etcd_restart.yml` and `etcd2_restart.yml` since I am conditionally calling them once at the end of main and not using it like handlers should be used.

ingvagabund:
> The end state is to deploy each role like a 'micro-service' that has granular control so one can spin up, spin down, individual roles without having damaging effects across the cluster.

That is reasonable. Each testing scenario can require different roles. E.g. I don't need to have etcd installed if I want to run smoke test on "kubernetes way" installed docker (though it can already done).
Anyway, it brings me the idea of running tests on individual roles "just" to test the pipes (smoke tests in individual parts to discover regression before running the "full" test). Each smoke test can run on different OS.

> All the roles should be broken down in a similar way, and broken out into their own repos.

Not sure if it is worth introducing new repos (or submodules) at this point. First, I would like to have a list of possible other scenarios for reusing the roles before we decide if it has benefit. Can you more elaborate on benefits of introducing new repo for each role?

> The ansible folder could have a repo called kube-ansible, or kube-deploy, and each role will have its own repo that will get pulled in by kube-deploy as specified in a requirements file.

ansible/roles/etcd/handlers/main.yml
```diff
@@ -4,11 +4,9 @@
 
 - name: restart etcd
```
ingvagabund: No longer needed if the ``restart etcd`` is triggered at the end of ``tasks/main.yml``

ansible/roles/etcd/handlers/main.yml
```diff
@@ -4,11 +4,9 @@
...
   service: name=etcd state=restarted
-  when: etcd_started.changed == false
 
 - name: restart etcd2
```
ingvagabund: No longer needed if the ``restart etcd`` is triggered at the end of ``tasks/main.yml``

stephenrlouie: removed

ansible/roles/etcd/handlers/main.yml
stephenrlouie: removed

stephenrlouie:
@ingvagabund Regarding your comment. I was looking at large ansible projects and used [openstack] (https://github.com/openstack?utf8=%E2%9C%93&query=ansible) as a reference. 

I talked to some folks on the openstack-ansible IRC and they recently went through this explosion of repos because one repo was hard to manage. Multiple repo's made it more approachable for a dev who needs to learn the role, make changes, commit, merge into individual repo's as opposed to one large one. No one had examples of workflows that have been made better because of multiple repo's, but the separate repo's forces isolation on the roles and hopefully means you could cherry-pick a combination of roles that you want in your deployment.

stephenrlouie:
I'll squash commits once I get a LGTM.

ingvagabund:
@redhat-k8s-bot test please

ansible/roles/etcd/tasks/install.yml
```diff
@@ -0,0 +1,26 @@
...
+  action: "{{ ansible_pkg_mgr }}"
+  args:
+        name: etcd
+        state: latest
+  when: etcd_source_type == "package-manager" 
+ 
+- name: Set the etcd_modified fact to true 
+  set_fact:
+    etcd_modified: true
+  when: etcd_source_type == "package-manager"
```
ingvagabund: It makes sense to set the ``etcd_modified: true`` only for ``package-manager`` as installing etcd from github just downloads the binary without any service file. So really nothing to [re]start.

ingvagabund:
@stephenrlouie LGTM, please squash.

stephenrlouie:
rebased

ingvagabund:
@redhat-k8s-bot test please

ingvagabund:
Error not related to the PR.

ingvagabund:
> I talked to some folks on the openstack-ansible IRC and they recently went through this explosion of repos because one repo was hard to manage. Multiple repo's made it more approachable for a dev who needs to learn the role, make changes, commit, merge into individual repo's as opposed to one large one.

Atm, there are 13 roles. When the number of roles grows it make sense to do that. On the other hand, breaking the roles into separate repositories increase the chance that something gets broken.

> No one had examples of workflows that have been made better because of multiple repo's, but the separate repo's forces isolation on the roles and hopefully means you could cherry-pick a combination of roles that you want in your deployment.

That is a valid argument. There are already existing repositories. What about to see their current state and combine solutions?

stephenrlouie:
> Atm, there are 13 roles. When the number of roles grows it make sense to do that. On the other hand, breaking the roles into separate repositories increase the chance that something gets broken.

Agreed, not sure when the right time is. But I assume its easier to make the change sooner than later.

> That is a valid argument. There are already existing repositories. What about to see their current state and combine solutions?

They use a [requirements file] (https://github.com/openstack/openstack-ansible/blob/master/ansible-role-requirements.yml) and run a [script] (https://github.com/openstack/openstack-ansible/blob/master/scripts/bootstrap-ansible.sh) to fetch all the roles and versions of the roles that you want. Even though it is a chore to separate out the roles, I like how they can cherry-pick any roles and any versions of those roles that they want. Definitely way more plug-able and consistent as compared to contrib/ansible

danehans:
openstack-ansible uses a requirements [file](https://github.com/openstack/openstack-ansible/blob/master/ansible-role-requirements.yml) similar to godep/govendor to manage the dep roles. I think separate repos for roles and a "master" repo that contains plays and the dep logic to pull in the appropriate roles is the best approach.

PR #992, opened 2016-05-16
cluster-autoscaler: Fix slicing pods into old pods and new pods

fgrzadkowski:
Previously after updating pod status all pods had `LastTransitionTime` after the threshold (time of newest node becoming available).

Additionally fix our makefile so that it's not required to set `REGISTRY` for every rule.

https://github.com/kubernetes/kubernetes/issues/24404

@mwielgus @jszczepkowski @piosz

### DIFF
--- a/cluster-autoscaler/cluster_autoscaler.go
+++ b/cluster-autoscaler/cluster_autoscaler.go
@@ -90,7 +90,7 @@ func main() {
 					continue
 				}
 
-				pods, err := unschedulablePodLister.List()
+				allPods, err := unschedulablePodLister.List()
 				if err != nil {
 					glog.Errorf("Failed to list unscheduled pods: %v", err)
 					continue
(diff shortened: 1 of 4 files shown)


cluster-autoscaler/cluster_autoscaler.go
```diff
@@ -99,11 +99,12 @@ func main() {
...
 				// the newest node became available for the scheduler.
 				allNodesAvailableTime := GetAllNodesAvailableTime(nodes)
-				resetOldPods(kubeClient, pods, allNodesAvailableTime)
+				oldPods, newPods := SlicePodsByPodScheduledTime(pods, allNodesAvailableTime)
+				ResetPodScheduledCondition(kubeClient, oldPods)
 
 				// From now on we only care about unschedulable pods that were marked after the newest
 				// node became available for the scheduler.
-				pods = filterOldPods(pods, allNodesAvailableTime)
+				pods = newPods
```
piosz: Do you need to introduce this new variable?

cluster-autoscaler/utils.go
```diff
@@ -147,40 +147,39 @@ func GetAllNodesAvailableTime(nodes []*kube_api.Node) time.Time {
...
+		if condition != nil {
+			if condition.LastTransitionTime.After(threshold) {
+				newPods = append(newPods, pod)
+			} else {
+				oldPods = append(oldPods, pod)
+			}
 		}
 	}
-	return result
+	return
```
piosz: Remove

piosz:
LGTM, just a nit

cluster-autoscaler/utils.go
fgrzadkowski: As discussed offline this is required.

cluster-autoscaler/cluster_autoscaler.go
fgrzadkowski: Done.

fgrzadkowski:
Apart from addressing your comments I'm adding a separate commit that fixes our makefile.

piosz:
LGTM

PR #996, opened 2016-05-16
Wild guess at inability to find LGTM

lavalamp:


lavalamp:
I did a small amount of debugging-- I can say that the Etag github returns doesn't change if you e.g. delete a comment from the PR.

eparis:
Were you looking at the events api?  I doesn't show comments at all...  Labels, assignments, subscriptions...

lavalamp:
I was running this in --dry-run mode and printing out the HTTP req/resp.

lavalamp:
AFAICT, you still get a cache hit if you change something like e.g. deleting a comment.

eparis:
On which API endpoint is my question. The etag is changing when I add/remove the label from this PR at https://api.github.com/repos/kubernetes/contrib/issues/996/events

lavalamp:
Oh, it's the issues/PR#/events api.

lavalamp:
hm, maybe it's only particular PRs. I'm no longer confident this change will help...

eparis:
oh, i have little confidence this PR will help. But it'll definitely give us some more debugging info.

lavalamp:
OK, yeah, adding/removing a label causes the Etag to change, but deleting a comment does not, and I observe a false cache hit in that case.

PR #998, opened 2016-05-16
SQ Health for the last day, not since restart.

spxtr:
I also renamed the E2E tab to not be "Google Internal"

@eparis I pointed this at you because I don't know who else to ask to review mungegithub stuff. Feel free to reassign.

### DIFF
--- a/mungegithub/submit-queue/www/index.html
+++ b/mungegithub/submit-queue/www/index.html
@@ -142,11 +142,11 @@
 
           <md-tab md-on-select="cntl.selectTab('e2e')">
             <md-tab-label>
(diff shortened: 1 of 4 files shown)


mungegithub/submit-queue/www/index.html
```diff
@@ -142,11 +142,11 @@ <h4 class="md-body-2">{{pr.Reason}}</h4>
 
           <md-tab md-on-select="cntl.selectTab('e2e')">
             <md-tab-label>
-              <span ng-class="{'redTab':cntl.failedBuild}">Google Internal E2E</span>
+              <span ng-class="{'redTab':cntl.failedBuild}">E2E Tests</span>
```
eparis: They are still "Google", even if not 'internal'.....

PR #1006, opened 2016-05-17
detect unneeded retests

lavalamp:
tested! works!

### DIFF
--- a/mungegithub/github/github.go
+++ b/mungegithub/github/github.go
@@ -678,9 +678,9 @@ func (obj *MungeObject) RemoveLabel(label string) error {
 	return nil
 }
 
-// Returns the head SHA and the base ref, so that you can get the base's sha in
-// a second step. Purpose: if head and base SHA are the same across two merge
-// attempts, we don't need to rerun tests.
+// GetHeadAndBase returns the head SHA and the base ref, so that you can get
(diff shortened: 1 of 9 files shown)


lavalamp:
@eparis ready for a review!

lavalamp:
I don't have much patience-- I built & pushed this just now... I promise to fix anything @eparis wants changed.

eparis:
If its already live we should just merge it, so we can at least find the git commit! (I totally understand, I want to do the same ALL THE TIME)

mungegithub/github/github.go
```diff
@@ -678,6 +678,39 @@ func (obj *MungeObject) RemoveLabel(label string) error {
 	return nil
 }
 
+// Returns the head SHA and the base ref, so that you can get the base's sha in
```
eparis: golint requires this comment to start with `// GetHeadAndBase .....`

lavalamp:
Fixed.

eparis:
let me write something to show 'RetestsAvoided' in the webUI....

lavalamp:
Sure, or we can wait to see if it actually ever avoids any...

lavalamp:
This got hit for the first time live! Every time this happens we've saved ~30 minutes.

screenshot 2016-05-17 at 10 33 19 pm

PR #1008, opened 2016-05-18
Update link in documentaion

itayariel:


itayariel:
I have signed the agreement

PR #1009, opened 2016-05-18
Cluster-autoscaler: PredicateChecker and better nodeInfo handling for ScaleDown

mwielgus:
cc: @piosz @fgrzadkowski


piosz:
LGTM

PR #1010, opened 2016-05-18
Cluster-autoscaler: Fast drain impl that doesn't talk to api server

mwielgus:
cc: @piosz @fgrzadkowski 

Depends on #1009

### DIFF
--- a/cluster-autoscaler/simulator/drain.go
+++ b/cluster-autoscaler/simulator/drain.go
@@ -1,0 +1,79 @@
+/*
+Copyright 2015 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


cluster-autoscaler/simulator/drain.go
```diff
@@ -0,0 +1,81 @@
...
+
+// FastGetPodsToMove returns a list of pods that should be moved elsewhere if the node
+// is drained. Raises error if there is an unreplicated pod and force option was not specified.
+// Based on kubectl drain code. It makes an assumption that RC, DS, Jobs and RS were deleted
+// along with their pods (no abandoned pods with dangling created-by annotation). Usefull for fast
+// checks.
+func FastGetPodsToMove(nodeInfo schedulercache.NodeInfo, force bool, decoder runtime.Decoder) ([]*api.Pod, error) {
+	pods := make([]*api.Pod, 0)
+	unreplicatedPodNames := []string{}
+	daemonSetPodNames := []string{}
```
piosz: Why do you need a list of daemon set?

piosz:
LGTM, one comment.

cluster-autoscaler/simulator/drain.go
mwielgus: Removed. It remained after the original drain.

PR #1011, opened 2016-05-18
ignore failures that don't happen twice in a row

lavalamp:
Actually parse the junit result, allow merging to continue as long as we were able to get the list of failed tests and it doesn't intersect with the same list for the previous run.

### DIFF
--- a/mungegithub/mungers/e2e/e2e.go
+++ b/mungegithub/mungers/e2e/e2e.go
@@ -17,22 +17,23 @@ limitations under the License.
 package e2e
 
 import (
-	"bufio"
+	"encoding/xml"
 	"fmt"
+	"io"
 	"net/http"
 	"strconv"
-	"strings"
 	"sync"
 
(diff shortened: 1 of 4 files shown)


lavalamp:
Inspired by @david-mcmahon

eparis:
@gmarek

lavalamp:
This also stops reading from jenkins and solely relies on GCS.

lavalamp:
@eparis any objections to turning this on? merge queue needs all the help it can get...

eparis:
Other than I haven't really reviewed it, my only objection would be that it effectively doubles the time required to notice a bad commit. And it just lets us paper over the flakes.

But given the tradeoffs, I'm good with it.

lavalamp:
@eparis gave an OK over chat. Going to merge & push.

lavalamp:
I broke the UI a little bit, but I guess there's no fear of not noticing (if you're looking at the dashboard)

gmarek:
LGTM. We probably can add the same logic to WeakStable.

PR #1012, opened 2016-05-18
Fix bugs

lavalamp:
Found a bug that will consume tokens until it triggers an "abuse prevention mechanism".


lavalamp:
Self-merging & pushing because this is eating our tokens.

PR #1013, opened 2016-05-18
Add a /metrics endpoint to submit-queue for external monitoring

rmmh:
/cc @eparis

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1013)

### DIFF
--- a/mungegithub/mungers/submit-queue.go
+++ b/mungegithub/mungers/submit-queue.go
@@ -341,6 +341,7 @@ func (sq *SubmitQueue) internalInitialize(config *github.Config, features *featu
 		handle("/merge-info", http.HandlerFunc(sq.serveMergeInfo))
 		handle("/priority-info", http.HandlerFunc(sq.servePriorityInfo))
 		handle("/health", makeHandler(sq.getHealth))
+		handle("/metrics", makeHandler(sq.getMetrics))
 		handle("/sq-stats", makeHandler(sq.getSQStats))
 		config.ServeDebugStats("/stats")
 		go http.ListenAndServe(config.Address, nil)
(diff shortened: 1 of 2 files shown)


mungegithub/mungers/submit-queue.go
```diff
@@ -647,6 +657,34 @@ func (sq *SubmitQueue) getQueueStatus() []byte {
...
+	sq.Lock()
+	defer sq.Unlock()
+
+	status := sq.e2e.GetBuildStatus()
+	blocked := false
+	for _, state := range status {
+		if state.Status != "Stable" {
+			blocked = true
+		}
+	}
```
eparis: I think this is wrong. You really want some combination of 
`sq.e2e.GCSBasedStable()` and `sq.e2e.GCSWeakStable()`
I believe that after @lavalamp changes this is a bit more nuanced.  Ignore `sq.e2e.Stable()` as that's dead and needs cleaned up...

mungegithub/mungers/submit-queue.go
```diff
@@ -647,6 +657,34 @@ func (sq *SubmitQueue) getQueueStatus() []byte {
...
+	type gauges struct {
+		PRCount     int
+		QueueLength int
+		Blocked     bool
+	}
+	return sq.marshal(struct {
+		Gauges gauges
+	}{Gauges: gauges{
+		PRCount:     len(sq.prStatus),
+		QueueLength: len(sq.e2e.GetBuildStatus()),
```
eparis: This doesn't make sense to me...  did you mean `len(sq.githubE2EQueue)` ?

mungegithub/mungers/submit-queue.go
rmmh: `sq.e2e.GCSWeakStable` does synchronous requests to GCS buckets, which is not what we want in this simple health check. Checking the build status like this replicates the return value of `GCSWeakStable` using the cached information.

eparis: I get the sync issue/question but GCSStable() is more complex than a single pass/fail, it only blocks if the same test fails twice in a row.

eparis: My question is, should this be more quickly/cleanly exposed in e2e/, rather than putting your type of brains into submit-queue.go

lavalamp: When my other PR is merged, there will be caching for GCSBasedStable.

Don't rewrite this logic. Cache it from the place that it's used and just repeat the cached value here. ...or wait. Use promethieus. Don't make your own metrics system. Just kick the metric from where the computation is done.

mungegithub/mungers/submit-queue.go
```diff
@@ -647,6 +657,34 @@ func (sq *SubmitQueue) getQueueStatus() []byte {
 	return sq.marshal(status)
 }
 
+// getMetrics returns a json representation of important stats about the
+// submit queue's health, for external monitoring/graphing/alerting.
+func (sq *SubmitQueue) getMetrics() []byte {
```
lavalamp: Do not do this. We use Promethieus in Kubernetes for metric publishing. Search in the main repository, there should be lots of examples. :)

rmmh: I'll modify this to use Prometheus's metrics exporting format.

eparis:
Is this coming back?

rmmh:
Yes, I'm working on it.

test-foxish:
recomputing cla status...

rmmh:
I'm going to close this until we figure out whether we're using Prometheus.

PR #1018, opened 2016-05-19
virtualbox as the default provider

stephenrlouie:
- My assumption is most people don't have an openstack cluster at their disposal to work with, so why is it the default, we should order these by which is the most likely vagrant deployment platform first and so on... If it's openstack -> libvirt -> virtualbox then we can close here, just wanted to bring it into question.


eparis:
i'll merge. but openstack should be below even libvirt.

PR #1022, opened 2016-05-20
unit tests take over an hour

lavalamp:
Submit queue just gave up on a good PR because of this.


lavalamp:
Self-merging.

lavalamp:
I will try to push this tonight, it got hit again.

lavalamp:
should be pushed now.

eparis:
@ixdy how do we fix the real problem here?

eparis:
I should say, I found the PR daniel filed above...

PR #1023, opened 2016-05-20
Assign issues based on the "fixes #<n>" in the PR body

asalkeld:
This is initially from a discussion on kubernetes-dev about the struggle to set the assignee
field in issues. It does not solve the problem of setting this field as the developer intends to
work on an issue, but at least fills it in once the PR is created. Ideas welcome...

### DIFF
--- a/mungegithub/mungers/assign-fixes.go
+++ b/mungegithub/mungers/assign-fixes.go
@@ -1,0 +1,95 @@
+/*
+Copyright 2016 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package mungers
+
+import (
+	"k8s.io/contrib/mungegithub/features"
+	"k8s.io/contrib/mungegithub/github"
+
+	"github.com/golang/glog"
+	"github.com/spf13/cobra"
+)
+
+// AssignFixesMunger will assign issues to users based on the config file
+// provided by --assignfixes-config.
+type AssignFixesMunger struct {
+	config              *github.Config
+	features            *features.Features
+	assignfixesReassign bool
+}
+
+func init() {
+	assignfixes := &AssignFixesMunger{}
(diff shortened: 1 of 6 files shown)


mungegithub/mungers/assign-fixes.go
```diff
@@ -0,0 +1,95 @@
...
+	if !obj.IsPR() {
+		return
+	}
+	// we need the PR for the "user" (creator of the PR not the assignee)
+	pr, err := obj.GetPR()
+	if err != nil {
+		glog.Infof("Couldn't get PR %v", obj.Issue.Number)
+		return
+	}
+	prOwner := describeUser(pr.User)
```
eparis: Does this sort of thing belong in github.go?

mungegithub/mungers/assign-fixes.go
```diff
@@ -0,0 +1,95 @@
...
+	a.config = config
+	return nil
+}
+
+// EachLoop is called at the start of every munge loop
+func (a *AssignFixesMunger) EachLoop() error { return nil }
+
+// AddFlags will add any request flags to the cobra `cmd`
+func (a *AssignFixesMunger) AddFlags(cmd *cobra.Command, config *github.Config) {
+	cmd.Flags().BoolVar(&a.assignfixesReassign, "fixes-issue-reassign", false, "Assign fixes Issues even if they're already assigned")
```
eparis: Do we need this?  We ahve it in the blunderbuss, but I don't remember it ever being used...

mungegithub/mungers/assign-fixes.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+	for _, fixesNum := range issuesFixed {
+		// "issue" is the issue referenced by the "fixes #<num>"
+		issueObj, err := a.config.GetObject(fixesNum)
+		if err != nil {
+			glog.Infof("Couldn't get issue %v", fixesNum)
+			continue
+		}
+		issue := issueObj.Issue
+		if !a.assignfixesReassign && issue.Assignee != nil {
```
eparis: And just bail here if it is assigned?

mungegithub/mungers/assign-fixes.go
asalkeld: Honestly not sure. It seemed like something that we *may* want to experiment with if people didn't like the re-assign. Happy to remove if you wish.

mungegithub/mungers/assign-fixes.go
asalkeld: So if you have "fixes 4 and fixes 5" and 4 is assigned, I wanted to still assign 5.

eparis:
Merged, although to turn this on you'll need to send a PR to update https://github.com/kubernetes/contrib/blob/master/mungegithub/submit-queue/deployment.yaml

PR #1048, opened 2016-05-24
cluster-autoscaler: Use all predicates in simulator

fgrzadkowski:
Ref https://github.com/kubernetes/kubernetes/issues/24404

@davidopp @piosz @mwielgus


mwielgus:
LGTM

PR #1054, opened 2016-05-24
[nginx-ingress-controller] Add ssl certificate checksum to template

### LINKED ISSUE #800: Nginx Ingress Controller - Config file not updated when a secret is updated
It seems that updating a secret does not trigger the config file to be updated. I have a script that is automating let's encrypt. But after the certificate is fetched and the Secret resource updated, the nginx configuration is not rebuilt (so hosts waiting on the certificate are updated). Is this intentional?

aledbf:
fixes #800

### DIFF
--- a/ingress/controllers/nginx/nginx.tmpl
+++ b/ingress/controllers/nginx/nginx.tmpl
@@ -155,6 +155,8 @@ http {
     server {
         listen 80{{ if $cfg.useProxyProtocol }} proxy_protocol{{ end }};
         {{ if $server.SSL }}listen 443{{ if $cfg.useProxyProtocol }} proxy_protocol{{ end }} ssl http2;
+        {{/* comment PEM sha is required to detect changes in the generated configuration and force a reload */}}
+        # PEM sha: {{ $server.SSLPemChecksum }}        
         ssl_certificate {{ $server.SSLCertificate }};
         ssl_certificate_key {{ $server.SSLCertificateKey }};{{ end }}
 
--- a/ingress/controllers/nginx/nginx/ssl.go
+++ b/ingress/controllers/nginx/nginx/ssl.go
@@ -17,7 +17,9 @@ limitations under the License.
 package nginx
 
 import (
+	"crypto/sha1"
 	"crypto/x509"
+	"encoding/hex"
 	"encoding/pem"
 	"fmt"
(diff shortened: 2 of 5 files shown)


simonswine:
cherry-picked with #1063 and works perfectly in my cluster

LGTM

ingress/controllers/nginx/nginx.tmpl
```diff
@@ -155,6 +155,7 @@ http {
     server {
         listen 80{{ if $cfg.useProxyProtocol }} proxy_protocol{{ end }};
         {{ if $server.SSL }}listen 443{{ if $cfg.useProxyProtocol }} proxy_protocol{{ end }} ssl http2;
+        # PEM sha: {{ $server.SSLPemChecksum }}
```
bprashanth: why comment this out?

ingress/controllers/nginx/nginx/ssl.go
```diff
@@ -26,28 +28,52 @@ import (
...
+		PemFileName:  pemFileName,
+		PemSHA:       nginx.pemSHA1(pemFileName),
+		CN:           cn,
+	}, nil
 }
 
 // CheckSSLCertificate checks if the certificate and key file are valid
 // returning the result of the validation and the list of hostnames
 // contained in the common name/s
-func (nginx *Manager) CheckSSLCertificate(pemFileName string) ([]string, error) {
```
bprashanth: please fix comment to match method name

ingress/controllers/nginx/nginx.tmpl
aledbf: Is just a comment. If the secret changes this comment will trigger the reload (diff with the current running cfg)

ingress/controllers/nginx/nginx/ssl.go
aledbf: done

ingress/controllers/nginx/nginx.tmpl
bprashanth: oh, hmm. suggest adding a commet about the significance of that comment.

aledbf: done

bprashanth:
Thanks, LGTM

PR #1056, opened 2016-05-24
make weakly stable builds disableable

mikedanese:
cc @eparis @lavalamp

### DIFF
--- a/mungegithub/submit-queue/deployment.yaml
+++ b/mungegithub/submit-queue/deployment.yaml
@@ -16,6 +16,7 @@ spec:
         - --token-file=/etc/secret-volume/token
         - --pr-mungers=blunderbuss,lgtm-after-commit,cherrypick-auto-approve,label-unapproved-picks,needs-rebase,ok-to-test,rebuild-request,path-label,size,stale-pending-ci,stale-green-ci,block-path,release-note-label,comment-deleter,submit-queue
         - --dry-run=true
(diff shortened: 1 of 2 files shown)


eparis:
should we do the same for the 'strong' stable builds?  (less likely to be 'empty')

eparis:
but this UI for the next person who comes along to have to remember one place to set the weak stable and another place to set the strong stable jobs is off-putting.

eparis:
also, FYI, I looked at doing something like this a while ago, but never finished doing it...
https://github.com/kubernetes/contrib/compare/master...eparis:zero-empty-string-slices

lavalamp:
I'm confused, doesn't just passing an empty string disable it? or a ","?

mungegithub/submit-queue/deployment.yaml
```diff
@@ -16,6 +16,7 @@ spec:
         - --token-file=/etc/secret-volume/token
         - --pr-mungers=blunderbuss,lgtm-after-commit,cherrypick-auto-approve,label-unapproved-picks,needs-rebase,ok-to-test,rebuild-request,path-label,size,stale-pending-ci,stale-green-ci,block-path,release-note-label,comment-deleter,submit-queue
```
lavalamp: Oh, I'm a bad person and I changed the list in prod without checking it in. If we're keeping this file up to date, can you add issue-cache and flake-manager?

lavalamp:
OK. Merging.

PR #1063, opened 2016-05-25
ingress: nginx controller watches referenced tls secrets

simonswine:
This enables the controller to reload certificates on changes in the secrets.

We are maintaining a list of referenced certificates in secrMetadata to only reload/sync the controller when something in referenced certificates changes.

### DIFF
--- a/ingress/controllers/nginx/controller.go
+++ b/ingress/controllers/nginx/controller.go
@@ -90,10 +90,12 @@ type loadBalancerController struct {
 	ingController  *framework.Controller
 	endpController *framework.Controller
 	svcController  *framework.Controller
+	secrController *framework.Controller
 	mapController  *framework.Controller
 	ingLister      StoreToIngressLister
 	svcLister      cache.StoreToServiceLister
 	endpLister     cache.StoreToEndpointsLister
+	secrLister     StoreToSecretsLister
 	mapLister      StoreToConfigmapLister
 	nginx          *nginx.Manager
 	podInfo        *podInfo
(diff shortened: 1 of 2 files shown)


simonswine:
@alanhartless @aledbf 

#800 #1054

ingress/controllers/nginx/controller.go
```diff
@@ -747,12 +802,13 @@ func (lbc *loadBalancerController) createServers(data []interface{}) map[string]
...
 	pems := make(map[string]string)
+	secrMetadata := make(map[string]bool)
 
 	for _, ingIf := range data {
 		ing := ingIf.(*extensions.Ingress)
-
 		for _, tls := range ing.Spec.TLS {
 			secretName := tls.SecretName
+			secrMetadata[fmt.Sprintf("%s/%s", ing.Namespace, secretName)] = true
 			secret, err := lbc.client.Secrets(ing.Namespace).Get(secretName)
```
aledbf: @simonswine can you change this to use `secrLister` instead of `api.Client`?

simonswine: @aledbf thanks for that hint, now using the `secrLister`

aledbf:
LGTM

ingress/controllers/nginx/controller.go
```diff
@@ -797,9 +860,24 @@ func (lbc *loadBalancerController) getPemsFromIngress(data []interface{}) map[st
 		}
 	}
 
+	lbc.secrMetadataLock.Lock()
```
bprashanth: do we really need to maintain this map/locks or can we just relist all ingresses on secret update and break on first name match ?

simonswine: @bprashanth I had the feeling that this would be too expensive, but I can modify the code accordingly

bprashanth: I'm fine if you think it's required. I usually try to avoid locking unless I absolutely need to. If you just need a locked map, suggest using cache.Store (or cache.ThreadSafeStore as the need may be:https://github.com/kubernetes/kubernetes/blob/master/pkg/client/cache/thread_safe_store.go).

bprashanth:
I see you already modified it per my last comment, @simonswine thanks! LGTM, will merge on green

simonswine:
@bprashanth, seems like it has happened in the same minute :)

bprashanth:
Merged, but no new image. If you need this in an image please send a pr with the version bump and I can push it out.

PR #1064, opened 2016-05-25
Enable CPU and Memory accounting for systemd services

ingvagabund:
Enable CPU and Memory accounting for systemd services by default when building from local sources.
For distribution packages (rpms, deb, etc.) let the installation mechanism to take care of accounting.

Quoting:
Fix system container detection in kubelet on systemd.

This fixed environments where CPU and Memory Accounting were not enabled on the unit
that launched the kubelet or docker from reporting the root cgroup when
monitoring usage stats for those components.

From: https://github.com/kubernetes/kubernetes/pull/25982

### DIFF
--- a/ansible/cluster.yml
+++ b/ansible/cluster.yml
@@ -18,6 +18,7 @@
   tags:
     - etcd
 
+# Install docker
 - hosts: all
   sudo: yes
   roles:
--- a/ansible/roles/common/files/kubernetes-accounting.conf
+++ b/ansible/roles/common/files/kubernetes-accounting.conf
@@ -1,0 +1,3 @@
+[Manager]
+DefaultCPUAccounting=yes
+DefaultMemoryAccounting=yes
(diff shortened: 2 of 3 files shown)


ingvagabund:
Or maybe enable the accounting for every installation way. Or make it configurable and introduce ``enable_systemd_accounting`` var.

ingvagabund:
Tested on cluster with RHEL images:

**On nodes**:

```
# cat /proc/$(pidof docker)/cgroup
10:perf_event:/
9:hugetlb:/
8:net_cls:/
7:blkio:/system.slice/docker.service
6:memory:/system.slice/docker.service
5:freezer:/
4:cpuacct,cpu:/system.slice/docker.service
3:devices:/system.slice/docker.service
2:cpuset:/
1:name=systemd:/system.slice/docker.service
```

**On master**:

```
# cat /proc/$(pidof docker)/cgroup
10:perf_event:/
9:cpuacct,cpu:/
8:freezer:/
7:net_cls:/
6:memory:/
5:hugetlb:/
4:devices:/system.slice
3:blkio:/
2:cpuset:/
1:name=systemd:/system.slice/docker.service
```

Issue #1045, opened 2016-05-24
merge bot doesn't read the junit XML files that the unit tests make


lavalamp:
...because they are named differently from the e2e tests.

I don't care if we change the unit test output or change the merge bot. @ixdy opinions?

lavalamp:
Actually I think we should fix the unit test output so it's consistent.

This is preventing the merge queue from filing unit test flakes.

ixdy:
My preference is that mergebot looks for anything named `junit_*.xml`.

I'm also curious how mergebot will handle unit tests, since we run each test there multiple times with slightly different options.

lavalamp:
I don't think merge bot can actually get a directory listing, now that I
think about it. Maybe it can?

lavalamp:
Chatted w/ @freehan-- he's going to figure out how to do a directory listing.

lavalamp:
Hm, it looks like merge bot still isn't filing issues about the unit test suite-- I suspect Jenkins is failing harder than it needs to? Example: http://kubekins.dls.corp.google.com/view/Critical%20Builds/job/kubernetes-test-go/12970/

@freehan let me know if you have time to figure out what jenkins is doing wrong. :)

lavalamp:
Or maybe it's as easy as merge queue needing to look for junit files inside completely failed builds?

freehan:
Investigating...

freehan:
Okay. It is due to the Junit xml parsing. It could not parse junit xml with multiple testsuites and fail silently. 

Will submit a fix soon.

ixdy:
Per http://llg.cubic.org/docs/junit/, the `<testsuites>` tag can be omitted if there's only a single `<testsuite>`. And it looks like `ginkgo` takes that option.

@spxtr @rmmh do our various test-infra tools handle this correctly?

freehan:
Hope there is no more corner cases. It is very annoying that golang xml parser fail silently. This may mask lots of problem.

rmmh:
Gubernator and the testgrid collector handle either <testsuite> or <testsuites>. Test-history only handles <testsuite>, which makes it skip the unit tests.

lavalamp:
Thanks @freehan!

spxtr:
Actually testgrid handles `<testsuites>` properly, but I still have the test-go tab turned off because it's pretty slow to render and it's tricky to make the tests group properly. Might be worth turning it back on again.

PR #1069, opened 2016-05-25
add util to list files in bucket with matching prefix

freehan:
First part of #1045

### DIFF
--- a/test-utils/utils/utils.go
+++ b/test-utils/utils/utils.go
@@ -22,27 +22,37 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"net/url"
+	"path"
 	"time"
 
 	"github.com/golang/glog"
 )
 
 const (
-	// GoogleBucketURL is an url go GCS bucket with results of Google-run tests.
-	GoogleBucketURL = "https://storage.googleapis.com/kubernetes-jenkins/logs"
-	successString   = "SUCCESS"
-	retries         = 3
-	retryWait       = 100 * time.Millisecond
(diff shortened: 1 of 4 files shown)


test-utils/utils/utils.go
```diff
@@ -160,3 +173,43 @@ func (u *Utils) CheckFinishedStatus(job string, buildNumber int) (bool, error) {
 	}
 	return result.Result == successString, nil
 }
+
+// ListFilesInBuild takes build info and list all file names with matching prefix
+// The returned file name included the complete path from bucket root
+func (u *Utils) ListFilesInBuild(job string, buildNumber int, prefix string) ([]string, error) {
+	ret, err := u.ListFilesWithPrefix(fmt.Sprintf("%v/%v/%v/%v", u.directory, job, buildNumber, prefix))
```
lavalamp: This file gets it wrong all over, but for this particular function path.Join seems clearly superior?

test-utils/utils/utils.go
```diff
@@ -61,16 +66,24 @@ func (u *Utils) getResponseWithRetry(url string) (*http.Response, error) {
 	return response, nil
 }
 
+func (u *Utils) GetGCSBucketURL() string {
```
lavalamp: Can we name these by their purpose? eg GetGCSFileURL and GetGCSDirectoryURL?

lavalamp: Also add a godoc comment since it won't pass vet otherwise.

lavalamp:
Looks good-- just a couple nits

PR #1078, opened 2016-05-26
add support for test util

freehan:


test-utils/utils/utils.go
```diff
@@ -73,11 +79,19 @@ func (u *Utils) getResponseWithRetry(url string) (*http.Response, error) {
 
 // GetGCSDirectoryURL returns the url of the bucket directory
 func (u *Utils) GetGCSDirectoryURL() string {
+	// return overrideURL if specified
+	if u.overrideURL != "" {
+		return u.overrideURL
```
lavalamp: You don't want to test having the bucket & directory appended?

test-utils/utils/utils.go
```diff
@@ -46,15 +46,21 @@ const (
...
+	overrideURL string
 }
 
 // NewUtils returnes new Utils struct for a given bucket name and subdirectory
 func NewUtils(bucket, directory string) *Utils {
 	return &Utils{bucket: bucket, directory: directory}
 }
 
+// NewTestUtils returnes new Utils struct for a given url pointing to a file server
+func NewTestUtils(url string) *Utils {
```
lavalamp: Not super thrilled about having test code mixed in. I guess I can live with it?

test-utils/utils/utils.go
lavalamp: Explained IRL. This is OK.

lavalamp:
LGTM

PR #1079, opened 2016-05-26
WIP: [nginx-ingress-controller] Add support for rewrite

### LINKED ISSUE #860: Nginx Ingress Controller - the /foo redirection is not working ?
Hi guys,

I work with the Ingress / Nginx Ingress Controller, but i can't get the /foo redirection working.

I deployed on a Kubernetes cluster a daemonset of Nginx ingress controller, default backend is the http-default-backend example.

This is my ingress object : 

```
# apimobile Ingress
kind: Ingress
metadata:
  name: apimobile-map
spec:
  rules:
  - host: azure.coursierprive.com
    http:
      paths:
      - path: /apimobile
        backend:
          serviceName: application-apimobile
          servicePort: 80
  - host: kube-20160428-120550-loadbalancer.northeurope.cloudapp.azure.com
    http:
      paths:
      - path: /
        backend:
          serviceName: application-apimobile
          servicePort: 80

```

My second rule is working but not the first one, if i try to reach my api on : azure.coursierprive.com/apimobile 

I face the : backend - 404 

I tried with Jenkins portal instead of my API but i face the same problem.

aledbf:
fixes #860

### DIFF
--- a/ingress/controllers/nginx/examples/rewrite/README.md
+++ b/ingress/controllers/nginx/examples/rewrite/README.md
@@ -1,0 +1,67 @@
+
+Create an Ingress rule with a rewrite annotation:
+```
+$ echo "
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  annotations:
+    ingress.kubernetes.io/rewrite-target: /
+  name: rewrite
+  namespace: default
+spec:
+  rules:
+  - host: rewrite.bar.com
+    http:
+      paths:
+      - backend:
+          serviceName: echoheaders
+          servicePort: 80
+        path: /something
+" | kubectl create -f -
+```
+
+Check the rewrite is working
+
+```
+$ curl -v http://172.17.4.99/something -H 'Host: rewrite.bar.com'
+*   Trying 172.17.4.99...
+* Connected to 172.17.4.99 (172.17.4.99) port 80 (#0)
+> GET /something HTTP/1.1
+> Host: rewrite.bar.com
+> User-Agent: curl/7.43.0
+> Accept: */*
+>
+< HTTP/1.1 200 OK
+< Server: nginx/1.11.0
+< Date: Tue, 31 May 2016 16:07:31 GMT
+< Content-Type: text/plain
+< Transfer-Encoding: chunked
+< Connection: keep-alive
+<
+CLIENT VALUES:
+client_address=10.2.56.9
+command=GET
+real path=/
+query=nil
+request_version=1.1
+request_uri=http://rewrite.bar.com:8080/
+
+SERVER VALUES:
+server_version=nginx: 1.9.11 - lua: 10001
+
+HEADERS RECEIVED:
+accept=*/*
+connection=close
+host=rewrite.bar.com
+user-agent=curl/7.43.0
+x-forwarded-for=10.2.56.1
+x-forwarded-host=rewrite.bar.com
+x-forwarded-port=80
+x-forwarded-proto=http
+x-real-ip=10.2.56.1
+BODY:
+* Connection #0 to host 172.17.4.99 left intact
+-no body in request-
+```
+
--- a/ingress/controllers/nginx/nginx/rewrite/main.go
+++ b/ingress/controllers/nginx/nginx/rewrite/main.go
@@ -17,68 +17,57 @@ limitations under the License.
 package rewrite
 
 import (
+	"errors"
 	"strconv"
 
 	"k8s.io/kubernetes/pkg/apis/extensions"
 )
 
 const (
-	rewrite = "ingress-nginx.kubernetes.io/rewrite-to"
-	fixUrls = "ingress-nginx.kubernetes.io/fix-urls"
+	rewriteTo  = "ingress.kubernetes.io/rewrite-target"
+	addBaseURL = "ingress.kubernetes.io/add-base-url"
 )
-
-// ErrMissingAnnotations is returned when the ingress rule
-// does not contains annotations related with redirect or strip prefix
-type ErrMissingAnnotations struct {
-	msg string
-}
-
-func (e ErrMissingAnnotations) Error() string {
-	return e.msg
-}
 
 // Redirect returns authentication configuration for an Ingress rule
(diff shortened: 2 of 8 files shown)


ingress/controllers/nginx/nginx/rewrite/main.go
```diff
@@ -0,0 +1,84 @@
...
+
+import (
+	"strconv"
+
+	"k8s.io/kubernetes/pkg/apis/extensions"
+)
+
+const (
+	rewriteTo  = "ingress-nginx.kubernetes.io/rewrite-to"
+	addBaseURL = "ingress-nginx.kubernetes.io/add-base-url"
```
bprashanth: please add an example in the readme with how these interact with the controller and ingresses in the cluster

ingress/controllers/nginx/nginx/rewrite/main.go
```diff
@@ -0,0 +1,84 @@
...
+
+	return false
+}
+
+func (a ingAnnotations) rewriteTo() string {
+	val, ok := a[rewriteTo]
+	if ok {
+		return val
+	}
+
```
bprashanth: please drop this dead line, and the one above the return of the previous method

ingress/controllers/nginx/nginx/rewrite/main.go
```diff
@@ -0,0 +1,84 @@
...
+	return ""
+}
+
+// ParseAnnotations parses the annotations contained in the ingress
+// rule used to rewrite the defined paths
+func ParseAnnotations(ing *extensions.Ingress) (*Redirect, error) {
+	if ing.GetAnnotations() == nil {
+		return &Redirect{}, ErrMissingAnnotations{"no annotations present"}
+	}
+
```
bprashanth: can you just inline the 2 local vars into the struct?

ingress/controllers/nginx/nginx/rewrite/main.go
```diff
@@ -0,0 +1,84 @@
...
+	rt := ingAnnotations(ing.GetAnnotations()).rewriteTo()
+	abu := ingAnnotations(ing.GetAnnotations()).addBaseURL()
+	return &Redirect{
+		To:         rt,
```
bprashanth: can you name this field something else? maybe base/target, since that matches the html concept?

ingress/controllers/nginx/nginx/template.go
```diff
@@ -100,3 +107,68 @@ func toCamelCase(src string) string {
 	}
 	return string(bytes.Join(chunks, nil))
 }
+
+func buildLocation(input interface{}) string {
```
bprashanth: please include a comment with an example output string so it's clear what role this function plays in producing the final template

ingress/controllers/nginx/nginx/template.go
```diff
@@ -100,3 +107,68 @@ func toCamelCase(src string) string {
...
+
+	path := location.Path
+	if len(location.Redirect.To) > 0 && location.Redirect.To != path {
+		return fmt.Sprintf("~* %s", path)
+	}
+
+	return path
+}
+
+func buildProxyPass(input interface{}) string {
```
bprashanth: same as comment above, an example like: buildProxyPass produces the proxy pass string, if the ingress has redirects (specified through the kubernetes.io/... annotation), a base tag is added, eg: for an annotation like `kubernetes.io...` you will get: `<base href="http://">`...

ingress/controllers/nginx/nginx/rewrite/main.go
```diff
@@ -0,0 +1,84 @@
...
+
+	"k8s.io/kubernetes/pkg/apis/extensions"
+)
+
+const (
+	rewriteTo  = "ingress-nginx.kubernetes.io/rewrite-to"
+	addBaseURL = "ingress-nginx.kubernetes.io/add-base-url"
+)
+
+// ErrMissingAnnotations is returned when the ingress rule
```
bprashanth: why define a custom error type if you're not actually checking for it?

ingress/controllers/nginx/nginx/rewrite/main.go
```diff
@@ -0,0 +1,84 @@
...
+
+package rewrite
+
+import (
+	"strconv"
+
+	"k8s.io/kubernetes/pkg/apis/extensions"
+)
+
+const (
```
bprashanth: can you express this as a single annotation, they're tied together right? maybe just: `ingress.kubernetes.io/rewrite: {base, target}` ? you'd have to unmarshal a map but that's simple.

Also why `ingress-nginx`? just ignore the annotations you don't care about and use the rest? typically the first part of the annotations prefix is the resource name and has nothing to do with the implementation, and all site specific data is included after the "/"

bprashanth:
Looks good, mostly just stylistic comments. Rewrites are so common that I think we should document this with ample examples (with without base, with regex etc).

ingress/controllers/nginx/nginx/rewrite/main.go
aledbf: yes, this is what's missing now

ingress/controllers/nginx/nginx/rewrite/main.go
aledbf: > Also why ingress-nginx?

No particular reason. Can I leave this change to the last PR before the release of the new version (other pr's use a similar prefix)

ingress/controllers/nginx/examples/rewrite/README.md
```diff
@@ -0,0 +1,67 @@
+
+Create an Ingress rule with a rewrite annotation:
```
bprashanth: can you also add an examble with base path?

bprashanth:
You have a merge conflict, LGTM but for the doc nit

ingress/controllers/nginx/examples/rewrite/README.md
aledbf: yes, I'm searching for a sample app :)

aledbf:
@bprashanth can we left the missing example for #1104?

PR #1070, opened 2016-05-25
Merge queue: look up old test results in case of downtime

lavalamp:
First commit is in #1061

PR #1081, opened 2016-05-27
Don't file individual issues when big groups of tests fail.

### LINKED ISSUE #1071: Submit queue: don't file individual flake issues when groups of tests fail in a suite
It's too noisy.

Alternatively, file each issue but also file an issue for the group and cross-link.

lavalamp:
Fixes #1071

Builds on #1070. Only the last commit is new.


eparis:
rebase :-(

lavalamp:
Rebased! ...but this could use some manual testing soak time still.

lavalamp:
Example: https://github.com/lavalamp/issue-testing/issues/107

lavalamp:
This did good things to my test repo overnight so I think I'm OK with pushing it live.

lavalamp:
Fixed lint error (don't use 'self'??? wtf!)

PR #1082, opened 2016-05-27
Enable the assign-fixes githubmunger

asalkeld:
The code for this munger has just been merged, this just enables it.

https://github.com/kubernetes/contrib/blob/master/mungegithub/mungers/assign-fixes.go

This munger will search for "fixes #<num>" in a PR message text and assign the referenced issues to the owner of the PR.

### DIFF
--- a/mungegithub/submit-queue/deployment.yaml
+++ b/mungegithub/submit-queue/deployment.yaml
@@ -14,7 +14,7 @@ spec:
         command:
         - /mungegithub
         - --token-file=/etc/secret-volume/token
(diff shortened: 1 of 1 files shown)


lavalamp:
Is it OK if we turn this on in a subsequent push? We've got a lot of new code just merged and I don't want a production outage... :)

In the meantime, can you adjust this PR to make the documentation a little clearer? I had to look at the PR adding this to figure out what it was supposed to be doing. Thanks!

asalkeld:
@lavalamp  sure, no rush on this at all.

eparis:
apparently I wasted too much time and now you need to rebase.

eparis:
The submit-queue just ran with this patch and, to the best of my knowledge assigned nothing. Is this broken or was there nothing to do?

PR #1084, opened 2016-05-27
Show retests avoided

eparis:
And some other minor display updates

### DIFF
--- a/mungegithub/submit-queue/www/script.js
+++ b/mungegithub/submit-queue/www/script.js
@@ -23,7 +23,6 @@ function SQCntl(dataService, $interval, $location) {
   self.queryNum = 0;
   self.selected = 1;
   self.OverallHealth = "";
-  self.StartTime = "";
   self.StatOrder = "-Count";
   self.location = $location;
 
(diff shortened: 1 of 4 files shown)


eparis:
@lavalamp

mungegithub/submit-queue/www/script.js
```diff
@@ -61,25 +61,43 @@ function SQCntl(dataService, $interval, $location) {
...
+    for (var i = 0; i < reloadFuncs.length; i++) {
+      var func = reloadFuncs[i];
+      if (self.functionLoading[func] == true) {
+        continue;
       }
-      self.tabLoaded[self.selected] = true;
+      self.functionLoading[func] = true;
+
+      var updateIntervalMinutes = reloadFunctions[func];
+      func();
```
lavalamp: I don't really know js... is this really a thing??

lavalamp: Oh, I see. It's calling func. Github's syntax highlighting confused me.

lavalamp:
I have no idea if this is reasonable js or not but I don't see anything obviously wrong. LGTM, merge if you're confident.

mungegithub/submit-queue/www/script.js
eparis: yeah, this calls `func` 'right now' and then the next line calls it 'every so often'

PR #1083, opened 2016-05-27
Introduce a flag to block merges for certain milestones

eparis:
This blocks merge for v1.4 and next-candidate. When we open for v1.4 we
should remove `v1.4` from the flags and remove `next-candidate` from all
PRs.

PR #1091, opened 2016-05-27
Introduce a flag to block merges for certain milestones

lavalamp:
This blocks merge for v1.4 and next-candidate. When we open for v1.4 we
should remove `v1.4` from the flags and remove `next-candidate` from all
PRs.


lavalamp:
Manual rebase of #1083

PR #1093, opened 2016-05-27
[nginx-ingress-controller] Add support for rate limiting in ingress rule locations

aledbf:


ingress/controllers/nginx/controller.go
```diff
@@ -580,6 +581,12 @@ func (lbc *loadBalancerController) getUpstreamServers(data []interface{}) ([]*ng
 				continue
 			}
 
+			rl, err := ratelimit.ParseAnnotations(ing)
+			glog.V(3).Infof("nginx rate limit %v", rl)
+			if err != nil {
+				glog.V(3).Infof("error reading rate limit annotation in Ingress %v/%v: %v", ing.GetNamespace(), ing.GetName(), err)
```
bprashanth: things like this feel event worthy, if it's easy enough to send an event. If the user thought they specified the right annotations but didn't, they should know their backend isn't getting ratelimited without having to debug a bunch and look around in the logs

ingress/controllers/nginx/nginx/ratelimit/main.go
```diff
@@ -0,0 +1,131 @@
...
+	"errors"
+	"fmt"
+	"strconv"
+
+	"k8s.io/kubernetes/pkg/apis/extensions"
+)
+
+const (
+	limitIp  = "ingress-nginx.kubernetes.io/limit-connections"
+	limitRps = "ingress-nginx.kubernetes.io/limit-rps"
```
bprashanth: same comment I made on the other pr about the format of annotations.
I like the modular way we're handling the annotations, but I think we should surface them all in one common place in the docs, so users know whats available without having to cd into each dir or find/grep for the `kubernetes.io` string

ingress/controllers/nginx/nginx/ratelimit/main.go
```diff
@@ -0,0 +1,131 @@
...
+// connections per second.
+// Note: Is possible to specify both limits
+type RateLimit struct {
+	// Connections indicates a limit with the number of connections per IP address
+	Connections Zone
+	// RPS indicates a limit with the number of connections per second
+	RPS Zone
+}
+
+// Zone returns information about the rate limit
```
bprashanth: can you be more specific about the meaning of "zone"? since it clashes with what CPs call a (availability-)zone

ingress/controllers/nginx/nginx/template.go
```diff
@@ -101,3 +103,56 @@ func toCamelCase(src string) string {
 	}
 	return string(bytes.Join(chunks, nil))
 }
+
+func buildRateLimitZones(input interface{}) []string {
```
bprashanth: same comment as I made in the other pr about comments with examples of the output where possible

bprashanth:
Mostly similar comments to those in the rewrite pr (https://github.com/kubernetes/contrib/pull/1079). I think we should err on the side of caution and leave as many explicit examples and send as many events as we can, when exposing site specific features through annotations, because it's so easy to get wrong. Both from a user perspective, and from a cross-platform perspective (i.e I look at how the nginx controller is handling redirects to make sure it's the same for other platforms)

ingress/controllers/nginx/nginx/ratelimit/main.go
aledbf: > but I think we should surface them all in one common place in the docs

agree, in #1104 I'm adding a table with all the annotations, description and link to the proper example

ingress/controllers/nginx/nginx/ratelimit/main.go
aledbf: done

ingress/controllers/nginx/nginx/template.go
aledbf: done

bprashanth:
LGTM but you have a merge conflict

aledbf:
done

ingress/controllers/nginx/nginx.tmpl
```diff
@@ -156,6 +156,12 @@ http {
     }
     {{ end }}
 
+    {{/* build all the required rate limit zones. Each annotation requires a dedicated zone */}}
+    {{/* 1MB -> 16 thousand 64-byte states or about 8 thousand 128-byte states */}}
+    {{ $zone := range (buildRateLimitZones .servers) }}
```
louis-paul: I'm late to the battle but shouldn't this be `{{ range $zone := (buildRateLimitZones .servers) }}`?

Should I make a PR?

aledbf: @louis-paul this is fixed in https://github.com/kubernetes/contrib/pull/1104. 
(just in case this PR is not published yet)

louis-paul: Great, thanks!

PR #1095, opened 2016-05-28
string casting doesn't work that way

lavalamp:
@freehan 

...going to push this shortly


freehan:
Thanks for fixing my bug...

lavalamp:
My fault for not catching in review!

PR #1105, opened 2016-05-31
Slight re-write to NewIssue dry-run handling

eparis:
print a message so you know if it triggered with --dry-run and count the
API tokens.


eparis:
@lavalamp

mungegithub/github/github.go
```diff
@@ -498,15 +498,16 @@ func (config *Config) GetObject(num int) (*MungeObject, error) {
 
 // NewIssue will file a new issue and return an object for it.
 func (config *Config) NewIssue(title, body string, labels []string) (*MungeObject, error) {
+	config.analytics.CreateIssue.Call(config, nil)
```
lavalamp: Oh, the resp is not important?

lavalamp:
LGTM

PR #1109, opened 2016-05-31
Do not update merge rate if we don't run e2e

apelisse:
If we don't run e2e tests, it makes it very easy to merge quickly, and
that makes the merge rate very high. Just do not include non-e2e merges
into the rate so that we keep it relevant.


ixdy:
lgtm

Issue #882, opened 2016-05-02
Add release-note-experimental to release-note label checking


david-mcmahon:
Can we add this one to the mix too?  More probably coming.

david-mcmahon:
I already added the label and am parsing for it in the release notes generation tool.  So we can just add enforcement when we get to it.

david-mcmahon:
cc @bgrant0607

david-mcmahon:
Just a run down here on what we currently have.  The release notes have a previously agreed upon list of [headings](https://github.com/kubernetes/release/pull/1/files#diff-d2b70fc121f3dab046b476fbfcd7f7abR298).  Here they are with associated labels in order of how they appear.  

Heading | Label
---------- | --------
Major Themes | _curated_
Other notable improvements | _curated_
Experimental Features | release-note-experimental
Action Required | release-note-action-required
Known Issues | _curated_
Provider-specific Notes | _curated_
Other notable changes | release-note

The idea is that the _curated_ headings only appear in the release notes for .0 releases and are populated by hand by the release owner (soon) after the initial notes are published during the release.

I believe the only one of these likely get a label is "known issues" - being discussed in another thread.  cc @roberthbailey

bgrant0607:
Looks reasonable to me.

david-mcmahon:
I just noticed that `release-note-experimental` is no longer on the label list.  Probably pulled because it was not enforced yet.  We can re-add when we add enforcement of it.

PR #1111, opened 2016-05-31
Add release-note-experimental as a valid label.

david-mcmahon:
ref #882

@eparis is there any doc I can follow on how to test and rollout changes in contrib?

### DIFF
--- a/mungegithub/mungers/release-note-label.go
+++ b/mungegithub/mungers/release-note-label.go
@@ -34,15 +34,16 @@ const (
 	releaseNote               = "release-note"
 	releaseNoteNone           = "release-note-none"
 	releaseNoteActionRequired = "release-note-action-required"
+	releaseNoteExperimental   = "release-note-experimental"
 
 	releaseNoteFormat = `Removing LGTM because the release note process has not been followed.
-One of the following labels is required %q, %q, or %q
-Please see: https://github.com/kubernetes/kubernetes/blob/master/docs/devel/pull-requests.md#release-notes`
-	parentReleaseNoteFormat = `The 'parent' PR of cherry-picks must have either the %q or %q label or this PR must follow the release note process.`
+One of the following labels is required %q, %q, %q or %q.
+Please see: https://github.com/kubernetes/kubernetes/blob/master/docs/devel/pull-requests.md#release-notes.`
+	parentReleaseNoteFormat = `The 'parent' PR of a cherry-pick PR must have one of the %q or %q labels, or this PR must follow the standard/parent release note labeling requirement. (release-note-experimental must be explicit for cherry-picks)`
 )
 
 var (
-	releaseNoteBody       = fmt.Sprintf(releaseNoteFormat, releaseNote, releaseNoteNone, releaseNoteActionRequired)
+	releaseNoteBody       = fmt.Sprintf(releaseNoteFormat, releaseNote, releaseNoteActionRequired, releaseNoteExperimental, releaseNoteNone)
 	parentReleaseNoteBody = fmt.Sprintf(parentReleaseNoteFormat, releaseNote, releaseNoteActionRequired)
 )
 
@@ -89,15 +90,16 @@ func (r *ReleaseNoteLabel) prMustFollowRelNoteProcess(obj *github.MungeObject) b
 
 	for _, parent := range parents {
 		// If the parent didn't set a release note, the CP must
-		if !parent.HasLabel(releaseNote) && !parent.HasLabel(releaseNoteActionRequired) {
+		if !parent.HasLabel(releaseNote) &&
+			!parent.HasLabel(releaseNoteActionRequired) {
 			if !obj.HasLabel(releaseNoteLabelNeeded) {
 				obj.WriteComment(parentReleaseNoteBody)
 			}
 			return true
 		}
 	}
-	// All of the parents set the releaseNote or releaseNoteActionRequired label, so
-	// this cherrypick PR needs to do nothing.
+	// All of the parents set the releaseNote or releaseNoteActionRequired label,
+	// so this cherrypick PR needs to do nothing.
 	return false
 }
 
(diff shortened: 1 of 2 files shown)


mungegithub/mungers/release-note-label.go
```diff
@@ -34,16 +34,17 @@ const (
...
+One of the following labels is required %q, %q, %q or %q.
+Please see: https://github.com/kubernetes/kubernetes/blob/master/docs/devel/pull-requests.md#release-notes.`
+	parentReleaseNoteFormat = `The 'parent' PR of a cherry-pick PR must have one of the %q, %q or %q labels or this PR must then follow the standard/parent release note labeling requirement.`
 )
 
 var (
-	releaseNoteBody       = fmt.Sprintf(releaseNoteFormat, releaseNote, releaseNoteNone, releaseNoteActionRequired)
-	parentReleaseNoteBody = fmt.Sprintf(parentReleaseNoteFormat, releaseNote, releaseNoteActionRequired)
+	releaseNoteBody       = fmt.Sprintf(releaseNoteFormat, releaseNote, releaseNoteActionRequired, releaseNoteExperimental, releaseNoteNone)
+	parentReleaseNoteBody = fmt.Sprintf(parentReleaseNoteFormat, releaseNote, releaseNoteActionRequired, releaseNoteExperimental)
```
eparis: Would `releaseNoteExperimental`really make sense on a release branch cherry pick?

mungegithub/mungers/release-note-label.go
```diff
@@ -89,15 +90,17 @@ func (r *ReleaseNoteLabel) prMustFollowRelNoteProcess(obj *github.MungeObject) b
 
 	for _, parent := range parents {
 		// If the parent didn't set a release note, the CP must
-		if !parent.HasLabel(releaseNote) && !parent.HasLabel(releaseNoteActionRequired) {
+		if !parent.HasLabel(releaseNote) &&
+       !parent.HasLabel(releaseNoteActionRequired) &&
```
eparis: this has to fail gofmt, right?

eparis:
`make help` gives a little description. I started to write a README the other day. I often do something like the following for locally testing specific changes:

`make mungegithub && ./mungegithub --dry-run --pr-mungers=$MUNGERS --kubernetes-dir=/path/to/dir --token-file=/path/to/token/file --min-pr-number=X --max-pr-number=Y`

And making sure the local logs look like it is doing the right thing.

Then I'll do a `REPO=docker.io/eparis APP=submit-queue make deploy` to test it against a real kube cluster how it would be deployed. If that all looks ok I'll do `KUBECONFIG=/path/to/production/kubeconfig READONLY=false REPO=docker.io/eparis APP=submit-queue make deploy` to deploy to production.

mungegithub/mungers/release-note-label.go
david-mcmahon: "experimental" is just one of the available release note types we want to cover per #882.  I suspect as a general rule we don't want to include "experimental" things on a release branch, but in practice there will be times we want them and want to leave that judgement up to the PR owner and reviewer.  Thoughts @bgrant0607 ?

david-mcmahon:
Got it.  Thanks for the writeup.  Some questions for the very uninitiated:

1. What is the purpose of the min/max pr args?  Do I just pick some range that includes a couple of PRs?  What is that doing exactly?  Can I leave out one or the other?
2. When I did run it, I'm not sure what to expect.  It stopped at this point, but I'm not sure if that's a good thing and I should wait?  I ^C'd to see if there were any logs yet, but no.  Is ~10min the expected run-time here?  And what exactly can I look for in the logs that indicates things went 'ok'?

```
I0531 17:13:19.021779   57813 mungegithub.go:59] Running mungers
I0531 17:13:35.329834   57813 github.go:223] Made 38 API calls since the last Reset 2.330136 calls/sec
I0531 17:13:35.329908   57813 mungegithub.go:74] Sleeping for 9m43.691871175s
```

eparis:
min-max is just to make it only tests the PRs you expect to be relevant. In this case no PRs have the experimental label, so you wouldn't expect this PR to do anything...

the bot runs in a loop, once every 10 minutes. So it tested all of the pr's between min/max (or all PRs if you left it out) and then waited for 10 minutes to test those PRs again.

Create a test PR, with the experimental label, then run it 'against' that PR, and read the logs.

david-mcmahon:
Label testing seems to do what it should, AFAICT.
As far as actual deployment, what is the /path/to part of `KUBECONFIG=/path/to/production/kubeconfig`?  I haven't played with any of this yet.

Do you usually check things in before deploying?  I guess it doesn't really make a difference for this.

eparis:
You'll have to talk with someone to get the 'production' kubeconfig. I know that at least @lavalamp @mikedanese and @brendandburns have it. We only put HEAD into production. So yes, it would need to be merged before you ran it with `READONLY=false`

mungegithub/mungers/release-note-label.go
eparis: Right. I would expect that a cherry-pick to a release branch where the parent was 'experimental' would be something we wouldn't automatically consider acceptable and we'd for the release branch owner to follow the release note process on the cherry-pick PR.  AKA I think you should revert https://github.com/kubernetes/contrib/pull/1111/files#diff-7294a858e5f70293980043d7fe211686R42
https://github.com/kubernetes/contrib/pull/1111/files#diff-7294a858e5f70293980043d7fe211686R94
https://github.com/kubernetes/contrib/pull/1111/files#diff-7294a858e5f70293980043d7fe211686R94

david-mcmahon:
This has been deployed.  LGTM when you get a chance. Thanks everyone for the assist!

eparis:
This is already deployed in production?

david-mcmahon:
@eparis yes.  It looks like we wanted this merged before however.  Can we merge this now or would we like to roll back?

eparis:
@david-mcmahon I want to know who helped you, so I can yell at them (and then merge this)

david-mcmahon:
You can yell at me. I misread the last comment and went ahead with the deployment.

eparis:
bad david. bad.  we've had a number of regressions where people committed (even minor) changes to production without merging. And the next person along accidentally reverted their fix.

If you self push to production, better to self merge. We can roll production back, but silent changes are stinky.

david-mcmahon:
This dev->deployment process could probably benefit from some releng tooling.
Guided processes that enforce policy through tooling certainly can avoid the pitfalls that are bound to happen when in the wrong hands.  I'll file an issue.

PR #1112, opened 2016-06-01
Make flake URLs point to Gubernator instead of storage.googleapis.com.

rmmh:
The previous URL was not useful (see kubernetes/kubernetes#26509).

individualFlakeSource/brokenJobSource.ID() now returns things like:
"/kubernetes-jenkins/logs/e2e-gce/123/\n", which allows some flexibility
to make the Body() return different URLs.

There are tests for compatibility with old ID() results.


mungegithub/mungers/flake-manager.go
```diff
@@ -127,6 +127,11 @@ func (p *FlakeManager) isIndividualFlake(f cache.Flake) bool {
 	return true
 }
 
+// makeGubernatorLink returns a URL to view the build results in a GCS path.
+func makeGubernatorLink(id string) string {
+	return "https://k8s-gubernator.appspot.com/build" + id
+}
```
eparis: Is there some link about what this app is? What it does? How it is maintained? Breadcrumbs for the next poor sap?

rmmh: https://github.com/kubernetes/test-infra/tree/master/gubernator

rmmh: Would you like a brief summary in the comment?

lavalamp: Yes please.

mungegithub/mungers/flake-manager.go
```diff
@@ -127,6 +127,11 @@ func (p *FlakeManager) isIndividualFlake(f cache.Flake) bool {
 	return true
 }
 
+// makeGubernatorLink returns a URL to view the build results in a GCS path.
+func makeGubernatorLink(id string) string {
```
lavalamp: Explain the format of 'id'. Don't call it 'id'.

mungegithub/mungers/flake-manager.go
lavalamp: Why is this an improvement over the direct link? It doesn't offer any more features, and it loads very slowly?

rmmh: The direct link is incorrect, linking to things like https://storage.googleapis.com/kubernetes-jenkins/logs/kubernetes-e2e-gce/17823/ which is a 404.

lavalamp: Oops. That's weird, I tested it, is that job an exception?

Why not just fix the link?

rmmh: All the links are broken.

Right now Gubernator doesn't have very much additional information versus what this puts in the issue, but it's intended to provide a better triaging experience than poking around in the GCS buckets directly.

lavalamp: Hm. Maybe they broke after I put them in.

Can we make gubernator load faster?

rmmh:
Added a comment describing gubernator, fixed the contrib/test-utils vending problem by using a symlink.

mungegithub/mungers/flake-manager.go
rmmh: Yes, but unfortunately GAE doesn't have libraries for asynchronous GCS downloads. I'll look into writing my own.

test-utils/utils/utils.go
```diff
@@ -96,10 +96,9 @@ func (u *Utils) GetGCSListURL() string {
 	return fmt.Sprintf(GCSListAPIURLTemplate, u.bucket)
 }
 
-// GetPathToJenkinsGoogleBucket reads data from Google project's GCS bucket for the given job and buildNumber.
-// Returns a response with file stored under a given (relative) path or an error.
-func (u *Utils) GetPathToJenkinsGoogleBucket(job string, buildNumber int, path string) string {
-	return fmt.Sprintf("%v/%v/%v/%v", u.GetGCSDirectoryURL(), job, buildNumber, path)
+// GetPathToJenkinsGoogleBucket returns a GCS path containing the artifacts for a given job and buildNumber.
```
lavalamp: Please fix this comment. Note that it only formats a path, and doesn't include a host.

mungegithub/mungers/flake-manager.go
```diff
@@ -127,6 +127,14 @@ func (p *FlakeManager) isIndividualFlake(f cache.Flake) bool {
 	return true
 }
 
+// makeGubernatorLink returns a URL to view the build results in a GCS path.
+//
+// Gubernator is a simple frontend that reads test result buckets to improve
+// test triaging. Its source code is in kubernetes/test-infra/gubernator
+func makeGubernatorLink(id string) string {
```
lavalamp: probably going to merge & push this shortly because I want this fix, but please come up with a better variable name than ID, and explain what its contents should be.

rmmh: Nits addressed.

lavalamp:
LGTM, thanks! I will push this soon.

PR #1114, opened 2016-06-01
Cluster-autoscaler: remove node infos without node

mwielgus:
Fixes e2e - scale up/scale down tests.

cc: @piosz @fgrzadkowski @jsz


piosz:
LGTM

PR #1124, opened 2016-06-02
[Addon resizer] Added ExponentialEstimator to pod nanny

piosz:
Once CA is enabled the cluster size is being changed pretty frequently which cases Heapster downtime. To prevent from this we want to adjust resources less.

cc @a-robinson @fgrzadkowski @jszczepkowski @mwielgus @roberthbailey


piosz:
The PR is not yet tested since there is no godeps in nanny directory, I'll fix it tomorrow.

roberthbailey:
How do we plan to control whether we launch the linear estimator vs. the logarithmic estimator? Do you suggest launching the logarithmic one if the cluster autoscaler is running in the cluster and the linear one otherwise?

addon-resizer/nanny/estimator.go
```diff
@@ -56,3 +56,18 @@ func (e LinearEstimator) scaleWithNodes(numNodes uint64) *api.ResourceRequiremen
...
+type LogarithmicEstimator struct {
+	linearEst LinearEstimator
+}
+
+func (e LogarithmicEstimator) scaleWithNodes(numNodes uint64) *api.ResourceRequirements {
+	n := uint64(16)
+	for n <= numNodes {
+		n = n * 3 / 2
+	}
+	return e.linearEst.scaleWithNodes(n)
```
Q-Lee: Clever: I wouldn't want to touch that awful fixed point library either. On the other hand it's weird for LogarithmicEstimator to depend on LinearEstimator. The math of LogarithmicEstimator is dead simple, and this kind of obfuscates it.

It's worth noting that we've taken the speter.net/go/exp/math/dec/inf dependency out of the API. My understanding is that there's some much saner library we can use.

piosz: I can do a copy/paste if you think it would be more readable;)

Regarding speter.net/go/exp/math/dec/inf the general problem with this project is that there is no godeps, so it's not possible to compile it:(

piosz:
@roberthbailey the easier way is to use logarithmic one also when CA is disabled, however if you are against it we can try to do as you described. WDYT?

roberthbailey:
So is the plan to switch logarithmic to be the default? My only concern is that if it balloons each cluster addon by 50% then our system overheads (especially for small clusters) may increase by an unreasonable degree. Is the extra 50% overhead present at small node sizes (1-3) or just at large node sizes (>50)?

mwielgus:
Please note that in Heapster (probably the biggest nanny client) we allocate about 4mb per node so for small clusters it doesn't make much difference. For 50 node cluster the max extra overhead wold be 100MB. However if the cluster is scaled up and down from 1 to 50 the extra overhead may be higher at some point (although the total consumption won't be bigger than at 50 nodes).
And if you are not comfortable with 50% we can probably try with 30%.

roberthbailey:
I'm more concerned with the extra overhead on small clusters. On large clusters, a few hundred MB doesn't concern me. But on a 1 node cluster, increasing the overhead by 50% could mean the difference between using up half a core and three quarters of a core and I don't think that an increase in reserved resources on a small cluster would be ok.

mwielgus:
Heapster uses flat 100m CPU.

mwielgus:
Afair Heapster formula for memory is 200 + 4 * nodes. So for small clusters even if we cap at 16 nodes the extra memory consumption would be 64MB (we can probably include some of that in the initial 200).

jszczepkowski:
I signed it!

fgrzadkowski:
I think this PR does not solve the problem entirely. Let's say that cluster autoscaler oscillates between 15 & 16 nodes. With this new algorithm we'd change resource requirements with the same frequency, with the difference, that we'd set resources for either 16 nodes (if 15 nodes in the cluster) or 24 nodes (if 16 nodes in the cluster).

I guess the advantage is that those unstable points are less often (only for 16*(3/2)^n nodes).

I think a good solution would be to just decrease resources only if we are at say XX% of the lower threshold. That would give us stabilization.

To address @roberthbailey concern we can reduce the 16 node constant and 150% multiplier. We can also combine this with first linear scaling in bigger chunks and above some threshold scaling exponentially. WDYT?

piosz:
@fgrzadkowski I proposed this attitude initially, but after talk with @mwielgus we decided to do is as above, because:
- we can get rid off of flapping but as cost of increased overhead - it would be (3/2)^2
- it would require significant redesign of addon-resizer
- current approach which is very simple improves the situation significantly

addon-resizer/nanny/estimator.go
```diff
@@ -56,3 +56,18 @@ func (e LinearEstimator) scaleWithNodes(numNodes uint64) *api.ResourceRequiremen
...
+
+// LogarithmicEstimator estimates the amount of resources in the way that
+// prevents from frequent updates but may end up with larger resource usage
+// than actually needed (though no more than 50%).
+type LogarithmicEstimator struct {
+	linearEst LinearEstimator
+}
+
+func (e LogarithmicEstimator) scaleWithNodes(numNodes uint64) *api.ResourceRequirements {
+	n := uint64(16)
```
Q-Lee: The object is very much a mathematical construct. From the object's POV this should be parameterized (even if it's just left as a default for the program).

addon-resizer/nanny/estimator.go
```diff
@@ -56,3 +56,18 @@ func (e LinearEstimator) scaleWithNodes(numNodes uint64) *api.ResourceRequiremen
...
+	for n <= numNodes {
+		n = n * 3 / 2
```
Q-Lee: I would love some control over the scale factor.

addon-resizer/nanny/estimator.go
Q-Lee: I think that's one more reason to tear it out like we did in k8s.

addon-resizer/nanny/estimator.go
piosz: Done

addon-resizer/nanny/estimator.go
piosz: Done

piosz:
@Q-Lee @mwielgus the PR is ready for review. PTAL
I added godeps to the project

@fgrzadkowski @roberthbailey let's postone

mwielgus:
I don't have any bigger concerns there, but please wait for @Q-Lee approval.

Q-Lee:
If we still experience a problem with flapping, it should be pretty easy to plumb thresholds into the exponential scaler. But I'm not too worried about that.

piosz:
Thanks @Q-Lee for the review!

PR #1128, opened 2016-06-02
add quotes for clarity

lavalamp:


eparis:
can you pass a slice to %q? isn't it a slice?

lavalamp:
It adds quotes around each element-- it's very handy :)

eparis:
^--- this guy had no idea!

Issue #953, opened 2016-05-11
ingress controller should health check straight to endpoints


bprashanth:
Scrape endpoints, look for a http readiness probe, make that the health check url. https://github.com/kubernetes/contrib/tree/master/ingress/controllers/gce#health-checks

PR #1133, opened 2016-06-03
Ubernetes multizone and custom healthchecks

### LINKED ISSUE #983: GCE Ingress fails to sync instance group in multi-az cluster
When an ingress is spun up in a multi-zone cluster, no nodes end up in the k8s-ig instance group. My logs show

`I0513 12:21:19.826319       1 instances.go:130] Node pool encountered a 404, ignoring: googleapi: Error 404: The resource 'projects/myproject/zones/us-east1-c/instances/gke-mazcluster-default-pool-67470956-1n15' was not found, notFound`

(gke-mazcluster-default-pool-67470956-1n15 is not found in us-east1-c because it is in us-east1-d)

I believe the underlying issue here relates to getting the gce cloudprovider without config:

In [ClusterManager.getGCEClient()](https://github.com/kubernetes/contrib/blob/master/ingress/controllers/gce/controller/cluster_manager.go#L193)

The GCE provider is retrieved with `nil` as the config param.
 `cloudInterface, err := cloudprovider.GetCloudProvider("gce", nil)`

Which means `cfg.Global.Multizone` would never get set, managedZones will not get set to nil and instead we will always be looking in the primary zone.

I believe the fix should be similar to kubernetes/kubernetes#23769 but I don't know how you might/if it's possible to get a reference to the kubelet and configured CloudProvider here vs. how it can be done in a plugin.

@bprashanth

bprashanth:
1. Adds multi zone support to glbc
2. Pipes health checks up from the readiness probe of a pod to GCLB

@kubernetes/goog-cluster fyi
Fixes https://github.com/kubernetes/contrib/issues/983 and https://github.com/kubernetes/contrib/issues/953


ingress/controllers/gce/backends/backends.go
```diff
@@ -201,18 +197,31 @@ func (b *Backends) List() ([]interface{}, error) {
 	return interList, nil
 }
 
+func getBackendsForIGs(igs []*compute.InstanceGroup) []*compute.Backend {
```
freehan: just curious. If not all ubernetes clusters are on gce, how do you handle it?   Or just does not support this case? 
Can GCE L7 even support compute backends which are not on gce?

bprashanth: nope, this is only "ubernetes-lite" i.e ubernetes deployment multi-az in gce.

freehan: okay I see.

ingress/controllers/gce/controller/utils.go
```diff
@@ -37,6 +38,10 @@ import (
 const (
 	allowHTTPKey    = "kubernetes.io/ingress.allow-http"
 	staticIPNameKey = "kubernetes.io/ingress.global-static-ip-name"
+
+	// Label key to denote which GCE zone a Kubernetes node is in.
+	zoneKey     = "failure-domain.beta.kubernetes.io/zone"
```
freehan: is there a naming convention for ubernetes? 
`federation.beta.kubernetes.io/zone` ?

ingress/controllers/gce/controller/utils.go
```diff
@@ -315,3 +320,40 @@ func (t *GCETranslator) toNodePorts(ings *extensions.IngressList) []int64 {
 	}
 	return knownPorts
 }
+
+func getZone(n api.Node) string {
```
freehan: Why return a empty zone instead of err? 
It may hide underlying error

ingress/controllers/gce/controller/utils.go
bprashanth: this is the zone label, not the ubernetes federation zone afaik. It should show up on every non-ubernetes node too with kubectl get no --show-labels.

ingress/controllers/gce/controller/utils.go
bprashanth: it will use the default zone if the key isn't found for some reason, i.e older clusters with newer image

ingress/controllers/gce/instances/instances.go
```diff
@@ -35,89 +36,162 @@ const (
...
 	if err != nil {
 		return nil, err
 	}
-	i.snapshotter.Add(name, ig)
+	i.snapshotter.Add(name, struct{}{})
 	return ig, nil
 }
 
-// Add adds the given instances to the Instance Group.
+func (i *Instances) splitNodesByZone(names []string) map[string][]string {
```
freehan: a comment above might be better

ingress/controllers/gce/backends/backends.go
freehan: getBackendsByIGs?

ingress/controllers/gce/backends/backends.go
```diff
@@ -201,18 +201,31 @@ func (b *Backends) List() ([]interface{}, error) {
...
+}
+
 // edgeHop checks the links of the given backend by executing an edge hop.
 // It fixes broken links.
-func (b *Backends) edgeHop(be *compute.BackendService, ig *compute.InstanceGroup) error {
-	if len(be.Backends) == 1 &&
-		utils.CompareLinks(be.Backends[0].Group, ig.SelfLink) {
-		return nil
+func (b *Backends) edgeHop(be *compute.BackendService, igs []*compute.InstanceGroup) error {
+	beIGs := sets.String{}
```
freehan: nit: s/beIGs/backendIGLinks

I feel like there are too many acronym used here.

freehan:
mostly nits. Feel free to update and self LGTM.

bprashanth:
Updated nits, merging

PR #1134, opened 2016-06-03
mungegithub: Delete all Jenkins code - we use GCS

eparis:


eparis:
@lavalamp should be easy to review, but I'd like a second set of eyes

lavalamp:
LGTM, merge at will

PR #1138, opened 2016-06-04
[keepalived-vip]: Update keepalived and godeps

aledbf:


aledbf:
ping @bprashanth

keepalived-vip/Makefile
```diff
@@ -6,7 +6,7 @@ PREFIX = gcr.io/google_containers/kube-keepalived-vip
 BUILD_IMAGE = build-keepalived
 
 controller: clean
-	CGO_ENABLED=0 GOOS=linux godep go build -a -ldflags '-w' -o kube-keepalived-vip
+	CGO_ENABLED=0 GOOS=linux go build -a -ldflags '-w' -o kube-keepalived-vip
```
bprashanth: this will use gopath right? or am i missing something

aledbf: yes

bprashanth: why do we want that? hermetic builds are good, no?

aledbf: I'm sorry, in go 1.6 uses the vendor dir (if exists)

eparis: Just FYI: godep also used GOPATH. It just used Godep/_workspace first. Just like 1.6+vendor will use vendor/ first.

bprashanth: Ah, yes

bprashanth:
LGTM

PR #1139, opened 2016-06-05
[404-server] Parse command line flags

louis-paul:
The command line flags were ignored, even though they are defined in the code. This made impossible changing the port of the server.


luxas:
@louis-paul Thanks for the PR! Seems that you had to wait absolutely too long, sorry!
Now when the bot is auto-assigning, things will get better.

LGTM, but I don't have the permission to push the new image 
(I'm looking at you @mikedanese @bprashanth @vishh :smile:)

@louis-paul But you should bump the `TAG` value before I merge it though.

@eparis The `size/*` labels don't exist in this repo, so the bot couldn't apply them.

louis-paul:
@luxas No problem, I updated the tag to 1.2.

bprashanth:
LGTM thanks, yep this got dropped because no one realized sorry

PR #1125, opened 2016-06-02
Multi-architecture dnsmasq container builds + rename container to kub…

girishkalele:
Fixed builds for architecture amd64 - don't have a solution for arm/arm64/ppc yet.

PR #1146, opened 2016-06-06
Add a tiny-glibc image

luxas:
Please review and push: @girishkalele 

cc @bprashanth

### DIFF
--- a/tiny-glibc/Makefile
+++ b/tiny-glibc/Makefile
@@ -1,0 +1,74 @@
+# Copyright 2016 The Kubernetes Authors All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Cross-platform image that is based on busybox, but includes glibc so dynamically linked binaries may run inside. It's only 11 MB!
+# make [ARCH=amd64] REGISTRY=[gcr.io/google_containers] (build|push)
+
+.PHONY:	build push
+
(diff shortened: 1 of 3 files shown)


girishkalele:
Ah I see - the base image I was using is missing glibc - I thought since I was using the same base image as the pause.c pause image, it would be compatible for dnsmasq C code and contain the needed C libraries.

tiny-glibc/Makefile
```diff
@@ -0,0 +1,72 @@
...
+	docker run -it --rm -v $(TEMP_DIR):/build $(BIND_QEMU_VOLUME_ARG) $(BUILDIMAGE) /bin/bash -c \
+		"apt-get update \
+		&& apt-get download libc6 \
+		&& mkdir libc \
+		&& dpkg-deb -x *.deb libc \
+		&& cd libc \
+		&& tar -czf /build/glibc.tar.gz ."
+
+	docker build -t $(REGISTRY)/$(IMAGE)-$(ARCH):$(TAG) $(TEMP_DIR)
+
```
girishkalele: And wipe out the temp directory here ? And probably after all targets (since the mktemp runs for all invocations).

girishkalele:
@luxas 

Not familiar with this cross-compilation stuff (is it running a qemu emulator inside Docker containers ?) but apart from my comment on the temp directory cleanup, this LGTM.

Just curious, do you have a multi-architecture e2e test framework ?
Perhaps using QEMU emulators to emulate the master and minion nodes ?

luxas:
Updated. Please push.
I've documented things in https://github.com/kubernetes/kubernetes/pull/26863

thockin:
Can you explain WHY in more detail?  This looks pretty brittle, and I *really* don't want to be in the business of maintaining a base-image in light of security updates and so forth.

As an alternative, there is already `busybox:glibc` on docker hub.  Can we start from there?  Can we convince THEM to build this for multi-arch automatically?

luxas:
WTF!!! I hadn't noticed there are `glibc` images already cross-compiled on the Hub!

Why: We must have small cross-compiled images with glibc (for dynamic linking) for dnsmasq, probably minikube/localkube/hyperkube and in some other places as well, but it now seems we've found the base images.

Thanks for questioning, I'll delete the folder directly. I just was thinking about helping #1125 as fast as possible.

PR #1147, opened 2016-06-06
reference previous issues for suites

lavalamp:
If we file a new issue for a broken test suite, link to previous issues for that suite.

### DIFF
--- a/mungegithub/mungers/flake-manager.go
+++ b/mungegithub/mungers/flake-manager.go
@@ -127,6 +127,17 @@ func (p *FlakeManager) isIndividualFlake(f cache.Flake) bool {
 	return true
 }
 
+func (p *FlakeManager) listPreviousIssues(title string) []string {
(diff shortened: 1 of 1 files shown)


lavalamp:
@eparis @fejta

fejta:
LGTM

mungegithub/mungers/flake-manager.go
```diff
@@ -127,6 +127,17 @@ func (p *FlakeManager) isIndividualFlake(f cache.Flake) bool {
 	return true
 }
 
+func (p *FlakeManager) listPreviousIssues(title string) []string {
+	if previousIssues := p.finder.AllIssuesForKey(title); len(previousIssues) > 0 {
```
eparis: given how this is used the test for len() > 0 seems useless. You could just return the empty []string{} but this is fine too.

PR #1155, opened 2016-06-08
nginx-ingress-controller: Use DefaultClientConfig to create clientConfig

whs:
This PR use the configuration generated from DefaultClientConfig, instead of trying to generate one so that the server URL and credentials can be specified from the command line.

The runtime pod detection haven't been improved, and probably will fail if you use remote server. To workaround this, use `--running-in-cluster=false` and manually load a load balancer service configuration.

An image with this PR is available at [willwill/nginx-ingress-controller](https://hub.docker.com/r/willwill/nginx-ingress-controller/) on Docker Hub.

### DIFF
--- a/ingress/controllers/nginx/main.go
+++ b/ingress/controllers/nginx/main.go
@@ -90,8 +90,8 @@ var (
 func main() {
 	var kubeClient *unversioned.Client
 	flags.AddGoFlagSet(flag.CommandLine)
-	flags.Parse(os.Args)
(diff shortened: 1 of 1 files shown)


whs:
I signed it!

whs:
I'm retracting this patch. It seems that it might break the original use case to the point of it being unsupported. (Haven't tested for that though, I decided to go with traefik instead)

I'll keep the docker image up just in case anyone tried to run this with insecured apiserver.

PR #1157, opened 2016-06-08
Cluster-autoscaler: basic metrics for ca

mwielgus:
cc: @piosz @jszczepkowski

### DIFF
--- a/cluster-autoscaler/cluster_autoscaler.go
+++ b/cluster-autoscaler/cluster_autoscaler.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"flag"
+	"net/http"
 	"net/url"
 	"os"
 	"time"
@@ -30,10 +31,12 @@ import (
 	kube_client "k8s.io/kubernetes/pkg/client/unversioned"
 
 	"github.com/golang/glog"
+	"github.com/prometheus/client_golang/prometheus"
 )
 
 var (
 	migConfigFlag           config.MigConfigFlag
+	address                 = flag.String("address", ":8080", "The address to expose prometheus metrics.")
 	kubernetes              = flag.String("kubernetes", "", "Kuberentes master location. Leave blank for default")
 	cloudConfig             = flag.String("cloud-config", "", "The path to the cloud provider configuration file.  Empty string for no configuration file.")
 	verifyUnschedulablePods = flag.Bool("verify-unschedulable-pods", true,
@@ -54,6 +57,12 @@ func main() {
 	flag.Var(&migConfigFlag, "nodes", "sets min,max size and url of a MIG to be controlled by Cluster Autoscaler. "+
 		"Can be used multiple times. Format: <min>:<max>:<migurl>")
 	flag.Parse()
+
+	http.Handle("/metrics", prometheus.Handler())
+	err := http.ListenAndServe(*address, nil)
+	if err != nil {
+		glog.Fatalf("Failed to start http server metrics: %v", err)
+	}
 
 	url, err := url.Parse(*kubernetes)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


cluster-autoscaler/cluster_autoscaler.go
```diff
@@ -50,11 +53,52 @@ var (
 		"How often scale down possiblity is check")
 )
 
+var (
```
wojtek-t: Can you please move metrics to a separate file (or even package)?
As an example:
https://github.com/kubernetes/kubernetes/blob/master/pkg/apiserver/metrics/metrics.go

cluster-autoscaler/cluster_autoscaler.go
```diff
@@ -50,11 +53,52 @@ var (
...
+	lastTimestamp = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: "cluster_autoscaler",
+			Subsystem: "main",
```
wojtek-t: I think we don't set subsystem anywhere else. But that's probably not a big problem.

cluster-autoscaler/cluster_autoscaler.go
```diff
@@ -169,7 +216,15 @@ func main() {
 				if len(unschedulablePodsToHelp) == 0 {
 					glog.V(1).Info("No unschedulable pods")
 				} else {
+					scaleUpStart := time.Now()
+					lastTimestamp.WithLabelValues("scale_up").Set(float64(scaleUpStart.Unix()))
 					scaledUp, err := ScaleUp(unschedulablePodsToHelp, nodes, migConfigs, gceManager, kubeClient, predicateChecker, recorder)
+
+					duration.WithLabelValues("scale_up").Observe(
```
wojtek-t: Can you opaque these 4 lines into function? The same lines (that different only with the string and start time) you have in 4 different places

cluster-autoscaler/cluster_autoscaler.go

cluster-autoscaler/cluster_autoscaler.go

cluster-autoscaler/cluster_autoscaler.go
```diff
@@ -215,7 +244,13 @@ func main() {
...
 					}
 				}
+				updateDuration("main", loopStart)
 			}
 		}
 	}
 }
+
+func updateDuration(label string, start time.Time) {
+	duration.WithLabelValues("main").Observe(durationToMicro(start))
```
wojtek-t: s/main/label/

same below

cluster-autoscaler/cluster_autoscaler.go
```diff
@@ -55,6 +58,12 @@ func main() {
 		"Can be used multiple times. Format: <min>:<max>:<migurl>")
 	flag.Parse()
 
+	http.Handle("/metrics", prometheus.Handler())
```
wojtek-t: git add on the new file please

cluster-autoscaler/cluster_autoscaler.go

cluster-autoscaler/cluster_autoscaler.go

cluster-autoscaler/cluster_autoscaler.go
```diff
@@ -109,6 +118,9 @@ func main() {
 		select {
 		case <-time.After(time.Minute):
 			{
+				loopStart := time.Now()
+				lastTimestamp.WithLabelValues("main").Set(float64(loopStart.Unix()))
```
wojtek-t: how about making this also a function, like:
updateLastTimestamp(labels, time)

wojtek-t:
One comment - other than that lgtm

cluster-autoscaler/cluster_autoscaler.go
mwielgus: Done

PR #1165, opened 2016-06-09
Cluster-autoscaler: use port 8085 for ca prometheus metrics

mwielgus:
CA uses host network and 8080 is already used.


piosz:
LGTM

PR #1169, opened 2016-06-09
cluster-autoscaler: Don't delete nodes with pods with local storage

### LINKED ISSUE #1162: cluster-autoscaler: We shouldn't delete nodes with pods with local storage
We shouldn't delete nodes which have pods with local storage like `emptyDir` or `hostPath`.

@piosz @jszczepkowski

fgrzadkowski:
Don't delete nodes with pods with local storage.

Add flags to control whether we fail on draining nodes with kube-system pods & pods with local storage.

Fixes https://github.com/kubernetes/contrib/issues/1162

@piosz @jszczepkowski

### DIFF
--- a/cluster-autoscaler/simulator/cluster.go
+++ b/cluster-autoscaler/simulator/cluster.go
@@ -17,6 +17,7 @@ limitations under the License.
 package simulator
 
 import (
+	"flag"
 	"fmt"
 	"math"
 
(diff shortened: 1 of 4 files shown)


cluster-autoscaler/simulator/cluster.go
```diff
@@ -29,6 +30,14 @@ import (
 	"github.com/golang/glog"
 )
 
+var (
+	failOnSystemPods = flag.Bool("fail-on-system-pods", true,
```
mwielgus: Maybe skip-nodes-with-system-pods? CA doesn't **fail** on system pods.

cluster-autoscaler/simulator/cluster.go
```diff
@@ -29,6 +30,14 @@ import (
 	"github.com/golang/glog"
 )
 
+var (
```
piosz: How about having those flags in the main file? I really don't like having flags spread across multiple files.

fgrzadkowski: But then we'd have to push through all of the classes/methods. If follow the same logic should also have all the constants in the main file?

I think it's much cleaner if we have in modules that they actually control.

cluster-autoscaler/simulator/cluster.go
fgrzadkowski: I'll it! Should we also use `skip-nodes-with-local-storate`?

davidopp:
I assume you've seen kubernetes/kubernetes#26667

fgrzadkowski:
@davidopp To be honest I haven't seen it, but it's even better! We have a faster/simpler version of drain logic that we are using before we call the real `kubectl drain`. This PR adds support there in the simpler version.

fgrzadkowski:
Done. PTAL.

mwielgus:
LGTM

PR #1170, opened 2016-06-09
cluster-autoscaler: Increase main loop frequency to 10s

### LINKED ISSUE #1163: cluster-autoscaler: Increase how often we try to scale up/down
Currently it's once a minute. That's very rare. We should aim for something like `10s`.

@piosz @jszczepkowski

fgrzadkowski:
Increase main loop frequency to 10s

Fixes https://github.com/kubernetes/contrib/issues/1163


mwielgus:
Please make it a flag

fgrzadkowski:
@mwielgus Done. PTAL.

mwielgus:
LGTM

PR #1171, opened 2016-06-09
cluster-autoscaler: Rename reservation to utilization

### LINKED ISSUE #1168: cluster-autoscaler: Rename reservation to utilization
In many places we are using word `reservation` which has a special meaning. We should rename it all to utilization!

fgrzadkowski:
Rename reservation to utilization

Fixes https://github.com/kubernetes/contrib/issues/1168


mwielgus:
LGTM

PR #1173, opened 2016-06-09
cluster-autoscaler: Rename scale down function to better inform what they do.

### LINKED ISSUE #1172: cluster-autoscaler: Rename scale-down function to better describe what they do
Currently number of functions have `utilization` in their name, even though we actually checking if the node can be removed and utilization is only an optimization to find good candidates. We should rename those functions for better readability.

fgrzadkowski:
Rename scale down function to better inform what they do.

Fixes https://github.com/kubernetes/contrib/issues/1172


mwielgus:
LGTM

PR #1174, opened 2016-06-09
cluster-autoscaler: Log that we are capping to MAX.

### LINKED ISSUE #1074: cluster-autoscaler: Log that we are capping by MAX in scale-up
I think that currently we don't do this. If we want to add 4 machines, but due MAX_NODES we can only add 1 node we are not logging anything (or I don't see it.)

fgrzadkowski:
Log that we are capping to MAX.

Fixes https://github.com/kubernetes/contrib/issues/1074


mwielgus:
LGTM

PR #1175, opened 2016-06-09
addon-resizer: Bumped version to 1.1

piosz:


Q-Lee:
If this causes a break, then I'll be a monkey's uncle.

PR #1189, opened 2016-06-10
DNSPerf container build and RC yaml

girishkalele:
This is the DNSPerf container used to run benchmarks and reliability tests against the kubedns pod

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1189)

### DIFF
--- a/dnsperf/Dockerfile
+++ b/dnsperf/Dockerfile
@@ -1,0 +1,34 @@
+FROM ubuntu
+MAINTAINER Abhishek Shah "abshah@google.com"
+
+RUN apt-get -q update
+
+#Install dependencies needed for building dnsperf
+RUN apt-get install -y build-essential
+RUN apt-get install -y bind9utils libbind-dev libkrb5-dev libssl-dev libcap-dev libxml2-dev libgeoip-dev
+
+#Install curl, wget is not required, but adding it in for convenience
+RUN apt-get install -y curl wget
+
+#Fetch dnsperf
+RUN curl ftp://ftp.nominum.com/pub/nominum/dnsperf/2.0.0.0/dnsperf-src-2.0.0.0-1.tar.gz -O
+RUN tar xfvz dnsperf-src-2.0.0.0-1.tar.gz
+
+#Get query data file for dnsperf
+RUN wget ftp://ftp.nominum.com/pub/nominum/dnsperf/data/queryfile-example-current.gz
+COPY queryfile-local-services queryfile-local-services
+RUN gunzip queryfile-example-current.gz
+RUN cat queryfile-example-current | head -10000 > queryfile-example-short
+
+#Build dnsperf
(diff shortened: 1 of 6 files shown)


girishkalele:
Looks like these checks run by Travis CI should be part of the pre-commit hook for the contrib repo.

From the travis-ci logs:
```
$ hack/for-go-proj.sh install
$ hack/verify-all.sh -v
$ hack/for-go-proj.sh test
```

thockin:
Will this be re-targetted at a standalone repo?  I am worried that we're trading a too-flat hierarchy inside one repo for a too-flat hierarchy in one org.

@bgrant0607 @gmarek

gmarek:
I wholeheartedly agree with @thockin - can we start by just having "dedicated" grab-bags, as @lavalamp put it with defined scople, and flatten it further it'll not be enough?

girishkalele:
This is probably a standalone benchmark that then potentially can get re-used for testing other environments.

What is a grab-bag ? I wasn't in on the original @lavalamp comment.

gmarek:
It's in other thread and it means a repo with everything (currently: contrib).

thockin:
The idea is that, rather than 100 individual test repos we would have one
"test stuff" repo that help a number of test projects.

gmarek:
That's what I wanted to have. I'll reiterate the idea on kubernetes-dev

lavalamp:
If you want that, definitely don't call the repo "kubemark" :)

Is there a clear difference between test-tools and test-infra (which we
already have)?

girishkalele:
This is not test-infra, so kubernetes/test-tools would probably be the place then (it doesn't exist yet).

thockin:
kubernetes/benchmarks ?

kubernetes/perf-tests ?

Do we need to get more proecise in the name?

gmarek:
It depends - I'm OK with both. Let's move back to kuberentes-dev.

thockin:
@gmarek did we get any consensus on this?  I am tired of waffling.  I propose `kubernetes/perftests` and we put kubemark, dns perf, network perf, etc there.  ACK and we can create repo tomorrow.

gmarek:
I agree. No one replied to my last proposal, so I assume everyone agreed (or are too busy to reply, but this means they don't care too much). `kubernetes/perftests` SGTM, but I'll be OOO till the end of the week, so I'll be able to create all boilerplate files only after I come back.

lavalamp:
If it's going to be a grab bag, why limit it with the name "perf"? Why not
`test-tools` or something?

thockin:
some amount of scope limiting.

thockin:
repo perf-tests has been created

girishkalele:
Moved pull request to new repository https://github.com/kubernetes/perf-tests/pull/2

PR #1193, opened 2016-06-10
File issues for failures even in non-blocking jobs

### LINKED ISSUE #1153: auto-file issues for test runs even if they don't block the merge queue
Now that we're autofiling issues and people are getting out of the habit of looking at the jekins jobs, we need to start filing issues for test failures on jobs that don't block the submit queue. e.g., https://github.com/kubernetes/kubernetes/issues/27011 if auto filed would have dozens if not hundreds of comments.

lavalamp:
Fixes #1153

### DIFF
--- a/mungegithub/mungers/e2e/e2e.go
+++ b/mungegithub/mungers/e2e/e2e.go
@@ -49,8 +49,9 @@ type BuildInfo struct {
 // RealE2ETester is the object which will get status from a google bucket
 // information about recent jobs
 type RealE2ETester struct {
-	BlockingJobNames   []string
-	WeakStableJobNames []string
+	BlockingJobNames    []string
+	NonBlockingJobNames []string
+	WeakStableJobNames  []string
 
 	sync.Mutex
 	BuildStatus          map[string]BuildInfo // protect by mutex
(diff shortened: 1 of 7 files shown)


mungegithub/mungers/e2e/e2e.go
```diff
@@ -49,8 +49,9 @@ type BuildInfo struct {
 // RealE2ETester is the object which will get status from a google bucket
 // information about recent jobs
 type RealE2ETester struct {
-	JobNames           []string
-	WeakStableJobNames []string
+	BlockingJobNames    []string
+	NonBlockingJobNames []string
+	WeakStableJobNames  []string
```
eparis: We really need to find a better way to get this info...  Exposed by something/somewhere else...

lavalamp: Agreed, I should be able to use this without the submit queue.
On Jun 10, 2016 8:39 PM, "Eric Paris" <notifications@github.com> wrote:

In mungegithub/mungers/e2e/e2e.go
<https://github.com/kubernetes/contrib/pull/1193#discussion_r66700157>:

> @@ -49,8 +49,9 @@ type BuildInfo struct {
>  // RealE2ETester is the object which will get status from a google bucket
>  // information about recent jobs
>  type RealE2ETester struct {
> -	JobNames           []string
> -	WeakStableJobNames []string
> +	BlockingJobNames    []string
> +	NonBlockingJobNames []string
> +	WeakStableJobNames  []string

We really need to find a better way to get this info... Exposed by
something/somewhere else...

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/kubernetes/contrib/pull/1193/files/5e9be452b0e037ef10404eef80538652d53dfd15#r66700157>,
or mute the thread
<https://github.com/notifications/unsubscribe/AAngli_Nc0f5F99B2OmwtQXGSDsgYpsoks5qKi36gaJpZM4IzX9w>
.

lavalamp:
@eparis any issues with this? I'd like to get it running later today if possible.

eparis:
Nope, looks good to me.

Can I get an EBS volume (and maybe a PV or enough info I can create the PV) today? Maybe 1G?

lavalamp:
Is a PD OK? I'll try to set one up and make a PR with config changes. (But will have to be this afternoon, I'm about to have a big block of meetings)

eparis:
Oh crap, I totally mean PD, yes

PR #1195, opened 2016-06-13
Enable operating on microvalues in addong resizer

wojtek-t:
Since we want to enable passing values smaller than e.g. 1 millicore per node, we need a solution other  than operating on MilliValues.

@gmarek

### DIFF
--- a/addon-resizer/Makefile
+++ b/addon-resizer/Makefile
@@ -25,7 +25,7 @@
 OUT_DIR = build
 PACKAGE = k8s.io/contrib/addon-resizer
 PREFIX = gcr.io/google_containers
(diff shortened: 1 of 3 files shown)


piosz:
cc @Q-Lee

piosz:
Looks good, please add a test for exponential estimator (should return 0.308) and some explanation to the PR why do need it (something like [this comment](https://github.com/kubernetes/contrib/pull/1195/files#diff-6a14b614a30610c5f437b2c717eeb57bR66))

wojtek-t:
@piosz - comment applied; PTAL

piosz:
LGTM

PR #1198, opened 2016-06-13
Adding kubernetes-e2e-gce-federation to list of merge blocking suites

nikhiljindal:
cc @kubernetes/sig-cluster-federation @lavalamp 
Will wait for the test to be green for a while before merging this.

Ref https://github.com/kubernetes/kubernetes/issues/26723

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1198)

### DIFF
--- a/mungegithub/mungers/submit-queue.go
+++ b/mungegithub/mungers/submit-queue.go
@@ -401,6 +401,7 @@ func (sq *SubmitQueue) AddFlags(cmd *cobra.Command, config *github.Config) {
 		"kubernetes-build",
 		"kubernetes-test-go",
 		"kubernetes-e2e-gce",
+		"kubernetes-e2e-gce-federation",
 		"kubernetes-e2e-gce-slow",
 		"kubernetes-e2e-gce-serial",
 		"kubernetes-e2e-gke-serial",


goltermann:
@nikhiljindal What issue is blocking these tests being green?

nikhiljindal:
Fixes waiting to be merged: https://github.com/kubernetes/kubernetes/pull/27407, https://github.com/kubernetes/kubernetes/pull/27411 and https://github.com/kubernetes/kubernetes/pull/27390.
Some of the other fixes are being discussed in https://github.com/kubernetes/kubernetes/issues/26762

goltermann:
@nikhiljindal each of the 4 blockers you mentioned have been merged.

nikhiljindal:
The tests are still not green.
Sent https://github.com/kubernetes/contrib/pull/1242 so that we have automatic issue filing for failing tests.

nikhiljindal:
We should be ready to merge this once https://github.com/kubernetes/kubernetes/issues/28615 is fixed.

quinton-hoole:
Yup, agreed.

lavalamp:
Do you have a presubmit test that covers this? I'm not turning on blocking post submits that aren't covered by presubmits.

nikhiljindal:
By presubmit, do you mean adding this suite to PR builder tests?
We have an issue open for it: https://github.com/kubernetes/test-infra/issues/155, but havent implemented it yet.

lavalamp:
Yes.

quinton-hoole:
Sorry, I typed up a long response to this yesterday, but must not have submitted it.  I will try again, but in summary:

1. we already have this problem (many test jobs, like -slow, -serial, and several others are already merge-blockers but not present in pre-submits.
2. There is some work to do before Federation e2e tests can be added to presubmit checks (see the issue @nikhiljindal quotes above).

quinton-hoole:
@lavalamp I would strongly suggest that we merge this PR, and address  kubernetes/test-infra#155 separately as the next step.  Unless I'm missing something?

quinton-hoole:
`belated comment from yesterday, that I accidentally failed to push the submit button on :-) `

@lavalamp There is currently no way to run e2e tests against multiple cluster deployments in the presubmit tests (at least that I'm aware of).  The presubmit tests currently spin up a single cluster, and run a predefined set of tests against that single cluster.

To add the federation tests to presubmit, we'd have to teach the pre-submit test infrastructure to spin up multiple "clusters" (in this case one of those would actually be a federation of clusters), and then run a different set of tests across each cluster, presumably in parallel, to keep elapsed time to a minimum).

Besides that, we already have this problem.  The following test jobs are merge-blockers but not included in presubmit tests:

```
		"kubernetes-e2e-gce-slow",		  	
  		"kubernetes-e2e-gce-serial",		  	
  		"kubernetes-e2e-gke-serial",		  	
 		"kubernetes-e2e-gke",		 		
 		"kubernetes-e2e-gke-slow",		 		
 		"kubernetes-e2e-gce-scalability",		 		
 		"kubernetes-kubemark-5-gce",		 		
```

lavalamp:
We chatted offline and seem to agree that getting a presubmit test running first is best for the whole team, even though it will cause federation some pain until it is in.

quinton-hoole:
Ack! I agree.

madhusudancs:
Is there a middle ground here? Can we run federation tests against a single cluster in the pre-submit. That should help us catch breaking changes in the Kubernetes code we depend on (code that we import as libraries from `k8s.io/kubernetes/pkg`). For example, this would have helped us catch https://github.com/kubernetes/kubernetes/pull/29123. 

I understand this is less useful than running the tests against multiple clusters, but it is not useless. And it helps us have something, instead of having nothing, until we resolve all the other issues.

lavalamp:
I'm fine with adding that as a presubmit. Warning, though-- if we get a
bunch of post-submit breaks even with that in place, then I'll kick the
post-submit job to nonblocking.

madhusudancs:
@lavalamp fair enough.

test-foxish:
recomputing cla status...

PR #1199, opened 2016-06-14
fix UI for flakes & nonblocking suites

lavalamp:
Tired of the UI being totally wrong and confusing. Turns out to be easy.

image


rmmh:
That yellow/gray contrast is pretty poor, could you use orange instead?

lavalamp:
done

lavalamp:
This is an improvement and I have to restart anyway.

PR #1201, opened 2016-06-14
make resolutions work in weak stable jobs

lavalamp:


lavalamp:
Merge queue is blocked right now due to this so I'm just going to go ahead and push and restart.

PR #1203, opened 2016-06-14
Display when the test are 'yellow'

eparis:
troubled

They aren't quite red..  So give a slightly different header and message
for ignorable flakes and manual overrides


eparis:
@lavalamp Ha, I did this yesterday and never sent a PR.

PR #1204, opened 2016-06-14
Add emergency stop button

lavalamp:


mungegithub/mungers/submit-queue.go
```diff
@@ -466,6 +506,10 @@ func (sq *SubmitQueue) e2eStable(aboutToMerge bool) bool {
 	wentUnstable := false
 
 	stable, ignorableFlakes := sq.e2e.GCSBasedStable()
+	if stable && sq.emergencyMergeStop() {
+		stable = false
+		glog.Errorf("E2E counts as not stable because an emergency merge stop is in progress.")
```
eparis: Is this going to be too spammy? Maybe it's ok.

eparis: I'm worried this is going to log as fast as the CPU will let it log if the tests are all green, there is something in the queue, but it has been stopped...

lavalamp: I will adjust.

lavalamp:
Updated, but I'm pretty sure it polls once every 30 seconds in that case instead of constantly.

lavalamp:
Merge queue is stopped so I'm going to push.

PR #1205, opened 2016-06-14
add auto prioritization for flake issues

### LINKED ISSUE #1206: Auto Prioritize Flake Issues
Now we have flake issue auto filing. The natural next step is to auto prioritize this issues.

freehan:
fixes: #1206


mungegithub/mungers/sync/issue-sync.go
```diff
@@ -201,7 +228,7 @@ func (s *IssueSyncer) updateIssue(obj *github.MungeObject, source IssueSource) e
 
 // createIssue makes a new issue for the given item. If we know about other
 // issues for the item, then they'll be referenced.
-func (s *IssueSyncer) createIssue(source IssueSource) (issueNumber int, err error) {
+func (s *IssueSyncer) createIssue(source IssueSource) (int, *github.MungeObject, error) {
```
rmmh: This should probably return `(issue *github.MungeObject, err error)`.

freehan: Fixed

mungegithub/mungers/sync/issue-sync.go
```diff
@@ -105,25 +131,26 @@ func (s *IssueSyncer) Sync(source IssueSource) error {
...
+			return fmt.Errorf("error making issue for %v: %v", source.ID, err)
+		}
+		issueNum := *obj.Issue.Number
+		s.finder.Created(source.Title(), issueNum)
 	}
-	s.finder.Created(source.Title(), n)
 	s.synced.Insert(source.ID())
-	return nil
+
+	return s.autoPrioritize(obj)
```
lavalamp: nit: I slightly prefer the old logic. You know when you first file an issue that there's only one occurrence, so you could just file it with the P3 label or whatever-- no need to auto prioritize?

mungegithub/mungers/sync/issue-sync.go
```diff
@@ -221,8 +248,53 @@ func (s *IssueSyncer) createIssue(source IssueSource) (issueNumber int, err erro
...
+func (s *IssueSyncer) autoPrioritize(obj *github.MungeObject) error {
+	comments, err := obj.ListComments()
+	if err != nil {
+		return fmt.Errorf("Failed to list comment of issue")
+	}
+	count := 0
+	for _, c := range comments {
+		// "Failed:" is a special string contained in flake issue filed by k8s-bot
+		// Please make sure it matches the body generated by IssueSource
+		if !robotUser.Has(*c.User.Login) || !strings.Contains(*c.Body, "Failed:") {
```
lavalamp: Maybe push this logic into each issue source? Especially that second condition could vary.

mungegithub/mungers/sync/issue-sync.go
freehan: Good point. Will get it fixed.

mungegithub/mungers/sync/issue-sync.go
freehan: Alright. Will do.

lavalamp:
OK, I think this is pretty close!

* Can you (or did you) try on a separate test repo? And/or unit testing will also satisfy me. I know I've been somewhat bad about testing some of the recent merge queue features.
* The documentation hasn't been updated yet, but if github gives the timestamp of each comment (and I think it does?), I think it might be easy to implement the actual flake priority policy? https://groups.google.com/forum/#!topic/kubernetes-dev/0_5vgSCWUyo I think what you wrote is a little more generous than the policy that @matchstick came up with. :)

freehan:
I have tried it with my own test repo. I can also add some unit tests. 

Alright. I will try to implement the threshold.

lavalamp:
Oh, great! let me know if you think it'll be difficult to implement the
policy.

freehan:
I am adding unit test.

mungegithub/mungers/flake-manager.go
```diff
@@ -270,5 +278,55 @@ func (p *brokenJobSource) Body(newIssue bool) string {
...
+	lastWeek := time.Now().Add(-1 * 7 * 24 * time.Hour)
+	// number of flakes happened in this month
+	monthCount := 0
+	// number of flakes happened in this week
+	weekCount := 0
+
+	for _, c := range comments {
+		// "Failed:" is a special string contained in flake issue filed by flake-manager
+		// Please make sure it matches the body generated by IssueSource.Body()
+		if !sync.RobotUser.Has(*c.User.Login) || !strings.Contains(*c.Body, "Failed:") {
```
matchstick: Will this match the text "Failed"? Or be more selective? I can see the word "Failed" being a pretty common string in an issue.

mungegithub/mungers/flake-manager.go
```diff
@@ -270,5 +278,55 @@ func (p *brokenJobSource) Body(newIssue bool) string {
...
+		if lastWeek.Before(o) {
+			weekCount += 1
+		}
+	}
+
+	// P2: By default
+	// P1: Flake happens more than once in last month.
+	// P0: Flake happens more than once in last week.
+	p := sync.PriorityP2
+	if weekCount >= 2 {
```
matchstick: Sometimes we are going to want to exercise human judgement and say "This is a P1 or P2 or not an issue" despite this algorithm. Can we search for some override to prevent the assignment from happening?

mungegithub/mungers/flake-manager.go
freehan: This matches the comments added by k8s-bot with text "Failed:". 

I could not figure out a better way to identify flake comments added by k8s-bot. Plus, k8s-bot rarely leave comment in issues. It usually comments PRs.

mungegithub/mungers/flake-manager.go
freehan: This is handled by issue syncer. During syncPriority, it will check if the existing priority is higher. If higher, then do nothing.

eparis: It's possible to see if the label was added by the bot, and only change it if it was last set by the bot. I think we do that in some other places...

eparis: https://github.com/kubernetes/contrib/blob/master/mungegithub/mungers/path_label.go#L155

lavalamp: I think we probably want the bot to increase the priority even if a human had previously set it lower. Since an additional occurrence could mean it is now more important.

lavalamp:
Needs a gofmt :)

mungegithub/mungers/flake-manager.go
lavalamp: Yeah, at the moment the text check isn't even needed, since the bot only comments on an issue to report a failure.

lavalamp:
LGTM

PR #1209, opened 2016-06-14
Separate non-blocking from blocking builds in E2E results page.

goltermann:
Changing http://submit-queue.k8s.io/#/e2e into:

nonblocking

Also making it easy to just run a static server by pulling down current data, then running from submit-queue/www (good idea @rmmh)

### DIFF
--- a/mungegithub/Makefile
+++ b/mungegithub/Makefile
@@ -69,6 +69,12 @@ secret:
 clean:
 	rm -f mungegithub $(APP)/local.deployment.yaml $(APP)/local.secret.yaml
 
+# pull down current public queue state, and run UI based off that data
+ui-stub:
+	@/bin/bash -c "wget -q -r -nH -P ./submit-queue/www http://submit-queue.k8s.io/{prs,github-e2e-queue,history,sq-stats,stats,users,health,google-internal-ci,priority-info,merge-info}; \
+	pushd ./submit-queue/www; \
+	python -m SimpleHTTPServer;"
+
 help:
 	@echo "ENVIRONMENT VARS:"
 	@echo " REPO=       repository for the docker image being build. Default: $(REPO)"
--- a/mungegithub/submit-queue/www/index.html
+++ b/mungegithub/submit-queue/www/index.html
@@ -148,12 +148,21 @@
               <md-toolbar class="md-whiteframe-z2">
                 <h2 class="md-toolbar-tools">End-to-End Results</h2>
               </md-toolbar>
+              <h2 class="md-title" ng-show="cntl.builds.length > 0">Queue-Blocking Builds</h2>
               <md-chips ng-model="cntl.builds" readonly="true">
                 <md-chip-template title="{{$chip.msg}}">
                   <span style="color: {{$chip.color}}">{{$chip.state}}</span> <a ng-href="http://ci-test.k8s.io/{{$chip.name}}/{{$chip.id}}/">{{$chip.name}}</a>
                   <span style="color: {{$chip.color}}">{{$chip.msg}}</span> <span title="Job stability">{{$chip.stability}}</span>
                 </md-chip-template>
               </md-chips>
+              <h2 class="md-title" ng-show="cntl.nonBlockingBuilds.length > 0">Non-Queue-Blocking Builds</h2>
+              <md-chips ng-model="cntl.nonBlockingBuilds" readonly="true">
+                <md-chip-template title="{{$chip.msg}}">
+                  <span style="color: {{$chip.color}}">{{$chip.state}}</span> <a ng-href="http://ci-test.k8s.io/{{$chip.name}}/{{$chip.id}}/">{{$chip.name}}</a>
+                  <span style="color: {{$chip.color}}">{{$chip.msg}}</span> <span title="Job stability">{{$chip.stability}}</span>
+                </md-chip-template>
+              </md-chips>
+              <br>
               <br>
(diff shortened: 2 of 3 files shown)


lavalamp:
Wow, looks nice! I'll let @eparis take a look, I don't know js ;)

mungegithub/Makefile
```diff
@@ -69,6 +69,12 @@ secret:
 clean:
 	rm -f mungegithub $(APP)/local.deployment.yaml $(APP)/local.secret.yaml
 
+# pull down current public queue state, and run UI based off that data
+ui-stub:
+	@/bin/bash -c "wget -r -nH -P ./submit-queue/www http://submit-queue.k8s.io/{prs,github-e2e-queue,history,sq-stats,stats,users,health,google-internal-ci,priority-info,merge-info,health}; \
+	pushd ./submit-queue/www; \
+	python -m SimpleHTTPServer;"
```
eparis: Now this looks cool....  However I'm unaware of the @/bin/bash format. How does that differ from $(shell ....)

eparis:
This is what I started to do yesterday when I got mad about the whole design and started to rip the e2e out entirely.  People just want me to have to work on rebases   :)   (I love it)

mungegithub/submit-queue/www/script.js
```diff
@@ -235,15 +238,24 @@ function SQCntl(dataService, $interval, $location) {
           failedBuild = true;
       }
       obj.stability = '';
-      result.push(obj);
+      if (!obj.msg) {
+        result.push(obj);
+      } else {
+        if (obj.msg.includes("[nonblocking]")) {
```
eparis: can/should we remove [nonblocking] from the obj.msg here? It's just a distraction at this point.

mungegithub/Makefile
goltermann: It's basically http://stackoverflow.com/questions/5193048/bin-sh-pushd-not-found.  If I understand correctly, $(shell ...) uses the default shell, which may not have pushd available.

mungegithub/submit-queue/www/script.js
goltermann: Fixed.

mungegithub/Makefile
lavalamp: TIL

mungegithub/submit-queue/www/script.js
```diff
@@ -235,15 +238,25 @@ function SQCntl(dataService, $interval, $location) {
           failedBuild = true;
       }
       obj.stability = '';
-      result.push(obj);
+      if (!obj.msg) {
+        result.push(obj);
+      } else {
+        if (obj.msg.includes('[nonblocking]')) {
+          obj.msg = ''
```
lavalamp: I suggest not clearing it, just remove the '[nonblocking]' prefix since the rest of the message may say something useful.

goltermann: I don't think there's a case where there's more to the message now, but that's good future proofing.  Fixed.

mungegithub/Makefile
```diff
@@ -69,6 +69,12 @@ secret:
 clean:
 	rm -f mungegithub $(APP)/local.deployment.yaml $(APP)/local.secret.yaml
 
+# pull down current public queue state, and run UI based off that data
+ui-stub:
+	@/bin/bash -c "wget -q -r -nH -P ./submit-queue/www http://submit-queue.k8s.io/{prs,github-e2e-queue,history,sq-stats,stats,users,health,google-internal-ci,priority-info,merge-info,health}; \
```
rmmh: this has health twice

lavalamp: It's twice as important as the others ;) ;)

mungegithub/submit-queue/www/index.html
```diff
@@ -148,12 +148,21 @@ <h4 class="md-body-2">{{pr.Reason}}</h4>
               <md-toolbar class="md-whiteframe-z2">
                 <h2 class="md-toolbar-tools">End-to-End Results</h2>
               </md-toolbar>
+              <h2 class="md-title" ng-show="cntl.builds.length > 0">Queue-Blocking Builds</h2>
               <md-chips ng-model="cntl.builds" readonly="true">
```
rmmh: Use md-transform-chip to do the filtering here, instead of splitting out builds explicitly. This is also more flexible if there are more categories in the future.

> * md-transform-chip	expression	
> An expression of form myFunction($chip) that when called expects one of the following return values:
> 
> an object representing the $chip input string
> undefined to simply add the $chip input string, or
> null to prevent the 

chip from being appended

mungegithub/submit-queue/www/script.js
```diff
@@ -255,6 +268,13 @@ function SQCntl(dataService, $interval, $location) {
         build.stability = Math.round(percentStable) + "%"
       }
     });
+    angular.forEach(nonBlockingBuilds, function(build) {
```
rmmh: don't copy this function's code, extract it and reuse it.

mungegithub/Makefile
goltermann: Fixed.

mungegithub/submit-queue/www/script.js
goltermann: Fixed.

mungegithub/submit-queue/www/index.html
goltermann: Spent some time on this and it sounds like it would work, but I couldn't get it working.  Can we refactor when needed in a follow-up PR?

eparis: I'm ok with that...

PR #1213, opened 2016-06-15
Fix stats on retest-not-required merges

### LINKED ISSUE #1212: submit queue: merge rate broken
Shows as zero even though it's merging things. #1208 changes this but I don't see how it could break it.

lavalamp:
Fixes #1212

### DIFF
--- a/mungegithub/mungers/submit-queue.go
+++ b/mungegithub/mungers/submit-queue.go
@@ -272,8 +272,15 @@ func getSmoothFactor(dur time.Duration) float64 {
 //    of guess-and-test and intuition. Someone who knows about this stuff
 //    is likely to laugh at the naivete. Point him to where someone intelligent
 //    has thought about this stuff and he will gladly do something smart.
+// Merges that took less than 5 minutes are ignored completely for the rate
+// calculation.
 func calcMergeRate(oldRate float64, last, now time.Time) float64 {
 	since := now.Sub(last)
+	if since <= 5*time.Minute {
+		// retest-not-required PR merges shouldn't affect our best
+		// guess about the rate.
+		return oldRate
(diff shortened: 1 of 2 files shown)


mungegithub/mungers/submit-queue.go
```diff
@@ -287,13 +287,13 @@ func calcMergeRate(oldRate float64, last, now time.Time) float64 {
...
 // returns 'Now()' and the rate.
-// Should be called once after every merge. Also updates sq.totalMerges.
+// Note that this does not update sq.lastMergeTime-- that should be done with
+// every merge, but this function should not be called on retest-not-required
+// merges.
 func (sq *SubmitQueue) updateMergeRate() {
 	now := sq.clock.Now()
 
 	sq.mergeRate = calcMergeRate(sq.mergeRate, sq.lastMergeTime, now)
-	sq.lastMergeTime = now
```
eparis: hmmmm, moving this is going to throw off the rate calculation when the next NON-retest-not-required comes through, no?  I was using the merge rate to figure out 'how many real PRs it can do a day' but now if the bot blocks for 4 hours, restarts and merges a retest-not-required and then in 30 minutes minutes it will calculate on the last PR having taken 30 minutes instead of 4:30...

eparis: I could also be convinced all my crazy math just isn't worth it and we should just keep a list of the last X hours or merges and show the average...

lavalamp: > throw off the rate calculation when the next NON-retest-not-required comes through

I think that's a feature, not a bug. 

I think your math makes numbers that are generally in the right ballpark and we should not throw it away yet :)

PR #1218, opened 2016-06-15
Fix typo "does no" -> "does not"

coryfklein:
My first contribution. I've signed the CLA.


eparis:
Can you update the git commit to use the same e-mail address as you signed the CLA?

coryfklein:
Will do.

PR #1222, opened 2016-06-16
make verify a blocking build

mikedanese:
cc @lavalamp

### DIFF
--- a/mungegithub/mungers/submit-queue.go
+++ b/mungegithub/mungers/submit-queue.go
@@ -51,11 +51,11 @@ const (
 	claYesLabel            = "cla: yes"
 	claHumanLabel          = "cla: human-approved"
 
-	jenkinsE2EContext  = "Jenkins GCE e2e"
-	jenkinsUnitContext = "Jenkins unit/integration"
-	jenkinsNodeContext = "Jenkins GCE Node e2e"
-	travisContext      = "continuous-integration/travis-ci/pr"
-	sqContext          = "Submit Queue"
+	jenkinsE2EContext    = "Jenkins GCE e2e"
+	jenkinsUnitContext   = "Jenkins unit/integration"
+	jenkinsVerifyContext = "Jenkins verification"
+	jenkinsNodeContext   = "Jenkins GCE Node e2e"
+	sqContext            = "Submit Queue"
 
 	retestNotRequiredMergePriority = -1 // used for retestNotRequiredLabel
 	defaultMergePriority           = 3  // when an issue is unlabeled
@@ -67,6 +67,13 @@ var (
 	_ = fmt.Print
 	// This MUST cause a RETEST of everything in the sq.RequiredRetestContexts
 	retestBody = fmt.Sprintf("@%s test this [submit-queue is verifying that this PR is safe to merge]", jenkinsBotName)
+
+	requiredContexts = []string{
+		jenkinsUnitContext,
+		jenkinsE2EContext,
+		jenkinsNodeContext,
+		jenkinsVerifyContext,
+	}
 )
 
 type submitStatus struct {
@@ -453,6 +460,7 @@ func (sq *SubmitQueue) AddFlags(cmd *cobra.Command, config *github.Config) {
 		"kubelet-gce-e2e-ci",
 		"kubernetes-build",
 		"kubernetes-test-go",
+		"kubernetes-verify-master",
 		"kubernetes-e2e-gce",
 		"kubernetes-e2e-gce-slow",
 		"kubernetes-e2e-gce-serial",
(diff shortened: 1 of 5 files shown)


mungegithub/mungers/submit-queue.go
```diff
@@ -51,11 +51,12 @@ const (
...
-	jenkinsE2EContext  = "Jenkins GCE e2e"
-	jenkinsUnitContext = "Jenkins unit/integration"
-	jenkinsNodeContext = "Jenkins GCE Node e2e"
-	travisContext      = "continuous-integration/travis-ci/pr"
-	sqContext          = "Submit Queue"
+	jenkinsE2EContext    = "Jenkins GCE e2e"
+	jenkinsUnitContext   = "Jenkins unit/integration"
+	jenkinsVerifyContext = "Jenkins verification"
+	jenkinsNodeContext   = "Jenkins GCE Node e2e"
+	travisContext        = "continuous-integration/travis-ci/pr"
```
spxtr: Can remove this while you're at it, but it's not a big deal.

mikedanese: done.

eparis:
the 'k8s-bot test this' comment will cause this to be re-tested?  Does this (and node e2e) need to be added to the ping-stale* mungers?

mungegithub/mungers/submit-queue.go
```diff
@@ -466,7 +467,7 @@ func (sq *SubmitQueue) AddFlags(cmd *cobra.Command, config *github.Config) {
 		[]string{},
 		"Comma separated list of jobs in Jenkins to use for stability testing that needs only weak success")
 	cmd.Flags().StringSliceVar(&sq.RequiredStatusContexts, "required-contexts", []string{}, "Comma separate list of status contexts required for a PR to be considered ok to merge")
-	cmd.Flags().StringSliceVar(&sq.RequiredRetestContexts, "required-retest-contexts", []string{jenkinsE2EContext, jenkinsUnitContext, jenkinsNodeContext}, "Comma separate list of statuses which will be retested and which must come back green after the `retest-body` comment is posted to a PR")
+	cmd.Flags().StringSliceVar(&sq.RequiredRetestContexts, "required-retest-contexts", []string{jenkinsE2EContext, jenkinsUnitContext, jenkinsNodeContext, jenkinsVerifyContext}, "Comma separate list of statuses which will be retested and which must come back green after the `retest-body` comment is posted to a PR")
```
lavalamp: Can you put these items on separate lines for readability?

mungegithub/mungers/submit-queue.go
```diff
@@ -51,11 +51,11 @@ const (
...
 	claHumanLabel          = "cla: human-approved"
 
-	jenkinsE2EContext  = "Jenkins GCE e2e"
-	jenkinsUnitContext = "Jenkins unit/integration"
-	jenkinsNodeContext = "Jenkins GCE Node e2e"
-	travisContext      = "continuous-integration/travis-ci/pr"
-	sqContext          = "Submit Queue"
+	jenkinsE2EContext    = "Jenkins GCE e2e"
+	jenkinsUnitContext   = "Jenkins unit/integration"
+	jenkinsVerifyContext = "Jenkins verification"
```
lavalamp: Is there a reason to have these constants listed here? @eparis

eparis: they are hard coded in the 'ping-*' mungers I think.

mikedanese: Should they all be added to the stale ping checks? Currently jenkinsNodeContext is missing from that list.

eparis: I think they should be, but doesn't have to be this PR.

eparis: actually maybe it does need to be. I think many/most things in the queue have never run verification, right? so they'll just fall out of the queue and won't ever come back...

mikedanese: Done.

mikedanese:
Ok tried to do a minimal cleanup. The required builds should really be plumbed through to the mungers eventually though.

mungegithub/mungers/submit-queue.go
```diff
@@ -67,6 +67,13 @@ var (
 	_ = fmt.Print
 	// This MUST cause a RETEST of everything in the sq.RequiredRetestContexts
 	retestBody = fmt.Sprintf("@%s test this [submit-queue is verifying that this PR is safe to merge]", jenkinsBotName)
+
+	requiredContexts = []string{
```
eparis: Can you call this 'requiredRetestContexts' ?

eparis: Actually this is fine, i'm just a little worried/uncomfortable about the usage in the 'required-retest-contexts' flag...  I want that part broken out somehow...

eparis:
@mikedanese @ixdy is the infrastructure going to be able to handle it when we hit it with 50+ 'verification' tests at once?  Or do we need to run some custom stale munger by hand over the next couple days to get caught up?

mungegithub/mungers/submit-queue.go
lavalamp: we need to make these all flags somehow eventually if we want to run this thing against other repos.

lavalamp:
LGTM

mikedanese:
@eparis We can watch it. I think it will be ok to do it and let the queue drain.

spxtr:
If you think there's going to be a huge rush of verify requests then I might suggest merging either early in the morning or late at night when PR Jenkins isn't busy.

eparis:
when it triggers a retest I believe it triggers `test this` which is actually all 4!

lavalamp:
Yeah, we've *got* to fix that...

Issue #799, opened 2016-04-18
Nginx Ingress Controller - Secure default backend and/or support for wildcards?


alanhartless:
Currently it's not possible to support wildcard subdomains for the Nginx ingress controller due to Kubernetes Ingress resources only allowing a legitimate domain for host. In the scenario that a wildcard domain SSL cert needs to be used for a lot of subdomains (think SaaS with custom subdomain names), it's a bit overkill to list every subdomain. 

I thought of two solutions. 

One - support via a flag to secure the default backend server with the default SSL certificate where the backend server can be something more than just a simple 404. This of course limits to a single cert and wouldn't be able to support multiple wildcard domains. 

Two - use a flag with a "wildcard" placeholder that's replaced with `*` when generating the hosts. The placeholder can be a little funny looking since the host is limited to alphanumeric keys (like `0wildcard0`, `wildcard-` or just `wildcard` if for sure not being used. 

I've implemented both for my uses but wanted to see if either (or both) would be something interested in  before I submit a PR (they feel a little hackish :-) ). Thoughts?

It would be GREAT if paths would support Nginx location regex but I know that's an Ingress limitation.

bprashanth:
This is valid:
https://github.com/kelseyhightower/talks/blob/master/kubecon-eu-2016/demo/ingress/ghost.yaml#L12
You can also specify a list of hosts, and a wildcard cert, or just a list of hosts without a cert for something like SNI.

So I guess the request is that we allow wildcard matches in the host field of the ingress *rules*?  

> Two - use a flag with a "wildcard" placeholder that's replaced with * when generating the hosts

This is the specce'd behavior actually: https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/extensions/types.go#L675

@alanhartless can you give me a yaml sketch of the proposed modifcations?

alanhartless:
@bprashanth Sorry for the long delay in replying!

`So I guess the request is that we allow wildcard matches in the host field of the ingress rules?`

That's exactly what I'm looking for. If I could define a wildcard host in the ingress rules that would result in  a nginx config similar to the following, no hack would be needed: 

```
 server {
        listen 80;
        listen 443 ssl http2;

        server_name *.example.com;
```

aledbf:
@alanhartless using `server_name *.example.com;` means only the first matched server (in the generated nginx.conf) will route traffic to the service, the rest is not going to be reachable. That's why you need to specify the host.

aledbf:
[...]

PR #1223, opened 2016-06-16
added domain name replacement for kleenestar. subdomain

from-nibly:
i set up a loop that looks for domains that start with kleenestar. Then replaces that prefix with *. on the nginx side. This is a intermediate solution while we try to fix the validation problem with ingress denying \* wild cards. 

issue link: https://github.com/kubernetes/contrib/issues/799

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1223)

### DIFF
--- a/ingress/controllers/nginx/nginx/template.go
+++ b/ingress/controllers/nginx/nginx/template.go
@@ -61,6 +61,14 @@ func (ngx *Manager) loadTemplate() {
 
 func (ngx *Manager) writeCfg(cfg config.Configuration, ingressCfg IngressConfig) (bool, error) {
 	conf := make(map[string]interface{})
+	//replace ^kleenestar\. with *.
+	r := regexp.MustCompile("^kleenestar\\.")
+	for i := 0; i < len(ingressCfg.Servers); i++ {
+		s := ingressCfg.Servers[i]
+		old := s.Name
+		s.Name = r.ReplaceAllString(s.Name, "*.")
+		glog.Infof("replacing [%v] with [%v]", old, s.Name)
+	}
 	conf["upstreams"] = ingressCfg.Upstreams
 	conf["servers"] = ingressCfg.Servers
 	conf["tcpUpstreams"] = ingressCfg.TCPUpstreams


aledbf:
@from-nibly please check [this comment](https://github.com/kubernetes/contrib/issues/799#issuecomment-230073069)

Besides that if you have more than one Ingress rule using the same host and path, only the first one will be used in the nginx.conf file (and a warning about this situation will appear in the logs).

from-nibly:
@aledbf our problem is that we do host name processing in our app. specifically we put the customer name at the front of the domain name. We need to be able to route all calls to <customerName>.ourservice.com to the same server. The server then determines what customer is being referred to using the host name. We are already using this pull request and it works great. 

I however realize that we don't want to actually have this merged because "kleenestar." -> "*." is really awful. Is there a reason though why an ingress resource should not be able to define a wild card at the beginning of a host name?

aledbf:
> We need to be able to route all calls to <customerName>.ourservice.com to the same server.

So you have 1 service (your app) to handle the traffic of *.ourservice.com?
If that's the case then Ingress is not the right abstraction (too complex for that). Just use a standard nginx container with an upstream pointing to the service you are using.

aledbf:
@from-nibly just in case I'm using the ingress controller in the same scenario `<customerName>.ourservice.com` but I have 1 Ingress rule per customer `acme.ourservice.com` that uses the service `acme-myproduct`

aledbf:
@from-nibly this is an example:
```
apiVersion: v1
items:
- apiVersion: extensions/v1beta1
  kind: Ingress
  metadata:
    name: demo-ps
    namespace: demo-ps
  spec:
    rules:
    - host: demo.my-product.cl
      http:
        paths:
        - backend:
            serviceName: demo-ps
            servicePort: 80
          path: /
  status:
    loadBalancer:
      ingress:
      - ip: 10.0.0.51
      - ip: 10.0.0.49
      - ip: 10.0.0.46
      - ip: 10.0.0.48
kind: List
metadata: {}
```

I have 1 namespace per customer with:
- 2 rc (web and worker node.js) with web scaled to 3
- if uses ssl a secret with the certificate
- configmap with the configuration (env variables). 

Right now I'm running more than 890 "instances" using this schema.

from-nibly:
@aledbf we actually have a multi-tenant system where multiple customers will get routed to a single namespace. The customer only exists in the data sense. We would like to be able to decouple creating a customer from interacting with the ingress controller as kubernetes will be one of may ways we deploy our app as we transition from our existing infrastructure to kubernetes.  Ingress is actually what we hope to use as I simplified our needs quite a bit.

we have about 30 services. One of them is our entry point into the system where everyone would enter via &lt;customer&gt;.webservice.ourservice.com but we also need to route to the individual services via &lt;serviceName&gt;.ourservice.com

Why should Kubernetes not be able to allow a *. wildcard at the front of a host name in the ingress resource? All proxy backends support it.

PR #1228, opened 2016-06-17
add build cop tools

lavalamp:
@eparis @mikedanese

### DIFF
--- a/mungegithub/build-cop/emergency.sh
+++ b/mungegithub/build-cop/emergency.sh
@@ -1,0 +1,37 @@
+#!/bin/bash
+
+# Copyright 2016 The Kubernetes Authors All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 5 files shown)


mungegithub/build-cop/emergency.sh
```diff
@@ -0,0 +1,35 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+STOP=${1:-"stop"}
```
mikedanese: what are the other options?

lavalamp: "resume"-- there's a message that prints it below.

lavalamp: You can also do "status" but I don't think that's useful so I didn't bother
to print it.

lavalamp:
I pushed a change to get rid of the sleep, which was buggy (of course).

lavalamp:
I guess no one is going to review this-- I'm merging. Tired of switching branches to get these tools.

PR #1233, opened 2016-06-17
[nginx-ingress-controller]: Avoid sync without a reachable master

### LINKED ISSUE #1226: Ingress Controller doesn't handle very well when master is down
I just had to update the box where my Kubernetes master is running.
While the box was rebooting, the Ingress Controller lost all the secrets (TLS), and the HTTP basic auth configuration. In the end, all the Ingress Resources were just listening on 80. No more TLS, no more HTTP Basic Auth (and probably all the other configurations depending on annotations).

When master is not available, it shouldn't really update its nginx.conf. Thoughts ?

aledbf:
fixes #1226

### DIFF
--- a/ingress/controllers/nginx/controller.go
+++ b/ingress/controllers/nginx/controller.go
@@ -421,13 +421,19 @@ func (lbc *loadBalancerController) sync(key string) {
 		return
 	}
 
-	var cfg *api.ConfigMap
+	// by default no custom configuration configmap
+	cfg := &api.ConfigMap{}
 
-	ns, name, _ := parseNsName(lbc.nxgConfigMap)
-	cfg, err := lbc.getConfigMap(ns, name)
-	if err != nil {
-		glog.V(3).Infof("unexpected error searching configmap %v: %v", lbc.nxgConfigMap, err)
-		cfg = &api.ConfigMap{}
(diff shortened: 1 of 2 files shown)


aledbf:
ping @bprashanth

ingress/controllers/nginx/controller.go
```diff
@@ -421,13 +421,17 @@ func (lbc *loadBalancerController) sync(key string) {
...
-	var cfg *api.ConfigMap
+	cfg := &api.ConfigMap{}
+	var err error
 
-	ns, name, _ := parseNsName(lbc.nxgConfigMap)
-	cfg, err := lbc.getConfigMap(ns, name)
-	if err != nil {
-		glog.V(3).Infof("unexpected error searching configmap %v: %v", lbc.nxgConfigMap, err)
-		cfg = &api.ConfigMap{}
+	if lbc.nxgConfigMap != "" {
```
bprashanth: please add a comment on when this might be empty in the abnormal case

bprashanth:
LGTM, just comment nit

ingress/controllers/nginx/controller.go

bprashanth:
Reverting, @aledbf I think we chaned a method signature with the delayerd queue that didn't rebase too well with this change

PR #1237, opened 2016-06-18
Move *-serial to nonblocking.

lavalamp:
Also remove the -subnet test which is apparently not useful. (per @gmarek)


lavalamp:
Pushing this live right now.

gmarek:
The problem with `-subnet`, if I understand the suite correctly is that it runs on 1.2 release branch. @roberthbailey 

Another discussion we probably should have is do we really need to run tests of old release branches. We don't make a lot of changes in them, and they are separate from changes to the master, so I don't think it's useful for them to be blockers. @lavalamp @fejta

fejta:
-subnet shouldn't be queue blocking? And/or if it is it should be listed at http://submit-queue.k8s.io/#/e2e

lavalamp:
it wasn't blocking, but it also wasn't useful. More noise than signal.

PR #1242, opened 2016-06-20
Adding federation e2e to non blocking test suite

nikhiljindal:
Adding federation e2e test suite to the list of tracked tests so that we at least have issues filed while we wait for https://github.com/kubernetes/contrib/pull/1198.

cc @lavalamp


lavalamp:
LGTM

lavalamp:
Will go out in next push

PR #1243, opened 2016-06-20
add link to health chart

lavalamp:
We really need to parameterize this, but until then, put it by the other link.

### DIFF
--- a/mungegithub/submit-queue/www/index.html
+++ b/mungegithub/submit-queue/www/index.html
@@ -168,7 +168,11 @@
               <br>
               <br>
               <h2 class="md-title" ng-show="cntl.OverallHealth.length > 0">Overall Health: {{ cntl.OverallHealth }}</h2>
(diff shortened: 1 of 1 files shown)


fejta:
Maybe also put this inline at http://submit-queue.k8s.io/#/info?

rmmh:
It's kind of ugly for an inline chart.

lavalamp:
I actually just put it inline now. I can't figure out the magic CSS incantation to make it shrink down nicely in the amount of time I'm willing to spend on CSS (hint: not much).

lavalamp:
OK inlined in two places now.

lavalamp:
Actually maybe inlining is bad unless it also refreshes occasionally?

rmmh:
It doesn't refresh automatically. You probably want `style="max-width: 800px"` or something on the element to resize it (but it might be ugly).

lavalamp:
I tried `max-width: 100%` and that didn't seem to work...

rmmh:
style="max-width:100%"  appears to work for me?

lavalamp:
hm, I'll try again

lavalamp:
Seems to work now. Not sure what I did before.

fejta:
LGTM

PR #1244, opened 2016-06-20
Update image on ingress example

bprashanth:
Just updating the image from something ancient to the one we have in gcr.io


bprashanth:
Green, merging since this is simple enough and the example isn't really used anywhere (though maybe I should include it in k8s.io documentation)

PR #1247, opened 2016-06-21
Update Ingress godep

bprashanth:
@aledbf fyi


zmerlynn:
LGTM.

Out of curiosity, can this tag be used for `release-1.2` as well, or is there an incompatibility there?

bprashanth:
It can, I'm not sure how I'm going to handle the 1.2 update. We don't really "release" cluster addons, the l7 controller runs on the master so it should just rev forward automatically on hosted offerings. If someone sets up a new cluster (with old 1.2 code) on gce/prem etc, they'll get an old l7. They can use kubectl edit to update the image on this old controller forward to 0.7.0. 

It *should be* safe to just update this tag to 1.2  because the resource didn't change, but I need to test it some more.

bprashanth:
> We don't really "release" cluster addons, the l7 controller runs on the master so it should just rev forward automatically on hosted offerings. 

I meant we don't branch, we release eg: https://github.com/kubernetes/contrib/releases, no reason we can't other than it's in beta so it's still in contrib/

zmerlynn:
We're about to cut a 1.2.5. If you get the PR into master for bumping `kubernetes/kubernetes/cluster/saltbase/salt/l7-gcp/glbc.manifest`, we can get it cherry-picked to `release-1.2`, and it'll go out when we update hosted masters in GKE.

zmerlynn:
Does this Travis ever work?

zmerlynn:
(Feel free to merge if not.)

bprashanth:
Yeah I think it broke maybe because of some change, looking. Thanks for lgtm will merge when i fix. I think we wouldn't have the famed "green badge" if it didn't work: https://github.com/kubernetes/contrib

bprashanth:
Green, merging

PR #1249, opened 2016-06-22
Reduce number of backends in basic ingress example

bprashanth:
Delete a backend service so we fall under default quota.


piosz:
LGTM

PR #1190, opened 2016-06-10
Write github cache to disk

eparis:
This seems to work! It will grow unbound (we can look into that)

Just a clean submit queue run created 39M of data.

A single run of all mungers is still running and is around 100M. I'll
update when I know what it looks like overnight.

PR #1254, opened 2016-06-22
Use memcached to persist github cache out-of-process.

rmmh:
Alternative implementation: #1190 (disk-backed)

Not sure which is better. This depends on another pod instead of an attached PD.

### DIFF
--- a/mungegithub/github/github.go
+++ b/mungegithub/github/github.go
@@ -36,6 +36,7 @@ import (
 	"github.com/golang/glog"
 	"github.com/google/go-github/github"
 	"github.com/gregjones/httpcache"
+	httpmemcache "github.com/gregjones/httpcache/memcache"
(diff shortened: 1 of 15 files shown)


eparis:
getting `vendor` right given the fact that we 'self vendor' `contrib/test-utils` or whatever if HARD.  Might I suggest.
```
rm -rf Godeps/ vendor/
git add --all
git commit
godep save ./...
rm -rf vendor/k8s.io/contrib/test-utils/
cd vendor/k8s.io/contrib/
ln -s ../../../../test-utils/
git add --all
git commit --amend
git rebase -i # to merge the result with your last attempt to update
```

lavalamp:
I think I'd rather set up the PD.

rmmh:
We went with the PD solution.

PR #1255, opened 2016-06-23
get logs for failed pods in node e2e tests

mnshaw:
@timstclair 

Initial commit, still work in progress.


timstclair:
I was just introduced to "k8s-gubernator" (e.g. https://k8s-gubernator.appspot.com/build/kubernetes-jenkins/logs/kubernetes-e2e-gce/19087/), which might actually be a better fit for this. I'll follow up with the owner's of that project and see how you might integrate with it. For now, if you want you could take a look at https://github.com/kubernetes/test-infra/tree/master/gubernator

eparis:
this PR seems pretty mis-formed?   43 commits from lots of different people across most of contrib...  Can you clean it up to only be the single change you are hoping to make?

PR #1258, opened 2016-06-23
[nginx-ingress-controller]: Avoid sync without a reachable master

aledbf:
fixes #1233


aledbf:
@bprashanth this fixes the revert

bprashanth:
LGTM

PR #1262, opened 2016-06-23
limit number of flakes we keep in memory

lavalamp:
Need this because I'm going to start reading flakes from PR builder runs, too.

### DIFF
--- a/mungegithub/mungers/flakesync/cache.go
+++ b/mungegithub/mungers/flakesync/cache.go
@@ -17,6 +17,7 @@ limitations under the License.
 package flakesync
 
 import (
+	"container/list"
 	"sort"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


mungegithub/mungers/flakesync/cache.go
```diff
@@ -115,7 +121,9 @@ func NewCache(getFunc ResultFunc) *Cache {
 	c := &Cache{
 		byJob:             jobMap{},
 		flakeQueue:        flakeMap{},
+		expireList:        list.New(),
```
eparis: I guess the memory usage of a `container/list` is sufficiently superior to a `slice` for this usage to justify the indirection? I don't care. PR looks fine.

lavalamp: oh, haha, this is probably worse. I was anchored because I thought of an LRU first. Doesn't matter much though.

Issue #1251, opened 2016-06-22
Fix Godeps in cluster-autoscaler


zmerlynn:
I'm about to put up an embarassing PR to fix https://github.com/kubernetes/kubernetes/issues/27821 that will manually alter the `vendor/` directory, because the `Godeps` are currently in conflict, but we need to get a release out:

```
$ godep go build ./...
# k8s.io/contrib/cluster-autoscaler/utils/gce
utils/gce/gce.go:65: cannot use gce.NewAltTokenSource(cfg.Global.TokenURL, cfg.Global.TokenBody) (type "k8s.io/kubernetes/vendor/golang.org/x/oauth2".TokenSource) as type "k8s.io/contrib/cluster-autoscaler/vendor/golang.org/x/oauth2".TokenSource in assignment:
        "k8s.io/kubernetes/vendor/golang.org/x/oauth2".TokenSource does not implement "k8s.io/contrib/cluster-autoscaler/vendor/golang.org/x/oauth2".TokenSource (wrong type for Token method)
                have Token() (*"k8s.io/kubernetes/vendor/golang.org/x/oauth2".Token, error)
                want Token() (*"k8s.io/contrib/cluster-autoscaler/vendor/golang.org/x/oauth2".Token, error)
# k8s.io/contrib/cluster-autoscaler/simulator
simulator/cluster.go:78: undefined: cmd.GetPodsForDeletionOnNodeDrain
simulator/nodes.go:31: undefined: cmd.GetPodsForDeletionOnNodeDrain
godep: go exit status 2
```

eparis:
Wait, why is kubernetes not being vendored in? Unlike GOPATH (which is what Godep used) using vendor is a much stronger all or nothing. You must vendor everything....

zmerlynn:
@eparis: It *is* being vendored in, AFAICT. That's after I updated the k8s dep.

eparis:
The fact that it says you have a `k8s.io/kubernetes/vendor/golang.org/x/oauth2` says that it is using `$GOPATH/src/k8s.io/kubernetes` instead of `$GOPATH/src/k8s.io/contrib/vendor/k8s.io/kubernetes`.  hmmmm

zmerlynn:
@eparis: The second error is the more obnoxious one, and looks like it'll take a PR to the main repo (unless I'm missing something). It's very possible I just screwed something up, too.

eparis:
"I think you did something wrong to get the first error, I don't get it"

[...]

PR #1263, opened 2016-06-23
[cluster-autoscaler]: Update godeps

aledbf:


aledbf:
fixes #1251

cluster-autoscaler/simulator/nodes.go
```diff
@@ -20,31 +20,28 @@ import (
...
 )
 
 // GetRequiredPodsForNode returns a list od pods that would appear on the node if the
-// node was just created (like deamonset and manifest-run pods). It reuses kubectl
-// drain command to get the list.
+// node was just created (like deamonset and manifest-run pods).
 func GetRequiredPodsForNode(nodename string, client *kube_client.Client) ([]*kube_api.Pod, error) {
-	podsToRemoveList, _, _, err := cmd.GetPodsForDeletionOnNodeDrain(client, nodename,
-		kube_api.Codecs.UniversalDecoder(), true, true)
+	allPodList, err := client.Pods(kube_api.NamespaceAll).List(
```
mwielgus: There was actually a reason why we were using the cmd drain instead of fast drain. Cmd drain does thorough checks whether all pods are actually replicated and have a running rc/rs. Fast drain checks only if a pod was created by rc/rs and blindly assumes that replication controller/replica set is still alive and the pod can be safely removed.

aledbf: @mwielgus after #26667 there's no `cmd.GetPodsForDeletionOnNodeDrain`.
Please advice how to procede

mwielgus: Probably the most reasonable option would be to temporarily duplicate the old code here. We have a plan to move drain operation from kubectl to server-side anyway.

aledbf: done

PR #1264, opened 2016-06-23
Add option to load test owners from a URL.

rmmh:
For example, a CSV file stored in a GCS bucket, updated by some other
process.

### DIFF
--- a/mungegithub/mungers/testowner/owner.go
+++ b/mungegithub/mungers/testowner/owner.go
@@ -18,13 +18,14 @@ package testowner
 
 import (
 	"encoding/csv"
+	"errors"
 	"io"
-	"os"
+	"net/http"
 	"regexp"
 	"strings"
-	"time"
 
 	"github.com/golang/glog"
+	"github.com/gregjones/httpcache"
 )
 
 var tagRegex = regexp.MustCompile(`\[.*?\]|\{.*?\}`)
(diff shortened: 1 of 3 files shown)


mungegithub/mungers/testowner/owner.go
```diff
@@ -107,33 +113,31 @@ func NewReloadingOwnerList(path string) (*ReloadingOwnerList, error) {
...
+		glog.Errorf("Unable to reload test owners at %s: %v", o.url, err)
 		// Process using the previous data.
 	}
 	return o.ownerList.TestOwner(testName)
 }
 
-func (o *ReloadingOwnerList) reload() error {
-	info, err := os.Stat(o.path)
+func (o *HTTPOwnerList) reload() error {
+	resp, err := o.client.Get(o.url)
```
fejta: How does this work? It sends an If-None-Match header with the etag from the last response?

https://cloud.google.com/storage/docs/xml-api/reference-headers#standard

If so then that is sweet!

fejta:
LGTM, will let someone else with go readability toss on the lgtm label

mungegithub/mungers/testowner/owner.go
rmmh: Yes, httpcache respects the HTTP caching RFCs. It also respects the max-age header, so it won't even do an HTTP request to check freshness if the current request is valid.

The munger doesn't open too many new issues, so this is all somewhat premature regardless.

fejta: In that case... After we copy the updated csv how long until munger gets the new config? Hopefully w/in an hour or two?

rmmh: You can set cache-control headers with gsutil-- I'm planning to set max-age=60 to make it verify contents after a minute.

lavalamp:
I don't know if this is a great plan. Why not just keep a local copy of the repo? And sync after a merge or every 15 minutes or something. I don't want to introduce another service. Remember we're going to need to run submit queue against other repositories-- I'd like to keep the operational complexity to a minimum.

eparis:
We get OWNERS data out of the git repo every ~10 minutes.  We could pull test owners from there similarly...

fejta:
@eparis @lavalamp is the request here to commit the CSV file somewhere and then download this on an intermittent basis directly from github rather than GCS?

lavalamp:
Right, the OWNERs files should all be in the repository that they apply to.
(We want to rerun this infrastructure against all of our repositories!)

Ideally this won't be a CSV forever, or the CSV will be generated from
comments or whatnot in the source files.

fejta:
Agreed that we want this to work for any arbitrary repo (including non kubernetes repos).

For the time being may we decouple automatically generating the list of tests and owners for those tests from this PR?

For the time being we're depending on a CSV file and I want to prioritize moving auto-assignment off of Ryan's workstation and/or doing it by hand.

We can then iteratively improve the way we store/generate these test owners.

lavalamp:
Yeah, I'm fine if we keep it as a CSV right now, but it should be checked
into and obtained from the repository.

rmmh:
How about pointing it at https://raw.githubusercontent.com/kubernetes/kubernetes/master/test/test_owners.csv (after that PR merges)?

lavalamp:
It seems we've merged another way of doing this.

Issue #1266, opened 2016-06-24
nginx-alpha Ingress Controller ProxyPass syntax issue


aLostEngineer:
The ProxyPass URL must have a trailing forward-slash to properly proxy to the root of another http service. Furthermore, the Ingress points must be declared with trailing slashes to match.

Example: with the current setup, if I have a ingress `/a` pointing to a backend nginx service, the Ingress service will keep the `/a` subdirectory appended, forcing the backend service to try to access `<service>/a`, instead of the webroot.

Adding a trailing forward slash to the Ingress point and nginx config will properly forward HTTP requests.

aledbf:
@aLostEngineer the behavior you are describing is correct. 
What you are proposing is a rewrite (which can break some apps). Please check this issues https://github.com/kubernetes/contrib/issues/679 and https://github.com/kubernetes/contrib/issues/860

aLostEngineer:
@aledbf is the ingress `rewrite` directive ready for deployment? If so, I can close the issues/PRs. From what it looks like [here](https://github.com/kubernetes/contrib/tree/master/ingress/controllers/nginx/examples/rewrite), it is. If so, maybe I should add an example to the nginx-alpha README instead?

aLostEngineer:
Closed PR #1267 and Issue #1268 until I am better informed on newer features.

aledbf:
>  is the ingress rewrite directive ready for deployment? 

Yes.

@aLostEngineer `nginx-alpha` and `nginx` are different implementations of an Ingress controllers.

aLostEngineer:
@aledbf Ok, thanks. I'll refactor our deploy to use the `rewrite` option in Ingress.

Closing.

bprashanth:
Yeah this is one area that we compromise usability on one backend for cross platform compat. Annotations/config map is the way to expose such features. Please do add examples (but to the main nginx controller, since nginx-alpha is a toy example that is intentionally kept as simple as possible).

PR #1267, opened 2016-06-24
Add trailing slashes to nginx-alpha ingress config

aLostEngineer:
This address most of #1266. Does not account for failure to add trailing slashes to ingress resource.


aLostEngineer:
CLA signed.

PR #1269, opened 2016-06-24
open for every milestone

lavalamp:
Pushing live now.


lavalamp:
@eparis

eparis:
we'll see if the cache keeps us from running out of tokens...

eparis:
we also need a message on kubernetes-dev telling people to label PRs they want in 1.3 with cherrypick-candidate and make sure the 1.3 milestone is set.

lavalamp:
Yeah, I'm working on the email right now.

lavalamp:
Huh, this barfed on trying to put the empty string into that arg. Pushing again...

eparis:
did I screw those up? probably should have unit tested :-(

PR #1270, opened 2016-06-24
fix deployment

lavalamp:
Live now.


lavalamp:
@eparis, did you expect it to work passing an empty string?

eparis:
I think i expected `=""` to work

PR #1272, opened 2016-06-24
submit-queue: update the 'details' link in github

### LINKED ISSUE #1271: submit queue "details" link on PRs no longer works
The submit queue adds a commit status with a details link to something like http://submit-queue.k8s.io/#/queue?prDisplay=27798&historyDisplay=27798.

This no longer seems to be a valid page.

eparis:
We were sending people to the front page of the submit-queue. But then
we changed the front page to be the queue instead of the 'all PRs'.

While this link it's particularly useful, at least seeing your PR most
of the time is better than not...

fixes #1271


lavalamp:
LGTM

PR #1274, opened 2016-06-25
Switch graph to use svg instead of png

fejta:
http://storage.googleapis.com/kubernetes-test-history/k8s-queue-health.svg exists (and the .png's content type is actually svg)


lavalamp:
lgtm

PR #1276, opened 2016-06-26
Add gcsweb

thockin:
This is a VERY simple webserver that offers selected GCS buckets for browsing.
The HTML is pathetic, but it seems to work (it's as fast as the GCE-hosted
browser (which is not fast), but far less pretty).

This is just the main.  There's no test for it.  The HTML needs help.  There's no Dockerfile or Makefile.

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1276)


test-foxish:
recomputing cla status...

thockin:
CLA is failing because of multiple authors.  Need manual override.  This should be moved from contrib to its own repo, though.

eparis:
cncf CLA was complaining about you before the multiple authors....

eparis:
@foxish this is a problem though, that the pinger is assuming (and posting) that the cncf failure is because @thockin didn't sign....

eparis:
@foxish can we make the cla pinger clean up stale messages?

foxish:
@eparis Yes, I'll add the stale comments cleanup to the CLA munger. But in this case, @thockin did sign and is authorized. It's the other author of a particular commit, @nebril who hasn't.

foxish:
@eparis Oh I see what you mean, it's pinging the wrong person. Will open an issue for this.

eparis:
@foxish but the pinger is still yelling at thockin.  i think we need to look at the status
`thockin authorized, but 1 other problem`
and output some more useful message.

foxish:
@eparis Yeah, just saw that. You're right, we should do something smarter. I don't know if there is a way to get the commit author's name who hasn't signed the CLA from the status. I'll ask the LF folks about this. Else, we might want to do a manual check among all commit authors.

ixdy:
@thockin do you want me to look at this PR? or do you want to reopen it somewhere else? (maybe under k8s.io?)

thockin:
Jeff,

What think about moving it to test-infra ?

PR #1278, opened 2016-06-28
Update the sidecar logging example with a more recent fluentd build

a-robinson:


a-robinson:
ping @cjcullen

cjcullen:
LGTM. Sorry I sat on this for so long.

PR #1282, opened 2016-06-28
Cluster-autoscaler: fix for multi-mig autoscaling

### LINKED ISSUE #1283: Cluster autoscaler requires all migs to be autoscaled
Otherwise it doesn't do any scale up or scale down.

mwielgus:

Fixes #1283

### DIFF
--- a/cluster-autoscaler/utils/gce/gce.go
+++ b/cluster-autoscaler/utils/gce/gce.go
@@ -19,6 +19,7 @@ package gce
 import (
 	"fmt"
 	"io"
+	"strings"
 	"sync"
 	"time"
 
@@ -40,9 +41,14 @@ const (
 	operationPollInterval = 100 * time.Millisecond
 )
 
+type migInformation struct {
+	config   *config.MigConfig
+	basename string
+}
+
 // GceManager is handles gce communication and data caching.
 type GceManager struct {
-	migs       []*config.MigConfig
+	migs       []*migInformation
 	service    *gce.Service
 	migCache   map[config.InstanceConfig]*config.MigConfig
 	cacheMutex sync.Mutex
@@ -74,8 +80,15 @@ func CreateGceManager(migs []*config.MigConfig, configReader io.Reader) (*GceMan
 		return nil, err
 	}
 
+	migInfos := make([]*migInformation, 0, len(migs))
+	for _, mig := range migs {
+		migInfos = append(migInfos, &migInformation{
+			config: mig,
+		})
+	}
+
 	manager := &GceManager{
-		migs:     migs,
+		migs:     migInfos,
 		service:  gceService,
 		migCache: map[config.InstanceConfig]*config.MigConfig{},
 	}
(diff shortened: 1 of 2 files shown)


fgrzadkowski:
@mwielgus Please explain what is the bug and how do we fix this.

cluster-autoscaler/utils/gce/gce.go
```diff
@@ -164,13 +177,23 @@ func (m *GceManager) GetMigForInstance(instance *config.InstanceConfig) (*config
...
 		return mig, nil
 	}
-	if err := m.regenerateCache(); err != nil {
-		return nil, fmt.Errorf("Error while looking for MIG for instance %+v, error: %v", *instance, err)
-	}
-	if mig, found := m.migCache[*instance]; found {
-		return mig, nil
+
+	for _, mig := range m.migs {
+
```
piosz: remove empty line

mwielgus:
Fixes: #1283

Previously we were rising an error if we didn't have mig config for all nodes in the cluster. This error was equivalent to cluster being not in stable state and preventing CA from any scale up or scale downs. Now we raise this error only if the node has the prefix corresponding to some mig base name but the mig doesn't have this node on the list. Otherwise we assume that the node belongs to non-autoscaled mig and continue with all activities.

piosz:
LGTM
Please create an issue and add TODO somewhere to solve this problem better as we talked offline.

cluster-autoscaler/utils/gce/gce.go
mwielgus: Done

cluster-autoscaler/utils/gce/gce.go
```diff
@@ -164,13 +177,22 @@ func (m *GceManager) GetMigForInstance(instance *config.InstanceConfig) (*config
...
-		return nil, fmt.Errorf("Error while looking for MIG for instance %+v, error: %v", *instance, err)
-	}
-	if mig, found := m.migCache[*instance]; found {
-		return mig, nil
+
+	for _, mig := range m.migs {
+		if mig.config.Project == instance.Project &&
+			mig.config.Zone == instance.Zone &&
+			strings.HasPrefix(instance.Name, mig.basename) {
+			if err := m.regenerateCache(); err != nil {
```
fgrzadkowski: I think it'd be readable if you restructure the code such that:

1. in a for loop check if we have any corresponding loop
2. if we have a corresponding mig regenerate cache and return mig
3. if don't have it return nils

cluster-autoscaler/utils/gce/gce.go
```diff
@@ -164,13 +177,22 @@ func (m *GceManager) GetMigForInstance(instance *config.InstanceConfig) (*config
...
+				return nil, fmt.Errorf("Error while looking for MIG for instance %+v, error: %v", *instance, err)
+			}
+			if mig, found := m.migCache[*instance]; found {
+				return mig, nil
+			}
+			return nil, fmt.Errorf("Instance %+v does not belong to any configured MIG", *instance)
+		}
 	}
-	return nil, fmt.Errorf("Instance %+v does not belong to any known MIG", *instance)
+	// Instance doesn't belong to any configured mig.
```
fgrzadkowski: Please explain it's ok.

cluster-autoscaler/utils/gce/gce.go
```diff
@@ -164,13 +177,22 @@ func (m *GceManager) GetMigForInstance(instance *config.InstanceConfig) (*config
...
-	if err := m.regenerateCache(); err != nil {
-		return nil, fmt.Errorf("Error while looking for MIG for instance %+v, error: %v", *instance, err)
-	}
-	if mig, found := m.migCache[*instance]; found {
-		return mig, nil
+
+	for _, mig := range m.migs {
+		if mig.config.Project == instance.Project &&
+			mig.config.Zone == instance.Zone &&
+			strings.HasPrefix(instance.Name, mig.basename) {
```
fgrzadkowski: During the first run of this loop basename will be empty and it will always be a prefix and we'll return a prefix. Am I missing something?

mwielgus:
@fgrzadkowski  sorry, too late, i will do it in a follow-up pr.

fgrzadkowski:
@mwielgus At the very least please answer to my question regarding base name in the first run.

cluster-autoscaler/utils/gce/gce.go
piosz: Cache is 'regenerated' in the beginning, so this could happen only in case of race condition (with very low probability).

PR #1284, opened 2016-06-29
Munger to add retest-not-required label

ciwang:
Munger checks if all files are .md files. If so, adds
retest-not-required label to avoid running e2e tests when not necessary.
Added retest-not-required option to --pr-mungers.

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1284)

### DIFF
--- a/mungegithub/mungers/retest-not-required.go
+++ b/mungegithub/mungers/retest-not-required.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2015 The Kubernetes Authors All rights reserved.
+Copyright 2016 The Kubernetes Authors All rights reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -34,12 +34,11 @@ import (
 type RetestMunger struct{}
 
 const (
-	retestNotRequiredLabel = "retest-not-required"
+	retestCommentBody = "Retest not required for this PR because of changed filetypes."
 )
 
 var (
-	markdownRE      = regexp.MustCompile(`.+\.md$`)
-	retestCommentRE = regexp.MustCompile("Retest not required for this PR because of changed filetypes.")
+	markdownRE = regexp.MustCompile(`.+\.md$`)
 )
 
 func init() {
@@ -71,11 +70,6 @@ func (r *RetestMunger) Munge(obj *github.MungeObject) {
 		return
 	}
 
-	pr, err := obj.GetPR()
-	if err != nil {
-		return
-	}
-
 	commits, err := obj.GetCommits()
 	if err != nil {
 		return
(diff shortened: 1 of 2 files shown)


mungegithub/mungers/retest-not-required.go
```diff
@@ -0,0 +1,109 @@
+/*
+Copyright 2015 The Kubernetes Authors All rights reserved.
```
eparis: 2015?

mungegithub/submit-queue/deployment.yaml
```diff
@@ -14,7 +14,7 @@ spec:
         command:
         - /mungegithub
         - --token-file=/etc/secret-volume/token
-        - --pr-mungers=blunderbuss,lgtm-after-commit,cherrypick-auto-approve,label-unapproved-picks,needs-rebase,ok-to-test,rebuild-request,path-label,size,stale-pending-ci,stale-green-ci,block-path,release-note-label,comment-deleter,submit-queue,issue-cacher,flake-manager,old-test-getter
+        - --pr-mungers=blunderbuss,lgtm-after-commit,cherrypick-auto-approve,label-unapproved-picks,needs-rebase,ok-to-test,rebuild-request,path-label,size,stale-pending-ci,stale-green-ci,block-path,release-note-label,comment-deleter,submit-queue,issue-cacher,flake-manager,old-test-getter,retest-not-required
```
eparis: run this before submit-queue.

mungegithub/mungers/retest-not-required.go
```diff
@@ -0,0 +1,109 @@
...
+)
+
+// RetestMunger will label a PR with retest-not-required if it only affects
+// .md files.
+// This may eventually be rewritten using a better heuristic for deciding
+// which filetypes require retesting.
+type RetestMunger struct{}
+
+const (
+	retestNotRequiredLabel = "retest-not-required"
```
eparis: don't we already have `retestNotRequiredLabel` in submit_queue.go ?

eparis: ha, guess so from the CI run, this doesn't build
```
# k8s.io/contrib/mungegithub/mungers
mungers/submit-queue.go:50: retestNotRequiredLabel redeclared in this block
	previous declaration at mungers/retest-not-required.go:37
godep: go exit status 2
```

mungegithub/mungers/retest-not-required.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	obj.AddLabel(retestNotRequiredLabel)
+}
+
+func (r *RetestMunger) isStaleComment(obj *github.MungeObject, comment githubapi.IssueComment) bool {
+	if !mergeBotComment(comment) {
+		return false
+	}
+	stale := retestCommentRE.MatchString(*comment.Body)
```
eparis: Where would this comment have come from?

eparis:
We have a 'label when any file matches' munger.

Should we have a similar 'label when all files match' munger? Or will they all need to be special case things, like this?

ciwang:
@eparis Checking for when all files are markdown is the simplest possible heuristic to determine that a retest is not required; and my thought it that this alone will save lots of testing resources but can eventually be tuned to include more types of changes.

mungegithub/mungers/retest-not-required.go
```diff
@@ -0,0 +1,107 @@
...
+
+// RequiredFeatures is a slice of 'features' that must be provided
+func (r *RetestMunger) RequiredFeatures() []string { return []string{} }
+
+// Munge is the workhorse the will actually make updates to the PR
+func (r *RetestMunger) Munge(obj *github.MungeObject) {
+	if !obj.IsPR() {
+		return
+	}
+
```
lavalamp: maybe check if the label is already applied first?

Also, I think there's one more step: if the label has been applied AND it was applied by the mergebot AND the PR now has a non-md file change in it, then we should probably remove the label. (just don't want to remove it if a human put it on manually)

lavalamp:
Wow, super fast, thanks! Just one suggestion to make this a bit safer :)

Can you add a test for this?

foxish:
ping @ciwang

saad-ali:
@ciwang's internship ended. Someone else will have to pick this up.

PR #1285, opened 2016-06-29
Ingress controller ux fixes

bprashanth:
I don't know if this will make 1.3, will try, if not I'll be pushing it out as a minor release cherrypick. 
These ux fixes actually make a difference:
- doesn't hot loop on failure
- doesn't spam activity log with unnecessary firewall updates
- allows non-gce ingress to coexist in the same cluster
- doesn't adopt complex http readiness probes as health checks (probes with headers/hosts/https)
- allows one to continue running the controller on older clusters as a pod (by fixing GCE client creation)

### DIFF
--- a/ingress/controllers/gce/controller/cluster_manager.go
+++ b/ingress/controllers/gce/controller/cluster_manager.go
@@ -241,23 +241,27 @@ func NewClusterManager(
 	defaultBackendNodePort int64,
 	defaultHealthCheckPath string) (*ClusterManager, error) {
 
-	var config *os.File
-	var err error
-	if configFilePath != "" {
-		glog.Infof("Reading config from path %v", configFilePath)
-		config, err = os.Open(configFilePath)
-		if err != nil {
-			return nil, err
-		}
-		defer config.Close()
-	}
-
 	// TODO: Make this more resilient. Currently we create the cloud client
 	// and pass it through to all the pools. This makes unittesting easier.
 	// However if the cloud client suddenly fails, we should try to re-create it
 	// and continue.
-	cloud := getGCEClient(config)
-	glog.Infof("Successfully loaded cloudprovider using config %q", configFilePath)
+	var cloud *gce.GCECloud
+	if configFilePath != "" {
+		glog.Infof("Reading config from path %v", configFilePath)
+		config, err := os.Open(configFilePath)
+		if err != nil {
+			return nil, err
+		}
+		defer config.Close()
+		cloud = getGCEClient(config)
+		glog.Infof("Successfully loaded cloudprovider using config %q", configFilePath)
+	} else {
+		// While you might be tempted to refactor so we simply assing nil to the
+		// config and only invoke getGCEClient once, that will not do the right
+		// thing because a nil check against an interface isn't true in golang.
+		cloud = getGCEClient(nil)
+		glog.Infof("Created GCE client without a confi file")
+	}
 
 	// Names are fundamental to the cluster, the uid allocator makes sure names don't collide.
 	cluster := ClusterManager{ClusterNamer: &utils.Namer{name}}
--- a/ingress/controllers/gce/controller/cluster_manager.go
+++ b/ingress/controllers/gce/controller/cluster_manager.go
@@ -260,7 +260,7 @@ func NewClusterManager(
 		// config and only invoke getGCEClient once, that will not do the right
 		// thing because a nil check against an interface isn't true in golang.
 		cloud = getGCEClient(nil)
-		glog.Infof("Created GCE client without a confi file")
+		glog.Infof("Created GCE client without a config file")
 	}
 
 	// Names are fundamental to the cluster, the uid allocator makes sure names don't collide.
--- a/ingress/controllers/gce/controller/controller.go
+++ b/ingress/controllers/gce/controller/controller.go
@@ -260,31 +260,27 @@ func (lbc *LoadBalancerController) storesSynced() bool {
 }
 
 // sync manages Ingress create/updates/deletes.
-func (lbc *LoadBalancerController) sync(key string) {
+func (lbc *LoadBalancerController) sync(key string) (err error) {
 	if !lbc.hasSynced() {
 		time.Sleep(storeSyncPollPeriod)
-		lbc.ingQueue.requeue(key, fmt.Errorf("Waiting for stores to sync"))
-		return
+		return fmt.Errorf("Waiting for stores to sync")
 	}
 	glog.V(3).Infof("Syncing %v", key)
 
 	paths, err := lbc.ingLister.List()
 	if err != nil {
-		lbc.ingQueue.requeue(key, err)
-		return
+		return err
 	}
 	nodePorts := lbc.tr.toNodePorts(&paths)
 	lbNames := lbc.ingLister.Store.ListKeys()
 	lbs, _ := lbc.ListRuntimeInfo()
 	nodeNames, err := lbc.getReadyNodeNames()
 	if err != nil {
-		lbc.ingQueue.requeue(key, err)
-		return
+		return err
 	}
 	obj, ingExists, err := lbc.ingLister.Store.GetByKey(key)
 	if err != nil {
-		lbc.ingQueue.requeue(key, err)
-		return
+		return err
 	}
 
 	// This performs a 2 phase checkpoint with the cloud:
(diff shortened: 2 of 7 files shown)


ingress/controllers/gce/controller/cluster_manager.go
```diff
@@ -241,24 +241,30 @@ func NewClusterManager(
...
 		}
 		defer config.Close()
+		cloud = getGCEClient(config)
+		glog.Infof("Successfully loaded cloudprovider using config %q", configFilePath)
+	} else {
+		// While you might be tempted to refactor so we simply assing nil to the
+		// config and only invoke getGCEClient once, that will not do the right
+		// thing because a nil check against an interface isn't true in golang.
+		cloud = getGCEClient(nil)
+		glog.Infof("Created GCE client without a confi file")
```
girishkalele: Typo

ingress/controllers/gce/controller/utils.go
```diff
@@ -70,6 +72,21 @@ func (ing ingAnnotations) staticIPName() string {
...
+	if !ok {
+		return ""
+	}
+	return val
+}
+
+// isGCEIngress returns true if the given Ingress either doesn't specify the
+// ingress.class annotation, or it's set to "gce".
+func isGCEIngress(ing *extensions.Ingress) bool {
+	class := ingAnnotations(ing.ObjectMeta.Annotations).ingressClass()
```
girishkalele: Check for nil here would be better than checking for nil after every type assertion.

ingress/controllers/gce/controller/controller.go
```diff
@@ -106,14 +106,28 @@ func NewLoadBalancerController(kubeClient *client.Client, clusterManager *Cluste
 	pathHandlers := framework.ResourceEventHandlerFuncs{
 		AddFunc: func(obj interface{}) {
 			addIng := obj.(*extensions.Ingress)
+			if !isGCEIngress(addIng) {
```
girishkalele: Check for nil after the type assertion ?

ingress/controllers/gce/controller/controller.go
```diff
@@ -260,31 +277,27 @@ func (lbc *LoadBalancerController) storesSynced() bool {
 }
 
 // sync manages Ingress create/updates/deletes.
-func (lbc *LoadBalancerController) sync(key string) {
+func (lbc *LoadBalancerController) sync(key string) (err error) {
 	if !lbc.hasSynced() {
 		time.Sleep(storeSyncPollPeriod)
-		lbc.ingQueue.requeue(key, fmt.Errorf("Waiting for stores to sync"))
-		return
+		return fmt.Errorf("Waiting for stores to sync")
```
girishkalele: Removed the requeue ?

ingress/controllers/gce/controller/cluster_manager.go
bprashanth: fixed

ingress/controllers/gce/controller/utils.go
bprashanth: if ing can't be nil, there are no pointers after that

ingress/controllers/gce/controller/controller.go
bprashanth: I can't get a nil interface here, if we do it's worth crashing for.

ingress/controllers/gce/controller/controller.go
bprashanth: Yeah I'm returning an error throughout this function so I can do a delayed requeue in the caller. That's the backoff fix.

ingress/controllers/gce/firewalls/firewalls.go
```diff
@@ -60,7 +62,21 @@ func (fr *FirewallRules) Sync(nodePorts []int64, nodeNames []string) error {
 		glog.Infof("Creating global l7 firewall rule %v", name)
 		return fr.cloud.CreateFirewall(suffix, "GCE L7 firewall rule", fr.srcRange, nodePorts, nodeNames)
 	}
-	glog.V(3).Infof("Firewall rule already %v exists, verifying for nodeports %v", name, nodePorts)
+
+	requiredPorts := sets.NewString()
+	for _, p := range nodePorts {
+		requiredPorts.Insert(strconv.Itoa(int(p)))
```
girishkalele: Looks odd that there is no Itoa for int64s - also interesting that nodeports is []int64 - seems too large for ports.

bprashanth: yeah, node ports is int64 because the gce api takes an int64 for the ports, our NodePort is an int32

ingress/controllers/gce/controller/utils.go
```diff
@@ -40,6 +40,8 @@ import (
 const (
 	allowHTTPKey    = "kubernetes.io/ingress.allow-http"
 	staticIPNameKey = "kubernetes.io/ingress.global-static-ip-name"
+	ingressClassKey = "kubernetes.io/ingress.class"
```
freehan: this is a new dedicated annotation right?
who is going to set this? Customer? kubectl?

ingress/controllers/gce/controller/cluster_manager.go
```diff
@@ -241,24 +241,30 @@ func NewClusterManager(
...
+	// TODO: Make this more resilient. Currently we create the cloud client
+	// and pass it through to all the pools. This makes unittesting easier.
+	// However if the cloud client suddenly fails, we should try to re-create it
+	// and continue.
+	var cloud *gce.GCECloud
 	var err error
 	if configFilePath != "" {
+		var config *os.File
 		glog.Infof("Reading config from path %v", configFilePath)
 		config, err = os.Open(configFilePath)
```
freehan: initialize config here?

ingress/controllers/gce/controller/utils.go
bprashanth: customer using kubectl, or at creation time. They set it when they create an Ingress that they'd like, eg: nginx to use. Running 2 of any controller (rc, endpoint etc) will cause to confusion unless they take a lock, and this is essentially the customer taking the lock on ingress creation.

bprashanth: Added comments

ingress/controllers/gce/controller/cluster_manager.go
bprashanth: done

bprashanth:
PTAL

girishkalele:
@freehan 

Changes LGTM to me with my limited experience with this feature/code. Will let you put in LGTM label once you've reviewed.

ingress/controllers/gce/controller/cluster_manager.go
```diff
@@ -241,24 +241,28 @@ func NewClusterManager(
...
 			return nil, err
 		}
 		defer config.Close()
+		cloud = getGCEClient(config)
+		glog.Infof("Successfully loaded cloudprovider using config %q", configFilePath)
+	} else {
+		// While you might be tempted to refactor so we simply assing nil to the
+		// config and only invoke getGCEClient once, that will not do the right
+		// thing because a nil check against an interface isn't true in golang.
+		cloud = getGCEClient(nil)
```
freehan: will this succeed with nil?

ingress/controllers/gce/controller/controller.go
```diff
@@ -106,14 +106,28 @@ func NewLoadBalancerController(kubeClient *client.Client, clusterManager *Cluste
...
+				return
+			}
 			lbc.recorder.Eventf(addIng, api.EventTypeNormal, "ADD", fmt.Sprintf("%s/%s", addIng.Namespace, addIng.Name))
 			lbc.ingQueue.enqueue(obj)
 		},
-		DeleteFunc: lbc.ingQueue.enqueue,
+		DeleteFunc: func(obj interface{}) {
+			delIng := obj.(*extensions.Ingress)
+			if !isGCEIngress(delIng) {
+				return
```
freehan: maybe log here?

ingress/controllers/gce/controller/controller.go
```diff
@@ -106,14 +106,28 @@ func NewLoadBalancerController(kubeClient *client.Client, clusterManager *Cluste
...
+			}
+			glog.Infof("Delete notification received for Ingress %v/%v", delIng.Namespace, delIng.Name)
+			lbc.ingQueue.enqueue(obj)
+		},
 		UpdateFunc: func(old, cur interface{}) {
+			curIng := cur.(*extensions.Ingress)
+			if !isGCEIngress(curIng) {
+				return
```
freehan: same here

ingress/controllers/gce/controller/cluster_manager.go
bprashanth: Yeah, it needs to be nil when running this as a pod in the cluster (vs with a valid gce config on the master)

ingress/controllers/gce/controller/controller.go
bprashanth: logging on every update is too noisy

ingress/controllers/gce/controller/controller.go
bprashanth: added

bprashanth:
@freehan done

ingress/controllers/gce/controller/controller.go
freehan: make sense

freehan:
LGTM

bprashanth:
Thanks, merging

PR #1288, opened 2016-06-29
Put did_install fact into group_vars

### LINKED ISSUE #1287: [ansible] Put did_install varible to group_vars//all.yml
I'm running testing on the same environment over and over, it takes lots of time to make sure the packages `kubernetes, docker, etcd, flannel` are latest each time. These packages are already latest so I want to skip checking latest package in the ansible script.

Run the default playbook will install openssl and PyYAML as well, it's not so obvious to user if they install package manually, so intend to not touch them.

gouyang:
Fixes #1287

### DIFF
--- a/ansible/group_vars/all.yml
+++ b/ansible/group_vars/all.yml
@@ -102,3 +102,6 @@ dns_replicas: 1
 
 # See kube documentation for apiserver runtime config options.  Example below enables HPA, deployments features.
 # apiserver_extra_args: "--runtime-config=extensions/v1beta1/deployments=true"
+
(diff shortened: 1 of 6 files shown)


ingvagabund:
``did_install`` variable is used per roles, not per entire playbook. Sharing the var across all roles will skip installation of some packages which are not present on the system.

If a ``node`` role installs ``kubernetes-node`` package, ``did_install`` is set globally to true. Once you start running ``master`` role, ``did_install`` is still set to true and the ``kubernetes-master`` does not get installed.

gouyang:
#1380 will cover this.

PR #1289, opened 2016-06-29
ingress: nginx suggest kube-lego for automated cert management

simonswine:
@aledbf @bprashanth 

Btw: after a release of version 0.8, I would be able to update the docs in the Kube-Lego project, to use the upstream nginx-ingress-controller image

### DIFF
--- a/ingress/controllers/nginx/README.md
+++ b/ingress/controllers/nginx/README.md
@@ -11,6 +11,7 @@ This is a nginx Ingress controller that uses [ConfigMap](https://github.com/kube
 * [HTTPS](#https)
   * [HTTPS enforcement](#server-side-https-enforcement)
   * [HSTS](#http-strict-transport-security)
+  * [Kube-Lego](#automated-certificate-management-with-kube-lego)
 * [TCP Services](#exposing-tcp-services)
 * [UDP Services](#exposing-udp-services)
 * [Proxy Protocol](#proxy-protocol)
@@ -148,6 +149,23 @@ By default the controller redirects (301) to HTTPS if there is a TLS Ingress rul
 
 To disable this behavior use `hsts=false` in the NGINX config map.
 
+
+### Automated Certificate Management with Kube-Lego
+
+[Kube-Lego] automatically requests missing certificates or expired from
+[Let's Encrypt] by monitoring ingress resources and its referenced secrets. To
+enable this for an ingress resource you have to add an annotation:
+
+```
+kubectl annotate ing ingress-demo kubernetes.io/tls-acme="true"
+```
+
+To setup Kube-Lego you can take a look at this [full example]. The first
+version to fully support Kube-Lego is nginx Ingress controller 0.8.
+
+[full example]:https://github.com/jetstack/kube-lego/tree/master/examples
+[Kube-Lego]:https://github.com/jetstack/kube-lego
+[Let's Encrypt]:https://letsencrypt.org
 
 ## Exposing TCP services
 


aledbf:
LGTM

Thanks @simonswine

ingress/controllers/nginx/README.md
```diff
@@ -149,6 +150,23 @@ By default the controller redirects (301) to HTTPS if there is a TLS Ingress rul
...
+enable this for an ingress resource you have to add an annotation:
+
+```
+kubectl annotate ing ingress-demo kubernetes.io/tls-acme="true"
+```
+
+To setup Kube-Lego you can take a look at this [full example]. The first
+version to fully support Kube-Lego is nginx Ingress controller 0.8.
+
+[full example]:https://github.com/jetstack/kube-lego/tree/master/examples
```
bprashanth: how is simonswine/nginx-ingress-controller:0.0.2 different from the controller in this repo? can we make them more similar, or will that be a lot of work? if they're more or less compatible, suggest clarifying (or just using 1 image everywhere).

aledbf: @bprashanth after 0.8 is only 1 image (this version contains 2 PR required for the let's encrypt workflow)

https://github.com/jetstack/kube-lego#modifications

bprashanth:
Great! thanks for the clarification. 

@aledbf @simonswine I have 2 follow up suggestions then:
1. Write an nginx controller e2e
2. Deploy the nginx controller+lego by default on clusters that don't have ingress and/or cert acquisition backends 

for 1. see https://github.com/kubernetes/kubernetes/blob/master/test/e2e/ingress.go#L214, that should work for all ingress controllers.

ingress/controllers/nginx/README.md
```diff
@@ -149,6 +150,23 @@ By default the controller redirects (301) to HTTPS if there is a TLS Ingress rul
...
 
 
+### Automated Certificate Management with Kube-Lego
+
+[Kube-Lego] automatically requests missing certificates or expired from
+[Let's Encrypt] by monitoring ingress resources and its referenced secrets. To
+enable this for an ingress resource you have to add an annotation:
+
+```
+kubectl annotate ing ingress-demo kubernetes.io/tls-acme="true"
```
bprashanth: what are the requirements of an ingress controller (if any) for this annotation to work? please inline, eg: will it work with cloudprovider loadbalancers too?

bprashanth:
LGTM just the one question

ingress/controllers/nginx/README.md
pieterlange: No, this currently only works with the nginx ingress. kube-lego does all the cert management (in kubernetes secrets). One could create a controller that would trigger on secret updates and pushes the certs to the appropriate cloud API though, but that's not currently included.

(i've been running kube-lego for a couple of weeks now, works great :+1: )

bprashanth: Yeah keeping it in a secret is exactly the sort of api i'd expect. My questions was: say I write a haproxy backend for ingress today, does it need to do anything special to augment kube-lego (like serve .well-known/acme-challenge/), or is kube-lego capable of satisfying the acme challenges on its own?

pieterlange: As far as i understand the implementation, kube-lego will create the appropriate ingress routes for acme challenges (so i think it'll fit into any compliant ingress controller), but for now it's up to the ingress controller to pick up on tls secrets (and updates).

bprashanth: The way it's currently written, I believe kube-lego will create an ingress for itself that exposes /well-known (just by looking at code, not tried). 

This will only work with Ingress controller that serve all Ingresses from the same ingress backend. On cloudproviders, each ingress is a different public vip, so if I provision example.com -> my ingress vip, and kube-lego creates another ingress for /well-know, it gets another IP (so example.com/well-known won't resolve).

We should just document that people should create a rule like:
```
      - backend:
          serviceName: kube-lego
          servicePort: 8080
        path: /.well-known/acme-challenge/*
```

then kube-lego just needs to talk acme. 

Creating a new ingress is a nice trick, for it to work we need ip/hostname claims.

pieterlange: > The way it's currently written, I believe kube-lego will create an ingress for itself that exposes /well-known (just by looking at code, not tried). 
This is correct.

>  On cloudproviders, each ingress is a different public vip,
I might be missing something here, but this is definitely not the case on AWS. I simply have an ELB pointing at all the nodes in the cluster (selecting the nodes running nginx-ingress through healthchecks).

bprashanth: Does AWS support Ingress? 
Currently there is no policy around how each backend handles a new ingress. Some like the nginx controller write out a single config with all the rules, others like the gce controller (and vshphere i believe) will allocate a vip per ingress. I don't know how we're planning to do it with aws yet (https://github.com/kubernetes/contrib/pull/841).

There should be something at the api layer that specifies if the Ingress "joins" an existing hostname/ip, or needs a new one. We don't have that yet.

aledbf:
@bprashanth please check this comment https://github.com/jetstack/kube-lego/issues/6#issuecomment-231753141

bprashanth:
I'll take a follow up pr to modify the doc with instructions on how to use kube-lego from different ingress backends.

PR #1295, opened 2016-06-30
Introduce etcd petset

ingvagabund:
Example of etcd cluster via petsets

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1295)

### DIFF
--- a/pets/etcd/README.md
+++ b/pets/etcd/README.md
@@ -35,19 +35,21 @@ b429c86e3cd4e077: name=etcd-2 peerURLs=http://etcd-2.etcd:2380 clientURLs=http:/
 ## Failover
 
 If any etcd member fails it gets re-joined eventually.
-You can test the scenario by deleting one of the pets:
+You can test the scenario by killing process of one of the pets:
 
 ```shell
-$ kubectl delete pod etcd-1
+$ ps aux | grep etcd-1
+$ kill -9 ETCD_1_PID
 ```
 
 ```shell
 $ kubectl get pods -l "app=etcd"
 NAME                 READY     STATUS        RESTARTS   AGE
 etcd-0               1/1       Running       0          54s
-etcd-1               1/1       Terminating   0          52s
 etcd-2               1/1       Running       0          51s
 ```
+
+After a while:
 
 ```shell
 $ kubectl get pods -l "app=etcd"
--- a/pets/etcd/README.md
+++ b/pets/etcd/README.md
@@ -77,10 +77,10 @@ Updated member with ID 7fd61f3f79d97779 in cluster
 
 1. provision one node cluster
 1. build HEAD of master branch
-1. run ``hack/local-cluster-up.sh`` with ``ENABLE_HOSTPATH_PROVISIONER`` set to ``true``:
+1. run ``hack/local-cluster-up.sh`` with both ``ENABLE_HOSTPATH_PROVISIONER`` and ``KUBE_ENABLE_CLUSTER_DNS`` to ``true``:
 
    ```shell
-   ENABLE_HOSTPATH_PROVISIONER=true ./hack/local-up-cluster.sh
+   KUBE_ENABLE_CLUSTER_DNS=true ENABLE_HOSTPATH_PROVISIONER=true ./hack/local-up-cluster.sh
    ```
 
 1. run skydns
@@ -88,8 +88,43 @@ Updated member with ID 7fd61f3f79d97779 in cluster
 
 ## Scaling
 
-TODO
+The etcd cluster can be scale up by running ``kubectl patch`` or ``kubectl edit``. For instance,
 
-## Limitations
+```sh
+$ kubectl get pods -l "app=etcd"
+NAME      READY     STATUS    RESTARTS   AGE
+etcd-0    1/1       Running   0          7m
+etcd-1    1/1       Running   0          7m
+etcd-2    1/1       Running   0          6m
 
-* no scaling
+$ kubectl patch petset/etcd -p '{"spec":{"replicas": 5}}'
+"etcd" patched
+
+$ kubectl get pods -l "app=etcd"
+NAME      READY     STATUS    RESTARTS   AGE
+etcd-0    1/1       Running   0          8m
+etcd-1    1/1       Running   0          8m
+etcd-2    1/1       Running   0          8m
+etcd-3    1/1       Running   0          4s
+etcd-4    1/1       Running   0          1s
+```
+
+Scaling-down is similar. For instance, changing the number of pets to ``4``:
+
+```sh
+$ kubectl edit petset/etcd
+petset "etcd" edited
+
+$ kubectl get pods -l "app=etcd"
+NAME      READY     STATUS    RESTARTS   AGE
+etcd-0    1/1       Running   0          8m
+etcd-1    1/1       Running   0          8m
+etcd-2    1/1       Running   0          8m
+etcd-3    1/1       Running   0          4s
+```
+
+Once a pet is terminated (either by running ``kubectl delete pod etcd-ID`` or scaling down),
+content of ``/var/run/etcd/`` directory is cleaned up.
+If any of the etcd pets restarts (e.g. caused by etcd failure or any other),
+the directory is kept untouched so the pet can recover from the failure.
+
--- a/pets/etcd/README.md
+++ b/pets/etcd/README.md
@@ -1,0 +1,95 @@
+# Etcd
+
+This example runs etcd through a petset.
+
+## Bootstrap
+
+Create the petset specified in ``etcd-petset.yaml``:
+
+```shell
+$ kubectl create -f etcd-petset.yaml
+service "etcd" created
+petset "etcd" created
+```
+
+The petset controller creates an etcd cluster of size 3.
+You can check all pets are running by
+
+```shell
+$ kubectl get pods -l "app=etcd"
+NAME      READY     STATUS    RESTARTS   AGE
+etcd-0    1/1       Running   0          3m
+etcd-1    1/1       Running   0          2m
+etcd-2    1/1       Running   0          3m
+```
+
+Listing all members from inside a pet:
+
+```shell
+$ etcdctl member list
+2e80f96756a54ca9: name=etcd-0 peerURLs=http://etcd-0.etcd:2380 clientURLs=http://etcd-0.etcd:2379
+7fd61f3f79d97779: name=etcd-1 peerURLs=http://etcd-1.etcd:2380 clientURLs=http://etcd-1.etcd:2379
+b429c86e3cd4e077: name=etcd-2 peerURLs=http://etcd-2.etcd:2380 clientURLs=http://etcd-2.etcd:2379
+```
+
+## Failover
+
+If any etcd member fails it gets re-joined eventually.
+You can test the scenario by deleting one of the pets:
+
+```shell
+$ kubectl delete pod etcd-1
+```
+
+```shell
+$ kubectl get pods -l "app=etcd"
+NAME                 READY     STATUS        RESTARTS   AGE
+etcd-0               1/1       Running       0          54s
+etcd-1               1/1       Terminating   0          52s
+etcd-2               1/1       Running       0          51s
+```
+
+```shell
+$ kubectl get pods -l "app=etcd"
+NAME                 READY     STATUS    RESTARTS   AGE
+etcd-0               1/1       Running   0          1m
+etcd-1               1/1       Running   0          20s
+etcd-2               1/1       Running   0          1m
+```
+
+You can check state of re-joining from ``etcd-1``'s logs:
+
+```shell
+$ kubectl logs etcd-1
+Waiting for etcd-0.etcd to come up
+Waiting for etcd-1.etcd to come up
+ping: bad address 'etcd-1.etcd'
+Waiting for etcd-1.etcd to come up
+Waiting for etcd-2.etcd to come up
+Re-joining etcd member
+Updated member with ID 7fd61f3f79d97779 in cluster
+2016-06-20 11:04:14.962169 I | etcdmain: etcd Version: 2.2.5
+2016-06-20 11:04:14.962287 I | etcdmain: Git SHA: bc9ddf2
+...
+```
+
+## Testing
+
+1. provision one node cluster
+1. build HEAD of master branch
+1. run ``hack/local-cluster-up.sh`` with ``ENABLE_HOSTPATH_PROVISIONER`` set to ``true``:
+
+   ```shell
+   ENABLE_HOSTPATH_PROVISIONER=true ./hack/local-up-cluster.sh
+   ```
+
+1. run skydns
+1. create the petset
+
+## Scaling
+
+TODO
+
+## Limitations
+
+* no scaling
--- a/pets/etcd/etcd-petset.yaml
+++ b/pets/etcd/etcd-petset.yaml
@@ -51,7 +51,7 @@ spec:
             exec:
               command:
                 - "/bin/sh"
-                - "-c"
+                - "-ec"
                 - |
                   EPS=""
                   for i in $(seq 0 $((${INITIAL_CLUSTER_SIZE} - 1))); do
@@ -73,7 +73,7 @@ spec:
                   fi
         command:
           - "/bin/sh"
-          - "-c"
+          - "-ec"
           - |
             HOSTNAME=$(hostname)
 
@@ -95,17 +95,6 @@ spec:
             member_hash() {
                 etcdctl member list | grep http://${HOSTNAME}.${PETSET_NAME}:2380 | cut -d':' -f1 | cut -d'[' -f1
             }
-
-            cleanup() {
-                export ETCDCTL_ENDPOINT=$(eps)
-
-                echo "Removing ${HOSTNAME} from etcd cluster"
-                etcdctl member remove $(member_hash)
-                exit 0
-            }
-
-            # Remove the member from the cluster before the container is terminated
-            trap cleanup TERM
 
             # re-joining after failure?
             if [ -e /var/run/etcd/default.etcd ]; then
--- a/pets/etcd/etcd-petset.yaml
+++ b/pets/etcd/etcd-petset.yaml
@@ -39,39 +39,81 @@ spec:
         - containerPort: 2379
           name: client
         env:
-        - name: CLUSTER_SIZE
+        - name: INITIAL_CLUSTER_SIZE
           value: "3"
         - name: PETSET_NAME
           value: "etcd"
         volumeMounts:
         - name: datadir
           mountPath: /var/run/etcd
+        lifecycle:
+          preStop:
+            exec:
+              command:
+                - "/bin/sh"
+                - "-c"
+                - |
+                  EPS=""
+                  for i in $(seq 0 $((${INITIAL_CLUSTER_SIZE} - 1))); do
+                      EPS="${EPS}${EPS:+,}http://${PETSET_NAME}-${i}.${PETSET_NAME}:2379"
+                  done
+
+                  HOSTNAME=$(hostname)
+
+                  member_hash() {
+                      etcdctl member list | grep http://${HOSTNAME}.${PETSET_NAME}:2380 | cut -d':' -f1 | cut -d'[' -f1
+                  }
+
+                  echo "Removing ${HOSTNAME} from etcd cluster"
+
+                  ETCDCTL_ENDPOINT=${EPS} etcdctl member remove $(member_hash)
+                  if [ $? -eq 0 ]; then
+                      # Remove everything otherwise the cluster will no longer scale-up
+                      rm -rf /var/run/etcd/*
+                  fi
         command:
           - "/bin/sh"
-          - "-ec"
+          - "-c"
           - |
-            for i in $(seq 0 $((${CLUSTER_SIZE} - 1))); do
-                while true; do
-                    echo "Waiting for ${PETSET_NAME}-${i}.${PETSET_NAME} to come up"
-                    ping -W 1 -c 1 ${PETSET_NAME}-${i}.${PETSET_NAME} > /dev/null && break
-                    sleep 1s
-                done
-            done
-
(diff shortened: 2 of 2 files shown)


ncdc:
@bprashanth @smarterclayton PTAL

pets/etcd/etcd-petset.yaml
```diff
@@ -0,0 +1,121 @@
...
+                echo "Re-joining etcd member"
+                member_id=$(cat /var/run/etcd/meta | cut -d':' -f1)
+
+                EPS=""
+                for i in $(seq 0 $((${CLUSTER_SIZE} - 1))); do
+                    EPS="${EPS}${EPS:+,}http://${PETSET_NAME}-${i}.${PETSET_NAME}:2379"
+                done
+
+                # re-join member
+                ETCDCTL_ENDPOINT=${EPS} etcdctl member update ${member_id} http://${HOSTNAME}.${PETSET_NAME}:2380
```
bprashanth: it would be great if you could do this in an init container and write the config to some hostPath shared location that etcd reads on startup

pets/etcd/etcd-petset.yaml
```diff
@@ -0,0 +1,121 @@
...
+            # store member id into PVC for later member replacement
+            collect_member() {
+                while ! etcdctl member list &>/dev/null; do sleep 1; done
+                etcdctl member list | grep ${HOSTNAME}.etcd > /var/run/etcd/meta
+            }
+
+            collect_member &
+
+            # join member
+            exec etcd --name ${HOSTNAME} \
```
bprashanth: ideally the entrypoint of your main etcd app container would be just this (like we do in the zookeeper example). Is that doable?

pets/etcd/etcd-petset.yaml
```diff
@@ -0,0 +1,121 @@
...
+        - name: datadir
+          mountPath: /var/run/etcd
+        command:
+          - "/bin/sh"
+          - "-ec"
+          - |
+            for i in $(seq 0 $((${CLUSTER_SIZE} - 1))); do
+                while true; do
+                    echo "Waiting for ${PETSET_NAME}-${i}.${PETSET_NAME} to come up"
+                    ping -W 1 -c 1 ${PETSET_NAME}-${i}.${PETSET_NAME} > /dev/null && break
```
bprashanth: do you need to wait for all peers or can you bootstrap the first one in standalone mode, then join them one by one? the disadvantage of doing this is:

1. that you can't specify a readiness probe (depending on the probe, it would probably deadlock here since you won't pass it till you get past this, but the petset won't scale till the pet is in ready). 

2. Scaling is harder

If you want to simply wait till you are up (and get a list of all previous peers), this simple script might help: https://github.com/kubernetes/contrib/tree/master/pets/peer-finder, it just does a dns resolution and sorts the list. You can pass your bash as --on-start or --on-change

pets/etcd/etcd-petset.yaml
```diff
@@ -0,0 +1,121 @@
...
+        image: gcr.io/google_containers/etcd-amd64:2.2.5
+        ports:
+        - containerPort: 2380
+          name: peer
+        - containerPort: 2379
+          name: client
+        env:
+        - name: CLUSTER_SIZE
+          value: "3"
+        - name: PETSET_NAME
```
bprashanth: you can figure this out from hostname?

pets/etcd/etcd-petset.yaml
```diff
@@ -0,0 +1,121 @@
...
+      containers:
+      - name: etcd
+        image: gcr.io/google_containers/etcd-amd64:2.2.5
+        ports:
+        - containerPort: 2380
+          name: peer
+        - containerPort: 2379
+          name: client
+        env:
+        - name: CLUSTER_SIZE
```
bprashanth: this makes scaling harder

bprashanth:
Seems like there're a lot of similarities with this and the zookeeper setup, would be nice to duplicate the pattern if possible (https://github.com/kubernetes/contrib/tree/master/pets/zookeeper). It would also make scaling easier.

pets/etcd/etcd-petset.yaml
ingvagabund: I could run ``$(hostname | rev | cut -d'-' -f2- | rev)``. Still, I would rather use the proposed ``IDENTITY`` parameter and then just updated the ``PETSET_NAME`` env instead of adding complexity to bootstrapping.

pets/etcd/etcd-petset.yaml
ingvagabund: Scaling is not supported atm. I need to know the number of members in advance. The cluster can be bootstrapped using DNS without knowing the cluster size. However, I need specific DNS entries for that [1].

[1] https://coreos.com/etcd/docs/latest/clustering.html#dns-discovery

pets/etcd/etcd-petset.yaml
ingvagabund: Currently, the static bootstrapping is used [1]. Thus, I need to wait for all peers to become reachable. It must fail the readiness probe before the cluster is ready.

[1] https://coreos.com/etcd/docs/latest/clustering.html#static

pets/etcd/etcd-petset.yaml
ingvagabund: With the current bootstrapping there are two entry points. One for the cluster being created for the first time. And second when a member is re-joined back. I would like to use the DNS discovery in future. Then, the script would become simpler.

pets/etcd/etcd-petset.yaml
ingvagabund: You can't. The ``/var/run/etcd/meta`` is generated after the cluster is bootstrapped. The file contains a member id which is know after the cluster started.

pets/etcd/etcd-petset.yaml
bprashanth: so I guess the SRV record discovery is the only way one grows an etcd cluster?

pets/etcd/etcd-petset.yaml
bprashanth: You get SRV records, can you use that: https://github.com/kubernetes/kubernetes.github.io/blob/release-1.3/docs/user-guide/petset.md#peer-discovery? or do we need modifications?

pets/etcd/etcd-petset.yaml
bprashanth: sure, this is a minor point

tschottdorf:
I tried running this for a while to no avail (all on 1.3) until I realized that I needed to manually create the PVCs expected by the cluster ahead of time. Is that the expected behavior? I would've expected that the necessary PVCs be auto-created (and automatically filled as long as a PV is available). Instead, creating the PetSet before the PVC does create a "zombie" PVC which is never filled, regardless of what I do.

Sorry about asking here, but I've looked around and wasn't able to find any resources on this. I'm sure they exist somewhere; appreciate a pointer.

bprashanth:
@ingvagabund if you actually want to invoke the dynamic provisioner you need to specify https://github.com/kubernetes/kubernetes.github.io/blob/release-1.3/docs/user-guide/petset.yaml#L43. 

@tschottdorf  see 4 in https://github.com/kubernetes/kubernetes.github.io/blob/release-1.3/docs/user-guide/petset.md#alpha-limitations and https://github.com/kubernetes/kubernetes.github.io/blob/release-1.3/docs/user-guide/petset.md#stable-storage

The provisioners are started based on your environment. From https://github.com/kubernetes/kubernetes/blob/release-1.3/examples/experimental/persistent-volume-provisioning/README.md
```
No configuration is required by the admin! 3 cloud providers will be provided in the alpha version of this feature: EBS, GCE, and Cinder.
```

pets/etcd/etcd-petset.yaml
ingvagabund: Yes, have not found any other way so far.

smarterclayton:
As noted on the cockroach thread you can do:

```
kind: PersistentVolume
apiVersion: v1
metadata:
  name: pv-host-1
spec:
  hostPath:
    path: /var/lib/random/dir-1
```

as necessary (you may have to set capacity and length).

bprashanth:
@ingvagabund did https://github.com/kubernetes/contrib/pull/1295#discussion_r69357495 work?

ingvagabund:
@bprashanth I will check that out. Sorry for the delay.

pets/etcd/etcd-petset.yaml
ingvagabund: Reading the ``peer-discovery`` and the etcd docs, I need at least ``_etcd-server._tcp.example.com`` srv record for each pet. Anyway, the DNS will only save some time at bootstrapping. It will not help with scaling.

In order to add a new member I need to distinguish between bootstrapping phase and after-bootstrapping one. At the same time it is not recommended to change to size from 3 to 5. It should be done one by one. At the same time if you decrease the size, random pet gets killed. I don't think there is a way for a pod to do "the last wish". At the same time if the pet is the etcd leader, it will take more time to elect new one.

Atm, member can be added/remove only via runtime configuration [1] as:

```sh
$ etcdctl member add infra3 http://10.0.1.13:2380
$ etcdctl member remove a8266ecf031671f3
```

[1] https://coreos.com/etcd/docs/latest/runtime-configuration.html

ingvagabund: Reading the scaling workflow [1], there could be a way. As pets are added/remove in deterministic order, one could:

1. rename ``CLUSTER_SIZE`` into ``INITIAL_CLUSTER_SIZE``
1. bootstrap the etcd cluster
1. if a new pet is created, the next pet has ``NAME-${INITIAL_CLUSTER_SIZE}`` name so based on that I can decide if the new pet is part of bootstrapping phase or not (for a new pet after bootstrapping, ``$ID >= $INITIAL_CLUSTER_SIZE``, where ``NAME-${ID}`` is the unique pet name, always holds).
1. the new pet is added via the runtime configuration process. If the pet dies, it follows the same process. Once the pet is running and ready, next pet (if there is any) is added the same way.
1. assuming there is a way to notify the pod about its removal, ``etcdctl member remove`` is run.

[1] https://github.com/kubernetes/kubernetes.github.io/blob/release-1.3/docs/user-guide/petset.md#scaling-a-pet-set

ingvagabund: Based on [1], each container should receive ``SIGTERM`` signal. Together with ``terminationGracePeriodSeconds`` a pet should have enough time to remove itself from the cluster.

[1] https://github.com/kubernetes/kubernetes/issues/6804

bprashanth: Sounds like it should work. Are you going to send a follow up or would you like me to wait till you prototype and modify this pr?

ingvagabund: Give me some time to extend the current PR so we don't have to do the review twice.

ingvagabund:
@bprashanth the etcd petset is now scalable. Though, I have hit some issues when scaling up and down:

* When running ``kubectl patch petset/etcd -p '{"spec":{"replicas": 3}}'`` from ``5`` to ``3`` replicas, pets ``etcd-4`` and ``etcd-5`` were not deleted.
* When running ``kubectl edit`` and changing number of replicas from ``13`` to ``5``, pets ``etcd-5`` and ``etcd-6`` were not deleted. After removing ``etcd-6``, both pets got deleted.

It does not have to be necessary related to the petset controller. Maybe something wrong in the specification.

pets/etcd/README.md
```diff
@@ -77,19 +77,54 @@ Updated member with ID 7fd61f3f79d97779 in cluster
...
+   KUBE_ENABLE_CLUSTER_DNS=true ENABLE_HOSTPATH_PROVISIONER=true ./hack/local-up-cluster.sh
    ```
 
 1. run skydns
 1. create the petset
 
 ## Scaling
 
-TODO
+The etcd cluster can be scale up by running ``kubectl patch`` or ``kubectl edit``. For instance,
```
sttts: kubectl scale?

ingvagabund:
Scaling up from ``3`` to ``13`` and back, the middle state of the petset is

```sh
NAME                 READY     STATUS    RESTARTS   AGE
etcd-0               1/1       Running   0          3m
etcd-1               1/1       Running   0          3m
etcd-11              1/1       Running   0          3m
etcd-12              1/1       Running   0          3m
etcd-2               1/1       Running   0          3m
etcd-3               1/1       Running   1          3m
```

Don't think pets ``etcd-12`` and ``etcd-11`` should exist if ``etcd-10`` is gone.

pets/etcd/README.md
ingvagabund:
```sh
$ kubectl scale --replicas=5 petset/etcd
error: no scaler has been implemented for {"apps" "PetSet"}
```

bprashanth:
Scaling will stop if a pet is unhealthy. It's possible that there are issues that need to be fixed when scaling halts because of an unhealth pet, and then the user scales up/down. The behavior currently is undefined in such situations. Scaling up should happen in order, once the petset is stable, scaling down should happen in reverse order. https://github.com/kubernetes/kubernetes/issues/30082

pets/etcd/etcd-petset.yaml
```diff
@@ -0,0 +1,205 @@
...
+            for i in $(seq 0 $((${INITIAL_CLUSTER_SIZE} - 1))); do
+                while true; do
+                    echo "Waiting for ${PETSET_NAME}-${i}.${PETSET_NAME} to come up"
+                    ping -W 1 -c 1 ${PETSET_NAME}-${i}.${PETSET_NAME} > /dev/null && break
+                    sleep 1s
+                done
+            done
+
+            PEERS=""
+            for i in $(seq 0 $((${INITIAL_CLUSTER_SIZE} - 1))); do
```
bprashanth: this stuff seems like exactly the type of thing to put in the init container (see zookeeper example:), any reason it should be in entrypoint? https://github.com/kubernetes/contrib/tree/master/pets/zookeeper/init. Putting it in init allows one to maintain both images independently, but I'm ok doing it this way if there are some limitations.

bprashanth: putting it in init also gives you a clear signal that bootstrapping/joining cluster failed

ingvagabund: > this stuff seems like exactly the type of thing to put in the init container (see zookeeper example:), any reason it should be in entrypoint?

There is nothing to install, setup/configure or bootstrap. Initial cluster bootstrapping consists of waiting for initial pets. The same holds for the adding post-bootstrap members. The script "just" adds new member and start etcd.

> Putting it in init allows one to maintain both images independently, but I'm ok doing it this way if there are some limitations. 

As all the code is basically written over etcd image, there is no need for additional one.

> putting it in init also gives you a clear signal that bootstrapping/joining cluster failed

Would not be ``postStart`` better? Basically, the main container command itself is to run ``etcd``. The current code "just" collects required envs, params, puts them together and start ``etcd``. So one either add member, remove member or update member. That is the only piece of logic worth putting into initialization part. Unfortunately, all three ``etcd`` commands differs in input options which are populated during initialization. So separation of initialization code and executing ``etcd`` itself does not make sense.

bprashanth:
I'm planning on waiting for consensus on https://groups.google.com/forum/#!topic/kubernetes-dev/MhQl2GBylgM (I think we're close) before doing a next pass. Hoping we can directly check this in there. Let me know if you'd rather check it in here and transition instead and I can do a review.

ingvagabund:
I am fine with that. At the same time (and as was commented in the discussion) we need some sort of CI to at least test 1) correctness and 2) regressions.

pets/etcd/etcd-petset.yaml
lachie83: I've looked into etcd DNS discovery for this purpose and it doesn't appear that the SRV records are configurable and PetSets won't lay down records that conform to what etcd is expecting.

Here's what they need to look like
```
_etcd-server-ssl._tcp.example.com
_etcd-client-ssl._tcp.example.com

File: pets/etcd/etcd-petset.yaml (hunk 7 above)
Comment (lachie83) [inline pets/etcd/etcd-petset.yaml]: Both are valid cases. I guess the question comes down to consistency

File: pets/etcd/etcd-petset.yaml (hunk 2 above)
Comment (bprashanth) [inline pets/etcd/etcd-petset.yaml]: @ingvagabund did you solve the scaling problem (I thought you said scaling worked now)? if not can you please open an issue in the main repo with the details and cc me?

Comment (ingvagabund) [inline pets/etcd/etcd-petset.yaml]: @bprashanth scaling is usually working up to some cases when I want to scale down. Some pets do not get deleted. Though, when describing the petset, it shows the correct number of pets. There are just some pet in the air no longer controlled by the petset controller.

File: pets/etcd/etcd-petset.yaml (hunk 5 above)
Comment (ingvagabund) [inline pets/etcd/etcd-petset.yaml]: Implemented

File: pets/etcd/etcd-petset.yaml (hunk 2 above)
Comment (ingvagabund) [inline pets/etcd/etcd-petset.yaml]: Back to the "Yes, have not found any other way so far.". It was incorrect, it does not matter if I use static discovery or the DNS one. It is used for bootstrapping only. Once than, the etcd cluster is scaled via runtime configuration.

Comment (lachie83) [inline pets/etcd/etcd-petset.yaml]: @ingvagabund - pods aren't currently terminated when a PetSet is deleted. See -- http://kubernetes.io/docs/user-guide/petset/#alpha-limitations

Comment (ingvagabund) [inline pets/etcd/etcd-petset.yaml]: @lachie83 when you scale up from 3 to 13 replicas and then back to 3, sometimes, some pets in interval <4;13> does not get deleted.

Comment (lachie83) [inline pets/etcd/etcd-petset.yaml]: My guess is that there is a bug in the PetSet resource. Let me see if I can dig a little further

Comment (bprashanth) [inline pets/etcd/etcd-petset.yaml]: I will try to get a repro post code freeze. I haven't tested rapid scaling up/down too much (https://github.com/kubernetes/kubernetes/issues/30082) because that's not the target use case for petset. You're meant to scale, wait for pets to join cluster/stabilize, re-scale. If a pet is unhealthy (includes failing readiness) the controller won't take any action, it'll wait for admin intervention.

Comment (bprashanth) [inline pets/etcd/etcd-petset.yaml]: In the meanwhile if either of you are interested in debugging the petset controller is quite simple: https://github.com/kubernetes/kubernetes/tree/master/pkg/controller/petset, this gets all pods based on labels https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/petset/pet_set.go#L304 this creates an iterator over a queue created with that list: https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/petset/pet_set.go#L304. All the pets are initially inserted in the "delete" state:https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/petset/iterator.go#L100 , and flipped to "sync" till spec.replicas:https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/petset/iterator.go#L128

There's some bts caching going on of course, but this loop should happen on watch and periodically every 30s (https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/petset/pet_set.go#L48) so even if the controller missed an edge it should converge.

**Comment by mdshuai** [comment]
@ingvagabund Is the etcd pod must run as root and in etcd namespaces?

File: pets/etcd/etcd-petset.yaml (hunk 2 above)
Comment (ingvagabund) [inline pets/etcd/etcd-petset.yaml]: Thanks for the links. It is true I did not wait 30s everytime it occurred. So maybe it is working as expected, it just needs more time to finish.

**Comment by ingvagabund** [comment]
> @ingvagabund Is the etcd pod must run as root and in etcd namespaces?

Etcd process runs as a root by default unless specified otherwise. Not aware of any etcd namespace requirement. What is your usecase here to test?

File: pets/etcd/etcd-petset.yaml [hunk 8]
```diff
@@ -0,0 +1,190 @@
...
+
+            # re-joining after failure?
+            if [ -e /var/run/etcd/default.etcd ]; then
+                echo "Re-joining etcd member"
+                member_id=$(cat /var/run/etcd/member_id)
+
+                # re-join member
+                ETCDCTL_ENDPOINT=$(eps) etcdctl member update ${member_id} http://${HOSTNAME}.${PETSET_NAME}:2380
+                exec etcd --name ${HOSTNAME} \
+                    --listen-peer-urls http://${HOSTNAME}.${PETSET_NAME}:2380 \
```
mdshuai: `{HOSTNAME}.${PETSET_NAME}` can't be resolve by dns. should be ``{HOSTNAME}.${SERVICE_NAME}``, in this example, you set petse-tname=svc-name so it can running well. Need add another env `SERVER_NAME` to replace some fields.

ingvagabund: Can you dump entries in your DNS? I have never had this problem.

pets/etcd/etcd-petset.yaml
```diff
@@ -0,0 +1,190 @@
...
+        command:
+          - "/bin/sh"
+          - "-ec"
+          - |
+            HOSTNAME=$(hostname)
+
+            # store member id into PVC for later member replacement
+            collect_member() {
+                while ! etcdctl member list &>/dev/null; do sleep 1; done
+                etcdctl member list | grep ${HOSTNAME}.etcd | cut -d':' -f1 | cut -d'[' -f1 > /var/run/etcd/member_id
```
mdshuai: this grep hardcode

pets/etcd/etcd-petset.yaml
mdshuai: In this design doc https://github.com/bprashanth/kubernetes.github.io/blob/4df7d96369964a267466bdffc73f7d343c77584f/docs/user-guide/petset.md#network-identity `As each pet is created, it gets a matching DNS subdomain, taking the form: $(petname).$(governing service domain), where the governing service is defined by the serviceName field on the Pet Set.`
If you set petsetname don't equal to svc name, then you can see this issue.

test-foxish:
recomputing cla status...

grodrigues3:

- **pets**

cc @lavalamp 

foxish:
Closing this, it should probably exist as an example outside contrib.

PR #1301, opened 2016-06-30
Cluster autoscaler: add handlign for nill mig configs (in case not all l node pools are autoscaled)

mwielgus:
With the previous fix #1282 a possibility for null pointer exceptions/panics was opened as GceManager started to return nills in case of nodes not belonging to an autoscaled mig. This PR fixes the NPE possibilities there and should be included in the release version ASAP.

cc: @piosz @jszczepkowski @fgrzadkowski

### DIFF
--- a/cluster-autoscaler/scale_down.go
+++ b/cluster-autoscaler/scale_down.go
@@ -133,6 +133,11 @@ func ScaleDown(
 				glog.Errorf("Error while checking mig config for instance %v: %v", instance, err)
 				continue
 			}
+			if migConfig == nil {
+				glog.V(4).Infof("Skipping %s - no mig config", node.Name)
+				continue
+			}
+
 			size, err := gceManager.GetMigSize(migConfig)
 			if err != nil {
 				glog.Errorf("Error while checking mig size for instance %v: %v", instance, err)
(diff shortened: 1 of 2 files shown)


piosz:
LGTM

fgrzadkowski:
@mwielgus Why this didn't come up during tests for #1282? AFAIU it should always panic if we have a non-autoscaled mig, right?

mwielgus:
Honestly - I don't know - i updated the cluster autoscaler on multi-node-pool cluster on @olekzabl cluster and it correctly scaled it down. @olekzabl do you still have this cluster running? 

I'm recreating the cluster so we can do the test again.

olekzabl:
The resources we've talked about are here:
* https://test-container.sandbox.googleapis.com/
* project: cloud-kubernetes-dev
* olekz-cluster49 (1.3.0-beta.1): 
  - default-pool of size 1 (autoscaling off), 
  - pool-1 of requested size 1-2 (autoscaling on), originally had size 3 for >2h but now has size 1
* olekz-cluster52 (1.3.0-beta.2):
  - default-pool of requested size 1-2 (autoscaling on), still has size 3 after >1d of running
* olekz-cluster53 (1.3.0-beta.2.191 - this was the version you advised me to request):
  - like ...-52; in error state since creation, and gcloud told me this version is somehow bad. (?)

PR #1306, opened 2016-07-01
Added check-labels munger and corresponding tests.

foxish:
https://github.com/kubernetes/contrib/issues/1423
The munger checks for changes in label config yaml file and creates labels as necessary in the repository.

### DIFF
--- a/mungegithub/github/github.go
+++ b/mungegithub/github/github.go
@@ -190,32 +190,34 @@ type analytics struct {
 	cachedAPICount     int       // how many api calls were answered by the local cache
 	apiPerSec          float64
 
-	AddLabels          analytic
-	RemoveLabels       analytic
-	ListCollaborators  analytic
-	GetIssue           analytic
-	CloseIssue         analytic
-	CreateIssue        analytic
-	ListIssues         analytic
-	ListIssueEvents    analytic
-	ListCommits        analytic
-	GetCommit          analytic
-	ListFiles          analytic
-	GetCombinedStatus  analytic
-	SetStatus          analytic
-	GetPR              analytic
-	AssignPR           analytic
-	ClosePR            analytic
-	OpenPR             analytic
-	GetContents        analytic
-	ListComments       analytic
-	ListReviewComments analytic
-	CreateComment      analytic
-	DeleteComment      analytic
-	Merge              analytic
-	GetUser            analytic
-	SetMilestone       analytic
-	ListMilestones     analytic
+	AddLabels            analytic
+	AddLabelToRepository analytic
+	RemoveLabels         analytic
+	ListCollaborators    analytic
+	GetIssue             analytic
+	CloseIssue           analytic
+	CreateIssue          analytic
+	ListIssues           analytic
+	ListIssueEvents      analytic
+	ListCommits          analytic
+	ListLabels           analytic
+	GetCommit            analytic
+	ListFiles            analytic
+	GetCombinedStatus    analytic
+	SetStatus            analytic
+	GetPR                analytic
+	AssignPR             analytic
+	ClosePR              analytic
+	OpenPR               analytic
+	GetContents          analytic
+	ListComments         analytic
+	ListReviewComments   analytic
+	CreateComment        analytic
+	DeleteComment        analytic
+	Merge                analytic
+	GetUser              analytic
+	SetMilestone         analytic
+	ListMilestones       analytic
 }
 
 func (a analytics) print() {
@@ -225,6 +227,7 @@ func (a analytics) print() {
 	w := new(tabwriter.Writer)
 	w.Init(buf, 0, 0, 1, ' ', tabwriter.AlignRight)
 	fmt.Fprintf(w, "AddLabels\t%d\t\n", a.AddLabels.Count)
+	fmt.Fprintf(w, "AddLabelToRepository\t%d\t\n", a.AddLabelToRepository.Count)
 	fmt.Fprintf(w, "RemoveLabels\t%d\t\n", a.RemoveLabels.Count)
 	fmt.Fprintf(w, "ListCollaborators\t%d\t\n", a.ListCollaborators.Count)
 	fmt.Fprintf(w, "GetIssue\t%d\t\n", a.GetIssue.Count)
@@ -233,6 +236,7 @@ func (a analytics) print() {
 	fmt.Fprintf(w, "ListIssues\t%d\t\n", a.ListIssues.Count)
 	fmt.Fprintf(w, "ListIssueEvents\t%d\t\n", a.ListIssueEvents.Count)
 	fmt.Fprintf(w, "ListCommits\t%d\t\n", a.ListCommits.Count)
+	fmt.Fprintf(w, "ListLabels\t%d\t\n", a.ListLabels.Count)
 	fmt.Fprintf(w, "GetCommit\t%d\t\n", a.GetCommit.Count)
 	fmt.Fprintf(w, "ListFiles\t%d\t\n", a.ListFiles.Count)
 	fmt.Fprintf(w, "GetCombinedStatus\t%d\t\n", a.GetCombinedStatus.Count)
@@ -1782,4 +1786,43 @@ func (config *Config) ListAllIssues(listOpts *github.IssueListByRepoOptions) ([]
 		page++
 	}
 	return allIssues, nil
+}
+
+// GetLabels grabs all labels from a particular repository so you don't have to
+// worry about paging.
+func (config *Config) GetLabels() ([]*github.Label, error) {
+	var listOpts github.ListOptions
+	var allLabels []*github.Label
+	page := 1
+	for {
+		glog.V(4).Infof("Fetching page %d of labels", page)
+		listOpts = github.ListOptions{PerPage: 100, Page: page}
+		labels, response, err := config.client.Issues.ListLabels(config.Org, config.Project, &listOpts)
+		config.analytics.ListLabels.Call(config, response)
+		if err != nil {
+			return nil, err
+		}
+		for i := range labels {
+			allLabels = append(allLabels, labels[i])
+		}
+		if response.LastPage == 0 || response.LastPage <= page {
+			break
+		}
+		page++
+	}
+	return allLabels, nil
+}
+
+// AddLabel adds a single github label to the repository.
+func (config *Config) AddLabel(label *github.Label) error {
+	config.analytics.AddLabelToRepository.Call(config, nil)
+	glog.Infof("Adding label %v to %v, %v", *label.Name, config.Org, config.Project)
+	if config.DryRun {
+		return nil
+	}
+	_, _, err := config.client.Issues.CreateLabel(config.Org, config.Project, label)
+	if err != nil {
+		return err
+	}
+	return nil
 }
--- a/mungegithub/mungers/check-labels.go
+++ b/mungegithub/mungers/check-labels.go
@@ -1,0 +1,142 @@
+/*
+Copyright 2015 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package mungers
+
+import (
+	"fmt"
+	"os"
+
+	"k8s.io/contrib/mungegithub/features"
+	githubhelper "k8s.io/contrib/mungegithub/github"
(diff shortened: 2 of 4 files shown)


foxish:
@janetkuo @lavalamp Please review.

mungegithub/mungers/check-labels_test.go
```diff
@@ -0,0 +1,118 @@
...
+		{
+			name:       "Identical labels in repository and file.",
+			repoLabels: fileLabels,
+			expected:   []*github.Label{},
+		},
+		{
+			name:       "Adding new labels with existing labels in repository.",
+			repoLabels: repoLabels,
+			expected:   fileLabels[0:2],
+		},
```
janetkuo: What if fileLabels contains a label with the same name but different color?

foxish: As far as I checked, the label name needs to be unique, or github won't let us add it, even with a different color. In check-labels.go:103, the name of the label is checked for uniqueness and we don't add it in the above case.

mungegithub/github/github.go
```diff
@@ -1692,3 +1696,65 @@ func (config *Config) ListAllIssues(listOpts *github.IssueListByRepoOptions) ([]
...
+		if response.LastPage == 0 || response.LastPage <= page {
+			break
+		}
+		page++
+	}
+	return allLabels, nil
+}
+
+// AddLabel adds a single github label to the repository.
+func (config *Config) AddLabel(label *github.Label) error {
```
eparis: I much prefer the analytics be collected and we print a log message for any 'write' type operation before the `DryRun` return.

mungegithub/mungers/check-labels.go
```diff
@@ -0,0 +1,140 @@
...
+	if len(file) == 0 {
+		glog.Fatalf("No --label-file= supplied, cannot check labels")
+	}
+	c.labelAccessor = config
+	return nil
+}
+
+// EachLoop is called at the start of every munge loop
+func (c *CheckLabelsMunger) EachLoop() error {
+	fileContents, sha, err := c.labelAccessor.GetFileContents(c.labelFileName)
```
eparis: We already have the whole master git tree which is used by the blunderbuss assignment. Do we need to use api calls to fetch this data via the github api?

mungegithub/mungers/check-labels.go
```diff
@@ -0,0 +1,140 @@
...
+	if err != nil {
+		glog.Errorf("Failed to read the check label config: %v", err)
+		return err
+	}
+	if c.lastSha != sha {
```
eparis: This gets run every couple of minutes. Is the extra code to not process the file really worth it?

mungegithub/mungers/check-labels.go
```diff
@@ -0,0 +1,140 @@
...
+			} else {
+				glog.Infof("Added label %s", *label.Name)
+			}
+		}
+	}
+}
+
+// AddFlags will add any request flags to the cobra `cmd`
+func (c *CheckLabelsMunger) AddFlags(cmd *cobra.Command, config *githubhelper.Config) {
+	cmd.Flags().StringVar(&c.labelFileName, "label-file", "", "Path from repository root to file containing"+
```
eparis: Most places we just default this to whatever we do in the core kube repo...

eparis:
Do you want this enabled in the core kube repo? If so you need to update the submit-queue/deployment.yaml...

mungegithub/mungers/check-labels_test.go
janetkuo: Yes add a test case for that sounds good.

mungegithub/mungers/check-labels.go
foxish: @eparis Since we are making mungegithub generic in order to work with other repositories as well, would it be fair to assume that all of them will have the same fetching of the master git tree? I designed it to use the API because then it is independent of whether the blunderbuss assigner is being run. However, if we will be running blunderbuss against other repositories in the same way, I'll go ahead and make the change to read the local file from the pull for blunderbuss.

lavalamp: Yes, that would be the idea-- we'd want all our repositories to be
consistent.

On Wed, Jul 6, 2016 at 3:44 PM, Anirudh Ramanathan <notifications@github.com

mungegithub/mungers/check-labels_test.go
foxish: Done.

mungegithub/github/github.go
foxish: Done.

foxish:
@eparis @janetkuo PTAL

mungegithub/mungers/check-labels.go
foxish: Done.

mungegithub/mungers/check-labels.go
foxish: When we process the file, we also pull all the labels from the repository, which needs multiple API calls. It is avoided when we see that the labels config file has not changed.

mungegithub/features/repo-updates.go
```diff
@@ -204,6 +206,21 @@ func (o *RepoInfo) LeafAssignees(path string) sets.String {
 	return peopleForPath(path, o.assignees, true)
 }
 
+// ReadFile returns the contents of a file in the checked out copy of the
+// repository
+func (o *RepoInfo) ReadFile(path string) (string, error) {
+	file, err := os.Open(filepath.Join(o.kubernetesDir, path))
```
janetkuo: Maybe I missed something, but why not just use `ioutil.ReadFile`?

foxish: Done.

foxish:
@lavalamp, @apelisse If this looks ok, I'd like to get this merged. We'll need it in the near future and the change just adds the munger but doesn't turn it on.

mungegithub/features/repo-updates.go
```diff
@@ -236,6 +237,17 @@ func (o *RepoInfo) LeafAssignees(path string) sets.String {
 	return peopleForPath(path, o.assignees, true)
 }
 
+// ReadFile returns the contents of a file in the checked out copy of the
+// repository
+func (o *RepoInfo) ReadFile(path string) (string, error) {
```
lavalamp: Generally it's best to return []byte rather than string.

mungegithub/features/repo-updates.go
```diff
@@ -236,6 +237,17 @@ func (o *RepoInfo) LeafAssignees(path string) sets.String {
...
+	bytes, err := ioutil.ReadFile(path)
```
lavalamp: Is this entire function useful? I was expecting it to modify the path. Why wouldn't people just call ioutil.ReadFile if they want to do this?

mungegithub/features/repo-updates.go
```diff
@@ -236,6 +237,17 @@ func (o *RepoInfo) LeafAssignees(path string) sets.String {
...
+	if err != nil {
+		glog.Errorf("%v", err)
```
lavalamp: I recommend not logging if you pass the error up the stack.

mungegithub/mungers/check-labels.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	if _, err := os.Stat(c.labelFilePath); os.IsNotExist(err) {
+		return fmt.Errorf("Failed to stat the check label config: %v", err)
+	}
+
+	return nil
+}
+
+func (c *CheckLabelsMunger) getSha(fileContents string) string {
```
lavalamp: SHA

lavalamp: Actually I'd call it getHash

mungegithub/mungers/check-labels.go
```diff
@@ -0,0 +1,160 @@
...
+	h := sha1.New()
+	h.Write([]byte(fileContents))
```
lavalamp: See, here you want a []byte :)

lavalamp: Consider canonicalizing the file before you hash it, so you don't trigger on e.g. sorts or whitespace changes

mungegithub/mungers/check-labels.go
```diff
@@ -0,0 +1,160 @@
...
+type labelAccessor interface {
+	AddLabel(label *github.Label) error
+	GetLabels() ([]*github.Label, error)
+}
+
+// CheckLabelsMunger will check that the labels specified in the labels yaml file
+// are created.
+type CheckLabelsMunger struct {
+	labelFilePath string
+	prevSha       string
```
lavalamp: prevHash

mungegithub/mungers/check-labels.go
```diff
@@ -0,0 +1,160 @@
...
+	labelAccessor labelAccessor
+	features      *features.Features
+	reader        func() (string, error)
```
lavalamp: readFunc

mungegithub/mungers/check-labels.go
```diff
@@ -0,0 +1,160 @@
...
+func (c *CheckLabelsMunger) RequiredFeatures() []string { return []string{features.RepoFeatureName} }
+
+// Initialize will initialize the munger.
+func (c *CheckLabelsMunger) Initialize(config *githubhelper.Config, features *features.Features) error {
+	if len(c.labelFilePath) == 0 {
+		glog.Fatalf("No --label-file= supplied, cannot check labels")
+	}
+	c.labelAccessor = config
+	c.features = features
+	c.reader = func() (string, error) {
```
lavalamp: Just `c.reader = c.features.Repos.ReadFile` should work

mungegithub/mungers/check-labels.go
```diff
@@ -0,0 +1,160 @@
...
+		if err != nil {
+			return err
+		}
+		c.createNewLabels(repoLabels, fileLabels["labels"])
+		c.prevSha = sha
+	}
+	return nil
+}
+
+func (c *CheckLabelsMunger) createNewLabels(repoLabels, fileLabels []*github.Label) {
```
lavalamp: document that this won't remove existing labels.

lavalamp: maybe "addMissingLabels"?

mungegithub/mungers/check-labels.go
```diff
@@ -0,0 +1,165 @@
...
+func (c *CheckLabelsMunger) AddFlags(cmd *cobra.Command, config *githubhelper.Config) {
+	cmd.Flags().StringVar(&c.labelFilePath, "label-file", "", "Path from repository root to file containing"+
+		" list of labels")
+}
+
+// Munge is unused by this munger.
+func (c *CheckLabelsMunger) Munge(obj *githubhelper.MungeObject) {}
+
+// This function can be used to write a check-labels config file.
+func writeLabelFile(c *CheckLabelsMunger, filename string) error {
```
lavalamp: does anything actually call this function? I recommend omitting it if not.

mungegithub/mungers/check-labels_test.go
```diff
@@ -0,0 +1,128 @@
...
+	"runtime"
+	"testing"
+
+	"github.com/golang/glog"
+	"github.com/google/go-github/github"
+)
+
+var (
+	_                              = fmt.Printf
+	_                              = glog.Errorf
```
lavalamp: There shouldn't be a need to actually check in these two lines.

mungegithub/mungers/check-labels_test.go
```diff
@@ -0,0 +1,128 @@
...
+	repoLabelNames []string        = []string{"team/abc", "team/def", "team/ghi", "label/rep"}
```
lavalamp: please just make []*github.Label variables rather than parallel []strings.

mungegithub/mungers/check-labels.go
foxish: Nothing calls it, but it is the way to write out valid labels-config yaml files when building the list from an existing repository. I thought it might be useful to include it.

lavalamp: Dead code rots quickly :)

Add it in another PR if you think it's important.

mungegithub/mungers/check-labels_test.go
foxish: I had trouble setting it directly because the `github.Label` expects `*strings`. I couldn't find a way to create and initialize a `*string` in a single step. Is there a way to do that? Else I could make it a single array of pairs of strings rather than two separate arrays perhaps.

eparis: The tests use `stringPtr()`

foxish: I should have looked harder. Works like a charm. Thanks @eparis.

mungegithub/features/repo-updates.go
foxish: There used to be path modification before, but not anymore. Removed the function to use ioutil.ReadFile directly.

foxish:
@lavalamp, addressed comments, except for canonicalizing the file, because the cost of a re-rerun in case it detects a change in case of spaces/sort order (which themselves seem rare) costs two API calls only. Do you think we should still do it?

lavalamp:
LGTM

Issue #1305, opened 2016-07-01
Commit a85ed07 : Break Coreos install with ansible 


ekozan:
```
TASK [etcd : Call restart etcd] ************************************************
skipping: [***]

TASK [etcd : Call restart etcd2] ***********************************************
included: /Development/Kube/contrib/ansible/roles/etcd/tasks/etcd2_restart.yml for ***

TASK [etcd : restart etcd2] ****************************************************
fatal: [***]: FAILED! => {"failed": true, "msg": "The conditional check 'etcd2_modified == true and etcd_started.changed == false' failed. The error was: error while evaluating conditional (etcd2_modified == true and etcd_started.changed == false): 'etcd_started' is undefined\n\nThe error appears to have been in '/Development/Kube/contrib/ansible/roles/etcd/tasks/main.yml': line 12, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Call restart etcd2\n  ^ here\n"}
```

ekozan:
@ping ingvagabund

ingvagabund:
@stephenrlouie ?

ingvagabund:
From https://github.com/kubernetes/contrib/blob/master/ansible/roles/etcd/tasks/etcd2-start.yml:

```yaml
- name: CoreOS | Start etcd2
  service: name=etcd2 state=started
  register: etcd2_started
```

s/etcd2_started/etcd_started.

@ekozan can you update the ``etcd2-start.yml`` manually and see if it works?

ingvagabund:
https://github.com/kubernetes/contrib/pull/1309

PR #1309, opened 2016-07-03
Fix incorrect variable in Call restart etcd2's conditional

merlin83:
Bugfix for incorrect variable name

Fixes: #1305

### DIFF
--- a/ansible/roles/etcd/tasks/main.yml
+++ b/ansible/roles/etcd/tasks/main.yml
@@ -11,4 +11,4 @@
 
 - name: Call restart etcd2
   include: etcd2_restart.yml
-  when: etcd2_modified == true and etcd_started.changed == false
+  when: etcd2_modified == true and etcd2_started.changed == false


ansible/roles/etcd/tasks/main.yml
```diff
@@ -11,4 +11,4 @@
 
 - name: Call restart etcd2
   include: etcd2_restart.yml
-  when: etcd2_modified == true and etcd_started.changed == false
+  when: etcd2_modified == true and etcd2_started.changed == false
```
ingvagabund: ``etcd_started`` variable is universal. Let's update start_etcd2.yaml and s/etcd2_started/etcd_started/ instead. Later, I would like to see both etcd and etcd2 unified, remove ``etcd2_*`` vars and use ``is_coreos`` (or other is_os vars) instead. Other distributions can introduce ``etcd3`` component as etcd-3.*.* is to be complete rewrite of etcd-2.*.*.

stephenrlouie: @ingvagabund I see where you want to take this for service wide variables, but I think keeping etcd2 might be more readable, since all the file names for the CoreOS etcd story is under etcd2. If you are looking for service wide variable names, the etcd role should be refactored to have one etcd-start.yml, one etcd-stop.yml.... and within those have all the statements contain the conditional `is_coreos` to distinguish between the etcd plays. I thought having a bunch of statements with identical conditionals was verbose.

I broke them down in my original refactor because I thought it was better to simplify all playbooks and make it easy to separate the different processes. I fought with that decision for a while and could see it going either way, single `etcd-started.yml` files or separate.


Visually this LGTM. I did not pick this error up in my virtualenv, I believe it's because I'm doing a clean vagrant up every time.

merlin83:
I signed it!

ansible/roles/etcd/tasks/main.yml
ingvagabund: @stephenrlouie ok, let's keep it that way and refactor it later if we find enough reasons to do that.

ingvagabund:
Thanks @merlin83 for the patch.

PR #1310, opened 2016-07-04
Ingress: Updated ingress docs, avoid nginx-alpha use.

kayrus:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1310)

### DIFF
--- a/ingress/controllers/README.md
+++ b/ingress/controllers/README.md
@@ -8,60 +8,86 @@ An Ingress Controller is a daemon, deployed as a Kubernetes Pod, that watches th
 
 ## Writing an Ingress Controller
 
-Writing an Ingress controller is simple. By way of example, the [nginx controller] (nginx-alpha) does the following:
+Writing an Ingress controller is simple. By way of example, the [nginx controller](nginx) does the following:
 * Poll until apiserver reports a new Ingress
-* Write the nginx config file based on a [go text/template](https://golang.org/pkg/text/template/)
+* Write the nginx config file based on a [go text/template](https://golang.org/pkg/text/template/).
 * Reload nginx
 
-Pay attention to how it denormalizes the Kubernetes Ingress object into an nginx config:
+By default nginx controller uses [this template](nginx/nginx.tmpl).
+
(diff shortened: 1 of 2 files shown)


kayrus:
I signed it!

kayrus:
I signed it!

aledbf:
@kayrus why this change? The code in `nginx-alpha` contains the most important pieces required to build an ingress controller. The other two implementations are too complex (if you are trying to build a custom one from)

kayrus:
@aledbf nginx-alpha doesn't contain `proxy_set_header Host $host;` inside the template. And when you are trying to switch to nginx with more complex config - it doesn't work out of the box. My approach was to resolve both problems.

aledbf:
> And when you are trying to switch to nginx with more complex config - it doesn't work out of the box

What do you mean?

kayrus:
@aledbf the amount of `maps` inside default nginx [template](https://github.com/kubernetes/contrib/blob/master/ingress/controllers/nginx/nginx.tmpl) (not alpha) doesn't allow to allocate memory for them. nginx even can not start, controller logs errors. then healthz after 60 sec timeout and pod is being recreated again and again. increasing the `map_hash_bucket_size` doesn't help.

aledbf:
@kayrus can you describe what/how are you running the ingress nginx controller?
I'm using that controller with more than 300 ingress rules without issues.

kayrus:
@aledbf simple k8s [webapp demo](https://github.com/kayrus/k8s-webapp). Here is the [ingress config](https://github.com/kayrus/k8s-webapp/blob/master/exposed_80_port.yaml) which I used.

bprashanth:
This readme should contain the absolute simplest nginx controller which is the alpha controller. I'm fine forking off another doc to containe a better example, but I don't want to complicate the instructions for writing a controller. I'm also fine adding a "known issues" section to nginx-alpha (i'd hoped the name was enough to surface that there were known issues).

test-foxish:
recomputing cla status...

Issue #1314, opened 2016-07-05
[nginx-ingress-controller] Nginx aborts reloads w/ certificate errors


fcvarela:
The ingress controller writes certificates to /etc/nginx-ssl on a goroutine and reload nginx on another. On a busy nginx controller, I've observed that nginx throws errors like:

`2016/07/05 11:32:56 [emerg] 13#13: PEM_read_bio_X509("/etc/nginx-ssl/xxx-yyy") failed (SSL: error:0906D066:PEM routines:PEM_read_bio:bad end line)`

After this happens any rc/pod/deployment restart that results in pods getting new IPs will make the entire upstream unavailable (503 on nginx) as the aborted reload renders it unaware of those changes.

Manual inspection of nginx.conf shows the upstream does have the new ips, that the certificates and keys are correct.

A second manual reload solves the issue.

This is caused by writing certificates/keys using ioutil (async) while nginx is trying to read them. I have a working and verified patch that changes it to tempfiles and atomic renames.

aledbf:
@fcvarela do you have some step to replicate this error?
(I'm trying to get this error in order to test your PR)

fcvarela:
Unfortunately no, it requires setting up a few hundred ingress rules spanning a few dozen hosts all using tls secrets. I set logging level to --v=5 and noticed the nginx error interleaved with logging from ssl.go updating pem files from secrets, indicating the nginx reload signal was triggered while files were being written in a non atomic way, hence the rename from tmp instead of asynchronous write in place.

aledbf:
@fcvarela how many certificates are you trying to use?
I made [this script](https://gist.github.com/aledbf/f8ce0e89c6051212f2e65ef3acc996d5) that creates 1000 certificates and ingress rules and I cannot reproduce this issue

fcvarela:
Not that many, we are using 12 certificates covering a few hundred hosts (one of the certs is a wildcard) and a few hundred ingress rules. My impression is that once the pem files are generated all will work fine. But if you have continuous activity generating ingress rules and adding hosts or containers crash-looping (and therefore registering/unregistering endpoints as upstream servers), then the pem generation will run continously. This problem happens when nginx is reloading from a previous iteration and new pem files are getting generated. You should see the problem occurring if you keep the ingress generation loop going indefinitely.

fcvarela:
Here's the error output (--v=5)
[...]

PR #1315, opened 2016-07-05
Addresses #1314 [nginx-ingress-controller ssl nginx reload abort]

fcvarela:
Replaces async file writes with atomic renaming of temp files.

### DIFF
--- a/ingress/controllers/nginx/nginx/ssl.go
+++ b/ingress/controllers/nginx/nginx/ssl.go
@@ -45,17 +45,28 @@ type SSLCert struct {
 
 // AddOrUpdateCertAndKey creates a .pem file wth the cert and the key with the specified name
 func (nginx *Manager) AddOrUpdateCertAndKey(name string, cert string, key string) (SSLCert, error) {
-	pemFileName := config.SSLDirectory + "/" + name + ".pem"
+	temporaryPemFileName := fmt.Sprintf("%v.pem", name)
+	pemFileName := fmt.Sprintf("%v/%v.pem", config.SSLDirectory, name)
 
-	pem, err := os.Create(pemFileName)
+	temporaryPemFile, err := ioutil.TempFile("", temporaryPemFileName)
 	if err != nil {
-		return SSLCert{}, fmt.Errorf("Couldn't create pem file %v: %v", pemFileName, err)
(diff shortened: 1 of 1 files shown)


ingress/controllers/nginx/nginx/ssl.go
```diff
@@ -45,17 +45,28 @@ type SSLCert struct {
 
 // AddOrUpdateCertAndKey creates a .pem file wth the cert and the key with the specified name
 func (nginx *Manager) AddOrUpdateCertAndKey(name string, cert string, key string) (SSLCert, error) {
+	tempPemFileName := name + ".pem"
 	pemFileName := config.SSLDirectory + "/" + name + ".pem"
```
aledbf: maybe this could be `pemFileName := fmt.Sprintf("%v/%v", config.SSLDirectory, tempPemFileName)`

fcvarela: Sure, I used the same format as the existing pemFileName to keep it consistent in method. Should I update that as well?

aledbf: @fcvarela yes, I'm trying to avoid string concatenations and use `fmt.Sprintf`

fcvarela:
I've added the suggested changes and rebased my ref, is that ok?

fcvarela:
Should I squash all three commits?

aledbf:
@fcvarela yes please

fcvarela:
Squashed!

aledbf:
@bprashanth please review

ingress/controllers/nginx/nginx/ssl.go
```diff
@@ -45,17 +45,28 @@ type SSLCert struct {
...
-		return SSLCert{}, fmt.Errorf("Couldn't write to pem file %v: %v", pemFileName, err)
+		return SSLCert{}, fmt.Errorf("Couldn't write to pem file %v: %v", temporaryPemFile.Name(), err)
+	}
+
+	err = temporaryPemFile.Close()
+	if err != nil {
+		return SSLCert{}, fmt.Errorf("Couldn't close temp pem file %v: %v", temporaryPemFile.Name(), err)
+	}
+
+	err = os.Rename(temporaryPemFile.Name(), pemFileName)
```
bprashanth: it would be even better if we validated the key/cert pair before moving it (though I think we already do so higher up the stack)

bprashanth:
LGTM

PR #1317, opened 2016-07-05
[nginx-ingress-controller] Fix duplicated real_ip_header

### LINKED ISSUE #1316: [nginx-ingress-controller] real_ip_header is duplicate when useProxyProtocol is enabled
It seems as though since https://github.com/kubernetes/contrib/commit/a152cc44707b8f214666ba7468ceb6575478333b, if useProxyProtocol is enabled, real_ip_header is set twice causing an error when reloading nginx: 

```
nginx: [emerg] "real_ip_header" directive is duplicate in /etc/nginx/nginx.conf:65
```

This prevents the ingress controller from starting in 0.8 if useProxyProtocol is enabled.

aledbf:
fixes #1316

### DIFF
--- a/ingress/controllers/nginx/nginx.tmpl
+++ b/ingress/controllers/nginx/nginx.tmpl
@@ -17,8 +17,14 @@ events {
 
 http {
     {{/* we use the value of the header X-Forwarded-For to be able to use the geo_ip module */}}
+    {{ if $cfg.useProxyProtocol -}}
+    set_real_ip_from    {{ $cfg.proxyRealIpCidr }};
+    real_ip_header      proxy_protocol;
+    {{ else }}
     real_ip_header      X-Forwarded-For;
     set_real_ip_from    0.0.0.0/0;
+    {{ end -}}
+    
     real_ip_recursive   on;
 
     {{/* databases used to determine the country depending on the client IP address */}}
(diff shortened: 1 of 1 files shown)


ingress/controllers/nginx/nginx.tmpl
```diff
@@ -17,8 +17,14 @@ events {
 
 http {
     {{/* we use the value of the header X-Forwarded-For to be able to use the geo_ip module */}}
+    {{ if $cfg.useProxyProtocol -}}
+    set_real_ip_from    {{ $cfg.proxyRealIpCidr }};
+    real_ip_header      proxy_protocol;
```
bprashanth: for my own eductaion, this actually sets the ip on the forwarded packet from the proxy protocol ascii block, right?

bprashanth:
I thought we were doing some static checking of the config before reload (and not reloading if the config was invalid)? if that's not the case, can you please file a follow up bug. LGTM.

ingress/controllers/nginx/nginx.tmpl
aledbf: yes

aledbf:
If reload returns an error nginx continues using the previous version

bprashanth:
ah so this won't cause downtime right? just a broken feature

aledbf:
exactly. This is an issue when you start with an error in nginx.conf not in a reload

bprashanth:
hmm, we can probably work around that too by persisting the last known good template in a volume of configmap, reading on boot up, and only overwriting it on success reload. But a better first step would be an e2e for the nginx controller :)

PR #1323, opened 2016-07-06
Cluster-autoscaler: cloud provider interface implementation for GCE

mwielgus:

Part of #1311

### DIFF
--- a/cluster-autoscaler/cloudprovider/cloud_provider.go
+++ b/cluster-autoscaler/cloudprovider/cloud_provider.go
@@ -27,12 +27,12 @@ type CloudProvider interface {
 	Name() string
 
 	// NodeGroups returns all node groups configured for this cloud provider.
-	NodeGroups() []*NodeGroup
+	NodeGroups() []NodeGroup
 
 	// NodeGroupForNode returns the node group for the given node, nil if the node
 	// should not be processed by cluster autoscaler, or non-nil error if such
 	// occurred.
-	NodeGroupForNode(*kube_api.Node) (*NodeGroup, error)
+	NodeGroupForNode(*kube_api.Node) (NodeGroup, error)
 }
 
 // NodeGroup contains configuration info and functions to control a set
@@ -44,8 +44,11 @@ type NodeGroup interface {
 	// MinSize returns minimum size of the node group.
 	MinSize() int
 
-	// Size returns the current size of the node group.
-	Size() (int, error)
+	// TargetSize returns the current target size of the node group. It is possible that the
+	// number of nodes in Kubernetes is different at the moment but should be equal
+	// to Size() once everything stabilizes (new nodes finish startup and registration or
+	// removed nodes are deleted completely)
+	TargetSize() (int, error)
 
 	// GetSampleNode returns a sample node that belongs to this node group, or error
 	// if occurred.
@@ -53,13 +56,13 @@ type NodeGroup interface {
 
 	// IncreaseSize increases the size of the node group. To delete a node you need
 	// to explicitly name it and use DeleteNode. This function should wait until
-	// the node appears in the Kubernetes cluster.
+	// node group size is updated.
 	IncreaseSize(delta int) error
 
-	// DeleteNode deletes the node from this node group. Error is returned either on
+	// DeleteNodes deletes nodes from this node group. Error is returned either on
 	// failure or if the given node doesn't belong to this node group. This function
-	// should wait until the node is removed from the Kubernetes cluster.
-	DeleteNode(*kube_api.Node) error
+	// should wait until node group size is updated.
+	DeleteNodes([]*kube_api.Node) error
 
 	// Id returns an unique identifier of the node group.
 	Id() string
--- a/cluster-autoscaler/cloudprovider/gce/gce_cloud_provider.go
+++ b/cluster-autoscaler/cloudprovider/gce/gce_cloud_provider.go
@@ -1,0 +1,267 @@
+/*
+Copyright 2016 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package gce
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+
+	"k8s.io/contrib/cluster-autoscaler/cloudprovider"
+	kube_api "k8s.io/kubernetes/pkg/api"
+)
+
+// NodeProvider is a function that provides a list of nodes.
+type NodeProvider func() ([]*kube_api.Node, error)
+
+// GceCloudProvider implements CloudProvider interface.
+type GceCloudProvider struct {
+	gceManager   *GceManager
+	migs         []*Mig
+	nodeProvider NodeProvider
+}
+
+// BuildGceCloudProvider builds CloudProvider implementation for GCE.
+func BuildGceCloudProvider(gceManager *GceManager, nodeProvider NodeProvider, specs []string) (*GceCloudProvider, error) {
+	gce := &GceCloudProvider{
+		gceManager:   gceManager,
+		migs:         make([]*Mig, 0),
+		nodeProvider: nodeProvider,
+	}
+	for _, spec := range specs {
+		if err := gce.addNodeGroup(spec); err != nil {
+			return nil, err
+		}
+	}
+	return gce, nil
+}
+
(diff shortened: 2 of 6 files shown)


cluster-autoscaler/cloudprovider/gce/gce_cloud_provider.go
```diff
@@ -0,0 +1,261 @@
...
+	return &GceCloudProvider{
+		gceManager:   gceManager,
+		migs:         make([]*Mig, 0),
+		nodeProvider: nodeProvider,
+	}
+}
+
+// AddNodeGroup adds node group defined in string spec. Format:
+// minNodes:maxNodes:migUrl
+func (gce *GceCloudProvider) AddNodeGroup(spec string) error {
```
piosz: Shouldn't this be passed in a constructor? Do you have any use case for adding NodeGroup in flight?

cluster-autoscaler/cloudprovider/gce/gce_cloud_provider.go
```diff
@@ -0,0 +1,261 @@
...
+	mig, err := buildMig(spec, gce.gceManager)
+	if err != nil {
+		return err
+	}
+	nodes, err := gce.nodeProvider()
+	if err != nil {
+		return err
+	}
+	for _, node := range nodes {
+		if belongs, err := mig.Belongs(node); err == nil && belongs {
```
piosz: As we discussed before we probably want to revisit the way how sample node is being chosen. Maybe add a TODO here to do it better?

cluster-autoscaler/cloudprovider/gce/gce_cloud_provider.go
```diff
@@ -0,0 +1,261 @@
...
+	return result
+}
+
+// NodeGroupForNode returns the node group for the given node.
+func (gce *GceCloudProvider) NodeGroupForNode(node *kube_api.Node) (cloudprovider.NodeGroup, error) {
+	ref, err := GceRefFromProviderId(node.Spec.ProviderID)
+	if err != nil {
+		return nil, err
+	}
+	mig, err := gce.gceManager.GetMigForInstance(ref)
```
piosz: optional: inline those two vars

cluster-autoscaler/cloudprovider/gce/gce_cloud_provider.go
mwielgus: Done.

cluster-autoscaler/cloudprovider/gce/gce_cloud_provider.go
```diff
@@ -0,0 +1,261 @@
...
+
+	gceManager *GceManager
+
+	minSize    int
+	maxSize    int
+	sampleNode *kube_api.Node
+}
+
+// Url builds GCE url for the MIG.
+func (mig *Mig) Url() string {
```
piosz: Why do you need this? It's equivalent to Id().

cluster-autoscaler/cloudprovider/gce/gce_cloud_provider.go
mwielgus: Done

cluster-autoscaler/cloudprovider/cloud_provider.go
```diff
@@ -44,7 +44,10 @@ type NodeGroup interface {
 	// MinSize returns minimum size of the node group.
 	MinSize() int
 
-	// Size returns the current size of the node group.
+	// Size returns the current target size of the node group. It is possible that the
+	// number of nodes in Kubernetes is different at the moment but should be equal
+	// to Size() once everything stabilizes (new nodes finish startup and registration or
+	// removed nodes are deleted completely)
 	Size() (int, error)
```
piosz: Maybe just TargetSize()?

cluster-autoscaler/cloudprovider/gce/gce_cloud_provider.go
mwielgus: Done.

cluster-autoscaler/cloudprovider/cloud_provider.go
mwielgus: I would rather leave it as is.

cluster-autoscaler/cloudprovider/gce/gce_cloud_provider.go
mwielgus: Ack

cluster-autoscaler/cloudprovider/cloud_provider.go
piosz: It's usually more convenient for users to have clear method name rather than ambiguous one and 4 lines of semantic description, but as you wish;)

cluster-autoscaler/cloudprovider/gce/gce_cloud_provider.go
```diff
@@ -0,0 +1,263 @@
...
+	if mig.sampleNode != nil {
+		return mig.sampleNode, nil
+	}
+	return nil, fmt.Errorf("no sample node available")
+}
+
+// IncreaseSize increases Mig size
+func (mig *Mig) IncreaseSize(delta int) error {
+	if delta <= 0 {
+		return fmt.Errorf("size increase must be positive")
```
piosz: so why not making delta uint?

cluster-autoscaler/cloudprovider/gce/gce_cloud_provider.go
```diff
@@ -0,0 +1,263 @@
...
+	return true, nil
+}
+
+// DeleteNode deletes the node from the group.
+func (mig *Mig) DeleteNode(node *kube_api.Node) error {
+	size, err := mig.gceManager.GetMigSize(mig)
+	if err != nil {
+		return err
+	}
+	if int(size) <= mig.MinSize() {
```
piosz: High level idea: how about making all var/methods which operates on size use the same type to avoid conversions in many places?

cluster-autoscaler/cloudprovider/gce/gce_cloud_provider.go
```diff
@@ -0,0 +1,263 @@
...
+		return false, fmt.Errorf("%s doesn't belong to a known mig", node.Name)
+	}
+	if targetMig.Id() != mig.Id() {
+		return false, nil
+	}
+	return true, nil
+}
+
+// DeleteNode deletes the node from the group.
+func (mig *Mig) DeleteNode(node *kube_api.Node) error {
```
piosz: DeleteNodes? As for now we only support scaling down by 1 node, but probably we want to change it soon

cluster-autoscaler/cloudprovider/gce/gce_cloud_provider.go
```diff
@@ -0,0 +1,263 @@
...
+func (mig *Mig) Id() string {
+	return GenerateMigUrl(mig.Project, mig.Zone, mig.Name)
+}
+
+// Debug returns a debug string for the Mig.
+func (mig *Mig) Debug() string {
+	return fmt.Sprintf("%s (%d:%d)", mig.Id(), mig.MinSize(), mig.MaxSize())
+}
+
+// Set adds a new configuration.
```
piosz: Weird comment

cluster-autoscaler/cloudprovider/gce/gce_cloud_provider.go
mwielgus: suppose that some has -1 in an int variable. Stupid int->uint conversion would make it some huge number that added to the current size couple lines below would effectively mean -1. In this way -1 would pass through the checks.

cluster-autoscaler/cloudprovider/cloud_provider.go
mwielgus: Ok, you convinced me.

cluster-autoscaler/cloudprovider/gce/gce_cloud_provider.go
mwielgus: Done.

cluster-autoscaler/cloudprovider/gce/gce_cloud_provider.go
mwielgus: Next PR.

cluster-autoscaler/cloudprovider/gce/gce_cloud_provider.go
mwielgus: Done.

piosz:
LGTM

cluster-autoscaler/cloudprovider/gce/gce_cloud_provider.go
```diff
@@ -0,0 +1,267 @@
...
+		return err
+	}
+	if int(size)+delta > mig.MaxSize() {
+		return fmt.Errorf("size increase to large - desired:%d max:%d", int(size)+delta, mig.MaxSize())
+	}
+	return mig.gceManager.SetMigSize(mig, size+int64(delta))
+}
+
+// Belongs retruns true if the given node belongs to the NodeGroup.
+func (mig *Mig) Belongs(node *kube_api.Node) (bool, error) {
```
pbitty: @mwielgus Does this method need to be exported?

PR #1324, opened 2016-07-06
GCE Ingress docs update

bprashanth:
Update docs for 1.3


ingress/controllers/gce/BETA_LIMITATIONS.md
```diff
@@ -158,23 +94,29 @@ GCE has a concept of [ephemeral](https://cloud.google.com/compute/docs/instances
...
+spec:
+  tls:
+  - secretName: tls-secret
+  backend:
+    serviceName: echoheaders-https
+    servicePort: 80
 ```
 
-* Nullify the RC (but don't delete it or the addon controller will "fix" it for you)
+As of Kubernetes 1.3, GLBC runs as a static pod on the master. If you want to totally disable it, you can ssh into the master node and delete the GLBC manifest file found at `/etc/kubernetes/manifests/glbc.manifest`. You can also disable it on GKE at cluster bring-up time through the `disable-addons` flag, eg:
```
girishkalele: Is the GKE disable/enable addons capability only accessible from the CLI ? Also, can they disable/enable addons for a running GKE cluster ?

ingress/controllers/gce/README.md
```diff
@@ -381,22 +382,7 @@ metadata:
 type: Opaque
```
girishkalele: Nit: This example yaml would read better if the data fields were at the end, and Kind, metadata.name fields were at the beginning.

ingress/controllers/gce/README.md
```diff
@@ -381,22 +382,7 @@ metadata:
...
-spec:
-  tls:
-  - secretName: testsecret
-  backend:
-    serviceName: s1
-    servicePort: 80
-```
-
-This creates 2 GCE forwarding rules that use a single static ip. Both `:80` and `:443` will direct traffic to your backend, which serves HTTP requests on the target port mentioned in the Service associated with the Ingress.
+Referencing this secret in an Ingress will tell the Ingress controller to secure the channel from the client to the loadbalancer using TLS.
```
girishkalele: Is the ingress controller hardcoded to pull the tls.crt and tls.key subfields or is this for the example at hand ?
Is there any logic that validates the TLS cert and private key are valid and not expired (in the case of the cert) when the yaml create is posted ?

ingress/controllers/gce/README.md
```diff
@@ -409,95 +395,121 @@ if ($http_x_forwarded_proto = "http") {
...
+Here's an example that demonstrates it, first lets create a self signed certificate:
 
 ```console
-$ kubectl create -f /tmp/tls.json
-$ kubectl describe secret tls-secret
-Name:		tls-secret
-Namespace:	default
-Labels:		<none>
-Annotations:	<none>
+$ openssl req -x509 -nodes -days 365 -newkey rsa:2048 -keyout /tmp/tls.key -out /tmp/tls.crt -subj "/CN=foobar.com"
```
girishkalele: Add a warning about the 365 day expiration cliff - some folks will convert this example to production use without realizing the expiration cliff exists.

ingress/controllers/gce/BETA_LIMITATIONS.md
bprashanth: > they disable/enable addons for a running GKE cluster ?

Currently they need to use the annotation to talk to another ingress controller. Users generally can't do anything about master components on GKE (eg: they can't turn off RCs).

ingress/controllers/gce/README.md
bprashanth: > Is the ingress controller hardcoded to pull the tls.crt and tls.key subfields or is this for the example at hand ?

Yeah, it's mentioned above: "The TLS secret must contain keys named `tls.crt` and `tls.key` that contain the certificate and private key to use for TLS, eg:". It needs to match: https://github.com/kubernetes/kubernetes/blob/master/pkg/api/types.go#L2696 (linked to that page). 

> Is there any logic that validates the TLS cert and private key are valid and not expired (in the case of the cert) when the yaml create is posted ?

That'll show up as an event at runtime. kubectl create secret will do some validation. 

> Would be good to spell out how to debug TLS failures - point them to a good FAQ.

I'll send a follow up. This part is just like they'd debug TLS failures outside GCE.

ingress/controllers/gce/README.md
bprashanth: done

ingress/controllers/gce/README.md
bprashanth: Added a line about expiration

bprashanth:
PTAL

girishkalele:
LGTM to me

bprashanth:
Cool thanks merging

PR #1326, opened 2016-07-07
Update the Makefile to allow local_dryrun to work

99BenGardner:


99BenGardner:
i signed the cla!

99BenGardner:
I signed it!

mungegithub/Makefile
```diff
@@ -63,8 +79,8 @@ endif
 
 # simple transformation of a github oauth secret file to a kubernetes secret
 secret:
-	@echo $(token)
-	sed -e 's|1234567890123456789012345678901234567890123456789012345=|$(token)|' $(APP)/secret.yaml > $(APP)/local.secret.yaml
+	@echo $(token64)
+	sed -e 's|1234567890123456789012345678901234567890123456789012345=|$(token64)|' $(APP)/secret.yaml > $(APP)/local.secret.yaml
```
eparis: I think we should update this to use `kubectl create secrete` instead of using `sed` but that's not your fault.

PR #1328, opened 2016-07-08
Cluster-autoscaler: use cloud provider interface in the code

mwielgus:
Plus remove SampleNode() function from CloudProvider interface.

Ref: #1311

cc: @iterion @piosz @fgrzadkowski @jszczepkowski


jszczepkowski:
LGTM

PR #1338, opened 2016-07-08
[nginx-ingress-controller]: Add HTTPS default backend

### LINKED ISSUE #1329: Nginx Ingress Controller - Default Backend when using HTTPS
NGINX redirects all HTTP traffic for undefined hosts to the default backend. But when using HTTPS the traffic all goes to the first defined TLS host in the generated `nginx.conf`. 

By default NGINX should direct all traffic regardless of HTTP or HTTPS to the default backend. When using HTTPS NGINX should serve a default self-signed certificate.

What do you think?

### LINKED ISSUE #1331: Nginx Ingress Controller - global default certificate
I want to be able to configure a certificate in the namespace of the Ingress Controller and it should be used for all Ingress rules, regardless of their namespace and their Ingress TLS configuration.
An Individual Ingress resource should still be able to overwrite it.

**Use-case:**
Wild card DNS and certificate are set up for *.mycorp.com. A user can now setup an ingress rule for myns.mycorp.com/app1 and it will be protected with TLS without having to configure certs or a TLS ingress rule.

**Implementation proposal:**
New optional command line parameter for the controller: a reference to the default secret.
If present and the cert matches spec.rules.host, TLS will be enabled.
If a rule specifies it's own cert, it will overwrite the default cert.
TCP and UDP services are not affected.

What do you think? Would it make sense to add this to the controller?

I tried to solve this via this Ingress resource that the doc mentions but I could not get it to work:
```
kind: Ingress
metadata:
  name: no-rules-map
spec:
  tls:
    secretName: testsecret
  backend:
    serviceName: s1
    servicePort: 80
```
How is this supposed to work?
Also all 30 seconds or so the following log message appears in the Ingress controller log:
```
I0708 14:26:54.489232       1 controller.go:933] Using the secret ingresssecret as source for the default SSL certificate
```

aledbf:
fixes #1331 
fixes #1329

### DIFF
--- a/hack/verify-flags/known-flags.txt
+++ b/hack/verify-flags/known-flags.txt
@@ -108,3 +108,4 @@ whitelist-override-label
 config-file-path
 required-retest-contexts
 retest-body
+default-ssl-certificate
--- a/ingress/controllers/nginx/README.md
+++ b/ingress/controllers/nginx/README.md
@@ -9,6 +9,7 @@ This is a nginx Ingress controller that uses [ConfigMap](https://github.com/kube
 * [Deployment](#deployment)
 * [HTTP](#http)
 * [HTTPS](#https)
+  * [Default SSL Certificate](#default-ssl-certificate)
   * [HTTPS enforcement](#server-side-https-enforcement)
   * [HSTS](#http-strict-transport-security)
 * [TCP Services](#exposing-tcp-services)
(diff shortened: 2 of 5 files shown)


bprashanth:
LGTM

gambol99:
Post this merge i'm getting the following error, with SSL no longer working ... 

2016/07/21 11:27:30 [error] 24#24: *224 no "ssl_certificate" is defined in server listening on SSL port while SSL handshaking, client: 10.10.4.1, server: 0.0.0.0:443
2016/07/21 11:27:30 [error] 24#24: *225 no "ssl_certificate" is defined in server listening on SSL port while SSL handshaking, client: 10.10.4.1, server: 0.0.0.0:443

I has to remove the section below to fix it

```yaml
{{/* Check for default SSL backend */}}
 +    {{ if and (eq $server.Name "_") (not $server.SSL) -}}
 +    server {
 +        server_name {{ $server.Name }};
 +        listen 443;
 +        # return protocol error.
 +        return 444;
 +    }
 +    {{ end }}

```

aledbf:
@gambol99 this issue was fixed in #1393.

PR #1339, opened 2016-07-08
Update election code, client and godeps

aledbf:


bprashanth:
@mikedanese review? Not sure who is point on this

mikedanese:
Sure, I can take a look.

election/Makefile
```diff
@@ -1,17 +1,22 @@
 all: push
 
 # 0.0 shouldn't clobber any released builds
-TAG = 0.0
+TAG = 0.5
```
mikedanese: Comment above is irrelevent if we want to make this change. How about
```Makefile
# 0.0 shouldn't clobber any released builds
# current latest is 0.5
TAG ?= 0.0
```
So tag will needed to be provided explicitly to deploy new version to gcr.io.

election/client/nodejs/Dockerfile
```diff
@@ -1,3 +1,5 @@
-FROM node:0.12
+FROM node:6.3-slim
```
mikedanese: Major bump :). Is this basically latest? (disclosure: I know nothing about node or javascript).

mikedanese:
Changes looks sensible. The diffstat (+824 −18,815) seems wrong... but If tests pass, let's merge. Thanks for keeping up to date. I'll do a push to gcr.io after merging.

election/client/nodejs/Dockerfile
aledbf: > Is this basically latest?

Yes, 6.3 is the latest stable version (check [this](https://nodejs.org/en/download/releases/) to see how old is 0.12 )

election/Makefile
aledbf: done

mikedanese:
pushed

PR #1340, opened 2016-07-11
Track issues added/removed/tested

fejta:
Added three integers, which increment in the following way:
- `issuesAdded` - whenever an item is added to the queue
- `issuesRemoved` - whenever an item is removed from the queue
- `issuesTested` -- whenever an item is tested

These values will allow us to track how rate we dequeue items. This rate usually differs from both the merge rate (we remove items without merging them) as well as the change in the size of the queue (we can add items as quickly as we remove them).

This also allows us to distinguish how quickly we tend to merge things from how quickly we test things (since we can test things but not merge them if they fail tests).

A couple additional small refactorings:
- Alphabetize `submitQueueStats`
- Encapsulate deleting issues from the queue into `deleteIssue()`

Test version visible at http://146.148.108.121/sq-stats

### DIFF
--- a/mungegithub/mungers/submit-queue.go
+++ b/mungegithub/mungers/submit-queue.go
@@ -125,15 +125,17 @@ type healthRecord struct {
 // information about the sq itself including how fast things are merging and
 // how long since the last merge
 type submitQueueStats struct {
-	StartTime          time.Time
-	LastMergeTime      time.Time
-	MergesSinceRestart int
-	MergeRate          float64
-	RetestsAvoided     int
+	Added              int // Number of items added to the queue since restart
 	FlakesIgnored      int
-
-	// Will be true if we've made at least one complete pass.
-	Initialized bool
+	Initialized        bool // true if we've made at least one complete pass
+	InstantMerges      int  // Number of commits without retests required
+	LastMergeTime      time.Time
+	MergeRate          float64
+	MergesSinceRestart int
+	Removed            int // Number of items dequeued since restart
+	RetestsAvoided     int
+	StartTime          time.Time
+	Tested             int // Number of e2e tests completed
 }
 
 // pull-request that has been tested as successful, but interrupted because head flaked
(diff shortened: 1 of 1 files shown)


eparis:
needs gofmt

fejta:
@eparis thanks! updated

mungegithub/mungers/submit-queue.go
```diff
@@ -125,15 +125,16 @@ type healthRecord struct {
...
-	MergesSinceRestart int
 	MergeRate          float64
+	MergesSinceRestart int
 	RetestsAvoided     int
-	FlakesIgnored      int
-
-	// Will be true if we've made at least one complete pass.
-	Initialized bool
+	StartTime          time.Time
+	IssuesAdded        int // Number of items added to the queue since restart
```
lavalamp: PRsAdded? And same everywhere. "Issues" is confusing, I think.

lavalamp: Or "QueueAdds"?

fejta: Done

mungegithub/mungers/submit-queue.go
```diff
@@ -915,6 +920,13 @@ func (sq *SubmitQueue) Munge(obj *github.MungeObject) {
 	return
 }
 
+func (sq *SubmitQueue) deleteIssue(obj *github.MungeObject) {
```
lavalamp: deleteQueueItem or deletePR?

mungegithub/mungers/submit-queue.go
```diff
@@ -1143,6 +1155,8 @@ func (sq *SubmitQueue) doGithubE2EAndMerge(obj *github.MungeObject) bool {
 			return true
 		}
 
+		atomic.AddInt32(&sq.prsTested, 1)
```
lavalamp: Should probably do this even if it timed out?

mungegithub/mungers/submit-queue.go
fejta: Done (sorry for missing that one)

mungegithub/mungers/submit-queue.go
fejta: Done

mungegithub/mungers/submit-queue.go
```diff
@@ -1129,6 +1141,7 @@ func (sq *SubmitQueue) doGithubE2EAndMerge(obj *github.MungeObject) bool {
 
 		// Wait for the retest to start
 		sq.SetMergeStatus(obj, ghE2EWaitingStart)
+		atomic.AddInt32(&sq.prsTested, 1)
```
lavalamp: Hm, I know I said that this was probably OK, but maybe we should wait for the tests to go pending at least, to know that it actually took affect? Up to you unless @eparis has a strong opinion one way or the other.

lavalamp:
LGTM

PR #1346, opened 2016-07-11
Load test owners from local kubernetes repo clone.

rmmh:
I tested this locally with `./mungegithub --dry-run --pr-mungers=flake-manager --test-owners-csv /tmp/owners.csv`

### DIFF
--- a/mungegithub/mungers/testowner/owner_test.go
+++ b/mungegithub/mungers/testowner/owner_test.go
@@ -54,7 +54,7 @@ func TestOwnerList(t *testing.T) {
 
 func TestOwnerListFromCsv(t *testing.T) {
 	r := bytes.NewReader([]byte(",,header nonsense,\n" +
-		",owner,suggested owner,test name\n" +
+		",owner,suggested owner,name\n" +
 		",foo,other,Test name\n" +
 		",bar,foo,other test\n"))
 	list, err := NewOwnerListFromCsv(r)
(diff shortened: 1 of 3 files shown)


mungegithub/mungers/testowner/owner_test.go
```diff
@@ -54,7 +54,7 @@ func TestOwnerList(t *testing.T) {
 
 func TestOwnerListFromCsv(t *testing.T) {
 	r := bytes.NewReader([]byte(",,header nonsense,\n" +
-		",owner,suggested owner,test name\n" +
+		",owner,suggested owner,name\n" +
```
lavalamp: Is there a reason not to use the enconding/csv package?

lavalamp:
LGTM, you can fix the comment in a followup if you like.

mungegithub/mungers/testowner/owner_test.go
rmmh: The file is small and simple enough to write out directly for the test case.

foxish:
The deployment file points to `--test-owners-csv=/gitrepos/kubernetes/test_owners.csv` which expects to find `test_owners.csv` at the root of the kubernetes repo. I don't think we have that yet, which makes my test deployment fail at the moment. 
Maybe we should make changes to the deployment after creating the corresponding file in the main repository?

lavalamp:
Indeed, that's the wrong path.
https://github.com/kubernetes/kubernetes/pull/28402

@rmmh, please fix :)

rmmh:
Fixed in #1364

Issue #1304, opened 2016-07-01
Make mungegithub work on all our repos


bgrant0607:
cc @kubernetes/contributor-experience

foxish:
A couple of questions I had, @lavalamp/@eparis, if you could answer these, it would help me out.

- How and where from are the [health charts for the past week](http://storage.googleapis.com/kubernetes-test-history/k8s-queue-health.svg) generated? 
- Can we assume that the underlying mechanism by which we run tests and read log files will remain the same for other repositories? In that case, it seems like it should be sufficient to remove from code into YAML files, parameters like the Jenkins bucket-name, to make the submit-queue generic enough for other repositories.

lavalamp:
That's in the test-infra repo, @fejta and I talked about getting that
running in the same pod as the munger.

And yes, I think we should mandate the same test result distribution
technique, just with a different bucket per repo (or maybe different
directory, but I think different bucket would be less confusing).

foxish:
There is a hardcoded "requiredContexts" [here](https://github.com/kubernetes/contrib/blob/master/mungegithub/mungers/submit-queue.go#L72), and a parameter we supply from the commandline [here](https://github.com/kubernetes/contrib/blob/master/mungegithub/mungers/submit-queue.go#L488). We use the former in some places (such as [this](https://github.com/kubernetes/contrib/blob/master/mungegithub/mungers/ok-to-test.go#L71) and [this](https://github.com/kubernetes/contrib/blob/master/mungegithub/mungers/stale-green-ci.go#L85)) and the latter in other places. This looks to me like a bug.

lavalamp:
Yeah, those should be exposed as flags.

Also default values to our flags should be moved into the deployment.yaml
file.

foxish:
Okay. Does it also make sense to remove defaults from code completely and just have them in the yaml file? Some places may need some additional error handling code in that case.

lavalamp:
Depending on the default, yes. Definitely anything with a string like
"kubernetes" in it. And yes, I wouldn't be surprised if more error handling
was needed.

foxish:
The first [PR](https://github.com/kubernetes/contrib/pull/1349). The first one which makes the submit-queue accept GCS parameters such as bucket-name and directory from the command-line. I'm unable to add reviewers, could you please add them for me?

bgrant0607:
@foxish Once this is working, it should assign PRs in contrib automatically. :-)

foxish:
[...]

PR #1349, opened 2016-07-12
Make submit-queue run against other projects

foxish:
#1304

Added feature to access google-cloud-storage and specify bucket
and directory parameters from the command-line.

### DIFF
--- a/mungegithub/features/features.go
+++ b/mungegithub/features/features.go
@@ -26,8 +26,9 @@ import (
 // Features are all features the code know about. Care should be taken
 // not to try to use a feature which isn't 'active'
 type Features struct {
-	Repos  *RepoInfo
-	active []feature
+	Repos   *RepoInfo
+	GCSInfo *GCSInfo
+	active  []feature
 }
 
 type feature interface {
@@ -47,7 +48,7 @@ func (f *Features) GetActive() []feature {
 // Initialize should be called with the set of all features needed by all (active) mungers
 func (f *Features) Initialize(requestedFeatures []string) error {
 	for _, name := range requestedFeatures {
-		glog.Infof("Initilizing feature: %v", name)
+		glog.Infof("Initializing feature: %v", name)
 		feat, found := featureMap[name]
 		if !found {
 			return fmt.Errorf("Could not find a feature named: %s", name)
@@ -59,6 +60,8 @@ func (f *Features) Initialize(requestedFeatures []string) error {
 		switch name {
 		case RepoFeatureName:
 			f.Repos = feat.(*RepoInfo)
+		case GCSFeature:
+			f.GCSInfo = feat.(*GCSInfo)
 		}
 	}
 	return nil
--- a/mungegithub/mungegithub.go
+++ b/mungegithub/mungegithub.go
@@ -95,13 +95,15 @@ func main() {
 			if len(config.PRMungersList) == 0 {
 				glog.Fatalf("must include at least one --pr-mungers")
 			}
-			err := mungers.InitializeMungers(config.PRMungersList, &config.Config, &config.Features)
-			if err != nil {
-				glog.Fatalf("unable to initialize requested mungers: %v", err)
+			if err := mungers.RegisterMungers(config.PRMungersList); err != nil {
+				glog.Fatalf("unable to find requested mungers: %v", err)
 			}
 			requestedFeatures := mungers.RequestedFeatures()
 			if err := config.Features.Initialize(requestedFeatures); err != nil {
 				return err
+			}
+			if err := mungers.InitializeMungers(&config.Config, &config.Features); err != nil {
+				glog.Fatalf("unable to initialize mungers: %v", err)
 			}
 			return doMungers(config)
 		},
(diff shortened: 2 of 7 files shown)


mungegithub/features/features.go
```diff
@@ -92,3 +95,13 @@ func RegisterFeature(feat feature) error {
 	glog.Infof("Registered %#v at %s", feat, feat.Name())
 	return nil
 }
+
+// GetFeature returns a feature when requested by name.
+func GetFeature(name string) feature {
+	var feat feature
```
apelisse: No need to define variable first, just use `:=`

mungegithub/mungers/submit-queue.go
```diff
@@ -381,8 +383,8 @@ func (sq *SubmitQueue) internalInitialize(config *github.Config, features *featu
 			gcs = utils.NewTestUtils("bucket", "logs", overrideUrl)
 		} else {
 			gcs = utils.NewWithPresubmitDetection(
-				utils.KubekinsBucket, utils.LogDir,
-				utils.PullKey, utils.PullLogDir,
+				sq.features.GCSInfo.BucketName, sq.features.GCSInfo.LogDir,
+				sq.features.GCSInfo.PullKey, sq.features.GCSInfo.PullLogDir,
```
apelisse: I assume this is the  main part of the feature: we want to be able to change these value. Does it need to be a feature though?

foxish: I had a discussion with @lavalamp about this. GCSInfo needs to be used by two different mungers, and supplied from the commandline. Because it could be used by several mungers, it should exist outside any individual munger in my opinion. The options were:
* to make them global flags populated when mungegithub is run, in the main function (but that makes them sort of opaque).
* Introduce a new struct, which would in turn be passed into every munger in its Initialize() method (this needed changes to the Mungers interface).
* Making it a feature, similar to the feature 'gitrepos' used to read the locally fetched repository in blunderbuss.

I went with the third because it seemed like the most natural way to approach it, and its similarity to how 'gitrepos' works. If there is a better mechanism I may have missed, please let me know. :)

mungegithub/features/features.go
foxish: Removing the 'GetFeature' function, as it is unused.

mungegithub/mungers/mungers.go
```diff
@@ -69,15 +69,22 @@ func RequestedFeatures() []string {
 	return out.List()
 }
 
-// InitializeMungers will call munger.Initialize() for all mungers requested
```
apelisse: How is this related? What problem is it trying to fix?

mungegithub/mungers/submit-queue.go
apelisse: OK sgtm

mungegithub/mungers/mungers.go
foxish: Previously, the mungers would first be validated & initialized, before the necessary features were initialized. The order was:

- mungers register & initialize
- features initialize

However, submit-queue.go would use the GCS feature during its initialization, so, it was necessary to initialize the features before the mungers. Hence, I broke it up such that the order becomes:

- mungers are registered
- list of features is obtained
- features initialize
- mungers initialize

apelisse: Cool, makes sense. Can you squash the three commits? Then LGTM

foxish: Done. Thank you for the quick review @apelisse.

mungegithub/features/features.go
```diff
@@ -26,7 +26,8 @@ import (
 // Features are all features the code know about. Care should be taken
 // not to try to use a feature which isn't 'active'
 type Features struct {
-	Repos  *RepoInfo
+	Repos *RepoInfo
+	*GCSInfo
```
lavalamp: `GCSInfo *GCSInfo` unless you're purposely embedding the methods.

foxish: Did not know that it embedded methods that way. Fixed.

lavalamp:
LGTM, thanks!

lavalamp:
@eparis I'll merge this next time I push. You might want to take a look.

mungegithub/mungers/submit-queue.go
eparis: I don't love the feature interface or how squirrely it is to use features correctly (I wrote it but someone smarter can probably rewrite it better) but I agree it seems like the right thing here...

mungegithub/mungegithub.go
```diff
@@ -105,7 +105,7 @@ func main() {
 			}
 			errInit := mungers.InitializeMungers(&config.Config, &config.Features)
 			if errInit != nil {
```
lavalamp: Please use the pattern:

```
if err := mungers.InitializeMungers(&config.Config, &config.Features); err != nil {
```

Here and above on line 99. This way you can use the same variable name while avoiding this type of bug (err is only declared in the body of the if).

foxish: Done. Thanks! :)

lavalamp:
Great, please squash your changes and we can merge this.

foxish:
Done.

PR #1351, opened 2016-07-12
[nginx-ingress-controller]: Avoid generation of invalid ssl certificates

aledbf:
This change avoids the generation of the pem file used by nginx in case of errors in the certificate


bprashanth:
conflict, otherwise LGTM

aledbf:
done

PR #1356, opened 2016-07-12
Atomic clone and updates for git-sync

gregory-lyons:
The git-sync sidecar container pulls Git content into a shared volume, but another container may begin accessing the incomplete repository while the initial pull is occurring. There has been some discussion of the matter here: https://github.com/kubernetes/kubernetes/issues/17676.

This change provides the option to make the initial clone/pull atomic (i.e. the target directory will not appear in the filesystem until the pull is complete). I've included atomic.md as a readme with more information about usage and implementation, but can merge that into the existing README.md instead if that is preferred. 

@thockin @huggsboson

### DIFF
--- a/git-sync/README.md
+++ b/git-sync/README.md
@@ -4,13 +4,17 @@ git-sync is a command that pull a git repository to a local directory.
 
 It can be used to source a container volume with the content of a git repo.
 
+In order to ensure that the git repository content is cloned and updated atomically, you cannot use a volume root directory as the directory for your local repository. 
+
+The local repository is created in a subdirectory of /git, with the subdirectory name specified by GIT_SYNC_DEST. 
+
 ## Usage
 
 ```
 # build the container
 docker build -t git-sync .
 # run the git-sync container
-docker run -d GIT_SYNC_REPO=https://github.com/GoogleCloudPlatform/kubernetes GIT_SYNC_DEST=/git -e GIT_SYNC_BRANCH=gh-pages -r HEAD GIT_SYNC_DEST=/git -v /git-data:/git git-sync
+docker run -d GIT_SYNC_REPO=https://github.com/GoogleCloudPlatform/kubernetes GIT_SYNC_DEST=/git -e GIT_SYNC_BRANCH=gh-pages -r HEAD -v /git-data:/git git-sync
 # run a nginx container to serve sync'ed content
 docker run -d -p 8080:80 -v /git-data:/usr/share/nginx/html nginx 
 ```
--- a/git-sync/main.go
+++ b/git-sync/main.go
@@ -30,12 +30,16 @@ import (
 	"strconv"
 	"strings"
 	"time"
+	"io/ioutil"
+	"path/filepath"
 )
+
+const volMount = "/git"
 
 var flRepo = flag.String("repo", envString("GIT_SYNC_REPO", ""), "git repo url")
 var flBranch = flag.String("branch", envString("GIT_SYNC_BRANCH", "master"), "git branch")
 var flRev = flag.String("rev", envString("GIT_SYNC_REV", "HEAD"), "git rev")
-var flDest = flag.String("dest", envString("GIT_SYNC_DEST", ""), "destination path")
+var flDest = flag.String("dest", envString("GIT_SYNC_DEST", ""), "destination subdirectory path within volume")
 var flWait = flag.Int("wait", envInt("GIT_SYNC_WAIT", 0), "number of seconds to wait before next sync")
 var flOneTime = flag.Bool("one-time", envBool("GIT_SYNC_ONE_TIME", false), "exit after the initial checkout")
 var flDepth = flag.Int("depth", envInt("GIT_SYNC_DEPTH", 0), "shallow clone with a history truncated to the specified number of commits")
@@ -134,32 +138,9 @@ func main() {
 	}
 }
 
-// syncRepo syncs the branch of a given repository to the destination at the given rev.
-func syncRepo(repo, dest, branch, rev string, depth int) error {
-	gitRepoPath := path.Join(dest, ".git")
-	_, err := os.Stat(gitRepoPath)
-	switch {
-	case os.IsNotExist(err):
-		// clone repo
-		args := []string{"clone", "--no-checkout", "-b", branch}
-		if depth != 0 {
-			args = append(args, "-depth")
-			args = append(args, string(depth))
-		}
-		args = append(args, repo)
-		args = append(args, dest)
-		output, err := runCommand("git", "", args)
-		if err != nil {
-			return err
-		}
-
-		log.Printf("clone %q: %s", repo, string(output))
-	case err != nil:
-		return fmt.Errorf("error checking if repo exist %q: %v", gitRepoPath, err)
-	}
-
+func gitPullIntoDir(dir, branch, rev string) error {
 	// fetch branch
-	output, err := runCommand("git", dest, []string{"pull", "origin", branch})
+	output, err := runCommand("git", dir, []string{"pull", "origin", branch})
 	if err != nil {
 		return err
 	}
@@ -167,7 +148,7 @@ func syncRepo(repo, dest, branch, rev string, depth int) error {
 	log.Printf("fetch %q: %s", branch, string(output))
 
 	// reset working copy
-	output, err = runCommand("git", dest, []string{"reset", "--hard", rev})
+	output, err = runCommand("git", dir, []string{"reset", "--hard", rev})
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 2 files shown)


gregory-lyons:
I'm with Box and am covered under Box's CLA

git-sync/atomic.md
```diff
@@ -0,0 +1,70 @@
...
+   containers: [
+   {
+       name: "git-sync",
+       ...
+       env: [{
+               name: "GIT_SYNC_REPO",
+               value: "https://github.com/kubernetes/kubernetes",
+           },
+           {
+              name: "GIT_SYNC_DEST",
```
huggsboson: I still don't fully understand the use case for this.  Feels like it should always be /git and then there are various if else's for if subdir is set, not set

git-sync/main.go
```diff
@@ -98,10 +101,14 @@ func main() {
...
 
+	if *flAtomic == true && *flSubDir == "" {
+		log.Fatal("error: GIT_SYNC_SUBDIR is required for atomic initial pull")
+	}
+
 	initialSync := true
 	failCount := 0
 	for {
-		if err := syncRepo(*flRepo, *flDest, *flBranch, *flRev, *flDepth); err != nil {
+		if err := syncRepo(*flRepo, *flDest, *flBranch, *flRev, *flSubDir, *flDepth, *flAtomic); err != nil {
```
huggsboson: this is looking pretty close to wanting to be pulled out into a struct with methods.

git-sync/main.go
```diff
@@ -127,19 +134,28 @@ func main() {
...
 // syncRepo syncs the branch of a given repository to the destination at the given rev.
-func syncRepo(repo, dest, branch, rev string, depth int) error {
-	gitRepoPath := path.Join(dest, ".git")
+func syncRepo(repo, dest, branch, rev, subdir string, depth int, atomic bool) error {
+	pullTarget := path.Join(dest, subdir)
+	gitRepoPath := path.Join(pullTarget, ".git")
 	_, err := os.Stat(gitRepoPath)
 	switch {
 	case os.IsNotExist(err):
+		if atomic {
```
huggsboson: if the repo does exist we won't create the atomic pull path?

thockin:
Thanks for this!  I promise I or someone will look at it ASAP.  I may not
get to it this week, though.

@mikedanese - got any free time?

git-sync/main.go
gregory-lyons: If the repo does exist, then we've already done our initial pull, so we don't need to pull into the tempdir. We just pull into the existing repo. This PR only covers making the first pull atomic.

git-sync/atomic.md
gregory-lyons: Yeah that makes sense. My thinking was that the user might want the volume root directory at somewhere besides /git, but I forgot that this is just a single-purpose sidecar and that another container can mount the shared volume on its filesystem wherever it wants anyway.

Proposed behavior: Shared emptyDir volume is mounted at /git by default, and we use that volume root directory as the repo unless optional GIT_SYNC_SUBDIR (or TARGET, DEST, etc.) is specified. Atomic pull requires the SUBDIR. How does that sound?

mikedanese:
Heh. Yes, i can take a look.

mikedanese:
Why do we need an option to enable this? Why don't we always want this?

gregory-lyons:
The idea of making it optional came from some discussion in https://github.com/kubernetes/kubernetes/issues/17676 regarding the GIT_SYNC_DEST environment variable and the necessity of putting the repo in a subdirectory of the volume. In the current implementation, I think the expected behavior would be for the repo to end up in the volume root, so requiring another subdirectory env var seemed like it would be a burden worth making optional.

However, I think with an implementation like the one I mentioned above (https://github.com/kubernetes/contrib/pull/1356#discussion_r70564309) we could make this atomic pull the default behavior. The user shouldn't care where the volume is mounted within this sidecar container, they should only care about the relative path of the Git repo within the volume (so they can access it from within the volume mount in whatever other container is using it).

We would need documentation that the required volume mountPath for git-sync is /git, but then we could combine the GIT_SYNC_DEST and GIT_SYNC_SUBDIR env vars into a single env var (probably still named GIT_SYNC_DEST?), which will be the relative path within the volume where the repo actually ends up.

thockin:
Unfortunately, the way that bind mounts work means that when you bind mount
a symlink, it will resolve the link to the final target.  The normal way to
do an atomic flip of multiple files is with a symlink, so while it might be
an atomic flip, anyone who bind mounted it will not see the update.

gregory-lyons:
1) @thockin just to clarify, you are saying that we won't be able to do atomic pull into the root of the mounted volume? I think we're in agreement that the subdir is necessary.

2) Given that the subdir is necessary, do we want to make this atomic pull the default behavior? We would require a GIT_SYNC_SUBDIR env var, but could get rid of the existing GIT_SYNC_DEST env var and put the data in /git/$GIT_SYNC_SUBDIR (where /git is the mount).

3) I actually implemented this without symlinks, using Go's os.Rename (which is implemented with a rename syscall and behaved atomically in some testing I did), but I see that for atomic update pulls (i.e. after the initial pull), we will want to use symlinks. I can go ahead and fix the implementation for this PR to make things easier for the followup PR.

gregory-lyons:
I've gone ahead and made some implementation changes in a followup commit, focused on making this the default behavior and using symlinks. When I started using symlinks, I found that it wouldn't be too difficult to go ahead and implement atomic updates as well, so I hope it's alright that I did that in this PR as well. I also split up some parts of the big syncRepo function into smaller functions.

thockin:
I have not had a chance to really look at the PR yet, so this is sort
of hand-waving.  I took a look at the doc, but I don't think it is
quite clear enough.  I will say, though, that we really should not
make an incompatible change unless we're willing to rev the major
version of the resulting container.

On Fri, Jul 15, 2016 at 10:37 AM, Gregory Lyons
<notifications@github.com> wrote:
>
> 1) @thockin just to clarify, you are saying that we won't be able to do atomic pull into the root of the mounted volume? I think we're in agreement that the subdir is necessary.

Well, we USED to be able to do this, so we need to either support that
or rev the major version.

> 2) Given that the subdir is necessary, do we want to make this atomic pull the default behavior? We would require a GIT_SYNC_SUBDIR env var, but could get rid of the existing GIT_SYNC_DEST env var and put the data in /git/$GIT_SYNC_SUBDIR (where /git is the mount).

Again, compatible or major rev

> 3) I actually implemented this without symlinks, using Go's os.Rename (which is implemented with a rename syscall and behaved atomically in some testing I did), but I see that for atomic update pulls (i.e. after the initial pull), we will want to use symlinks. I can go ahead and fix the implementation for this PR to make things easier for the followup PR.

Yeah, I think you NEED to use symlinks.  What makes this max ugly is
that you really don't want to re-pull if there are no changes, so we
should check if there are changes, then pull to a tmpdir, then
symlink.  It has to be JUST RIGHT or it will break for users.

thockin:
Just to be more specific, I think we need to do something like peek at
`origin` and see if it is the same hash as HEAD and, if not clone to a
new directory, make a symlink, rename it, clean up.  Or maybe `git
fetch origin`, compare that to HEAD, and then clone from the current
dir to a new dir or something (git is magical, I am sure there's a
way...)

gregory-lyons:
I think it's worth updating the major version. In the current version, GIT_SYNC_DEST isn't very useful anyway, it only ever points at the root of the volume mount or somewhere inside of it. My thought is that we simplify things by choosing the mount location for the user (/git or something similar) and using GIT_SYNC_DEST to specify a subdir path within it (users shouldn't care where the volume root is mounted within git-sync, but they do care about where their repo is located within the volume).

With regards to implementation, I've already built it along the lines you're suggesting (updated in my more recent commit). Here is pseudocode for the looping syncRepo function (actual implementation uses helper functions):

```
if target repo does not exist
        clone/pull into tempdir
        make temp symlink to tempdir
        rename temp symlink to target repo name
else if target repo needs update
        copy .git from target repo to tempdir
        pull into tempdir
        make temp symlink to tempdir
        rename temp symlink to target repo name
        remove symlink's previous target
```

Pseudocode for checking if the repo needs an update:
```
git remote update
localHead = git rev-parse HEAD
remoteHead = git rev-parse @{u}
return localHead != remoteHead
```

gregory-lyons:
@mikedanese @thockin anyone had a chance to take a look at the implementation yet?

gregory-lyons:
Added a commit to fix a bug regarding absolute/relative paths. Anyone had a chance to take a look or think any more about default vs. optional behavior? @mikedanese @thockin

gregory-lyons:
@bgrant0607 Any chance you could help me out with getting someone to take a look at this?

mikedanese:
Your pseudo code seems to do approximately what git workspace does. Can you add some {unit,integration} tests for this?

mikedanese:
CI is not passing either.

mikedanese:
To clarify on work tree
```
git fetch upstream master
git worktree add ../rev-$(date '+%y%m%d%H%M') FETCH_HEAD
(switch symplinks)
(rm and prune old worktrees
```
This has the advantage of not required a copy of the git database which can become very large.

thockin:
Awesome trick!

bgrant0607:
@foxish may also be interested.

gregory-lyons:
@mikedanese I've been working on an implementation with worktree. Something to note is that before we can do "git worktree add", we need to have a .git directory that persists (worktrees are stored in .git/worktrees).

This brings back the conversation of what to do about "GIT_SYNC_DEST" given that we need the symlink to be inside of the volume and point to subdirectory within the volume (can't do swapping if the symlink points to the volume itself). 

Here's my proposal: 
1) GIT_SYNC_DEST should specify the name of the subdir (within the volume) that will "contain" the repo - this will be the name of the symlink that another program would wait for
2) Do the initial clone (with no checkout) in the volume root, so if /git is the mounted volume root we end up with /git/.git
3) "Pulls" are done with @mikedanese's pseudocode, by adding a worktree to a new subdir in the volume root, swapping the DEST symlink to point to the new worktree, and then pruning the old worktree

Volume directory will look like:
```
/git/
   .git/
   rev-$(date '+%y%m%d%H%M')/
   targetLink/
```
where targetLink is a symlink that points to the current rev-.../ directory

gregory-lyons:
Took a stab at using git worktree in https://github.com/kubernetes/contrib/pull/1356/commits/88cfdeaddf3f0e5c27ae7025cb8b0fda7700cf26

Pseudocode is now:
```
if target repo does not exist
        clone into /git/.git (no checkout)
else if target repo does not need update
        return
endif
git fetch origin <branch>
git worktree add /git/rev-<random-int> origin/<branch>
(switch symlinks)
(rm and prune old worktrees)
```

I used random int for the rev-.../ naming instead of $(date '+%y%m%d%H%M'), because depending on how frequently and quickly we are polling the repo, we could end up with a collision if basing the naming off timing.

git-sync/main.go
```diff
@@ -134,40 +138,74 @@ func main() {
...
+	}
+
+	// newDir is /git/rev-..., we need to change it to relative path.
+	// Volume in other container may not be mounted at /git, so the symlink can't point to /git.
+	newDirRelative, err := filepath.Rel(volMount, newDir)
+	if err != nil {
+		return fmt.Errorf("error converting to relative path: %v", err)
+	}
+
+	if _, err := runCommand("ln", volMount, []string{"-snf", newDirRelative, "tmp-link"}); err != nil {
```
mikedanese: mv an ln can probably be done without exec a subprocess but we can leave for a followup

mikedanese:
Looks good, we should add tests to this sometime...

mikedanese:
Box, Inc. is a corporate CLA signer.

mikedanese:
It's very weird that FETCH_HEAD does not work. Glad this does though. Thanks!

git-sync/README.md
```diff
@@ -4,13 +4,17 @@ git-sync is a command that pull a git repository to a local directory.
...
+The local repository is created in a subdirectory of /git, with the subdirectory name specified by GIT_SYNC_DEST. 
+
 ## Usage
 
 ```
 # build the container
 docker build -t git-sync .
 # run the git-sync container
-docker run -d GIT_SYNC_REPO=https://github.com/GoogleCloudPlatform/kubernetes GIT_SYNC_DEST=/git -e GIT_SYNC_BRANCH=gh-pages -r HEAD GIT_SYNC_DEST=/git -v /git-data:/git git-sync
+docker run -d GIT_SYNC_REPO=https://github.com/GoogleCloudPlatform/kubernetes GIT_SYNC_DEST=/git -e GIT_SYNC_BRANCH=gh-pages -r HEAD -v /git-data:/git git-sync
```
thockin: This does not read as a legit docker command.  Can you please make sure this is cut-and-paste-able?

thockin:
I think I owe you a shirt - didn't I put a bounty on this?  Send me your
address and size by email :)

thockin:
`docker run -d -e GIT_SYNC_REPO=https://github.com/thockin/dbg -e GIT_SYNC_DEST=/git -e GIT_SYNC_BRANCH=master -v /tmp/git:/git myrepo/git-sync:1.0`

Trying to use this, I get a tmp-link -> rev.<something> and an error "2016/08/19 04:35:29 error syncing repo: error replacing symlink: error running command "mv -T tmp-link /git" : exit status 1: mv: cannot overwrite directory '/git' with non-directory"

`docker run -d -e GIT_SYNC_REPO=https://github.com/thockin/dbg -e GIT_SYNC_DEST=/git/dbg -e GIT_SYNC_BRANCH=master -v /tmp/git:/git myrepo/git-sync:1.0`

I get "error syncing repo: error running command "git clone --no-checkout -b master https://github.com/thockin/dbg /git" : exit status 128: fatal: destination path '/git' already exists and is not an empty directory."

TL;DR: the docs don't actually give an example of this that can be used.

PR #1357, opened 2016-07-12
[mungegithub] better error checking and error returns

shinyben:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1357)

### DIFF
--- a/mungegithub/github/github.go
+++ b/mungegithub/github/github.go
@@ -19,7 +19,6 @@ package github
 import (
 	"bytes"
 	"encoding/json"
-	"errors"
 	goflag "flag"
 	"fmt"
 	"io/ioutil"
@@ -459,22 +458,23 @@ func (config *Config) getIssue(num int) (*github.Issue, error) {
 
 // Refresh will refresh the Issue (and PR if this is a PR)
 // (not the commits or events)
-func (obj *MungeObject) Refresh() error {
+func (obj *MungeObject) Refresh() bool {
 	num := *obj.Issue.Number
 	issue, err := obj.config.getIssue(num)
 	if err != nil {
-		return err
+		glog.Errorf("Error in Refresh")
+		return false
 	}
 	obj.Issue = issue
 	if !obj.IsPR() {
-		return nil
+		return true
 	}
 	pr, err := obj.config.getPR(*obj.Issue.Number)
 	if err != nil {
-		return err
+		return false
 	}
 	obj.pr = pr
-	return nil
+	return true
 }
 
 // ListMilestones will return all milestones of the given `state`
@@ -482,12 +482,12 @@ func (config *Config) ListMilestones(state string) ([]github.Milestone, bool) {
 	listopts := github.MilestoneListOptions{
 		State: state,
 	}
-	milestones, resp, ok := config.client.Issues.ListMilestones(config.Org, config.Project, &listopts)
+	milestones, resp, err := config.client.Issues.ListMilestones(config.Org, config.Project, &listopts)
 	config.analytics.ListMilestones.Call(config, resp)
-	if !ok {
-		glog.Errorf("Error getting milestones of state %q: %v", state, err)
+	if err != nil {
+		glog.Errorf("Error getting milestones of state")
 	}
-	return milestones, ok
+	return milestones, true
 }
 
 // GetObject will return an object (with only the issue filled in)
@@ -1024,10 +1024,9 @@ func computeStatus(combinedStatus *github.CombinedStatus, requiredContexts []str
 
 func (obj *MungeObject) getCombinedStatus() (status *github.CombinedStatus, ok bool) {
 	config := obj.config
-	pr, err := obj.GetPR()
-	if err != nil {
-		glog.Errorf("Error in getCombinedStatus: %v", err)
-		return nil, false
+	pr, ok := obj.GetPR()
+	if !ok {
+		return nil, true
 	}
 	if pr.Head == nil {
 		glog.Errorf("pr.Head is nil in getCombinedStatus for PR# %d", *obj.Issue.Number)
@@ -1044,7 +1043,7 @@ func (obj *MungeObject) getCombinedStatus() (status *github.CombinedStatus, ok b
 }
 
 // SetStatus allowes you to set the Github Status
-func (obj *MungeObject) SetStatus(state, url, description, context string) error {
+func (obj *MungeObject) SetStatus(state, url, description, context string) bool {
 	config := obj.config
 	status := &github.RepoStatus{
 		State:       &state,
@@ -1052,21 +1051,22 @@ func (obj *MungeObject) SetStatus(state, url, description, context string) error
 		Description: &description,
 		Context:     &context,
 	}
-	pr, err := obj.GetPR()
-	if err != nil {
-		return err
+	pr, ok := obj.GetPR()
+	if !ok {
+		glog.Errorf("Error in SetStatus")
+		return false
 	}
 	ref := *pr.Head.SHA
 	glog.Infof("PR %d setting %q Github status to %q", *obj.Issue.Number, context, description)
 	config.analytics.SetStatus.Call(config, nil)
 	if config.DryRun {
-		return nil
+		return true
 	}
-	_, _, err = config.client.Repositories.CreateStatus(config.Org, config.Project, ref, status)
+	_, _, err := config.client.Repositories.CreateStatus(config.Org, config.Project, ref, status)
 	if err != nil {
 		glog.Errorf("Unable to set status. PR %d Ref: %q: %v", *obj.Issue.Number, ref, err)
 	}
-	return err
+	return false
 }
 
 // GetStatus returns the actual requested status, or nil if not found
@@ -1200,9 +1200,9 @@ func (obj *MungeObject) WaitForNotPending(requiredContexts []string) error {
 }
 
 // GetCommits returns all of the commits for a given PR
-func (obj *MungeObject) GetCommits() ([]github.RepositoryCommit, error) {
+func (obj *MungeObject) GetCommits() ([]github.RepositoryCommit, bool) {
 	if obj.commits != nil {
-		return obj.commits, nil
+		return obj.commits, true
 	}
 	config := obj.config
 	commits := []github.RepositoryCommit{}
@@ -1212,7 +1212,7 @@ func (obj *MungeObject) GetCommits() ([]github.RepositoryCommit, error) {
 		config.analytics.ListCommits.Call(config, response)
 		if err != nil {
 			glog.Errorf("Error commits for PR %d: %v", *obj.Issue.Number, err)
-			return nil, err
+			return nil, false
 		}
 		commits = append(commits, commitsPage...)
 		if response.LastPage == 0 || response.LastPage <= page {
@@ -1236,23 +1236,25 @@ func (obj *MungeObject) GetCommits() ([]github.RepositoryCommit, error) {
 		filledCommits = append(filledCommits, *commit)
 	}
 	obj.commits = filledCommits
-	return filledCommits, nil
+	return filledCommits, true
 }
 
 // GetPR will return the PR of the object.
-func (obj *MungeObject) GetPR() (*github.PullRequest, error) {
+func (obj *MungeObject) GetPR() (*github.PullRequest, bool) {
 	if obj.pr != nil {
-		return obj.pr, nil
+		return obj.pr, true
 	}
 	if !obj.IsPR() {
-		return nil, fmt.Errorf("Issue: %d is not a PR", *obj.Issue.Number)
+		fmt.Errorf("Issue: %d is not a PR", *obj.Issue.Number)
+		return nil, false
 	}
 	pr, err := obj.config.getPR(*obj.Issue.Number)
 	if err != nil {
-		return nil, err
+		fmt.Errorf("Error in GetPR")
+		return nil, false
 	}
 	obj.pr = pr
-	return pr, nil
+	return pr, true
 }
 
 // AssignPR will assign `prNum` to the `owner` where the `owner` is asignee's github login
@@ -1294,54 +1296,57 @@ func (obj *MungeObject) CloseIssuef(format string, args ...interface{}) error {
 }
 
 // ClosePR will close the Given PR
-func (obj *MungeObject) ClosePR() error {
+func (obj *MungeObject) ClosePR() bool {
 	config := obj.config
-	pr, err := obj.GetPR()
-	if err != nil {
-		return err
+	pr, ok := obj.GetPR()
+	if !ok {
+		glog.Errorf("Error in ClosePR")
+		return false
 	}
 	config.analytics.ClosePR.Call(config, nil)
 	glog.Infof("Closing PR# %d", *pr.Number)
 	if config.DryRun {
-		return nil
+		return true
 	}
 	state := "closed"
 	pr.State = &state
 	if _, _, err := config.client.PullRequests.Edit(config.Org, config.Project, *pr.Number, pr); err != nil {
 		glog.Errorf("Failed to close pr %d: %v", *pr.Number, err)
-		return err
+		return false
 	}
-	return nil
+	return true
 }
 
 // OpenPR will attempt to open the given PR.
 // It will attempt to reopen the pr `numTries` before returning an error
 // and giving up.
-func (obj *MungeObject) OpenPR(numTries int) error {
+//CHECK
+func (obj *MungeObject) OpenPR(numTries int) bool {
 	config := obj.config
-	pr, err := obj.GetPR()
-	if err != nil {
-		return err
+	pr, ok := obj.GetPR()
+	if !ok {
+		glog.Errorf("Error in OpenPR")
+		return false
 	}
 	config.analytics.OpenPR.Call(config, nil)
 	glog.Infof("Opening PR# %d", *pr.Number)
 	if config.DryRun {
-		return nil
+		return true
 	}
 	state := "open"
 	pr.State = &state
 	// Try pretty hard to re-open, since it's pretty bad if we accidentally leave a PR closed
 	for tries := 0; tries < numTries; tries++ {
-		if _, _, err = config.client.PullRequests.Edit(config.Org, config.Project, *pr.Number, pr); err == nil {
-			return nil
+		if _, _, err := config.client.PullRequests.Edit(config.Org, config.Project, *pr.Number, pr); err == nil {
+			return true
 		}
-		glog.Warningf("failed to re-open pr %d: %v", *pr.Number, err)
+		glog.Warningf("failed to re-open pr %d", *pr.Number)
 		time.Sleep(5 * time.Second)
 	}
-	if err != nil {
-		glog.Errorf("failed to re-open pr %d after %d tries, giving up: %v", *pr.Number, numTries, err)
+	if !ok {
+		glog.Errorf("failed to re-open pr %d after %d tries, giving up", *pr.Number, numTries)
 	}
-	return err
+	return false
 }
 
 // GetFileContents will return the contents of the `file` in the repo at `sha`
@@ -1391,13 +1396,13 @@ func (obj *MungeObject) MergeCommit() (*string, bool) {
 
 // MergePR will merge the given PR, duh
 // "who" is who is doing the merging, like "submit-queue"
-func (obj *MungeObject) MergePR(who string) error {
+func (obj *MungeObject) MergePR(who string) bool {
 	config := obj.config
 	prNum := *obj.Issue.Number
 	config.analytics.Merge.Call(config, nil)
 	glog.Infof("Merging PR# %d", prNum)
 	if config.DryRun {
-		return nil
+		return true
 	}
 	mergeBody := fmt.Sprintf("Automatic merge from %s", who)
 	obj.WriteComment(mergeBody)
@@ -1414,8 +1419,9 @@ func (obj *MungeObject) MergePR(who string) error {
 
 	// Get the text of the first commit
 	firstCommit := ""
-	if commits, err := obj.GetCommits(); err != nil {
-		return err
+	if commits, ok := obj.GetCommits(); !ok {
+		glog.Errorf("Error in MergePR")
+		return false
 	} else if commits[0].Commit.Message != nil {
 		firstCommit = *commits[0].Commit.Message
 	}
@@ -1445,9 +1451,9 @@ func (obj *MungeObject) MergePR(who string) error {
 	}
 	if err != nil {
 		glog.Errorf("Failed to merge PR: %d: %v", prNum, err)
-		return err
+		return false
 	}
-	return nil
+	return true
 }
 
 // GetPRFixesList returns a list of issue numbers that are referenced in the PR body.
@@ -1565,51 +1571,54 @@ func (obj *MungeObject) DeleteComment(comment *github.IssueComment) error {
 // PR again if github did not respond the first time. So the hopefully github
 // will have a response the second time. If we have no answer twice, we return
 // false
-func (obj *MungeObject) IsMergeable() (bool, error) {
+func (obj *MungeObject) IsMergeable() (bool, bool) {
 	if !obj.IsPR() {
-		return false, nil
+		return false, true
 	}
-	pr, err := obj.GetPR()
-	if err != nil {
-		return false, err
+	pr, ok := obj.GetPR()
+	if !ok {
+		glog.Errorf("Error in IsMergable")
+		return false, false
 	}
 	prNum := *pr.Number
 	if pr.Mergeable == nil {
 		glog.V(4).Infof("Waiting for mergeability on %q %d", *pr.Title, *pr.Number)
 		// TODO: determine what a good empirical setting for this is.
 		time.Sleep(2 * time.Second)
-		err := obj.Refresh()
-		if err != nil {
-			glog.Errorf("Unable to refresh PR# %d: %v", prNum, err)
-			return false, err
+		ok = obj.Refresh()
+		if !ok {
+			glog.Errorf("Unable to refresh PR# %d", prNum)
+			return false, false
 		}
-		pr, err = obj.GetPR()
-		if err != nil {
-			glog.Errorf("Unable to get PR# %d: %v", prNum, err)
-			return false, err
+		pr, ok = obj.GetPR()
+		if !ok {
+			glog.Errorf("Unable to get PR# %d", prNum)
+			return false, false
 		}
 	}
 	if pr.Mergeable == nil {
-		err := fmt.Errorf("no mergeability information for %q %d, Skipping", *pr.Title, *pr.Number)
-		glog.Errorf("%v", err)
-		return false, err
+		glog.Errorf("No mergeability information for %q %d, Skipping", *pr.Title, *pr.Number)
+		return false, false
 	}
-	return *pr.Mergeable, nil
+	return *pr.Mergeable, true
 }
 
 // IsMerged returns if the issue in question was already merged
-func (obj *MungeObject) IsMerged() (bool, error) {
+func (obj *MungeObject) IsMerged() (bool, bool) {
 	if !obj.IsPR() {
-		return false, fmt.Errorf("Issue: %d is not a PR and is thus 'merged' is indeterminate", *obj.Issue.Number)
+		glog.Errorf("Issue: %d is not a PR and is thus 'merged' is indeterminate", *obj.Issue.Number)
+		return false, false
 	}
-	pr, err := obj.GetPR()
-	if err != nil {
-		return false, err
+	pr, ok := obj.GetPR()
+	if !ok {
+		glog.Errorf("Error in IsMerged")
+		return false, false
 	}
 	if pr.Merged != nil {
-		return *pr.Merged, nil
+		return *pr.Merged, true
 	}
-	return false, fmt.Errorf("Unable to determine if PR %d was merged", *obj.Issue.Number)
+	glog.Errorf("Unable to determine if PR %d was merged", *obj.Issue.Number)
+	return false, false
 }
 
 // MergedAt returns the time an issue was merged (for nil if unmerged)
@@ -1617,9 +1626,9 @@ func (obj *MungeObject) MergedAt() (*time.Time, bool) {
 	if !obj.IsPR() {
 		return nil, true
 	}
-	pr, err := obj.GetPR()
-	if err != nil {
-		glog.Errorf("Error in MergedAt: %v", err)
+	pr, ok := obj.GetPR()
+	if !ok {
+		glog.Errorf("Error in MergedAt")
 		return nil, false
 	}
 	return pr.MergedAt, true
--- a/mungegithub/github/github.go
+++ b/mungegithub/github/github.go
@@ -19,6 +19,7 @@ package github
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	goflag "flag"
 	"fmt"
 	"io/ioutil"
@@ -477,7 +478,7 @@ func (obj *MungeObject) Refresh() error {
 }
 
 // ListMilestones will return all milestones of the given `state`
-func (config *Config) ListMilestones(state string) []github.Milestone {
+func (config *Config) ListMilestones(state string) ([]github.Milestone, error) {
 	listopts := github.MilestoneListOptions{
 		State: state,
 	}
@@ -486,7 +487,7 @@ func (config *Config) ListMilestones(state string) []github.Milestone {
 	if err != nil {
 		glog.Errorf("Error getting milestones of state %q: %v", state, err)
 	}
-	return milestones
+	return milestones, err
 }
 
 // GetObject will return an object (with only the issue filled in)
@@ -535,35 +536,39 @@ func (config *Config) NewIssue(title, body string, labels []string, owner string
 
 // Branch returns the branch the PR is for. Return "" if this is not a PR or
 // it does not have the required information.
-func (obj *MungeObject) Branch() string {
+func (obj *MungeObject) Branch() (string, error) {
 	pr, err := obj.GetPR()
 	if err != nil {
-		return ""
+		return "", err
 	}
 	if pr.Base != nil && pr.Base.Ref != nil {
-		return *pr.Base.Ref
+		return *pr.Base.Ref, nil
 	}
-	return ""
+	return "", errors.New("Unable to determine git branch")
 }
 
 // IsForBranch return true if the object is a PR for a branch with the given
 // name. It return false if it is not a pr, it isn't against the given branch,
 // or we can't tell
-func (obj *MungeObject) IsForBranch(branch string) bool {
-	objBranch := obj.Branch()
-	if objBranch == branch {
-		return true
+func (obj *MungeObject) IsForBranch(branch string) (bool, error) {
+	objBranch, err := obj.Branch()
+	if err != nil {
+		return false, err
 	}
-	return false
+	if objBranch == branch {
+		return true, nil
+	}
+	return false, nil
 }
 
 // LastModifiedTime returns the time the last commit was made
 // BUG: this should probably return the last time a git push happened or something like that.
-func (obj *MungeObject) LastModifiedTime() *time.Time {
+func (obj *MungeObject) LastModifiedTime() (*time.Time, bool) {
 	var lastModified *time.Time
 	commits, err := obj.GetCommits()
 	if err != nil {
-		return lastModified
+		glog.Errorf("Error in LastModifiedTime, unable to get commits: %v", err)
+		return lastModified, false
 	}
 	for _, commit := range commits {
 		if commit.Commit == nil || commit.Commit.Committer == nil || commit.Commit.Committer.Date == nil {
@@ -574,7 +579,7 @@ func (obj *MungeObject) LastModifiedTime() *time.Time {
 			lastModified = commit.Commit.Committer.Date
 		}
 	}
-	return lastModified
+	return lastModified, true
 }
 
 // labelEvent returns the most recent event where the given label was added to an issue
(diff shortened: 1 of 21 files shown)


mungegithub/github/github.go
```diff
@@ -726,39 +747,42 @@ func (obj *MungeObject) RemoveLabel(label string) error {
...
+		return "", false, err
 	}
-	return *commit.SHA, true
+	return *commit.SHA, true, err
 }
 
 // SetMilestone will set the milestone to the value specified
 func (obj *MungeObject) SetMilestone(title string) error {
-	milestones := obj.config.ListMilestones("all")
+	milestones, _ := obj.config.ListMilestones("all")
```
eparis: Let's just return the error here

mungegithub/github/github.go
```diff
@@ -535,35 +545,40 @@ func (config *Config) NewIssue(title, body string, labels []string, owner string
...
 // LastModifiedTime returns the time the last commit was made
 // BUG: this should probably return the last time a git push happened or something like that.
-func (obj *MungeObject) LastModifiedTime() *time.Time {
+//CHECK-UPDATED
+func (obj *MungeObject) LastModifiedTime() (*time.Time, error) {
 	var lastModified *time.Time
 	commits, err := obj.GetCommits()
 	if err != nil {
-		return lastModified
+		glog.Errorf("Error: %v", err)
```
eparis: Can you state this is in LastModifiedTime in the error message?

mungegithub/github/github_test.go
```diff
@@ -414,7 +414,7 @@ func TestGetLastModified(t *testing.T) {
 			config: config,
 			Issue:  github_test.Issue("bob", 1, nil, true),
 		}
-		ts := obj.LastModifiedTime()
+		ts, _ := obj.LastModifiedTime()
```
eparis: I would say that err != nil is also a t.Errorf()

mungegithub/mungers/lgtm_after_commit.go
```diff
@@ -66,7 +66,7 @@ func (LGTMAfterCommitMunger) Munge(obj *github.MungeObject) {
 		return
 	}
 
-	lastModified := obj.LastModifiedTime()
+	lastModified, _ := obj.LastModifiedTime()
```
eparis: Can you split the error checking https://github.com/kubernetes/contrib/pull/1357/files#diff-d54eba901e3ff2426a7c69c8f323f129R72 to include `if err != nil || lastModified == nil`

mungegithub/mungers/submit-queue.go
```diff
@@ -876,7 +876,7 @@ func (sq *SubmitQueue) validForMerge(obj *github.MungeObject) bool {
 	}
 
 	// PR cannot change since LGTM was added
-	lastModifiedTime := obj.LastModifiedTime()
+	lastModifiedTime, _ := obj.LastModifiedTime()
```
eparis: Similar error checking changes as suggested in lgtm_after_commit.go

mungegithub/mungers/submit-queue_test.go
```diff
@@ -359,7 +359,7 @@ func TestValidateLGTMAfterPush(t *testing.T) {
 			t.Errorf("Unexpected error getting events commits: %v", err)
 		}
 
-		lastModifiedTime := obj.LastModifiedTime()
+		lastModifiedTime, _ := obj.LastModifiedTime()
 		lgtmTime := obj.LabelTime(lgtmLabel)
 
 		if lastModifiedTime == nil || lgtmTime == nil {
```
eparis: And similar error handling as the previous...

mungegithub/github/github.go
```diff
@@ -574,7 +589,7 @@ func (obj *MungeObject) LastModifiedTime() *time.Time {
 			lastModified = commit.Commit.Committer.Date
 		}
 	}
-	return lastModified
+	return lastModified, err
```
eparis: return nil here instead of err (even thought err is nil)   just normal style

mungegithub/github/github.go
```diff
@@ -726,40 +747,45 @@ func (obj *MungeObject) RemoveLabel(label string) error {
 // GetHeadAndBase returns the head SHA and the base ref, so that you can get
 // the base's sha in a second step. Purpose: if head and base SHA are the same
 // across two merge attempts, we don't need to rerun tests.
-func (obj *MungeObject) GetHeadAndBase() (headSHA, baseRef string, ok bool) {
+//CHECK-UPDATED
+func (obj *MungeObject) GetHeadAndBase() (headSHA, baseRef string, ok bool, err error) {
```
eparis: Revert this change. I believe that `ok` is supposed to indicate if there was an error. Maybe we should use that pattern more often than actually returning the error. What are your thoughts?

mungegithub/github/github.go
```diff
@@ -726,40 +747,45 @@ func (obj *MungeObject) RemoveLabel(label string) error {
...
 	}
 	baseRef = *pr.Base.Ref
-	return headSHA, baseRef, true
+	return headSHA, baseRef, true, err
 }
 
 // GetSHAFromRef returns the current SHA of the given ref (i.e., branch).
-func (obj *MungeObject) GetSHAFromRef(ref string) (sha string, ok bool) {
+//CHECK-UPDATED
+func (obj *MungeObject) GetSHAFromRef(ref string) (sha string, ok bool, err error) {
```
eparis: Same here, I think `ok` satisfies the need of higher level callers...

mungegithub/github/github.go
```diff
@@ -1347,17 +1381,19 @@ func (obj *MungeObject) GetFileContents(file, sha string) (string, error) {
...
 // MergeCommit will return the sha of the merge. PRs which have not merged
 // (or if we hit an error) will return nil
-func (obj *MungeObject) MergeCommit() *string {
+//CHECK-UPDATED
+func (obj *MungeObject) MergeCommit() (*string, error) {
 	events, err := obj.GetEvents()
 	if err != nil {
-		return nil
+		return nil, err
+		glog.Errorf("Error: %v", err)
```
eparis: You can't return and then call `glog`.  So `glog` first. Include more info in the error message.

mungegithub/mungers/cherrypick-clear-after-merge.go
```diff
@@ -104,7 +104,7 @@ func (c *ClearPickAfterMerge) foundLog(branch string, logString string) (bool, s
 
 // Can we find a commit in the changelog that looks like it was done using git cherry-pick -m1 -x ?
 func (c *ClearPickAfterMerge) foundByPickDashX(obj *github.MungeObject, branch string) bool {
-	sha := obj.MergeCommit()
+	sha, _ := obj.MergeCommit()
 	if sha == nil {
```
eparis: On `err` return with no log.  On `sha == nil` print error then return...

mungegithub/mungers/cherrypick-queue.go
```diff
@@ -243,7 +243,7 @@ func (c *CherrypickQueue) serveRaw(res http.ResponseWriter, req *http.Request) {
 	sortedQueue := []rawReadyInfo{}
 	for _, key := range keyOrder {
 		obj := queue[key]
-		sha := obj.MergeCommit()
+		sha, _ := obj.MergeCommit()
 		if sha == nil {
```
eparis: I think you can/should check `if sha == nil || err != nil`

mungegithub/github/github.go
```diff
@@ -1584,15 +1621,17 @@ func (obj *MungeObject) IsMerged() (bool, error) {
...
+//CHECK-UPDATED
+func (obj *MungeObject) MergedAt() (*time.Time, error) {
 	if !obj.IsPR() {
-		return nil
+		return nil, errors.New("The object was not a PR")
 	}
 	pr, err := obj.GetPR()
 	if err != nil {
-		return nil
+		glog.Errorf("Error: %v", err)
```
eparis: More context

mungegithub/github/github.go
```diff
@@ -1584,15 +1621,17 @@ func (obj *MungeObject) IsMerged() (bool, error) {
...
+		return nil, err
 	}
-	return pr.MergedAt
+	return pr.MergedAt, err
```
eparis: `return pr.MergedAt, nil`

mungegithub/mungers/cherrypick-queue.go
```diff
@@ -148,7 +148,7 @@ func (c *CherrypickQueue) Munge(obj *github.MungeObject) {
 }
 
 func mergeTime(obj *github.MungeObject) time.Time {
-	t := obj.MergedAt()
+	t, _ := obj.MergedAt()
 	if t == nil {
```
eparis: `if t == nil || err != nil`

mungegithub/github/github_test.go
```diff
@@ -414,8 +414,8 @@ func TestGetLastModified(t *testing.T) {
 			config: config,
 			Issue:  github_test.Issue("bob", 1, nil, true),
 		}
-		ts := obj.LastModifiedTime()
-		if !ts.Equal(*test.expectedTime) {
+		ts, _ := obj.LastModifiedTime()
+		if !ts.Equal(*test.expectedTime) || err != nil {
```
eparis: Does this compile? Shouldn't the line above be `ts, err :=...`

mungegithub/mungers/cherrypick-auto-approve.go
```diff
@@ -118,7 +119,7 @@ func (c *CherrypickAutoApprove) Munge(obj *github.MungeObject) {
 
 	major := 0
 	minor := 0
-	branch := obj.Branch()
+	branch, _ := obj.Branch()
```
eparis: should check and return on err

mungegithub/mungers/cherrypick-auto-approve.go
```diff
@@ -140,7 +141,8 @@ func (c *CherrypickAutoApprove) Munge(obj *github.MungeObject) {
 		// not auto approve
 		parentMilestone := parent.ReleaseMilestone()
 		if parentMilestone != branchImpliedMilestone {
-			glog.Errorf("%d: parentReleaseMilestone=%q but branch is %q", *obj.Issue.Number, parentMilestone, obj.Branch())
+			boolean, _ := obj.Branch()
```
eparis: why call this `boolean`    isn't it a string (a branch name?)

mungegithub/mungers/cherrypick-label-unapproved.go
```diff
@@ -68,8 +68,9 @@ func (LabelUnapprovedPicks) Munge(obj *github.MungeObject) {
 	if !obj.IsPR() {
 		return
 	}
-
-	if obj.IsForBranch("master") {
+	//true=return
+	boolean, err := obj.IsForBranch("master")
+	if boolean || err != nil {
```
eparis: check err first

mungegithub/mungers/release-note-label.go
```diff
@@ -78,7 +78,9 @@ func (r *ReleaseNoteLabel) EachLoop() error { return nil }
 func (r *ReleaseNoteLabel) AddFlags(cmd *cobra.Command, config *github.Config) {}
 
 func (r *ReleaseNoteLabel) prMustFollowRelNoteProcess(obj *github.MungeObject) bool {
-	if obj.IsForBranch("master") {
+	boolean, err := obj.IsForBranch("master")
+
+	if boolean || err != nil {
```
eparis: check err first

mungegithub/github/github.go
99BenGardner: That would be useful in cases in which people don't need to access any values from the error-I really don't know how often/in which situations that would be

eparis: @lavalamp what do you think of doing more `ok bool` returns from `github/github.go` fucntions instead of `error` returns? Should we move towards that style? We seem to very very rarely ever care what the error is in the caller...

lavalamp: If the caller has no reasonable actions that could be conditionalized on the error, then I think returning the error is mostly just verbosity.

In the case of this function, if it fails an optimization is missed. The caller isn't going to bother to retry, so there's not much point in expressing the difference between various categories of failures. If anything, you could add that as a comment here.

mungegithub/github/github.go
lavalamp: I agree in this case, also.

mungegithub/github/github.go
lavalamp: IMO if you're returning an error, it's bad form to also glog it. One or the other.

lavalamp:
I'm not sure this is a good error handling strategy for this case.

I'd rather just have the merge queue stop or wait for github to come back / be reachable again, and I don't know that the submit queue code would be improved by adding that check in the many places it would require.

Just because you *can* return errors everywhere, doesn't mean you *should*. :) I think the point here is to make the submit queue code very readable and robust, which means we want to implement consistent error handling in the github access layer, I think. I can see two classes of functions in this layer:
1. Functions that must succeed, and should block in a retry loop until github comes back.
2. Functions that are optional/optimizations and failing is acceptable.

eparis:
In "IsForBranch()" the lack of error handling makes it impossible for the caller. The caller was doing `if !IsForBranch(master)`. Github went down. Things for master can back as `not for master` and the bot went friggin nuts on some PRs. That is why I asked Ben to start looking at this area.

I think propagating `error` is unnecessary in most places. But giving `ok` makes a lot of sense for future proofing. Some place we use a `*` and `nil` to return `errors`. Some places we use `error` (obviously). Some places we use `ok`.

I'm leaning towards trying to move the entire `github/github.go` API to use `ok` on most public functions. Prints of the failure happens in the `github.go` layer. Callers now know if they can trust their results.

lavalamp:
I think that's fine with me.

mungegithub/github/github.go
```diff
@@ -477,7 +478,7 @@ func (obj *MungeObject) Refresh() error {
 }
 
 // ListMilestones will return all milestones of the given `state`
-func (config *Config) ListMilestones(state string) []github.Milestone {
+func (config *Config) ListMilestones(state string) ([]github.Milestone, error) {
```
eparis: bool, not error

99BenGardner:
@eparis

mungegithub/github/github.go
```diff
@@ -458,35 +458,36 @@ func (config *Config) getIssue(num int) (*github.Issue, error) {
 
 // Refresh will refresh the Issue (and PR if this is a PR)
 // (not the commits or events)
-func (obj *MungeObject) Refresh() error {
+func (obj *MungeObject) Refresh() bool {
 	num := *obj.Issue.Number
 	issue, err := obj.config.getIssue(num)
 	if err != nil {
-		return err
+		glog.Errorf("Error in Refresh")
```
eparis: `glog.Errorf("Error in Refresh: %v", err)`

mungegithub/github/github.go
```diff
@@ -458,35 +458,36 @@ func (config *Config) getIssue(num int) (*github.Issue, error) {
...
 	}
 	obj.Issue = issue
 	if !obj.IsPR() {
-		return nil
+		return true
 	}
 	pr, err := obj.config.getPR(*obj.Issue.Number)
 	if err != nil {
-		return err
+		return false
```
eparis: I'm assuming that `getPR` already called `glog.Errorf()`?

mungegithub/github/github.go
```diff
@@ -458,35 +458,36 @@ func (config *Config) getIssue(num int) (*github.Issue, error) {
...
-func (config *Config) ListMilestones(state string) []github.Milestone {
+func (config *Config) ListMilestones(state string) ([]github.Milestone, bool) {
 	listopts := github.MilestoneListOptions{
 		State: state,
 	}
 	milestones, resp, err := config.client.Issues.ListMilestones(config.Org, config.Project, &listopts)
 	config.analytics.ListMilestones.Call(config, resp)
 	if err != nil {
-		glog.Errorf("Error getting milestones of state %q: %v", state, err)
+		glog.Errorf("Error getting milestones of state")
```
eparis: Why did you remove the additional print information?  please put it back.

mungegithub/github/github.go
```diff
@@ -535,35 +536,41 @@ func (config *Config) NewIssue(title, body string, labels []string, owner string
...
 // Branch returns the branch the PR is for. Return "" if this is not a PR or
 // it does not have the required information.
-func (obj *MungeObject) Branch() string {
-	pr, err := obj.GetPR()
-	if err != nil {
-		return ""
+func (obj *MungeObject) Branch() (string, bool) {
+	pr, ok := obj.GetPR()
+	if !ok {
+		glog.Errorf("Error in Branch, GetPR failed on obj")
```
eparis: Did `GetPR` already do glog.Errorf?  If so, no need here....

mungegithub/github/github.go
```diff
@@ -535,35 +536,41 @@ func (config *Config) NewIssue(title, body string, labels []string, owner string
...
 
 // IsForBranch return true if the object is a PR for a branch with the given
 // name. It return false if it is not a pr, it isn't against the given branch,
 // or we can't tell
-func (obj *MungeObject) IsForBranch(branch string) bool {
-	objBranch := obj.Branch()
+func (obj *MungeObject) IsForBranch(branch string) (bool, bool) {
+	objBranch, ok := obj.Branch()
+	if !ok {
+		glog.Errorf("Errof in IsForBranch, obj.Branch failed")
```
eparis: Did `Branch()` call glog? If so, no need here...

mungegithub/github/github.go
```diff
@@ -574,7 +581,7 @@ func (obj *MungeObject) LastModifiedTime() *time.Time {
 			lastModified = commit.Commit.Committer.Date
 		}
 	}
-	return lastModified
+	return lastModified, ok
```
eparis: How about `true` instead of `ok`

mungegithub/github/github.go
```diff
@@ -727,8 +736,9 @@ func (obj *MungeObject) RemoveLabel(label string) error {
 // the base's sha in a second step. Purpose: if head and base SHA are the same
 // across two merge attempts, we don't need to rerun tests.
 func (obj *MungeObject) GetHeadAndBase() (headSHA, baseRef string, ok bool) {
-	pr, err := obj.GetPR()
-	if err != nil {
+	pr, ok := obj.GetPR()
+	if !ok {
+		glog.Errorf("Error in GetHeadAndBase, obj.GetPR failed")
```
eparis: Do we need this glog? Did `GetPR` do it?

mungegithub/github/github.go
```diff
@@ -773,60 +786,65 @@ func (obj *MungeObject) SetMilestone(title string) error {
...
 // milestones of the form 'vX.Y' where X and Y are integeters. Return the maximum
 // possible time if there is no milestone or the milestone doesn't look like a
 // release milestone
-func (obj *MungeObject) ReleaseMilestoneDue() time.Time {
+func (obj *MungeObject) ReleaseMilestoneDue() (time.Time, bool) {
 	milestone := obj.Issue.Milestone
 	if milestone == nil {
-		return maxTime
+		glog.Errorf("Error in ReleaseMilestoneDue, nil milestone")
+		return maxTime, false
```
eparis: I don't think this is an error...

mungegithub/github/github.go
```diff
@@ -773,60 +786,65 @@ func (obj *MungeObject) SetMilestone(title string) error {
...
 	title := milestone.Title
 	if title == nil {
-		return maxTime
+		glog.Errorf("Error in ReleaseMilestoneDue, nil milestone.Title")
+		return maxTime, false
 	}
 	if !releaseMilestoneRE.MatchString(*title) {
-		return maxTime
+		glog.Errorf("Error in ReleaseMilestoneDue, does not match *title")
+		return maxTime, false
```
eparis: This is not an error

mungegithub/github/github.go
```diff
@@ -773,60 +786,65 @@ func (obj *MungeObject) SetMilestone(title string) error {
...
 	}
 	if milestone.DueOn == nil {
-		return maxTime
+		glog.Errorf("Error in ReleaseMilestoneDue, nil milestone.DueOn")
+		return maxTime, false
```
eparis: Not an error.

mungegithub/github/github.go
```diff
@@ -1004,64 +1022,67 @@ func computeStatus(combinedStatus *github.CombinedStatus, requiredContexts []str
...
 
-func (obj *MungeObject) getCombinedStatus() (status *github.CombinedStatus) {
+func (obj *MungeObject) getCombinedStatus() (status *github.CombinedStatus, ok bool) {
 	config := obj.config
-	pr, err := obj.GetPR()
-	if err != nil {
-		return nil
+	pr, ok := obj.GetPR()
+	if !ok {
+		return nil, true
```
eparis: false?

mungegithub/github/github.go
```diff
@@ -1004,64 +1022,67 @@ func computeStatus(combinedStatus *github.CombinedStatus, requiredContexts []str
...
 		TargetURL:   &url,
 		Description: &description,
 		Context:     &context,
 	}
-	pr, err := obj.GetPR()
-	if err != nil {
-		return err
+	pr, ok := obj.GetPR()
+	if !ok {
+		glog.Errorf("Error in SetStatus")
```
eparis: add   `%v, err` to this so we get more info.

mungegithub/github/github.go
```diff
@@ -1004,64 +1022,67 @@ func computeStatus(combinedStatus *github.CombinedStatus, requiredContexts []str
...
 	glog.Infof("PR %d setting %q Github status to %q", *obj.Issue.Number, context, description)
 	config.analytics.SetStatus.Call(config, nil)
 	if config.DryRun {
-		return nil
+		return true
 	}
-	_, _, err = config.client.Repositories.CreateStatus(config.Org, config.Project, ref, status)
+	_, _, err := config.client.Repositories.CreateStatus(config.Org, config.Project, ref, status)
 	if err != nil {
 		glog.Errorf("Unable to set status. PR %d Ref: %q: %v", *obj.Issue.Number, ref, err)
```
eparis: return false?

mungegithub/github/github.go
```diff
@@ -1004,64 +1022,67 @@ func computeStatus(combinedStatus *github.CombinedStatus, requiredContexts []str
...
 }
 
 // GetStatus returns the actual requested status, or nil if not found
-func (obj *MungeObject) GetStatus(context string) *github.RepoStatus {
-	combinedStatus := obj.getCombinedStatus()
-	if combinedStatus == nil {
-		return nil
+func (obj *MungeObject) GetStatus(context string) (*github.RepoStatus, bool) {
+	combinedStatus, ok := obj.getCombinedStatus()
+	if combinedStatus == nil || ok == false {
```
eparis: Break this up. `if !ok { return false } else if combinedStatus == nil {return true}`   if we already got a `glog` from getCominbedStatus you don't need to do another here.

mungegithub/github/github.go
```diff
@@ -1004,64 +1022,67 @@ func computeStatus(combinedStatus *github.CombinedStatus, requiredContexts []str
...
 	}
 	for _, status := range combinedStatus.Statuses {
 		if *status.Context == context {
-			return &status
+			return &status, true
 		}
 	}
-	return nil
+	glog.Errorf("Error in GetStatus, contexts never matched")
+	return nil, false
```
eparis: Lets go with `nil, true` and no glog.

mungegithub/github/github.go
```diff
@@ -1070,33 +1091,34 @@ func (obj *MungeObject) GetStatus(context string) *github.RepoStatus {
 //    * If any is 'error', the PR is in 'error'
 //    * If any is 'failure', the PR is 'failure'
 //    * Otherwise the PR is 'success'
-func (obj *MungeObject) GetStatusState(requiredContexts []string) string {
-	combinedStatus := obj.getCombinedStatus()
+func (obj *MungeObject) GetStatusState(requiredContexts []string) (string, bool) {
+	combinedStatus, ok := obj.getCombinedStatus()
 	if combinedStatus == nil {
```
eparis: you want to return here if `!ok`

mungegithub/github/github.go
```diff
@@ -1070,33 +1091,34 @@ func (obj *MungeObject) GetStatus(context string) *github.RepoStatus {
...
-	return computeStatus(combinedStatus, requiredContexts)
+	return computeStatus(combinedStatus, requiredContexts), ok
 }
 
 // IsStatusSuccess makes sure that the combined status for all commits in a PR is 'success'
-func (obj *MungeObject) IsStatusSuccess(requiredContexts []string) bool {
-	status := obj.GetStatusState(requiredContexts)
+func (obj *MungeObject) IsStatusSuccess(requiredContexts []string) (bool, bool) {
+	status, ok := obj.GetStatusState(requiredContexts)
 	if status == "success" {
```
eparis: Again, check `ok`...

mungegithub/github/github.go
```diff
@@ -1070,33 +1091,34 @@ func (obj *MungeObject) GetStatus(context string) *github.RepoStatus {
...
 
 // GetStatusTime returns when the status was set
-func (obj *MungeObject) GetStatusTime(context string) *time.Time {
-	status := obj.GetStatus(context)
-	if status == nil {
-		return nil
+func (obj *MungeObject) GetStatusTime(context string) (*time.Time, bool) {
+	status, ok := obj.GetStatus(context)
+	if status == nil || ok == false {
+		glog.Errorf("Error in GetStatusTime, nil status")
```
eparis: no need for glog if GetStatus printed one.

mungegithub/github/github.go
```diff
@@ -1208,23 +1236,25 @@ func (obj *MungeObject) GetCommits() ([]github.RepositoryCommit, error) {
...
 // GetPR will return the PR of the object.
-func (obj *MungeObject) GetPR() (*github.PullRequest, error) {
+func (obj *MungeObject) GetPR() (*github.PullRequest, bool) {
 	if obj.pr != nil {
-		return obj.pr, nil
+		return obj.pr, true
 	}
 	if !obj.IsPR() {
-		return nil, fmt.Errorf("Issue: %d is not a PR", *obj.Issue.Number)
+		fmt.Errorf("Issue: %d is not a PR", *obj.Issue.Number)
```
eparis: `glog` not `fmt`

mungegithub/github/github.go
```diff
@@ -1208,23 +1236,25 @@ func (obj *MungeObject) GetCommits() ([]github.RepositoryCommit, error) {
...
+		return nil, false
 	}
 	pr, err := obj.config.getPR(*obj.Issue.Number)
 	if err != nil {
-		return nil, err
+		fmt.Errorf("Error in GetPR")
```
eparis: `glog` not `fmt`

mungegithub/github/github.go
```diff
@@ -1266,54 +1296,57 @@ func (obj *MungeObject) CloseIssuef(format string, args ...interface{}) error {
...
 // ClosePR will close the Given PR
-func (obj *MungeObject) ClosePR() error {
+func (obj *MungeObject) ClosePR() bool {
 	config := obj.config
-	pr, err := obj.GetPR()
-	if err != nil {
-		return err
+	pr, ok := obj.GetPR()
+	if !ok {
+		glog.Errorf("Error in ClosePR")
```
eparis: There error was with `GetPR` but if it already glog'd no need to do it again.

mungegithub/github/github.go
```diff
@@ -1266,54 +1296,57 @@ func (obj *MungeObject) CloseIssuef(format string, args ...interface{}) error {
...
 	}
-	return nil
+	return true
 }
 
 // OpenPR will attempt to open the given PR.
 // It will attempt to reopen the pr `numTries` before returning an error
 // and giving up.
-func (obj *MungeObject) OpenPR(numTries int) error {
+//CHECK
```
eparis: no `//CHECK`

mungegithub/github/github.go
```diff
@@ -1266,54 +1296,57 @@ func (obj *MungeObject) CloseIssuef(format string, args ...interface{}) error {
...
 	pr.State = &state
 	// Try pretty hard to re-open, since it's pretty bad if we accidentally leave a PR closed
 	for tries := 0; tries < numTries; tries++ {
-		if _, _, err = config.client.PullRequests.Edit(config.Org, config.Project, *pr.Number, pr); err == nil {
-			return nil
+		if _, _, err := config.client.PullRequests.Edit(config.Org, config.Project, *pr.Number, pr); err == nil {
+			return true
 		}
-		glog.Warningf("failed to re-open pr %d: %v", *pr.Number, err)
+		glog.Warningf("failed to re-open pr %d", *pr.Number)
```
eparis: This was a real bug... 
Can you re-add the `err` but also break that if line (https://github.com/kubernetes/contrib/pull/1357/files#diff-6ff804e5140f204c3e7068a3ba138ae5R1340) into two lines so `err` is valid here?

mungegithub/github/github.go
```diff
@@ -1347,28 +1380,29 @@ func (obj *MungeObject) GetFileContents(file, sha string) (string, error) {
...
+		return nil, false
 	}
 	for _, event := range events {
 		if *event.Event == "merged" {
-			return event.CommitID
+			return event.CommitID, true
 		}
 	}
-	return nil
+	return nil, false
```
eparis: `true`

mungegithub/github/github.go
```diff
@@ -1385,8 +1419,9 @@ func (obj *MungeObject) MergePR(who string) error {
 
 	// Get the text of the first commit
 	firstCommit := ""
-	if commits, err := obj.GetCommits(); err != nil {
-		return err
+	if commits, ok := obj.GetCommits(); !ok {
+		glog.Errorf("Error in MergePR")
```
eparis: Error was in GetCommits() but if probably already `glog`'d

mungegithub/github/github.go
```diff
@@ -1416,20 +1451,20 @@ func (obj *MungeObject) MergePR(who string) error {
...
-		return err
+		return false
 	}
-	return nil
+	return true
 }
 
 // GetPRFixesList returns a list of issue numbers that are referenced in the PR body.
-func (obj *MungeObject) GetPRFixesList() []int {
+func (obj *MungeObject) GetPRFixesList() ([]int, bool) {
```
eparis: If this function can never fail, lets not bother with the `bool`

mungegithub/github/github.go
```diff
@@ -1536,63 +1571,67 @@ func (obj *MungeObject) DeleteComment(comment *github.IssueComment) error {
...
 	if !obj.IsPR() {
-		return false, nil
+		return false, true
 	}
-	pr, err := obj.GetPR()
-	if err != nil {
-		return false, err
+	pr, ok := obj.GetPR()
+	if !ok {
+		glog.Errorf("Error in IsMergable")
```
eparis: do we need a second glog?

mungegithub/github/github.go
```diff
@@ -1536,63 +1571,67 @@ func (obj *MungeObject) DeleteComment(comment *github.IssueComment) error {
...
 		glog.V(4).Infof("Waiting for mergeability on %q %d", *pr.Title, *pr.Number)
 		// TODO: determine what a good empirical setting for this is.
 		time.Sleep(2 * time.Second)
-		err := obj.Refresh()
-		if err != nil {
-			glog.Errorf("Unable to refresh PR# %d: %v", prNum, err)
-			return false, err
+		ok = obj.Refresh()
+		if !ok {
+			glog.Errorf("Unable to refresh PR# %d", prNum)
```
eparis: 2nd glog

mungegithub/github/github.go
```diff
@@ -1536,63 +1571,67 @@ func (obj *MungeObject) DeleteComment(comment *github.IssueComment) error {
...
+			return false, false
 		}
-		pr, err = obj.GetPR()
-		if err != nil {
-			glog.Errorf("Unable to get PR# %d: %v", prNum, err)
-			return false, err
+		pr, ok = obj.GetPR()
+		if !ok {
+			glog.Errorf("Unable to get PR# %d", prNum)
```
eparis: 2nd

mungegithub/github/github.go
```diff
@@ -1536,63 +1571,67 @@ func (obj *MungeObject) DeleteComment(comment *github.IssueComment) error {
...
-		return false, fmt.Errorf("Issue: %d is not a PR and is thus 'merged' is indeterminate", *obj.Issue.Number)
+		glog.Errorf("Issue: %d is not a PR and is thus 'merged' is indeterminate", *obj.Issue.Number)
+		return false, false
 	}
-	pr, err := obj.GetPR()
-	if err != nil {
-		return false, err
+	pr, ok := obj.GetPR()
+	if !ok {
+		glog.Errorf("Error in IsMerged")
```
eparis: 2nd

mungegithub/github/github.go
```diff
@@ -1536,63 +1571,67 @@ func (obj *MungeObject) DeleteComment(comment *github.IssueComment) error {
...
+	glog.Errorf("Unable to determine if PR %d was merged", *obj.Issue.Number)
+	return false, false
 }
 
 // MergedAt returns the time an issue was merged (for nil if unmerged)
-func (obj *MungeObject) MergedAt() *time.Time {
+func (obj *MungeObject) MergedAt() (*time.Time, bool) {
 	if !obj.IsPR() {
-		return nil
+		return nil, true
```
eparis: we called this a `false` above, lets call it false here.

mungegithub/github/github.go
```diff
@@ -1536,63 +1571,67 @@ func (obj *MungeObject) DeleteComment(comment *github.IssueComment) error {
...
 	}
-	pr, err := obj.GetPR()
-	if err != nil {
-		return nil
+	pr, ok := obj.GetPR()
+	if !ok {
+		glog.Errorf("Error in MergedAt")
```
eparis: 2nd

mungegithub/mungers/cherrypick-auto-approve.go
```diff
@@ -138,9 +145,10 @@ func (c *CherrypickAutoApprove) Munge(obj *github.MungeObject) {
 		// If the parent was for milestone v1.2 but this PR has
 		// comments saying it was 'on branch release-1.1' we should
 		// not auto approve
-		parentMilestone := parent.ReleaseMilestone()
-		if parentMilestone != branchImpliedMilestone {
-			glog.Errorf("%d: parentReleaseMilestone=%q but branch is %q", *obj.Issue.Number, parentMilestone, obj.Branch())
+		parentMilestone, ok := parent.ReleaseMilestone()
+		if parentMilestone != branchImpliedMilestone || !ok {
+			branch, _ := obj.Branch()
```
eparis: why do I have to do this again? Didn't we get branch in https://github.com/kubernetes/contrib/pull/1357/files#diff-b60bc1c6cc564c1b15eacccaca226ce6R126

mungegithub/mungers/submit-queue.go
```diff
@@ -1153,7 +1153,7 @@ func (sq *SubmitQueue) doGithubE2EAndMerge(obj *github.MungeObject) bool {
 		// Wait for the retest to start
 		sq.SetMergeStatus(obj, ghE2EWaitingStart)
 		atomic.AddInt32(&sq.prsTested, 1)
-		err = obj.WaitForPending(sq.RequiredRetestContexts)
+		err := obj.WaitForPending(sq.RequiredRetestContexts)
```
eparis: we didn't convert WaitForPending to a bool?  I guess that's fine.

eparis:
@99BenGardner

mungegithub/github/github.go
99BenGardner: Yep!

mungegithub/github/github.go
99BenGardner: It does

mungegithub/github/github.go
99BenGardner: It does

mungegithub/github/github.go
99BenGardner: Removed

mungegithub/github/github.go
99BenGardner: How would I do this if GetPr() no longer returns the error?

mungegithub/mungers/cherrypick-auto-approve.go
99BenGardner: Do you want me to take it out? It does look a little redundant

mungegithub/github/github.go
99BenGardner: If you mean below the Errorf on line 1063, it already does

mungegithub/github/github.go
eparis: Actually we shouldn't `glog` at all since GetPR already did.

mungegithub/github/github.go
```diff
@@ -1004,64 +1016,69 @@ func computeStatus(combinedStatus *github.CombinedStatus, requiredContexts []str
...
+		return true
 	}
-	_, _, err = config.client.Repositories.CreateStatus(config.Org, config.Project, ref, status)
+	_, _, err := config.client.Repositories.CreateStatus(config.Org, config.Project, ref, status)
 	if err != nil {
 		glog.Errorf("Unable to set status. PR %d Ref: %q: %v", *obj.Issue.Number, ref, err)
+		return false
 	}
-	return err
+	return false
```
eparis: return true

mungegithub/mungers/cherrypick-auto-approve.go
eparis: remove it.

apelisse:
I think it should be the other way around. Caller should log the err (if it matters) rather than having callee call it and return `ok`. Would potentially remove the glog dependency from `github/`

apelisse:
@eparis @99BenGardner What should we do about this? This is still blocking resolution of #1445 (which has been somewhat painful to recover from).

eparis:
I'll get it in monday. promise promise.

test-foxish:
recomputing cla status...

grodrigues3:

- **mungegithub**

cc @bprashanth 

PR #1360, opened 2016-07-13
Remove gce-scalability from blockers list

gmarek:
cc @eparis @lavalamp


lavalamp:
Why? Are we turning this job down?

wojtek-t:
@lavalamp - I'm also not 100% convinced. But the rationale behind this is that we have kubemark500 that is blocking.

lavalamp:
If we're confident in kubemark 500, then maybe we should also make
scalability a once-per-day thing, since it's expensive?

gmarek:
We'll be running it twice daily, but we don't want to block on that.

lavalamp:
OK, so you're running it less often. LGTM

PR #1361, opened 2016-07-13
[ansible] Add master_cluster_hostname param

talset:
- In order to add access to the api with an internal and public fqdn
  We added 2 variables in order to update the server.crt
- Add master_cluster_hostname variable
- Add kube_master_cluster_public_hostname variable
  
  TODO : we should use the internal name to configure the nodes

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1361)

### DIFF
--- a/ansible/group_vars/all.yml
+++ b/ansible/group_vars/all.yml
@@ -37,6 +37,12 @@ insecure_registrys:
 #https_proxy: "http://proxy.example.com:3128"
 #no_proxy: "127.0.0.1,localhost,docker-registry.somecorporation.com"
 
+# Internal fqdn used for the cluster (certificat only)
+# master_cluster_hostname: int-kubernetes
+
+# External fqdn used for the cluster (certificat only)
+# master_cluster_public_hostname: public-kubernetes
+
 # The port that the Kubernetes apiserver component listens on.
 kube_master_api_port: 443
 
--- a/ansible/roles/kubernetes/defaults/main.yml
+++ b/ansible/roles/kubernetes/defaults/main.yml
@@ -39,6 +39,12 @@ kube_cert_dir: "{{ kube_config_dir }}/certs"
 # The IP(s) for which the certificate will be valid
 kube_cert_ip: "{{ hostvars[inventory_hostname]['ansible_all_ipv4_addresses'] | join (',IP:') }}"
 
+# The internal name for the cluster
+kube_master_cluster_hostname: "{{ master_cluster_hostname | default('int-kubernetes') }}"
+
+# The public name for the cluster
+kube_master_cluster_public_hostname: "{{ master_cluster_public_hostname | default('public-kubernetes') }}"
+
 # This is where all of the bearer tokens will be stored
 kube_token_dir: "{{ kube_config_dir }}/tokens"
 
--- a/ansible/roles/kubernetes/files/make-ca-cert.sh
+++ b/ansible/roles/kubernetes/files/make-ca-cert.sh
@@ -34,6 +34,8 @@ set -o pipefail
 
 cert_ip="${MASTER_IP:="${1}"}"
 master_name="${MASTER_NAME:="kubernetes"}"
+master_cluster_hostname="${CLUSTER_MASTER_HOSTNAME:="public-kubernetes"}"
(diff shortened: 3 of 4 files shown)


ingvagabund:
@jbeda WDYT?

ingvagabund:
@eparis

test-foxish:
recomputing cla status...

talset:
I signed it!

grodrigues3:

- **ansible**

cc @eparis 

lavalamp:
@foxish @apelisse @grodrigues3 any idea why the bot wants me to approve changes to the ansible directory? I'm not in its OWNERS file.

foxish:
@lavalamp, The approval munger is currently turned off and will be turned on again by @grodrigues3 shortly. This issue where it would pick owners from a level higher (/ instead of ansible/) should be fixed now by https://github.com/kubernetes/contrib/pull/2194.

lavalamp:
ah, thanks for letting me know!

On Mon, Jan 9, 2017 at 1:59 PM, Anirudh Ramanathan <notifications@github.com

galexrt:
@talset Isn't that what #2356 does but without the extra `master_cluster_hostname` variable, as it just uses the Ansible detected IPs and hostnames?

talset:
@galexrt The goal here instead of using masters hostnames directly is to be able to have a VIP and public VIP fqdn to reach the kube-api.  (like in openshift : https://github.com/openshift/openshift-ansible/blob/master/inventory/byo/hosts.ose.example#L217-L225)

We should also use this internal vip fqdn to configure the kubeblet (api endpoint)

galexrt:
@talset Isn't this something that's already implemented? See https://github.com/kubernetes/contrib/blob/master/ansible/inventory/group_vars/all.yml#L14-L20

talset:
```Isn't this something that's already partly implemented```
Not one year ago and it is still incomplete.

Currently the master_cluster_hostname is used for internal and external. In many case we doesn't have the same dns for internal (between master/nodes) vip and external vip (from the client). For both dns it have to be part of the signed name of the certificat.

Regarding :
  * https://github.com/kubernetes/contrib/blob/master/ansible/roles/master/templates/scheduler.kubeconfig.j2#L8
   * https://github.com/kubernetes/contrib/blob/master/ansible/roles/master/templates/controller-manager.kubeconfig.j2#L8

It should use the internal "internal vip" name, not the first master. So we should replace it by cluster_hostname

**Comment by galexrt** [comment]
@talset The two files (`scheduler.kubeconfig`, `controller-manager.kubeconfig`) I mentioned are going to be fixed in this PR: https://github.com/kubernetes/contrib/pull/2450

***

Do you mean that the `master_cluster_hostname` should be part of the certificate on the masters?

Because that is already done from what I see, the `master_cluster_hostname` is already added to certificate of the masters: https://github.com/kubernetes/contrib/blob/master/ansible/roles/kubernetes/files/make-ca-cert.sh#L109-L111

**Comment by talset** [comment]
@galexrt Thanks, it is good that the cluster_hostname is now part of the certif. In my PR I expect that both cluster_hostname and public_cluster_hostname be part of the certif in order to allow to access to the api.
If not, and you try to access the api via your public dns, you will have an error because the domain is not part of the certificate.

**Comment by galexrt** [comment]
Is it really needed that two "extra" names are added to the certificate?

I personally find it reasonable to only add `master_cluster_hostname` and the other hostnames to the certificate. Isn't the `master_cluster_hostname` enough?

**Comment by talset** [comment]
The cluster_hostname is already present (to be able to reach the private/internal vip kube api).
The goal of public_cluster_hostname is to have the same thing but for public access to the api.

If the public vip domain is not part of the certificat, when you will try to access to the api via kubectl you will have something like that :

```
Unable to connect to the server: x509: certificate is valid for cluster_hostname, master0, master1 ..., not public_cluster_hostname 
```

galexrt:
@talset Do you want to continue this PR or should I add this feature in a new PR?

talset:
This PR is kind of deprecated for the cluster_hostname now. The only missing thing is to be able to add a second cluster_hostname (public) in order to sign the certificate with.

We could do that in a different PR.

talset:
I going to update my PR to handle only the public part for the certif

talset:
Closing the PR.

Public cluster hostname is here : https://github.com/kubernetes/contrib/pull/2452

PR #1363, opened 2016-07-13
Use existing uid if one is found

bprashanth:
Without this if we create some ingresses we will get eg: a forwarding rule like "foo-uid". Now if we restart 
the ingress controller, and while it's down delete the configmap where it stores its uid, it will come back, see an existing ingress, but wrongly record the uid as "empty string". This will cause the ingress to ignore the old forwarding rule, backends etc.

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1363)


ingress/controllers/gce/main.go
```diff
@@ -266,6 +267,15 @@ func getClusterUID(kubeClient *client.Client, name string) (string, error) {
 	}
 	for _, ing := range ings.Items {
 		if len(ing.Status.LoadBalancer.Ingress) != 0 {
+			fwRuleName, ok := ing.Annotations[fmt.Sprintf("%v/forwarding-rule", utils.K8sAnnotationPrefix)]
+			if ok {
+				uidList := strings.Split(fwRuleName, clusterNameDelimiter)
```
girishkalele: Kind of wary of this parsing logic here - the code/utility that created the fwRuleName string in the annotation should provide a parse function to convert it back into its constituent components ?

bprashanth:
@girishkalele sorry for the delay, PTAL. I did what you suggested (moved the parsing logic into functions closer to the source of setting the annotations), added unittests, polled the UID config map to pick up changes, and added some docs

ingress/controllers/gce/BETA_LIMITATIONS.md
```diff
@@ -120,3 +122,42 @@ gcloud container clusters create mycluster --network "default" --num-nodes 1 \
...
+```
+
+You can pick a different UID, but this requires you to:
+
+1. Delete existing Ingresses
+2. Edit the configmap using `kubectl edit`
+3. Recreate the same Ingress
+
+After step 3 the Ingress should come up using the new UID as the suffix of all cloud resources. You can't simply change the UID if you have existing Ingresses, because
+renaming a cloud resource requires a delete/create cycle that the Ingress controller does not currently automate. Note that the UID in step 1 might be the empty string,
```
girishkalele: s/be the empty/be an empty/

ingress/controllers/gce/BETA_LIMITATIONS.md
```diff
@@ -120,3 +122,42 @@ gcloud container clusters create mycluster --network "default" --num-nodes 1 \
...
+if you had a working Ingress before upgrading to Kubernetes 1.3.
+
```
girishkalele: Do these instructions apply to everyone with ingresses configured and upgrading to 1.3 ? Is the UID going to change from "" to something non-empty when we upgrade to 1.3 ?

bprashanth: If its "" it'll remain "". We search for existing loadbalancers and avoid autoassigning uids to them. 
The bug was that if it's --cluster-uid=random-string, we think it's "" on finding the existing loadbalancer.

mml:
Is there anything preventing the user from choosing a UI containing '-'?

---

Reviewed 1 of 1 files at r1, 7 of 7 files at r2.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[ingress/controllers/gce/BETA_LIMITATIONS.md, line 127 \[r2\]](https://reviewable.kubernetes.io:443/reviews/kubernetes/contrib/1363#-KOqa4hvf-j-R9OD4OIB:-KOqa4hwViPZduP6TuYR:b-hkht4r) ([raw file](https://github.com/kubernetes/contrib/blob/bcf4aae917726c5a45481fab57fb0895d86b84f1/ingress/controllers/gce/BETA_LIMITATIONS.md#L127)):*
> stores in a config map

"config map" or "configmap"?

---

*[ingress/controllers/gce/main.go, line 245 \[r2\]](https://reviewable.kubernetes.io:443/reviews/kubernetes/contrib/1363#-KOqa_QLtqiAsV3nnF4M:-KOqa_QLtqiAsV3nnF4N:bblx6zv) ([raw file](https://github.com/kubernetes/contrib/blob/bcf4aae917726c5a45481fab57fb0895d86b84f1/ingress/controllers/gce/main.go#L245)):*
> ```Go
> 	vault := storage.NewConfigMapVault(kubeClient, api.NamespaceSystem, uidConfigMapName)
> 
> 	// Start a goroutine to poll the cluster UID config map
> ```

Why don't we watch instead of poll?

---


*Comments from [Reviewable](https://reviewable.kubernetes.io:443/reviews/kubernetes/contrib/1363#-:-KOqc6cjSSAvYKIYsdVa:bx4774q)*

bprashanth:
The name delimiter is `--`, and no we don't validate that map. It's probably worth doing, and sending a warning/event, and documenting it. I don't expect users to set this uid. It's going to be machine generated in a majority of the cases.

bprashanth:
*[ingress/controllers/gce/main.go, line 245 \[r2\]](https://reviewable.kubernetes.io:443/reviews/kubernetes/contrib/1363#-KOqa_QLtqiAsV3nnF4M:-KOqcvyTlDX0-0sHHdlE:bjruj6r) ([raw file](https://github.com/kubernetes/contrib/blob/bcf4aae917726c5a45481fab57fb0895d86b84f1/ingress/controllers/gce/main.go#L245)):*
<details><summary><i>Previously, mml (Matt Liggett) wrote…</i></summary>
> Why don't we watch instead of poll?
</details>
too many degrees of freedom already. Thsi controller watches 5 resources. Might as well just replicate etcd at this point. This one is simple enough, and I know exactly what will change, that I don't want the overhead of a watch connection. 

---


*Comments from [Reviewable](https://reviewable.kubernetes.io:443/reviews/kubernetes/contrib/1363)*

mml:
Let's say the UID gets set to "foo-bar".  We emit the warning, but what does the controller do?  Does it stick with the previously set value?  What if there is no previous value?

---

Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[ingress/controllers/gce/main.go, line 245 \[r2\]](https://reviewable.kubernetes.io:443/reviews/kubernetes/contrib/1363#-KOqa_QLtqiAsV3nnF4M:-KOqhdYWnMaepda0Jl0p:bte4z6a) ([raw file](https://github.com/kubernetes/contrib/blob/bcf4aae917726c5a45481fab57fb0895d86b84f1/ingress/controllers/gce/main.go#L245)):*
<details><summary><i>Previously, bprashanth (Prashanth B) wrote…</i></summary>
> too many degrees of freedom already. Thsi controller watches 5 resources. Might as well just replicate etcd at this point. This one is simple enough, and I know exactly what will change, that I don't want the overhead of a watch connection. 
</details>
Please add a comment saying the same thing.

---


*Comments from [Reviewable](https://reviewable.kubernetes.io:443/reviews/kubernetes/contrib/1363#-:-KOqhngNQE3ZjHRrsbXo:b-z1db79)*

bprashanth:
if it gets set to "foo--bar" the controller takes bar. The spec is everything after the last delimiter, which default to "--". If the user noticies this and wants to change things, they'd either delete the ingress and modify the uid, or keep the existing uid but update the ingress to uid=bar. 

We could also have the controller hang, but why? I don't think it's going to be that detrimental. This is just like overlapping selectors, you can shoot yourself in the foot and nothing will stop you, but the system won't do it if left alone.  We can't validate config maps, and I don't think this is worth crashlooping over. Open to suggestions but don't want to rathole.

---







*Comments from [Reviewable](https://reviewable.kubernetes.io:443/reviews/kubernetes/contrib/1363#-:-KOqiYYi2GQQzz7PUhRW:bo7h3xp)*

bprashanth:
The recommended way to modify uid is to delete ingress, because we can't rename resource without delete/create cycled anyway and if we put that on the user they can pick a time when traffic is low. The alternative is to log something every 5s and keep using whatever the uid was (it can never be "nothing" post 1.3 because we pick a uid by default, it can only be nothing if they have a really old cluster, in which case it will continue being nothing). 

---

Review status: all files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.kubernetes.io:443/reviews/kubernetes/contrib/1363#-:-KOqkBwwDBg3dkoGe6s_:b-vy8zfe)*

bprashanth:
PTAL, updated language in the README and added a unittests for the name containing delimiter case.

bprashanth:
*[ingress/controllers/gce/main.go, line 245 \[r2\]](https://reviewable.kubernetes.io:443/reviews/kubernetes/contrib/1363#-KOqa_QLtqiAsV3nnF4M:-KOv0jCKwa1juiRtFMRe:b-896fix) ([raw file](https://github.com/kubernetes/contrib/blob/bcf4aae917726c5a45481fab57fb0895d86b84f1/ingress/controllers/gce/main.go#L245)):*
<details><summary><i>Previously, mml (Matt Liggett) wrote…</i></summary>
> Please add a comment saying the same thing.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.kubernetes.io:443/reviews/kubernetes/contrib/1363)*

bprashanth:
*[ingress/controllers/gce/BETA_LIMITATIONS.md, line 127 \[r2\]](https://reviewable.kubernetes.io:443/reviews/kubernetes/contrib/1363#-KOqa4hvf-j-R9OD4OIB:-KOv0lmOqWSGtD9-7-u2:b-896fix) ([raw file](https://github.com/kubernetes/contrib/blob/bcf4aae917726c5a45481fab57fb0895d86b84f1/ingress/controllers/gce/BETA_LIMITATIONS.md#L127)):*
<details><summary><i>Previously, mml (Matt Liggett) wrote…</i></summary>
> > stores in a config map
> 
> "config map" or "configmap"?
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.kubernetes.io:443/reviews/kubernetes/contrib/1363)*

girishkalele:
Reviewed 1 of 1 files at r1, 2 of 4 files at r3.
Review status: 6 of 8 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.kubernetes.io:443/reviews/kubernetes/contrib/1363)*

girishkalele:
Reviewed 2 of 7 files at r2.
Review status: 6 of 8 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.kubernetes.io:443/reviews/kubernetes/contrib/1363)*

bprashanth:
LGTM so i can get some test miles on this?

girishkalele:
Reviewed 2 of 4 files at r3.
Review status: 7 of 8 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.kubernetes.io:443/reviews/kubernetes/contrib/1363)*

girishkalele:
Reviewed 1 of 7 files at r2.
Review status: 7 of 8 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.kubernetes.io:443/reviews/kubernetes/contrib/1363)*

girishkalele:
:lgtm: 

---

Review status: 7 of 8 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.kubernetes.io:443/reviews/kubernetes/contrib/1363#-:-KOzqBUwEEPk0JMIeD7p:bg7igkv)*

PR #1365, opened 2016-07-14
Import kubeform - a Kubernetes Terraform plugin.

madhusudancs:
This should probably be in a separate repository, but upon @bgrant0607's suggestion I am putting it here until we find a better home.


madhusudancs:
Superseded by PR #1366.

PR #1341, opened 2016-07-11
Cluster-autoscaler: scaledown node usage trackers

mwielgus:
Right now on successful scale down we reset utilization information on all nodes, even if they had nothing in common during the simulation and utilization calculation. This PR brings in a data structure that remembers which nodes were found useful for which nodes during scale down simulations so that only nodes that used the same other nodes for simulated pod rescheduling are reset.

cc: @piosz @fgrzadkowski @jszczepkowski

PR #1370, opened 2016-07-14
Cluster-autoscaler: use node usage tracker for better utilization resets

mwielgus:
Follow up of #1341.

There remaining work is:
- Better node selection in rescheduling checks so that the possible reschedulings are evenly spread across the nodes.
- Include number of nodes that would be cleared in node to remove selection.


cluster-autoscaler/simulator/tracker.go
```diff
@@ -0,0 +1,169 @@
...
+		}
+	}
+	for _, key := range toDelete {
+		delete(tracker.usage, key)
+	}
+}
+
+// RemoveNodeFromTracker removes node from tracker and also cleans the passed utilization map.
+func RemoveNodeFromTracker(tracker *UsageTracker, node string, utilization map[string]time.Time) {
+	keysToRemove := make([]string, 0)
```
jszczepkowski: nit: use StringKeySet from pkg/util/sets/string.go

jszczepkowski:
LGTM

PR #1373, opened 2016-07-15
Skip retest for docs

### LINKED ISSUE #915: Automatically add retest-not-required to PRs that only touch docs
cc @fejta @eparis

apelisse:
Fixes #915.

Creates a new mungers that will check changed files, and label as "retest-not-required" if it contains only ".md" files.

### DIFF
--- a/mungegithub/mungers/docs-no-retest.go
+++ b/mungegithub/mungers/docs-no-retest.go
@@ -1,0 +1,96 @@
+/*
+Copyright 2015 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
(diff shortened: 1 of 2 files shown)


bgrant0607:
@grodrigues3 PTAL

grodrigues3:
Have two separate labels for indicating skips-retest.  One applied by the bot and one by humans.   This avoids cases where a doc file was edited and the label gets applied followed by coding being added on the same PR and the label not being removed.

apelisse:
@grodrigues3 Addressed, PTAL

mungegithub/mungers/docs-no-retest.go
```diff
@@ -0,0 +1,96 @@
...
+	"k8s.io/contrib/mungegithub/features"
+	"k8s.io/contrib/mungegithub/github"
+)
+
+const (
+	labelSkipRetest = "retest-not-required-docs-only"
+)
+
+var (
+	ignoreFilesRegex = regexp.MustCompile(".*\\.md$")
```
grodrigues3: are docs always markdown?  Will there ever be any other extension we want to apply this to?

apelisse: Good point. This has been discussed quickly in #915.
Getting this done for '.md' would probably fix 80% of the problem with 20% of the effort :-)

grodrigues3:
LGTM

eparis:
I would have prefered a different fix.  One in which the box removed `retest-not-required` if the bot added that label and now the PR was not docs only. But alas, this is good enough.

PR #1374, opened 2016-07-15
Simplify size calculation

apelisse:
I found an optimization in the way we compute the size of a pull-request.

Please let me know if this makes sense as it's pretty much untested (sorry, not sure how to test easily).
This depends on a commit common with #1373

### DIFF
--- a/mungegithub/github/github.go
+++ b/mungegithub/github/github.go
@@ -200,6 +200,7 @@ type analytics struct {
 	ListIssueEvents    analytic
 	ListCommits        analytic
 	GetCommit          analytic
+	ListFiles          analytic
 	GetCombinedStatus  analytic
 	SetStatus          analytic
 	GetPR              analytic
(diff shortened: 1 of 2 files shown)


apelisse:
@thockin @eparis You guys are the main contributors to this file

thockin:
I have lost a lot of context.  It looks OK.  Can you run it under dry-run and see it makes sense?  There's low impact to getting it wrong.  It already gets it wrong a lot anyway.

Long way of saying LGTM if you think it is better

apelisse:
Yeah, I ended-up running in dry-run, it provided very different result (e.g. Removing label XS -> Adding XXL), so that got me interested. Here's what's going on:

Let's say Jane submits a pull-request with 10 commits. Each of these commits change the same single line. Each commit will have 1 addition, 1 deletion. Same for the pull-request full stats.

If we think this file is generated, we will remove additions and deletions for this file for each commit from the full pull-request stats. IOW, we will end-up with -9 additions, -9 deletions ...
If there are other changes in the PR, this will negatively (literally) impact the size calculation.

The new way computes the number of changes based on the entire set of commit of the pull-request.

One extreme example of this happening is this pull-request:
https://github.com/kubernetes/kubernetes/pull/21942

It starts with 2466 additions and 122697 deletions. After discounting changes made to generated files, we end-up with ... -56407 additions and -50603 deletions. Marking it as a very very small pull-request :-)
With the patch, we end-up with 915 additions and 6719 deletions (marking it as XXL)

Also, the pull-request has 64 commits. So every time the mungebot tries to compute the size of this PR, it will (not considering pagination):
- Request list of commits (1 request)
- Request each individual commit (64 requests)

The new implementation will only request for changed files (1 request)

That's a significant reduction in the number of request.
And it will change some labels.

(Very very) long way of saying that I think it's better :-)

thockin:
SGTM

PR #1375, opened 2016-07-15
Added new feature to handle test options.

foxish:
- The required-retest-contexts are used by several mungers. Previously, it was [a global variable](https://github.com/kubernetes/contrib/issues/1304#issuecomment-231817807) which would get referenced irrespective of the parameters supplied via the command-line.
- We abstract the retest-context flag out as a feature that the mungers that want to can access them.
- In future, if multiple mungers start to use any other test related parameters, they could be added into the test-options feature.

### DIFF
--- a/mungegithub/features/features.go
+++ b/mungegithub/features/features.go
@@ -26,9 +26,10 @@ import (
 // Features are all features the code know about. Care should be taken
 // not to try to use a feature which isn't 'active'
 type Features struct {
-	Repos   *RepoInfo
-	GCSInfo *GCSInfo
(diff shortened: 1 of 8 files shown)


foxish:
@apelisse @lavalamp PTAL

apelisse:
This is somewhat a nit. But please use an empty line after the first line of your commit message. It makes it much easier to read in the console.

lavalamp:
looks OK to me. @eparis?

lavalamp:
LGTM

PR #1376, opened 2016-07-15
Add close-stale-pr to munger list

apelisse:
Makefile has a list of mungers to run, and this one has been omitted.


lavalamp:
Let's turn this on on monday-- I will warn people that this is coming in an email.

PR #1379, opened 2016-07-16
[nginx-ingress-controller] improve nginx performance

aledbf:
Changing container `/proc` values with a privileged sidecar.
NGINX reads `net.core.somaxconn` to increase the size of backlog queue of pending connections

### DIFF
--- a/ingress/controllers/nginx/examples/sysctl/change-proc-values-rc.yaml
+++ b/ingress/controllers/nginx/examples/sysctl/change-proc-values-rc.yaml
@@ -1,0 +1,123 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: default-http-backend
+  labels:
+    k8s-app: default-http-backend
+spec:
+  ports:
+  - port: 80
+    targetPort: 8080
+    protocol: TCP
+    name: http
+  selector:
+    k8s-app: default-http-backend
+---
+apiVersion: v1
+kind: ReplicationController
+metadata:
+  name: default-http-backend
+spec:
+  replicas: 1
+  selector:
+    k8s-app: default-http-backend
+  template:
+    metadata:
+      labels:
+        k8s-app: default-http-backend
+    spec:
+      terminationGracePeriodSeconds: 60
+      containers:
+      - name: default-http-backend
+        # Any image is permissable as long as:
+        # 1. It serves a 404 page at /
+        # 2. It serves 200 on a /healthz endpoint
+        image: gcr.io/google_containers/defaultbackend:1.0
+        livenessProbe:
+          httpGet:
+            path: /healthz
+            port: 8080
+            scheme: HTTP
+          initialDelaySeconds: 30
+          timeoutSeconds: 5
+        ports:
+        - containerPort: 8080
+        resources:
+          limits:
+            cpu: 10m
+            memory: 20Mi
+          requests:
+            cpu: 10m
+            memory: 20Mi
+---
+apiVersion: v1
+kind: ReplicationController
+metadata:
+  name: nginx-ingress-controller
+  labels:
+    k8s-app: nginx-ingress-lb
+spec:
+  replicas: 1
+  selector:
+    k8s-app: nginx-ingress-lb
+  template:
+    metadata:
+      labels:
+        k8s-app: nginx-ingress-lb
+        name: nginx-ingress-lb
+    spec:     
+      terminationGracePeriodSeconds: 60
+      containers:
+      - image: alpine:3.4
+        name: sysctl-buddy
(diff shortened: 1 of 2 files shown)


aledbf:
ping @bprashanth

ingress/controllers/nginx/examples/sysctl/change-proc-values-rc.yaml
```diff
@@ -0,0 +1,114 @@
...
+      terminationGracePeriodSeconds: 60
+      containers:
+      - image: alpine:3.4
+        name: sysctl-buddy
+        command:
+        - /bin/sh
+        args:
+        - -c
+        - sysctl -w net.core.somaxconn=32768
+          && sysctl -w net.ipv4.ip_local_port_range='1024 65535'
```
bprashanth: just make this a while true loop and you get free reconciliation?

bprashanth: eg: 
```
command:
- /bin/sh
- -c
- |
  while true; do sysctl && sysctl; sleep 10; done
```

aledbf: done

ingress/controllers/nginx/examples/sysctl/change-proc-values-rc.yaml
```diff
@@ -0,0 +1,112 @@
...
+    spec:     
+      terminationGracePeriodSeconds: 60
+      containers:
+      - image: alpine:3.4
+        name: sysctl-buddy
+        command:
+        - /bin/sh
+        - -c
+        - |
+          while true; do sysctl -w net.core.somaxconn=32768 && sysctl -w net.ipv4.ip_local_port_range='1024 65535'; sleep 10; done
```
bprashanth: you can actually split this into lines now that you have the "|", for readability. 
```
while true; do 
  sysctl stuff
done
```
Also aren't there a couple of other nice tcp level tunables? maybe we should consider a custom entrypoint script that people can extend?

bprashanth: It would also be good to note down what are per ns sysctls vs shared between host and ns. I don't think any of the tcp tunables are per netns actually. I do think the ones in /sys/net/core are per net ns.

aledbf: done

ingress/controllers/nginx/examples/sysctl/change-proc-values-rc.yaml
```diff
@@ -0,0 +1,123 @@
...
+        # Network Namespace: Sysctls beginning with net.*
+        # 
+        # kubectl <podname> -c sysctl-buddy -- sysctl -A | grep net
+        command:
+        - /bin/sh
+        - -c
+        - |
+          while true; do 
+            sysctl -w net.core.somaxconn=32768
+            sysctl -w net.ipv4.ip_local_port_range='1024 65535'
```
bprashanth: what about backlog queue size?  is that not per ns?

aledbf: > what about backlog queue size? is that not per ns?

No

```
sysctl -A |grep "net.core.netdev_max_backlog"
sysctl: error reading key 'net.ipv6.conf.all.stable_secret': I/O error
sysctl: error reading key 'net.ipv6.conf.default.stable_secret': I/O error
sysctl: error reading key 'net.ipv6.conf.eth0.stable_secret': I/O error
sysctl: error reading key 'net.ipv6.conf.lo.stable_secret': I/O error
```

bprashanth:
LGTM

ingress/controllers/nginx/examples/sysctl/change-proc-values-rc.yaml
```diff
@@ -0,0 +1,123 @@
...
+    spec:     
+      terminationGracePeriodSeconds: 60
+      containers:
+      - image: alpine:3.4
+        name: sysctl-buddy
+        # using kubectl exec you can check which other parameters is possible to change
+        # IPC Namespace:     kernel.msgmax, kernel.msgmnb, kernel.msgmni, kernel.sem, kernel.shmall, 
+        #                    kernel.shmmax, kernel.shmmni, kernel.shm_rmid_forced and Sysctls 
+        #                    beginning with fs.mqueue.*
+        # Network Namespace: Sysctls beginning with net.*
```
bprashanth: actually this is not true right? dont think fin timeout is per ns? or is it

PR #1380, opened 2016-07-16
Introduce etcd playbook

ingvagabund:
Following effort in https://github.com/kubernetes/contrib/pull/967, this PR introduces `etcd-setup` playbook. Changes:
- introduce localhost inventory
- merge `etcd_modified` and `etcd2_modified`, update `restart etcd` task so it can be run as self-standing task


ingvagabund:
@stephenrlouie can you review?

ingvagabund:
@redhat-k8s-bot test please

ingvagabund:
I will rebase the PR once https://github.com/kubernetes/contrib/pull/967 gets merged.

ingvagabund:
@gouyang

ansible/README.md
```diff
@@ -14,7 +14,7 @@ can be real hardware, VMs, things in a public cloud, etc. Anything that you can
 
 ### Configure inventory
 
-Add the system information gathered above into the 'inventory' file, or create a new inventory file for the cluster.
+Add the system information gathered above into the 'inventory' file, or create a new inventory file for the cluster. Place the 'inventory' file into the `./inventories` directory.
```
stephenrlouie: Added in #967

ansible/playbooks/etcd-setup.yml
```diff
@@ -0,0 +1,10 @@
+---
```
stephenrlouie: Why is this at the top level playbooks directory, will each role have a setup play here? Are you envisioning having a `deploy-cluster.yml` script that would call `<role>-setup.yml`. To expand we could have a `reset-cluster.yml` call `<role>-reset.yml's` In each `<role>-reset.yml` we use the appropriate tags or includes to enforce the right order and control of tasks executed within that role.

We need to figure out the exactness of how to run something like `etcd-setup.yml` or an `etcd-upgrade.yml`. Maybe we just call specific playbooks in the roles themselves tasks with these alternate flows. We can just call the role with an upgrade tag. I think the later would be easiest to do with what ansible provides and what is existing in contrib.

stephenrlouie:
> Following effort in #967, this PR introduces etcd-setup and etcd-update playbooks.

I don't see an etcd-update here. 

Restart changes look good. Lots of overlap with #967, but once thats merged and this is rebased it should be a much smaller PR.

ansible/README.md
ingvagabund: I have cherry picked your commit. Once #967 gets merged I will rebase this PR and remove your commit from it.

ingvagabund:
> > Following effort in #967, this PR introduces etcd-setup and etcd-update playbooks.
> 
> I don't see an etcd-update here.

I have removed it as there is no way to set tags within a playbook [1]

[1] http://stackoverflow.com/questions/25674649/ansible-in-a-playbook-filter-a-role-by-tags-without-passing-at-the-command-l?answertab=votes#tab-top

ansible/playbooks/etcd-setup.yml
ingvagabund: > Why is this at the top level playbooks directory, will each role have a setup play here?

I am following openstack-ansible. Though, I did not absorb every aspect for it.

> Are you envisioning having a deploy-cluster.yml script that would call <role>-setup.yml.

Yes

> To expand we could have a reset-cluster.yml call <role>-reset.yml's In each <role>-reset.yml we use the appropriate tags or includes to enforce the right order and control of tasks executed within that role.

That was the idea. The ``playbooks`` directory would contain ``<role>-setup.yml``, ``<role>-update.yml``, ``<role>-reset.yml`` and others. Each playbook would set appropriate tags (e.g. for ``<role>-update.yml`` it would be ``install`` and ``reset``) and only tagged tasks would be run. However, there is no way to choose tags from within a playbook or role. As based on [1] you can only use ``--tags`` and ``--skip-tags`` flags of the ``ansible-playbook`` command.

> We need to figure out the exactness of how to run something like etcd-setup.yml or an etcd-upgrade.yml. 

Completely agree with that.

> Maybe we just call specific playbooks in the roles themselves tasks with these alternate flows. We can just call the role with an upgrade tag. I think the later would be easiest to do with what ansible provides and what is existing in contrib.

IIUC, you would like to call specific tasks inside a role? By default, ``main.yml`` task is called. Is there a way to call other tasks like ``etcd-start.yml``?

[1] http://stackoverflow.com/questions/25674649/ansible-in-a-playbook-filter-a-role-by-tags-without-passing-at-the-command-l?answertab=votes#tab-top

ingvagabund: "If you find yourself building a role with lots of tags and you want to call subsets of the role at different times, you should consider just splitting that role into multiple roles." [1].

Meaning, we need to break the ``etcd`` role out into ``etcd-start``, ``etcd-configure``, ``etcd-restart`` and other roles if we want to support more scenarios than just an ``etcd-setup``. 

@stephenrlouie Is this, what you meant when you were talking about "The end state is to deploy each role like a 'micro-service' that has granular control so one can spin up, spin down, individual roles without having damaging effects across the cluster." ?

[1] http://docs.ansible.com/ansible/playbooks_roles.html#roles

ingvagabund: It is not hard to imagine how many playbooks there is going to be if we take this path. If we break the etcd roles into multiple roles we get overwhelmed. At the same time it will be more difficult to maintain a subset of etcd like playbooks than once playbook with tagged tasks. Instead of creating ``etcd-start``, ``etcd-configure`` and other playbooks included in ``etcd-setup`` one, I would suggest to shift the responsibility of what task gets run into scripts which set appropriate tags.

E.g. 
```sh
$ cat scripts/deploy-etcd.sh
ansible-playbook -i ${INVENTORY} playbooks/etcd-setup.yml --tags all
```

```sh
$ cat scripts/update-etcd.sh
ansible-playbook -i ${INVENTORY} playbooks/etcd-setup.yml --tags "install,restart"
```

```sh
$ cat scripts/restart-etcd.sh
ansible-playbook -i ${INVENTORY} playbooks/etcd-setup.yml --tags "restart"
```

On the level of the entire cluster:

```sh
$ cat scripts/deploy-cluster.sh
ansible-playbook -i ${INVENTORY} playbooks/deploy-cluster.yml
```

```sh
$ cat scripts/update-cluster.sh
ansible-playbook -i ${INVENTORY} playbooks/deploy-cluster.yml --tags="install,restart"
```

ansible/roles/etcd/tasks/install.yml
```diff
@@ -4,7 +4,7 @@
     etcd_source_type: "package-manager"
   when: not is_atomic and ansible_distribution != "Ubuntu" and not is_coreos
 
-- name: Install etcd
+- name: Install etcd via package manager
   action: "{{ ansible_pkg_mgr }}"
```
sttts: why not `package:` (http://docs.ansible.com/ansible/package_module.html)?

ansible/playbooks/deploy-cluster.yml
```diff
@@ -91,3 +91,11 @@
   sudo: yes
   roles:
     - { role: contiv, contiv_role: netplugin, when: networking == 'contiv' }
+
+# install runtime dependencies
```
sttts: why isn't the role called runtime-dependencies?

ansible/roles/etcd/tasks/install.yml
```diff
@@ -4,7 +4,7 @@
     etcd_source_type: "package-manager"
   when: not is_atomic and ansible_distribution != "Ubuntu" and not is_coreos
 
-- name: Install etcd
+- name: Install etcd via package manager
```
sttts: bad commit message. First line should be understandable alone. Put the "... to support the following scenarios" in the third line.

ansible/roles/etcd/tasks/install.yml
ingvagabund: ``package`` is available since ansible-2.0. The code was written before that.

ansible/playbooks/deploy-cluster.yml
ingvagabund: Sadly, the command is not ultimately correct. I have already remove that role when refactoring kubernetes roles https://github.com/kubernetes/contrib/pull/1444. So this can be skipped.

ansible/playbooks/etcd-setup.yml
sttts: I usually make the main playbook idem-potent, doing everything which makes sense at least. So you can run it again and again (= deploy + configure). Restarts are done with notifiers (compare https://github.com/sttts/compute-platform/blob/master/roles/mesos/tasks/master.yml#L8).

Those steps which do not fit in this scheme (or are dangerous to do automatically), I split them out in special playbooks.

ingvagabund: As word ``setup`` is very close to work ``configure``, I would use ``<role>-deploy`` instead the setup form. At the same time ``<action>-<role>`` sounds more friendly than ``<role>-setup``. As we have ``deploy-cluster`` already, let's align with that.

ingvagabund: > 
> 
> I usually make the main playbook idem-potent, doing everything which makes sense at least. So you can run it again and again (= deploy + configure). Restarts are done with notifiers (compare https://github.com/sttts/compute-platform/blob/master/roles/mesos/tasks/master.yml#L8).

One can not run notifier without running a task that actually notifies a notifier. So this way is no way. Hard to say how many use cases there are which require only restart. I can imagine someone installing/updating etcd or docker manually (followed by manual configuration) without actually running install/configure tasks. So keeping the restart step separate out of notifiers make sense.

> 
> Those steps which do not fit in this scheme (or are dangerous to do automatically), I split them out in special playbooks.
> 

There is adhoc directory exactly for that (or at least that is my impression)

ingvagabund: Anyway, these are new scripts which does not have to be perfect right away. People can play with it and open issues/PR for improvements. Just wanted to prepare a general template which can be filled in remaining roles.

ingvagabund:
@redhat-k8s-bot test please

sttts:
looks good to me now

PR #1384, opened 2016-07-18
Unset Authorization header when proxying

### LINKED ISSUE #1383: [nginx-ingress-controller] Basic auth header is not stripped
If you use basic auth in an ingress, then the `Authorization` header is passed along to the proxied upstream, which can result in incorrect behaviour if the upstream doesn't expect one.

For example, [Drone](https://github.com/drone/drone) will ignore its session cookie if the auth header is specified.

The correct behaviour is to strip the header and not pass it to the upstream.

Until this is fixed, I've been trying to find a workaround by injecting a custom Nginx directive, but I can't a way. Is there one?

atombender:
Fixes #1383.


aledbf:
LGTM

PR #1385, opened 2016-07-18
Removing defaults from code and making command-line options explicit in deployment.yaml.

foxish:
Continuing work on #1304. 

We have a large number of defaults specified for various flags according to the main kubernetes repository. 
The defaults don't make sense as we make mungegithub work with other repositories.

- The first commit adds logging to print the values of various command-line parameters.
- The second removes defaults and includes them in the deployment.yaml file.

### DIFF
--- a/mungegithub/features/repo-updates.go
+++ b/mungegithub/features/repo-updates.go
@@ -135,8 +135,9 @@ func (o *RepoInfo) updateRepoUsers() error {
 
 // Initialize will initialize the munger
 func (o *RepoInfo) Initialize() error {
-	o.enabled = true
+	glog.Infof("kubernetes-dir: %#v\n", o.kubernetesDir)
 
+	o.enabled = true
 	if len(o.kubernetesDir) == 0 {
 		glog.Fatalf("--kubernetes-dir is required with selected munger(s)")
 	}
--- a/mungegithub/features/repo-updates.go
+++ b/mungegithub/features/repo-updates.go
@@ -168,7 +168,7 @@ func (o *RepoInfo) EachLoop() error {
 
 // AddFlags will add any request flags to the cobra `cmd`
 func (o *RepoInfo) AddFlags(cmd *cobra.Command) {
-	cmd.Flags().StringVar(&o.kubernetesDir, "kubernetes-dir", "./gitrepos/kubernetes", "Path to git checkout of kubernetes tree")
+	cmd.Flags().StringVar(&o.kubernetesDir, "kubernetes-dir", "", "Path to git checkout of kubernetes tree")
 }
 
 // GitCommand will execute the git command with the `args`
--- a/mungegithub/github/github.go
+++ b/mungegithub/github/github.go
@@ -301,9 +301,9 @@ func (config *Config) AddRootFlags(cmd *cobra.Command) {
 	cmd.PersistentFlags().IntVar(&config.MinPRNumber, "min-pr-number", 0, "The minimum PR to start with")
 	cmd.PersistentFlags().IntVar(&config.MaxPRNumber, "max-pr-number", maxInt, "The maximum PR to start with")
 	cmd.PersistentFlags().BoolVar(&config.DryRun, "dry-run", true, "If true, don't actually merge anything")
-	cmd.PersistentFlags().StringVar(&config.Org, "organization", "kubernetes", "The github organization to scan")
-	cmd.PersistentFlags().StringVar(&config.Project, "project", "kubernetes", "The github project to scan")
-	cmd.PersistentFlags().StringVar(&config.state, "state", "open", "State of PRs to process: 'open', 'all', etc")
+	cmd.PersistentFlags().StringVar(&config.Org, "organization", "", "The github organization to scan")
+	cmd.PersistentFlags().StringVar(&config.Project, "project", "", "The github project to scan")
+	cmd.PersistentFlags().StringVar(&config.state, "state", "", "State of PRs to process: 'open', 'all', etc")
 	cmd.PersistentFlags().StringSliceVar(&config.labels, "labels", []string{}, "CSV list of label which should be set on processed PRs. Unset is all labels.")
 	cmd.PersistentFlags().StringVar(&config.Address, "address", ":8080", "The address to listen on for HTTP Status")
 	cmd.PersistentFlags().StringVar(&config.WWWRoot, "www", "www", "Path to static web files to serve from the webserver")
@@ -329,7 +329,6 @@ func (config *Config) PreExecute() error {
 	glog.Infof("http-cache-dir: %#v\n", config.HTTPCacheDir)
 	glog.Infof("http-cache-size: %#v\n", config.HTTPCacheSize)
 
-	
 	if len(config.Org) == 0 {
 		glog.Fatalf("--organization is required.")
 	}
--- a/mungegithub/github/github.go
+++ b/mungegithub/github/github.go
@@ -315,6 +315,21 @@ func (config *Config) AddRootFlags(cmd *cobra.Command) {
 // PreExecute will initialize the Config. It MUST be run before the config
 // may be used to get information from Github
 func (config *Config) PreExecute() error {
+	glog.Infof("token: %#v\n", config.Token)
+	glog.Infof("token-file: %#v\n", config.TokenFile)
+	glog.Infof("min-pr-number: %#v\n", config.MinPRNumber)
+	glog.Infof("max-pr-number: %#v\n", config.MaxPRNumber)
+	glog.Infof("dry-run: %#v\n", config.DryRun)
+	glog.Infof("organization: %#v\n", config.Org)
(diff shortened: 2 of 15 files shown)


foxish:
@apelisse @lavalamp PTAL

apelisse:
That seems good to me overall.

I'm not very familiar with the "flag" library. Looks like right now all these flags are not "mandatory" but have an empty default now (still optional). Shouldn't we make it mandatory? If you don't specify a project, it shouldn't try to run with empty project, but fail right-away.

foxish:
@apelisse I can't seem to find any such functions that work in the cobra library. [MarkFlagRequired](https://github.com/spf13/cobra/blob/master/bash_completions.go#L588) sounded like the right thing, but it doesn't seem to work, and there is no documentation to be found. In the case of project name, there is a check being performed [here](https://github.com/foxish/contrib/blob/50f8dd5afe37e08b4e9e93677a563019cffe7a1e/mungegithub/github/github.go#L332). But I think you are right in that some places don't have enough sanity checks. I'll add some more error checking code.

apelisse:
@foxish Yes, after looking into it quickly, there doesn't seem to be an idiomatic way of doing it.
I think that checking like it is done there is the way to go.

Thanks for doing it :-)

foxish:
@apelisse, Added some sanity checking. PTAL.

apelisse:
@foxish LGTM

foxish:
The flags need to be added to [cherrypick/deployment.yaml](https://github.com/kubernetes/contrib/blob/0caf0a1fc593296caa40052770cb07b3f76d3114/mungegithub/cherrypick/deployment.yaml) and [shame_entrypoint.sh](https://github.com/kubernetes/contrib/blob/0caf0a1fc593296caa40052770cb07b3f76d3114/mungegithub/shame_entrypoint.sh). @lavalamp, are there any other places that invoke the mungegithub binary?

lavalamp:
@eparis?

eparis:
I think I'd love a patch to cobra/pflag which actually makes `MarkFlagRequired` fail in some way. In any case, the only callers I know are `cherrypick/deployment.yaml`, `submit-queue/deployment.yaml`, and the `shame mailer`.

mungegithub/util/functions.go
```diff
@@ -0,0 +1,27 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+// CleanStringSlice will take a string slice with a single empty value and return an empty slice.
+// This is extremely useful for StringSlice flags, so the user can do --flag="" and instead
+// of getting []string{""} they will get []string{}
+func CleanStringSlice(in []string) []string {
```
eparis: Any thoughts if this function belongs in pflag instead of in our code?

mungegithub/submit-queue/deployment.yaml
```diff
@@ -14,13 +14,34 @@ spec:
...
-        - --number-of-old-test-results=5
         - --http-cache-dir=/cache/httpcache
+        - --path-label-config=path-label.txt
+        - --block-path-config=block-path.yaml
+        - --path-label-config=path-label.txt
+        - --kubernetes-dir=/gitrepos/kubernetes
         - --test-owners-csv=/gitrepos/kubernetes/test/test_owners.csv
+        - --number-of-old-test-results=5
+        - --fixes-issue-reassign=false
+        - --blunderbuss-reassign=false
```
eparis: do these really need set? Do we need to set the defaults?

mungegithub/submit-queue/deployment.yaml
```diff
@@ -14,13 +14,34 @@ spec:
...
+        - --generated-files-config=generated-files.txt
+        - --nonblocking-jenkins-jobs=kubernetes-e2e-gke-staging,kubernetes-e2e-gke-staging-parallel,kubernetes-e2e-gce-serial,kubernetes-e2e-gke-serial,kubernetes-e2e-gke-test,kubernetes-e2e-gce-examples,kubernetes-e2e-gce-federation,kubernetes-e2e-gce-scalability,kubernetes-soak-continuous-e2e-gce,kubernetes-soak-continuous-e2e-gke
+        - --jenkins-jobs=kubelet-gce-e2e-ci,kubernetes-build,kubernetes-test-go,kubernetes-verify-master,kubernetes-e2e-gce,kubernetes-e2e-gce-slow,kubernetes-e2e-gke,kubernetes-e2e-gke-slow,kubernetes-kubemark-5-gce
+        - --presubmit-jobs=kubernetes-pull-build-test-e2e-gce
+        - --weak-stable-jobs=kubernetes-kubemark-500-gce
+        - --required-retest-contexts="Jenkins GCE e2e","Jenkins unit/integration","Jenkins verification","Jenkins GCE Node e2e"
+        - --fake-e2e=false
```
eparis: Do we need this default?

mungegithub/mungers/submit-queue.go
```diff
@@ -359,13 +350,23 @@ func (sq *SubmitQueue) internalInitialize(config *github.Config, features *featu
...
+	glog.V(10).Infof("jenkins-jobs: %#v\n", sq.BlockingJobNames)
+	glog.V(10).Infof("nonblocking-jenkins-jobs: %#v\n", sq.NonBlockingJobNames)
+	glog.V(10).Infof("presubmit-jobs: %#v\n", sq.PresubmitJobNames)
+	glog.V(10).Infof("weak-stable-jobs: %#v\n", sq.WeakStableJobNames)
+	glog.V(10).Infof("required-contexts: %#v\n", sq.RequiredStatusContexts)
+	glog.V(10).Infof("required-retest-contexts: %#v\n", sq.RequiredRetestContexts)
+	glog.V(10).Infof("do-not-merge-milestones: %#v\n", sq.doNotMergeMilestones)
+	glog.V(10).Infof("admin-port: %#v\n", sq.adminPort)
+	glog.V(10).Infof("retest-body: %#v\n", sq.retestBody)
+	glog.V(10).Infof("fake-e2e: %#v\n", sq.FakeE2E)
```
eparis: To be honest, I think I'd rather these just be `glog.Infof()` instead of `V(10)`...

mungegithub/submit-queue/deployment.yaml
foxish: I was trying to bring out and make explicit all the command-line options that we might need to change as we run on other repositories. I think relying on defaults in code could make it a little difficult to understand behavior later.

mungegithub/mungers/submit-queue.go
foxish: That makes sense, and it only prints once in the beginning anyway. I'll make that change.

mungegithub/submit-queue/deployment.yaml
foxish: I was unsure what `fake-e2e` was used for. It looks like it's limited to testing only, I'll get rid of it.

eparis: --fake-e2e was a poor man way to make `go test` work.

foxish: Done.

mungegithub/mungers/submit-queue.go
foxish: Done.

foxish:
@lavalamp, PTAL. I've addressed comments by @eparis. If it's all ok, I'd like to get this merged the next time we push.

mungegithub/features/repo-updates.go
```diff
@@ -167,7 +168,7 @@ func (o *RepoInfo) EachLoop() error {
 
 // AddFlags will add any request flags to the cobra `cmd`
 func (o *RepoInfo) AddFlags(cmd *cobra.Command) {
-	cmd.Flags().StringVar(&o.kubernetesDir, "kubernetes-dir", "./gitrepos/kubernetes", "Path to git checkout of kubernetes tree")
+	cmd.Flags().StringVar(&o.kubernetesDir, "kubernetes-dir", "", "Path to git checkout of kubernetes tree")
```
lavalamp: Maybe in a follow-up we can change this variable name to repoDir?

mungegithub/github/github.go
```diff
@@ -315,6 +315,20 @@ func (config *Config) AddRootFlags(cmd *cobra.Command) {
 // PreExecute will initialize the Config. It MUST be run before the config
 // may be used to get information from Github
 func (config *Config) PreExecute() error {
+	glog.Infof("token: %#v\n", config.Token)
```
lavalamp: Maybe in a follow up you can convert the config struct to yaml and print that. Harder for people to forget to add a line here when they add an option?

mungegithub/features/repo-updates.go
foxish: Yes, that PR is in flight as well. :) https://github.com/kubernetes/contrib/pull/1394/

mungegithub/github/github.go
foxish: That makes sense. Will do in a follow-up.

PR #1386, opened 2016-07-18
Remove non-blocking builds stability % from SQ

apelisse:
The stability of non-blocking builds is always 100% because they are
never blocking the submit-queue. As this is useless, let's just not
display it.


lavalamp:
LGTM

PR #1387, opened 2016-07-18
Run close-stale-pr in mungegithub

apelisse:
Fixed deployment.yaml (and removed it from Makefile where it is not used ...)

### DIFF
--- a/mungegithub/Makefile
+++ b/mungegithub/Makefile
@@ -17,7 +17,7 @@ CACHEDIR ?= /tmp/mungegithub-cache
 WWW ?= $(PWD)/$(APP)/www
 
 ifeq ($(APP),submit-queue)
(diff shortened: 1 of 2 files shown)


lavalamp:
lgtm, sorry, should have caught it in review the first time :(

eparis:
The `make_local_dryrun` in the Makefile is awkward at best.  Should we get rid of it? Or should we move the mungers to be described in the Makefile entirely, instead of in `deployment.yaml` ?

apelisse:
@eparis I think I agree. It should be at most at one place. And it's easier to move them from Makefile to deployment.xml than the other way around (especially as we already do it for image name).

On the other hand, I don't completely understand why we have this feature in the first place. I pushed a new munger today, and forgot to declare it again. Why don't we just have *all* by default?

PR #1394, opened 2016-07-19
Removing the use of gitrepo volume.

foxish:
Issue #1304 

The gitrepo volume is difficult to parameterize as we move towards configmaps to deploy mungegithub to different repositories. This change uses an emptyDir instead for the /gitrepos mount, and moves the repository cloning logic to within mungegithub.

### DIFF
--- a/mungegithub/features/repo-updates.go
+++ b/mungegithub/features/repo-updates.go
@@ -20,9 +20,11 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"path"
 	"path/filepath"
 	"strings"
 
+	"k8s.io/contrib/mungegithub/github"
 	"k8s.io/kubernetes/pkg/util/sets"
 	"k8s.io/kubernetes/pkg/util/yaml"
 
@@ -43,10 +45,12 @@ type assignmentConfig struct {
 
 // RepoInfo provides information about users in OWNERS files in a git repo
 type RepoInfo struct {
-	enabled       bool
-	kubernetesDir string
-	assignees     map[string]sets.String
-	//owners     map[string]sets.String
+	enabled    bool
+	projectDir string
+	baseDir    string
+	assignees  map[string]sets.String
+	config     *github.Config
+	//owners   map[string]sets.String
 }
 
 func init() {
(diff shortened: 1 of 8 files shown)


foxish:
@apelisse @eparis @lavalamp PTAL

apelisse:
What is wrong with the volume?
We already have all the options in `deployment.yaml` along with the volume, it should be just a matter of changing this single file?

foxish:
@apelisse The gitrepos feature previously relied on being given a checked out copy of the repository which it [updated in every iteration](https://github.com/foxish/contrib/blob/0caf0a1fc593296caa40052770cb07b3f76d3114/mungegithub/features/repo-updates.go#L114). This was done by the gitRepo volume, which did the work of cloning the master branch of kubernetes. Now we just give it an emptyDir and we perform cloning of the repository in code [here](https://github.com/foxish/contrib/blob/0caf0a1fc593296caa40052770cb07b3f76d3114/mungegithub/features/repo-updates.go#L156) instead. It uses the --project and --organization values that we supply to find the right repository to clone. 

Does that make sense?

apelisse:
Yeah I guess it's the same in the end, but we don't have to repeat the org and project in the deployment file. Cool LGTM

mungegithub/features/repo-updates.go
```diff
@@ -26,8 +25,11 @@ import (
 	"k8s.io/kubernetes/pkg/util/sets"
 	"k8s.io/kubernetes/pkg/util/yaml"
 
+	"fmt"
```
lavalamp: "fmt" should be in the top section

mungegithub/features/repo-updates.go
```diff
@@ -26,8 +25,11 @@ import (
...
 	"github.com/golang/glog"
 	"github.com/spf13/cobra"
+	"k8s.io/contrib/mungegithub/github"
+	"path"
```
lavalamp: as should path-- the three blocks are go std library imports, in-tree imports, out-of-tree (vendored) imports.

lavalamp:
@eparis can you spot-check?

mungegithub/features/repo-updates.go
foxish: Done.

mungegithub/features/repo-updates.go
foxish: Done.

eparis:
(have not looked at code yet) what does this get us? As long as kubernetes provides a `gitRepo` volume type if it is insufficient we should fix that type. no?

foxish:
The `gitRepo` volume type is not very easy to parameterize for deployments scaling to multiple repositories using [configMaps](https://github.com/kubernetes/contrib/pull/1395). It would need script hackery to replace the repository URL and branch parameters that a `gitRepo` volume takes, for each deployment. This way, it clones the right repository based on the project and organization flags that we supply to it anyway.

lavalamp:
@eparis, the goal here is to be able to parameterize by config map, and keep the same deployment file for each instance of the submit queue.

@foxish I think this will work and we can merge it, but when we talked I was actually trying to say that we should implement this as a side-car container. I.e., make a completely separate container image that does the syncing. That would be potentially useful to multiple people. Anyway, not a big deal at the moment.

LGTM

PR #1395, opened 2016-07-19
Deploying with configmaps

foxish:
Issue #1304 

This is the proposed structure and flow of how we will deploy mungegithub to multiple repositories with configmaps. The directory structure of contrib/mungegithub/submit-queue/ will look like the following:
```
.
├── deployment
│   ├── k8s-contrib
│   │   ├── configmap.yaml
│   │   ├── pvc.yaml
│   │   ├── pv.yaml
│   │   ├── secret.yaml
│   │   └── service.yaml
│   ├── k8s-kubernetes
│   │   ├── configmap.yaml
│   │   ├── pvc.yaml
│   │   ├── pv.yaml
│   │   ├── secret.yaml
│   │   └── service.yaml
│   └── k8s-docs
│   │   ├── configmap.yaml
│   │   ├── pvc.yaml
│   │   ├── pv.yaml
│   │   ├── secret.yaml
│   │   └── service.yaml
├── deployment.yaml
└── www
```

The makefile's deploy target would take an additional argument, the `CONFIGMAP`.  The deployment command for example, for k8s-kubernetes, the main repository like:

```
REPO=<> APP=submit-queue CONFIGMAP=k8s-kubernetes make deploy 
```

make deploy will work after one:
- creates the corresponding secret.
- creates PVC & PV if necessary.


foxish:
@lavalamp @eparis PTAL and let me know if I'm missing something. This PR is incomplete and is waiting for [1385](https://github.com/kubernetes/contrib/pull/1385) and [1394](https://github.com/kubernetes/contrib/pull/1394) to be merged first. For now, I'd like to get opinions on whether this is a good way to go.

eparis:
Just to make sure I understand it owuld be possible for me to run `APP=cherrypick CONFIGMAP=k8s-contrib` and that combination be impossible (because there is no `contrib/mungeighutb/cherrypick/k8s-contrib` directory?  (that's fine, just want to make sure I understand)

Auto-creation of the `secret`, `pv`, etc is going to be hard, but a (maybe) good goal...

foxish:
My thoughts were to create those directories and the required files on the cherrypick/ after we have finalized on the structure for submit-queue/. The file currently in the PR are definitely a WIP, but yes, the eventual goal would be make it possible to run the same against both `APP=cherrypick` and `APP=submit-queue`. I do not know yet if we want to turn up cherry-pick against the other repos right away. @lavalamp may know better.

eparis:
thus far, no, we shouldn't be turning up cherrypick. But maybe, someone will want to eventually.  Maybe.

foxish:
Okay. Then it would just involve restructuring the existing cherrypick/ directory with just a single `k8s-kubernetes`, and as you said, running `APP=cherrypick CONFIGMAP=k8s-contrib` wouldn't be possible.

foxish:
Looking through the way ingress resources work, I got to https://github.com/kubernetes/kubernetes/issues/17088. According to that, we are running the instances of the submit-queue in different namespaces, to point to each of those services directly from an ingress resource wouldn't be possible. Tim's [comment](https://github.com/kubernetes/kubernetes/issues/17088#issuecomment-157187876) appears to be one approach to make it work. Is there some other approach I am missing? Would creating a service pointing to another service make sense to make it work with ingress?

cc @lavalamp @bgrant0607

bgrant0607:
@foxish The main reason to run in different namespaces was to avoid substitution in the manifest files. However, it looks like substitution is being done anyway. Given that, it may be easier to run them all in the same namespace.

foxish:
Running all the services/deployments in the same namespace. 
There is quite a bit of string substitution in there and some of it could be eliminated when we get the binary to consume YAML directly rather than using environment variables and command-line arguments.  
I've tested the changes on my local cluster with ingress and it works as expected when turned up on contrib and docs repos.

PR #1399, opened 2016-07-20
[cluster-autoscaler] Add leader election to cluster autoscaler

fgrzadkowski:
Please note that this changes how we can pass flags to CA. In particular `-v` is incorrect and should be `--v`. Command lines in kubernetes scripts will have to be updated when releasing CA.

https://github.com/kubernetes/kubernetes/issues/21124

@jszczepkowski


fgrzadkowski:
@mwielgus Can you please take a look?

cluster-autoscaler/cluster_autoscaler.go
```diff
@@ -72,31 +75,29 @@ var (
...
+}
+
+func createEventRecorder(kubeClient *kube_client.Client) kube_record.EventRecorder {
+	eventBroadcaster := kube_record.NewBroadcaster()
+	eventBroadcaster.StartLogging(glog.Infof)
+	eventBroadcaster.StartRecordingToSink(kubeClient.Events(""))
+	return eventBroadcaster.NewRecorder(kube_api.EventSource{Component: "cluster-autoscaler"})
+}
+
+func run(_ <-chan struct{}) {
```
mwielgus: what is the purpose of the argument

mwielgus:
Just add comment to run. Otherwise LGTM

fgrzadkowski:
Added comment.

PR #1400, opened 2016-07-20
Cluster-autoscaler: shuffle nodes in scale down

mwielgus:
So that candidate nodes use different nodes to move their pods. So that removal of one nodes doesn't reset utilization counters on all other pods. Please take a look at #1341 #1370 for more context. 

### DIFF
--- a/cluster-autoscaler/simulator/cluster_test.go
+++ b/cluster-autoscaler/simulator/cluster_test.go
@@ -134,3 +134,19 @@ func TestFindNone(t *testing.T) {
 		time.Now())
 	assert.NoError(t, err)
 }
+
+func TestShuffleNodes(t *testing.T) {
+	nodes := []*kube_api.Node{
+		BuildTestNode("n1", 0, 0),
+		BuildTestNode("n2", 0, 0),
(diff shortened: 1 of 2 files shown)


cluster-autoscaler/simulator/cluster_test.go
```diff
@@ -134,3 +134,19 @@ func TestFindNone(t *testing.T) {
...
 	assert.NoError(t, err)
 }
+
+func TestShuffleNodes(t *testing.T) {
+	nodes := []*kube_api.Node{
+		BuildTestNode("n1", 0, 0),
+		BuildTestNode("n2", 0, 0),
+		BuildTestNode("n3", 0, 0)}
+	gotPermutation := false
+	for i := 0; i < 10000; i++ {
```
piosz: I can see potential flakiness here.

piosz:
LGTM

cluster-autoscaler/simulator/cluster_test.go
mwielgus: Yeah. If random is random then the chance of flakiness is (5/6)^10000. That is approximately 1e-800 %.

PR #1404, opened 2016-07-20
cross namespace tls secrets

gambol99:
- adding an extra commad line option --cross-permitted which is a comma separated list of cross namespace secrets a ingress resource can request. This permits us to reuse certificates from one namespace across others. A use case being keeping a wildcard certificate in one namespace, but allow others to use it.
- shifting the tag to version 0.8.2
- fixed the bug in the ssl.go, should not use temporaryPemFile.Name() on error, it will be nil

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1404)

### DIFF
--- a/hack/verify-flags/known-flags.txt
+++ b/hack/verify-flags/known-flags.txt
@@ -10,6 +10,7 @@ cloud-config
 cloud-provider
 cluster-uid
 configuration-name
+cross-permitted
 current-release-pr
 custom-error-service
 default-backend-service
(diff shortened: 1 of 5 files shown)


hack/verify-flags/known-flags.txt
```diff
@@ -10,6 +10,7 @@ cloud-config
 cloud-provider
 cluster-uid
 configuration-name
+cross-permitted
```
gambol99: admittedly perhaps a better name?

aledbf:
> adding an extra commad line option --cross-permitted which is a comma separated list of cross namespace secrets a ingress resource can request.

@gambol99 I think this should be changes in the Ingress spec, not in a ingress controller. With this change is not possible to switch implementations (like GCE)

ping @bprashanth

PR #1412, opened 2016-07-20
Update go-github for mungegithub

apelisse:
Update to a new version, that will allow us to use Assignee**s**.
Because go-github made breaking changes to their API, the diff is kind of ugly. I've tried to carefully replace everywhere (thanks to go compiler). `make test` is passing for me.

### DIFF
--- a/mungegithub/Godeps/Godeps.json
+++ b/mungegithub/Godeps/Godeps.json
@@ -34,7 +34,7 @@
 		},
 		{
 			"ImportPath": "github.com/google/go-github/github",
-			"Rev": "63fbbb283ce4913a5ac1b6de7abae50dbf594a04"
+			"Rev": "b5e5babef39c18002f177a134fc49dc5013374ba"
 		},
 		{
 			"ImportPath": "github.com/google/go-querystring/query",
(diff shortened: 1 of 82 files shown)


apelisse:
cc @bgrant0607

bgrant0607:
@apelisse For future reference, 2 separate commits, one to update the godep and one for the fixes to make it work, would have been much easier to review.

Looks pretty mechanical, but I don't know if any deep copies are needed.

Anyway, LGTM, AFAICT.

apelisse:
Thanks for the review.
I can't convince myself that I should commit broken code ...

bgrant0607:
@apelisse How is this broken? Or do you mean that the other PR is broken?

apelisse:
Nothing is broken. I was referring to the "2 seperate commits". If I split the commit in two, one of the commits would be broken.

bgrant0607:
@apelisse You could always squash before merging.

apelisse:
@bgrant0607 OK I'll do it like this next time!

lavalamp:
Please split this into two commits. It's OK that they won't compile independently. It's more important to enable reviewers to see the changes made in our code separate from the dep changes.

apelisse:
@lavalamp done

lavalamp:
Looks safe. LGTM

PR #1414, opened 2016-07-21
Parameterized submit-queue health/status pages.

foxish:
Issue #1304 

Mungegithub generates several status pages showing queue health and PR merge queue status.
The URLs in these pages, as well as some contents need to be parameterizable as we run against other repositories.

### DIFF
--- a/mungegithub/submit-queue/www/index.html
+++ b/mungegithub/submit-queue/www/index.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en">
+<html lang="en" ng-app="SubmitQueueModule" ng-controller="SQCntl as cntl">
 
 <head>
   <link rel="stylesheet" href="https://cdn.gitcdn.link/cdn/angular/bower-material/v1.0.7/angular-material.min.css" type="text/css">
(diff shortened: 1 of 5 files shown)


foxish:
@eparis @apelisse PTAL

apelisse:
I think we've reached the point (maybe a long time ago) where we clearly have too many options. Should we consider making it a config file?

mungegithub/submit-queue/www/index.html
```diff
@@ -168,10 +168,10 @@ <h2 class="md-title" ng-show="cntl.testResults.nonBlockingBuilds.length > 0">Non
               <br>
               <br>
               <h2 class="md-title" ng-show="cntl.OverallHealth.length > 0">Overall Health: {{ cntl.OverallHealth }}</h2>
-              <p>Health percents are the fraction of the time that a given job is stable for the last day, or since the submit queue restarted ({{ cntl.sqStats.StartTime | date:'medium'}})</span>, whichever is shorter. The <a href="http://storage.googleapis.com/kubernetes-test-history/static/index.html"><strong>24-Hour Test Report</strong></a> shows more detail, along with a list of flaky and broken tests in merge-blocking jobs.
+              <p>Health percents are the fraction of the time that a given job is stable for the last day, or since the submit queue restarted ({{ cntl.sqStats.StartTime | date:'medium'}})</span>, whichever is shorter. The <a ng-href="{{cntl.metadata.HistoryUrl}}"><strong>24-Hour Test Report</strong></a> shows more detail, along with a list of flaky and broken tests in merge-blocking jobs.
               </p>
               <p>
-                <img src="http://storage.googleapis.com/kubernetes-test-history/k8s-queue-health.svg" style="max-width: 100%"/>
```
apelisse: I think it's good we remove this hardcode.

apelisse:
Otherwise, LGTM

foxish:
@apelisse I agree. We should definitely make config files out of some of these options as they are getting unwieldy to manage. It would have to work well with the configmaps and the new method of deployment as well, which are in the pipeline. I added an issue for it and will work on it after we get the submit-queue deployed.

lavalamp:
LGTM

PR #1420, opened 2016-07-22
mungegithub: Fix comment regex to NOT repost

apelisse:
The regex to find that we have already added a warning is broken
(because of a change in punctuation). This is causing the mungebot to
repost the warning every time.

Fixing the regex will remove previous comments and prevent the bot from
posting too much.


lavalamp:
LGTM

PR #1427, opened 2016-07-23
[WIP] DNS nanny - Autoscale replication controller based on number of nodes

girishkalele:
Adaption of the addon-resizer nanny worked out pretty well.


girishkalele:
cc @thockin @roberthbailey

roberthbailey:
@Q-Lee since you wrote the addon-resizer can you take a look at this?

girishkalele:
@Q-Lee 

I think the failsafe checks need a little change, instead of continuing with the old params, I think it will be more correct that if we fail to parse the config, we log.Fatal and just bail out of the container.
This will immediately surface to the operator that the configmap is not formatted correctly.

fgrzadkowski:
Why aren't we using `addon-resizer` for this?

fgrzadkowski:
If we decide to write something new, please don't use a generic name like `pod-autoscaler`. It suggests that it's the recommended way for doing things, and I believe we all agree it's not. If this is intended to only solve dns problem, then maybe `dns-nanny` or `dns-resizer`?

girishkalele:
@fgrzadkowski 

The existing addon-resizer works so differently - it doesn't resize replicas, but rather the cpu and memory limits of the pod itself and it works on pods and deployments, not on RCs - that it would have been pretty much two different functionalities in the same binary with more complicated flags than to make a new one for DNS RC replica resizing.

I will change the name to be DNS specific (although the new one is written as a generic sidecar container that can resize any RC it belongs to, so it can be dropped into any RC and it will auto-scale it based on the configmap params)

dns-rc-autoscaler/README.md
```diff
@@ -0,0 +1,88 @@
+# pod-autoscaler
+
+This container image watches over the number of schedulable nodes in the cluster and resizes 
+the number of replicas in its parent object. Works only for pods which are children of RCs or RSs objects.
```
thockin: a short explanation of HOW would be good.

dns-rc-autoscaler/nanny/kubernetes_client.go
```diff
@@ -0,0 +1,137 @@
...
+
+	api "k8s.io/kubernetes/pkg/api"
+	client "k8s.io/kubernetes/pkg/client/clientset_generated/release_1_3"
+)
+
+const replicationControllerKind = "ReplicationController"
+const createdByAnnotationName = "kubernetes.io/created-by"
+
+type kubernetesClient struct {
+	namespace, pod string
```
thockin: for clarity, please do these on separate lines.

thockin:
You're going to have to break this into multiple commits.  One for your
code and another for deps, at least.  Github basically can't load this.

As a name, can I propose "horizontal-self-scaler" ?

dns-rc-autoscaler/nanny/kubernetes_client.go
```diff
@@ -0,0 +1,137 @@
...
+//    kubernetes.io/created-by: |
+//      {"kind":"SerializedReference","apiVersion":"v1","reference":
+//      {"kind":"ReplicationController","namespace":"default",
+//      "name":"pod-autoscaler-nanny",
+//      "uid":"7ccf55b9-5073-11e6-8321-42010af00002",
+//      "apiVersion":"v1",
+//      "resourceVersion":"20339"}}
+//
+
+func (k *kubernetesClient) GetParentRc(namespace, podname string) (string, error) {
```
thockin: General note: if you give a function a public name (leading capital letter) you had better have a GoDoc comment explaining it.  Given that this is all one package (which is fine) I doubt this needs a public name.  Same elsewhere.

dns-rc-autoscaler/nanny/kubernetes_client.go
```diff
@@ -0,0 +1,137 @@
...
+	if !ok || len(createdBy) == 0 {
+		// Empty or missing created-by annotation would mean this is a
+		// standalone pod and does not have an RC/RS managing it
+		return "", fmt.Errorf("Pod does not have a created-by annotation")
+	}
+	type RCInner struct {
+		Kind string `json:"kind"`
+		Name string `json:"name"`
+	}
+	type AnnotationT struct {
```
thockin: The trailing T here is not very idiomatic Go

dns-rc-autoscaler/nanny/kubernetes_client.go
```diff
@@ -0,0 +1,137 @@
...
+		Reference RCInner `json:"reference"`
+	}
+	var annot AnnotationT
+	err = json.Unmarshal([]byte(createdBy), &annot)
```
thockin: Isn't this annotation already a serialized ObjectReference?  Is that type not exposed by our client?

dns-rc-autoscaler/nanny/kubernetes_client.go
```diff
@@ -0,0 +1,137 @@
...
+	if err != nil {
+		return "", fmt.Errorf("Failed to parse created-by annotation (%s)", err)
+	}
+	if annot.Reference.Kind != "ReplicationController" {
```
thockin: what about ReplicaSet?  Looking at an example, it doesn't look like Deployment puts the CreatedBy annotation on an RS, so you can't go any higher than that.  We will have to add that if we want to make this work with deployments.

@erictune any objections to adding CreaedBy to RSs under Deployments?

dns-rc-autoscaler/nanny/kubernetes_client.go
```diff
@@ -0,0 +1,137 @@
...
+
+	nodes, err := k.clientset.CoreClient.Nodes().List(opt)
+	if err != nil {
+		return 0, 0, err
+	}
+	var schedulableNodes uint64
+	var totalNodes = uint64(len(nodes.Items))
+	for _, node := range nodes.Items {
+		if !node.Spec.Unschedulable {
+			schedulableNodes = schedulableNodes + 1
```
thockin: schedulableNodes++

dns-rc-autoscaler/nanny/kubernetes_client.go
```diff
@@ -0,0 +1,137 @@
...
+func (k *kubernetesClient) PodReplicas() (uint64, error) {
+	replicationController, err := k.GetParentRc(k.namespace, k.pod)
+	if err != nil {
+		return 0, err
+	}
+	rc, err := k.clientset.CoreClient.ReplicationControllers(k.namespace).Get(replicationController)
+	if err != nil {
+		return 0, err
+	}
+	return uint64(*rc.Spec.Replicas), nil
```
thockin: why uint64?  That seems like overkill..

dns-rc-autoscaler/nanny/main/pod_nanny.go
```diff
@@ -0,0 +1,91 @@
...
+	"log"
+	"os"
+	"time"
+
+	flag "github.com/spf13/pflag"
+
+	"k8s.io/contrib/dns-rc-autoscaler/nanny"
+
+	client "k8s.io/kubernetes/pkg/client/clientset_generated/release_1_3"
+	"k8s.io/kubernetes/pkg/client/restclient"
```
thockin: There is a versioned client which should have a LOT less deps...

dns-rc-autoscaler/nanny/main/pod_nanny.go
```diff
@@ -0,0 +1,91 @@
...
+)
+
+var (
+	// Flags to identify the container to nanny.
+	podNamespace = flag.String("namespace", os.Getenv("MY_POD_NAMESPACE"), "The namespace of the ward. This defaults to the nanny pod's own namespace.")
```
thockin: A nit on the comment.  It doesn't default to the self-namespace.  That is done outside of the pod, so the flag should only document the semantics.

Additionally, the comment should mention the env var as a fallback if the flag is not specified.

dns-rc-autoscaler/nanny/main/pod_nanny.go
```diff
@@ -0,0 +1,91 @@
...
+	podName      = flag.String("pod", os.Getenv("MY_POD_NAME"), "The name of the pod to watch. This defaults to the nanny's own pod.")
```
thockin: same

dns-rc-autoscaler/nanny/main/pod_nanny.go
```diff
@@ -0,0 +1,91 @@
...
+	configFile   = flag.String("paramsFile", "", "Params file for the node scaling parameters")
+	verbose      = flag.Bool("verbose", false, "Turn on verbose logging to stdout")
+	// Flags to control runtime behavior.
+	pollPeriod = time.Millisecond * time.Duration(*flag.Int("poll-period", 10000, "The time, in milliseconds, to poll the dependent container."))
```
thockin: either make this a time.Duration flag or put the units in the name (poll-period-seconds).  Also could better document what this is polling for.

dns-rc-autoscaler/nanny/nanny_lib.go
```diff
@@ -0,0 +1,90 @@
...
+	CountNodes() (uint64, uint64, error)
+	PodReplicas() (uint64, error)
+	UpdateReplicas(uint64) error
+	GetParentRc(namespace, podname string) (string, error)
+}
+
+// PollAPIServer periodically counts the number of nodes, estimates the expected
+// number of replicas, compares them to the actual replicas, and
+// updates the parent ReplicationController with the expected replicas if necessary.
+func PollAPIServer(k8s KubernetesClient, scaler Scaler, pollPeriod time.Duration, verbose bool) {
```
thockin: why poll?  Watch should work for this, no?  I guess poll is simpler, if less common.

thockin: Let me know when you're ready for another round.  Overall this is really
cool. I could see putting this in a whole new repo.  Or we could make one
repo for the various side-car apps we are building.

dns-rc-autoscaler/nanny/kubernetes_client.go
erictune: No objection.

erictune:
Comments:

- This is cool.  I think it is something other people could fork and adapt easily to their needs.
- The nanny-as-sidecar-of-thing-it-is-controlling pattern is one I haven't seen before.   Interesting.  
  - Is there any issue running the scaler in the same pod as the thing being scaled?  If the thing being scaled is overloaded, will the nanny be able to run?
- I think @bgrant0607 is trying to keep new things out of contrib.  Brian, where does this belong?

dns-rc-autoscaler/nanny/kubernetes_client.go
```diff
@@ -0,0 +1,137 @@
...
+		if !node.Spec.Unschedulable {
+			schedulableNodes = schedulableNodes + 1
+		}
+	}
+	return totalNodes, schedulableNodes, nil
+}
+
+// Get number of replicas configured in the parent replication controller
+func (k *kubernetesClient) PodReplicas() (uint64, error) {
+	replicationController, err := k.GetParentRc(k.namespace, k.pod)
```
erictune: Looks we are using an RC currently for DNS.  Do we want to use a Deployment sometime in the future? 

If so, how does this work with a  deployment?  Say I change the image.  New RS created.   Now there are two pools of pods with different parent RSes .  So, now the nanny is asking for N pods for each RS.  So the number of DNS pods doubled when I wanted to just start a rolling update of the image?

erictune:
If I understand correctly, every DNS pod will be running this sidecar, and they are all working towards the same scale goal, so no coordination needed?

Probably worth documenting that.

Are there any scenarios where the sidecars might be working towards a different scale goal, and if so, would that cause a "controller battle"?

dns-rc-autoscaler/nanny/kubernetes_client.go
girishkalele: I created a new issue to track this feature request #29922.

I will work on that and drop in the same logic as used for populating the annotation in Pods.

girishkalele:
@erictune 

Will document that.

The central configmap parameters coordinate the scaling goals of all the nanny sidecars. They re-read the configmap every poll interval before making a change so they should have the same scaling goals.

In the case of multiple masters and partitioning, the listing of nodes might be different in each partition, are the API masters in each partition able to service write requests (to scale the RC) ?

dns-rc-autoscaler/nanny/kubernetes_client.go
girishkalele: Good case - haven't tested behavior under a Deployment and rolling update - is there an annotation that a pod can read, either in its own pod object or its parent RS, to find if the parent RS is the previous version that is being scaled down ?

bgrant0607:
@girishkalele 
I would prefer that this go into its own repo. If you want to merge this first here and then move it, I'm fine with that. `horizontal-self-scaler` is an acceptable name.

cc @erictune @thockin

erictune:
I like that this uses 10s polling.  I think all new controllers should use polling unless there is a demonstrated performance need for watching.

erictune:
I bet this can be done in 15 lines of shell, using kubectl.

girishkalele:
@bgrant0607 

I will move the changes - new repo would be github.com/kubernetes/horizontal-self-scaler.

(using https://github.com/kubernetes/kubernetes-template-project as the first commit)

Q-Lee:
Is there any chance you can break this into 2 commits (godeps and code)?

dns-rc-autoscaler/nanny/kubernetes_client.go
```diff
@@ -0,0 +1,137 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"log"
+
+	api "k8s.io/kubernetes/pkg/api"
+	client "k8s.io/kubernetes/pkg/client/clientset_generated/release_1_3"
+)
+
+const replicationControllerKind = "ReplicationController"
```
Q-Lee: const block?

dns-rc-autoscaler/nanny/kubernetes_client.go
Q-Lee: Do what one separate lines?

dns-rc-autoscaler/nanny/kubernetes_client.go
Q-Lee: This is [unfortunately] an int32 in the API. I wouldn't blame you if you make it a uint32, but a uint64 only introduces the possibility of error.

dns-rc-autoscaler/nanny/kubernetes_client.go
```diff
@@ -0,0 +1,137 @@
...
+	// First, get the parent replication controller.
+	replicationController, err := k.GetParentRc(k.namespace, k.pod)
+	if err != nil {
+		return err
+	}
+	rc, err := k.clientset.CoreClient.ReplicationControllers(k.namespace).Get(replicationController)
+	if err != nil {
+		return err
+	}
+	*rc.Spec.Replicas = int32(replicas)
```
Q-Lee: Change to int32?

dns-rc-autoscaler/nanny/kubernetes_client.go
```diff
@@ -0,0 +1,137 @@
...
+	if *rc.Spec.Replicas == 0 {
+		log.Fatalf("Cannot update to 0 replicas")
```
Q-Lee: Can't you validate this at start up?

dns-rc-autoscaler/nanny/main/pod_nanny.go
Q-Lee: Last week I think I saw the addon-resizer ignoring this flag. Can you confirm that this actually works?

dns-rc-autoscaler/nanny/main/pod_nanny.go
```diff
@@ -0,0 +1,91 @@
...
+	// Log all sanity check errors before returning a single error string
+	if errorsFound {
+		return errors.New("Failed to validate all input parameters")
+	}
+	return nil
+}
+
+func main() {
+	// First log our starting config, and then set up.
+	log.Printf("Invoked by %v\n", os.Args)
```
Q-Lee: Does the log need to be initialized anywhere?

dns-rc-autoscaler/nanny/main/pod_nanny.go
```diff
@@ -0,0 +1,91 @@
...
+	flag.Parse()
+	// Perform further validation of flags.
+	if err := sanityCheckParametersAndEnvironment(); err != nil {
+		log.Fatal(err)
+	}
+	// Set up work objects.
+	config, err := restclient.InClusterConfig()
```
Q-Lee: The addon-resizer had some memory usage issues awhile back. I'm going to guess this was derserializing the NodeList. What does it take to use protos?

dns-rc-autoscaler/nanny/main/pod_nanny.go
```diff
@@ -0,0 +1,91 @@
...
+	if err != nil {
+		log.Fatal(err)
+	}
+	clientset, err := client.NewForConfig(config)
+	if err != nil {
+		log.Fatal(err)
+	}
+	k8s := nanny.NewKubernetesClient(*podNamespace, *podName, clientset)
+	log.Printf("Looking for parent/owner of pod %s/%s\n", *podNamespace, *podName)
```
Q-Lee: Is there a log.Infof?

dns-rc-autoscaler/nanny/nanny_lib.go
Q-Lee: I'm going to guess he polls because the addon-resizer polls. KISS was the recommendation from the API guys, but the client has improved a lot since then.

dns-rc-autoscaler/nanny/nanny_lib.go
```diff
@@ -0,0 +1,90 @@
...
+			time.Sleep(pollPeriod)
+		}
+
+		// Query the apiserver for the number of nodes.
+		total, num, err := k8s.CountNodes()
+		if err != nil {
+			log.Println(err)
+			continue
+		}
+		if verbose {
```
Q-Lee: Refactor the conditional into a function.

dns-rc-autoscaler/nanny/nanny_lib.go
```diff
@@ -0,0 +1,90 @@
...
+		if replicas == expReplicas {
+			if verbose {
+				log.Println("Replicas are within the expected limits.")
+			}
+			continue
+		}
+		log.Printf("Replicas are not within the expected limits: updating the parent replication controller to %d replicas\n",
+			expReplicas)
+		if err := k8s.UpdateReplicas(expReplicas); err != nil {
+			log.Println(err)
```
Q-Lee: log.Errorf

girishkalele:
@Q-Lee 

Will do...I am in the process of moving it to a new repo, will send out the PR for that one.

dns-rc-autoscaler/nanny/kubernetes_client.go
girishkalele: Yes, this nanny library will also address ReplicaSets->Deployments once we have the Created-By annotation via https://github.com/kubernetes/kubernetes/issues/29922

dns-rc-autoscaler/nanny/main/pod_nanny.go
girishkalele: I moved away from the glog library (that addon-resizer uses) to the standard log package to reduce the command line flags (glog injects quite a few command line flags) and dependencies.

Log package defaults to stdout so all logs show up in container logs.

dns-rc-autoscaler/nanny/kubernetes_client.go
thockin: It's OK for now to just document that this only works with RCs, and then come back to this after it is landed.

dns-rc-autoscaler/nanny/main/pod_nanny.go
girishkalele: Good point - IIUC, you suggest we iterate over the protobuf response, count the number of nodes and discard without memory intensive deserialization ? I will investigate feasibility.

dns-rc-autoscaler/nanny/scaler.go
```diff
@@ -0,0 +1,106 @@
...
+	ConfigFile string
+	Verbose    bool
+}
+
+const paramsSeparator = "nodes.gte."
+
+func (s Scaler) scaleWithNodes(numCurrentNodes uint64) uint64 {
+	newMap, err := ParseScalerParamsFile(s.ConfigFile)
+	if err != nil {
+		log.Fatalf("Parse failure: The configmap volume file is malformed (%s)\n", err)
```
Q-Lee: Can you validate this at start up?

dns-rc-autoscaler/nanny/scaler.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	// construct a search ladder from the map
```
Q-Lee: If you do linear search, then just use the map. If you're going to sort, then do binary search. If you spend O(nlogn) building a searchable data structure, then do it during initialization and amortize over multiple queries.

thockin: As much as I love optimiziation, this is small-n, so linear will almost
always be faster.

dns-rc-autoscaler/nanny/scaler.go
```diff
@@ -0,0 +1,106 @@
...
+		keys = append(keys, int(k))
+	}
+	sort.Ints(keys)
+	var neededReplicas uint64 = 1
+
+	// Walk the search ladder to get the correct number of replicas
+	// for the current number of nodes
+	for _, nodeCount := range keys {
+		replicas := newMap[uint64(nodeCount)]
+		if uint64(nodeCount) > numCurrentNodes {
```
Q-Lee: If you keep a max that is <= numCurrentNodes, then you can do a linear scan without sorting (O(n) time). If you keep the sorted list, then you can bisect the list (O(log(n)).

dns-rc-autoscaler/nanny/scaler.go
```diff
@@ -0,0 +1,106 @@
...
+// Scaler determines the number of replicas to run
+type Scaler struct {
+	ConfigFile string
+	Verbose    bool
+}
+
+const paramsSeparator = "nodes.gte."
+
+func (s Scaler) scaleWithNodes(numCurrentNodes uint64) uint64 {
+	newMap, err := ParseScalerParamsFile(s.ConfigFile)
```
Q-Lee: Can this just be parsed at start up? The resulting map could then be verified and kept as a member.

dns-rc-autoscaler/nanny/scaler.go
```diff
@@ -0,0 +1,106 @@
...
+		}
+		neededReplicas = replicas
+	}
+	return neededReplicas
+}
+
+// ParseScalerParamsFile Parse the scaler params file every time around, since it is a ConfigMap volume.
+func ParseScalerParamsFile(filename string) (map[uint64]uint64, error) {
+	NodeToReplicasMap := make(map[uint64]uint64)
+	if _, err := os.Stat(filename); err != nil {
```
Q-Lee: Why not just call os.Open? It should return an error if the file does not exist, and if you rely on Stat => Open, then you could fall victim to time of check vs time of read bugs.

dns-rc-autoscaler/nanny/scaler.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	f, err := os.Open(filename)
+	if err != nil {
+		return nil, err
+	}
+	defer f.Close()
+	scanner := bufio.NewScanner(f)
+	var n = 1
+	for scanner.Scan() {
```
Q-Lee: For small files, just use https://golang.org/pkg/io/ioutil/#ReadFile

dns-rc-autoscaler/nanny/scaler.go
```diff
@@ -0,0 +1,106 @@
...
+		l := scanner.Text()
+		// Each line should be of the form nodes.gte.<N>=<M>
+		parts := strings.Split(l, "=")
+		if len(parts) < 2 {
+			return nil, fmt.Errorf("Error parsing line %d in params file %s", n, filename)
+		}
+		tokens := strings.Split(parts[0], paramsSeparator)
```
Q-Lee: Use a standard config language, and let a library parse for you. It looks like the standard library understands yaml: https://godoc.org/gopkg.in/yaml.v2

dns-rc-autoscaler/nanny/kubernetes_client.go
girishkalele: Yes, it is - moving to SerializedReference.

erictune:
@mml this PR is trying to solve a problem that you have thought about: given a pod, find its controller (you in the context of disruption budget).  You might have useful comments.

erictune:
@bgrant0607 said not to use `created-by`, and to prefer to use  Owner References. (which I think is `GetOwnerReferences` from `pkg/api/meta/meta.go`.  @lavalamp and @caesarxuchao should know about Owner References.

dns-rc-autoscaler/nanny/main/pod_nanny.go
Q-Lee: Nothing that complex. We're currently using JSON, but we can optionally speak protobufs over the wire. The protobuf protocol is a little more succinct, so I'm hoping it's less memory intensive.

dns-rc-autoscaler/nanny/scaler.go
Q-Lee: Yes, but he's doing a sort (so he doesn't need to keep track of a max). I'm fairly ambivalent about linear scan vs binary search when n ~ 2k, thus the conditionals.

Q-Lee:
The addon-resizer was written with the intent to be extended in this fashion if the use case ever arose. It should be straight forward to extend it in this fashion.

lavalamp:
I'm not a fan of doing this from every DNS pod in the cluster.

lavalamp:
We hope owner references will be in beta and on-by-default in 1.4 but it's not certain at this point.

girishkalele:
I created https://github.com/kubernetes/kubernetes/issues/29922 to request addition of the Created-By annotation to ReplicaSets but we can close that if the OwnerReferences will work for going from ReplicaSet -> parent Deployments.

dns-rc-autoscaler/nanny/scaler.go
girishkalele: Yes, will move to YAML and/or JSON.

thockin:
Have a simpler solution?

lavalamp:
Run one sizing thing on the master node. Alternatively, run one of these
pods in its own RS.

Q-Lee:
There should be no assumptions in the code that this container must scale its own pod. Inferring defaults is fine, but I need to be able to explicitly override that behavior. And if owner references are not a sure thing for 1.4, then we may consider dropping the inference altogether in this release.

@thockin - I think dan has a good point. If we extend the addon-resizer to another type of scaling, then we could also change it to scale multiple pods. Then it can consume a yaml object per scale, and a single pod could watch all of our addons.

The only down side is that this pod would require a deployment every time an addon is enabled or disabled.

thockin:
I figured running as a sidecar was cute, but I guess it could just be a
standalone pod that takes a string argument like
"replicationcontrollers/kube-dns-v19" or "deployments/kube-dns".

What do you think Girish?

For the sake of simplicity, do we really want to jam this together with the
existing vertical sizer?  They are pretty different things.

girishkalele:
@thockin That works too - we can add another RC that provides one replica of the horizontal resizer and point it towards the rc/deployment it should be scaling.

The addon-resizer scales the fluentd pod vertically and the DNS use-case is horizontal scaling. My first look at it, I figured combining them would make it overly complex for a single component, and would be better as two independent modules.
@Q-Lee suggested we can combine them into a single all-purpose scaler. Perhaps we can look at the complexity of the configuration to combine them and decide if combining is simpler.

thockin:
I'd say - IF we want to combine them, we do it later.  Let's get something
implemented.  If this just takes an arg, it's even simpler.

dns-rc-autoscaler/nanny/kubernetes_client.go
mikedanese: Also https://github.com/kubernetes/kubernetes/pull/29551 will help a lot when it's complete.

girishkalele:
To summarize a few of the points:

1. Current config file format is ugly, switch to a standard format - switching to JSON. (Embedding yaml strings inside the configmap yaml seems to give me indentation headaches).
2. Use ConfigMap ResourceVersion to optimize refresh logic.
3. Remove created-by and/or OwnerReference magic and point to the target RC/RS/Deployment via cmd line args. _This scaler will run as an RC independent from the controller it is scaling._
4. Implement horizontal scaler independently for now, combine them later if possible.
6. Close this PR and create a PR from the new repo.

girishkalele:
Moved out of contrib repo.

Issue #1389, opened 2016-07-19
Finish OWNERS implementation


bgrant0607:
Described here:
https://github.com/kubernetes/contrib/pull/474#issuecomment-226249631

cc @eparis @davidopp @philips @mike-saparov @hongchaodeng @kubernetes/contributor-experience 

@philips Please add your suggestions here.

bgrant0607:
Copying from #474:

Previous workflow:
* a reviewer, who didn't necessarily have write access (and therefore label access), would be assigned
* the reviewer would eventually type "LGTM" into a comment
* the author would (often) squash commits and/or rebase
* the reviewer would then apply the `lgtm` label if they had label access, or would notify someone in kubernetes-maintainers to do that if they did not; the author could self-apply `lgtm` if they had label access

Proposed workflow:
* a reviewer, who isn't necessarily an approver, would be assigned
* the reviewer would eventually type "LGTM" into a comment
* the author would (often) squash commits and/or rebase
* the reviewer would then type `I approve` if they were an approver, or would notify an approver to do that if they were not; the author could self-approve if they were an approver

It's pretty much the same, except that it bypasses github ACLs.

We've already taken direct merge powers away from most people in kubernetes-maintainers.

On approvers vs. reviewers:

We're bottlenecked on review bandwidth. We need to grow the reviewer pool, but we also need to move to multi-phase PR reviews in order to take load off of bottleneck people.

A reviewer just has to be someone who could provide constructive feedback on a PR review -- someone who understands our project's coding conventions, ideally the code of that component, etc. We should be more liberal with inclusion in the reviewers list.

An approver is someone with the authority to determine whether a PR should be merged or not.

We'll want a progression by which we can grow contributors into reviewers and reviewers into approvers.

A PR might have a single primary reviewer but require multiple approvers (e.g., an API approver and a Kubelet approver, for a Node API change), or multiple reviewers (e.g., people familiar with different parts of the repo) and a single approver (e.g., me).

As an example, I want to be removed from the PR auto-assigner, but would like to retain my approval powers -- I should be an escalation point, not on the front lines of reviews. 

We should find a home for this:
https://github.com/coreos/findowner

bgrant0607:
@hongchaodeng @xiang90 Do you have any more questions about this?

hongchaodeng:
@bgrant0607 
No questions from me so far. Will try to get this done by next Monday.

hongchaodeng:
[...]

PR #1428, opened 2016-07-23
lgtm handler: find lgtm comment and add label

hongchaodeng:
ref: #1389

This PR replaces LGTMAfterCommit munger with LGTMHandler that

```
// LGTMHandler will
// - apply the LGTM label if reviewer has commented "LGTM, or
// - notify reviewer if non-reviewer has commented "LGTM", or
// - remove the LGTM label from an PR which has been updated since the reviewer added LGTM
```


mungegithub/mungers/lgtm_handler.go
```diff
@@ -83,7 +98,46 @@ func (LGTMAfterCommitMunger) Munge(obj *github.MungeObject) {
...
+
+func foundLGTMFromReviewer(comments []*githubapi.IssueComment, reviewers []*githubapi.User, lastModified *time.Time) bool {
+	for _, c := range comments {
+		if lastModified == nil || c.CreatedAt == nil || (*lastModified).After(*c.CreatedAt) {
+			continue
+		}
+		if !isReviewer(c.User, reviewers) {
+			continue
+		}
+		if strings.ToLower(strings.Trim(*c.Body, " ")) == "lgtm" {
```
xiang90: contains instead of equal to?

xiang90: after some thought, probably equal to is better. sometimes, we say lgtm after XXX to tell author that we are getting really close.

hongchaodeng: Yep. Just want to keep the first version as simple and explicit as possible.

xiang90:
LGTM. Not sure how we should test it. But we should test this before get it merged.

bgrant0607:
cc @kubernetes/contributor-experience

mungegithub/mungers/lgtm_handler.go
jfrazelle: There is a `strings.TrimSpace` method

mungegithub/mungers/lgtm_handler.go
```diff
@@ -83,7 +98,46 @@ func (LGTMAfterCommitMunger) Munge(obj *github.MungeObject) {
...
+	}
+	return false
+}
+
+func foundLGTMFromReviewer(comments []*githubapi.IssueComment, reviewers []*githubapi.User, lastModified *time.Time) bool {
+	for _, c := range comments {
+		if lastModified == nil || c.CreatedAt == nil || (*lastModified).After(*c.CreatedAt) {
+			continue
+		}
+		if !isReviewer(c.User, reviewers) {
```
bgrant0607: It would be useful to explicitly comment that an LGTM from a reviewer is still needed, and notify the assignee if that person is a reviewer.

cc @apelisse

hongchaodeng: > It would be useful to explicitly comment that an LGTM from a reviewer is still needed

Contributors can ping the reviewer for review?

> and notify the assignee if that person is a reviewer.

If not a reviewer, why is a person assigned?
Assigning will generate an notification by github.

apelisse: What I think @bgrant0607 meant is that: if somebody that is NOT the reviewer says "LGTM", it would be nice to mention the reviewer, rather than just ignore

hongchaodeng:
@bgrant0607 @apelisse 
Addressed the comment to add the ability to notify reviewers.

hongchaodeng:
How to move forward this? Let me know if anything else I need to do.

@apelisse @bgrant0607

bgrant0607:
@hongchaodeng Sorry, we're in all-day meetings Tues through Thurs, and eparis is OOO.

@xiang90 Testing: We usually run the bot manually and make it print what it would do ("dry run") in order to test it manually. I haven't thought about how to regression-test it, but am open to suggestions.

mungegithub/mungers/lgtm_handler.go
```diff
@@ -0,0 +1,166 @@
...
+			// TODO: support more complex policies for multiple reviewers.
+			// See https://github.com/kubernetes/contrib/issues/1389#issuecomment-235161164
+			if isReviewer(user, reviewers) {
+				obj.AddLabel(lgtmLabel)
+				return
+			}
+		}
+		if len(users) != 0 {
+			// if somebody that is NOT the reviewer says "LGTM", it would be nice to mention the reviewer, rather than just ignore.
+			obj.WriteComment(fmt.Sprintf("LGTM from users (%v). Can reviewers (%v) review this?", users, reviewers))
```
apelisse: This needs to mention people by using the @ format

mungegithub/mungers/lgtm_handler.go
bgrant0607: @hongchaodeng Yes, I just don't want someone to think they put lgtm and then wonder why the label didn't get added.

Also note that PRs can be manually assigned, and might be assigned to someone not in the OWNERS file. Such reviews can still be helpful, but someone in the OWNERS file should still be required to sign off.

mungegithub/mungers/lgtm_handler.go
```diff
@@ -0,0 +1,166 @@
...
+	if lastModified.After(*lgtmTime) {
+		glog.Infof("PR: %d lgtm:%s  lastModified:%s", *obj.Issue.Number, lgtmTime.String(), lastModified.String())
+		if err := obj.WriteComment(lgtmRemovedBody); err != nil {
+			return
+		}
+		obj.RemoveLabel(lgtmLabel)
+	}
+}
+
+func findLGTMCommenters(comments []*githubapi.IssueComment, lastModified *time.Time) (res []*githubapi.User) {
```
apelisse: Are comments sorted? It'd be nice to leave the loop early if possible

mungegithub/mungers/lgtm_handler.go
```diff
@@ -0,0 +1,166 @@
...
+func (LGTMHandler) AddFlags(cmd *cobra.Command, config *github.Config) {}
+
+// Munge is the workhorse the will actually make updates to the PR
+func (h LGTMHandler) Munge(obj *github.MungeObject) {
+	if !obj.IsPR() {
+		return
+	}
+
+	lastModified := obj.LastModifiedTime()
+	if !obj.HasLabel(lgtmLabel) {
```
apelisse: I would personally split the function in two different functions: 1 for adding the label, 1 for removing it. They have very different use-cases.

mungegithub/mungers/lgtm_handler.go
bgrant0607: @hongchaodeng Please add this convention to the README: All bot notifications should explicitly notify the person/people who need to take action.

bgrant0607: Also, I'd notify the assignee, if the assignee is a reviewer, and auto-assign a reviewer if the assignee is not a reviewer.

mungegithub/mungers/lgtm_handler.go
```diff
@@ -0,0 +1,166 @@
...
+		}
+		comments, err := obj.ListComments()
+		if err != nil {
+			glog.Errorf("unexpected error getting comments: %v", err)
+		}
+		users := findLGTMCommenters(comments, lastModified)
+		for _, user := range users {
+			// TODO: support more complex policies for multiple reviewers.
+			// See https://github.com/kubernetes/contrib/issues/1389#issuecomment-235161164
+			if isReviewer(user, reviewers) {
```
bgrant0607: An approver should be acceptable, also.

bgrant0607:
@hongchaodeng Please also update the README to describe the new behavior.

hongchaodeng:
@bgrant0607 Updated readme

mungegithub/mungers/lgtm_handler.go
hongchaodeng: > mention people by using the @ format

@apelisse Fixed

> notify the assignee, if the assignee is a reviewer

@bgrant0607 Currently the assignee is the reviewer/owner. We should change this behavior. Can we add a TODO here and handle it in separate PR?

> auto-assign a reviewer if the assignee is not a reviewer.

@bgrant0607 This is outside the scope of LGTM handler. Should we put all logic of reviewer assignment into a single place -- which I assumed BlunderbussMunger be. It also has the same problem as above.

mungegithub/mungers/lgtm_handler.go
hongchaodeng: > Are comments sorted?

https://developer.github.com/v3/issues/comments/#list-comments-on-an-issue
"Issue Comments are ordered by ascending ID."

> It'd be nice to leave the loop early if possible

If we sort it ourselves, looping is O(N) and sorting is O(N*lgN). Instead, why don't we have metrics system?

mungegithub/mungers/lgtm_handler.go
hongchaodeng: > split the function in two different functions: 1 for adding the label, 1 for removing it

@apelisse Fixed

mungegithub/mungers/lgtm_handler.go
hongchaodeng: > An approver should be acceptable, also.

@bgrant0607 In my mental world, we haven't defined approver yet. I will try to implement the find approver logic according to this [comment](https://github.com/kubernetes/contrib/issues/1389#issuecomment-235161164) later, of course, in separate PR.
Let's add a TODO here.

mungegithub/mungers/lgtm_handler.go
hongchaodeng: > Currently the assignee is the reviewer/owner. We should change this behavior. Can we add a TODO here and handle it in separate PR?

@bgrant0607 I mean "we probably would change this behavior". I have added a note there saying "if assignees = reviewers not true anymore..."

bgrant0607: @hongchaodeng We can figure out reassignment in a subsequent PR. In this PR, it's mainly important to determine when to apply the lgtm label, and to make it clear why the lgtm label isn't applied in other cases. Beyond that, I'm fine with deferring improvements and extensions to subsequent PRs.

hongchaodeng:
> it's mainly important to determine when to apply the lgtm label, and to make it clear why the lgtm label isn't applied in other cases.

@bgrant0607 Yes. These were addressed in this PR. I have updated the top comment to reflect that, also squashed commits. 

We should move forward this. What else?

> We usually run the bot manually and make it print what it would do ("dry run") in order to test it manually.

Is there anyone in Google who do the QA/testing? Or do I need to test it? If latter, any pointer to testing documentation?

apelisse:
@hongchaodeng I'll try to have a look. We don't have testers, it's up to you to write a test. There are many (yet probably not enough) examples in `mungegithub/mungers/*_test.go`.

I'll try to have a look and move it forward as soon as I can.

Thanks,

hongchaodeng:
Rebased.

@apelisse Thanks for taking a look. Let's move this forward and care about others later. Let me know if there is anything you have strong opinions about.

apelisse:
I think this is great. Can you please fix the broken CI test and squash the rebase commit?

Also, we should use the same piece of code to mention users and build the assignee list. I'm also doing that in `close-stale-pr.go`. But it shouldn't prevent this from moving forward.

hongchaodeng:
@apelisse 
Thanks. Squashed.
The previous error was caused by undeleted lgtm_after_commit.go file after rebasing.

apelisse:
@hongchaodeng Thanks! I'll merge and deploy next week

bgrant0607:
@hongchaodeng @apelisse Great! Thanks!

Please document this in https://github.com/kubernetes/kubernetes/blob/master/docs/devel/automation.md, and send an announcement to kubernetes-dev.

hongchaodeng:
@bgrant0607 Sure. Thanks for notifying.

hongchaodeng:
Ann: https://groups.google.com/forum/#!topic/kubernetes-dev/OZuk1QFmIa4

lavalamp:
@hongchaodeng sorry, but we reverted this (because there wasn't a way to turn it off with a flag).

To turn it back on, we need to turn this logic into something that comprehensively considers every lgtm-affecting variable and delivers a verdict (and only once!).

That is, it needs to be able to make messages like "I would have applied the LGTM label, but the release note process hasn't been followed" instead of fighting with itself.

Speaking of, I think we need to add some shortcuts to turn LGTM off, and to turn other labels on/off. Maybe something like `bot:+lgtm,+release-note-none`.

lavalamp:
In particular, I need to be able to say something like "lgtm after you fix nits" and not have the bot interpret that as an lgtm.

hongchaodeng:
@lavalamp @bgrant0607 @apelisse 
I'm sorry for not replying in time.
Let me try to fix this. To me it sounds like there are LGTM logic in other places. Can we put them into just one handler? What are the other places?

foxish:
Part of this issue has been fixed in https://github.com/kubernetes/contrib/pull/1518, where we no longer remove LGTM, and instead add do-not-merge. Also, it would make sense to make LGTM an explicit command, such as `/LGTM`, which is unambiguous and less likely to lead to situations like: https://github.com/kubernetes/kubernetes/pull/29999#issuecomment-237404442.

foxish:
@hongchaodeng We definitely should have this implemented as a separate munger, rather than extend an existing one the way it currently does, so as to ensure that we have a way to turn it on/off decoupled from other mungers. I'd be happy to review your PR once these changes are made.

rmmh:
Note that Reviewable's LGTMs [look different](https://github.com/kubernetes/kubernetes/pull/26027#issuecomment-239527443). I don't know if it adds the label.

foxish:
@rmmh That is a good point. Perhaps we should still have the reviewer explicitly do a `/LGTM` in the comments, as a bot command. That would be the easiest and the intent is unmistakable in my opinion.

lavalamp:
I'm in favor of something explicit. `/lgtm`--not quoted with backticks & on
its own line--is fine.

lavalamp:
We should also let the PR author say "release-note: none" in the PR
message. Otherwise adding an lgtm command won't in practice help that much.

hongchaodeng:
"/lgtm" sounds good. Will do it.
"release-note: none" is a good point too. Will do it after lgtm thing. It's a separate issue.
Initially, I would just assume a single line with only "/lgtm" or "xxx" without anything else. If there is more feedback or strong demands, we could handle more later.

eparis:
Why not just "any line with ONLY `lgtm` or `:lgtm:`.

Also, lgtm without 'release-note-none` hopefully won't fight, but you right, we need a non-write way to se tthose...

apelisse:
@eparis We are trying to define more standard way to communicate (in both direction) with the bot right now. `/LGTM` would be a command you give to the bot. And on the other side, the bot would have specific messages `[NEEDS-REBASE]` to notify users of what's going on etc. This should make it easier to regex.

eparis:
I'm all for the bot having to be 'magic' so its actions can be easily captured and reasoned about in the future. But arbitrary non-obvious style requirements on humans stink. I will get it wrong. Repeatedly. And I'm going to use it a lot.

Humans talking to bots need to be VERY human friendly. Not all of us are particularly bright.. I think the bot should respond to me lgtm if I added it via reviewable or by hand, without any extra syntax...

foxish:
@eparis The problem with that may be that people use the LGTM comment in an overloaded fashion, sometimes meaning LGTM after nits, or LGTM after comments addressed. 

Like in the release note issue that surfaced at https://github.com/kubernetes/kubernetes/pull/29999#issuecomment-237404442, the bot started to add the LGTM label when it wasn't intended (the LGTM comment has since been deleted). It would be better if we knew that it is a command specifically intended for the bot IMO.

rmmh:
I agree with eparis.

LGTM on a line by itself, or the LGTM comment from Reviewable should both be sufficient.

Having the bot not interpret LGTM there's other things on the line "LGTM after fixing nits" would be okay. The bot could also reply to those messages with "if you want to apply the LGTM label, please respond with LGTM on a line by itself".

Another affordance for humans would be accepting punctuation.

Adds LGTM label:
"lgtm", "LGTM", "lgtm.", "LGTM!", "LGTM!\n\nthanks for fixing this", "`:lgtm:</img>`"

Doesn't add LGTM label, responds with usage hint (ambiguous directive):
"lgtm after nits", "seems fine, lgtm"

Does nothing:
"looks good to me", "unrelated comment"

The usage hint could be as simple as:
"@commenter: to mark a PR as LGTM, write "LGTM" on a line by itself."


In kubernetes/kubernetes#29999, people left these comments, which would be classified as ambiguous:
`oops. obviously, I mean  lgtm after nits.`
`@taufen Just one comment. Otherwise lgtm`

apelisse:
`/` is not arbitrary. It's the "de facto" command prefix since forever (IRC probably made it popular in the 90s, many games also use it). It makes it easy to write code that says "Look in the comments for the last commands"/"Do we have a LGTM command ...". It's consistent, it's always the same, it's easier to parse, it's a single piece of code (`FindCommand("LGTM")`) It can have arguments (maybe useful for some commands? `/RETEST #Flakyissue`). I think it's less awkward than: `retest this issue: #Flakyissue`.

Also, I completely agree that we should provide help if we detect something that could be a command but is not, mistyped, etc

bgrant0607:
@eparis @rmmh 

We have a lot of inconsistent phrases for communicating with our bots: `ok to test`, `@k8s-bot: test this issue: #....`, `I signed it!`, and so on. It's very confusing, especially the longer they are.

We also have the problems with ambiguity mentioned above.

And the need to be able to easily identify the commands, for multiple purposes: labeling, metrics collection, dashboard, etc.

We're about to add lots more commands. I would strongly prefer a consistent, terse syntax. The proposed syntax was inspired by various chatbots. I'm open to suggestions, but I don't think plain english is the way to go. Think about documenting the rules and examples from the comment above. That's way, way too complicated.

Hypothetical commands:
* /lgtm
* /approve
* /ack
* /ptal
* /norelnote
* /hascla
* /dibs (assign this to me)
* /help (dump possible commands)

We may also want to make it possible to invoke the same commands from elsewhere, such as a CLI or even slack.

bgrant0607:
Sentiment analysis may be a fun college project, but we shouldn't need it in order for our bot to work.

rmmh:
If we want the Reviewable flow to work well, we should support its LGTM format. LGTMs happen an order of magnitude more than anything else.

Finding "LGTM" on a line by itself is very far from sentiment analysis.

rmmh:
Opening a PR in Reviewable, reading through all the changes, and clicking the "LGTM" button makes sense.

Having to go back to Github and commenting "/lgtm" to explicitly instruct the bot is a needless extra step.

lavalamp:
Is there other identifying notation in comments from reviewable? How about
accepting both "/lgtm" from a human or the official reviewable LGTM message?

apelisse:
We definitely intend to be compatible with Reviewable and special case this if needed.

rmmh:
[A Reviewable LGTM](https://github.com/kubernetes/kubernetes/pull/26027#issuecomment-239527443) has these distinctive lines:

    :lgtm:
    *Comments from [Reviewable](https://reviewable.kubernetes.io:443/reviews/kubernetes/kubernetes/26027#-:-KP-2syrokgBDbUGJ_fL:bnfp4nl)*

bgrant0607:
@rmmh Reviewable's documentation claims that the text inserted upon lgtm can be customized. Can we reconfigure it to make it compatible with this syntax?

rmmh:
Maybe? I looked and didn't see that.

foxish:
We now have the bot honoring `/lgtm` and `/lgtm` cancel on this repo.
Just ran a quick test [here](https://github.com/kubernetes/contrib/pull/1558), and there shouldn't be races between different mungers (hopefully).

Thanks @hongchaodeng for helping with this effort.

hongchaodeng:
Thanks @foxish for helping too!

Now we are gonna move to "release-note" or "approve" in next step.

PR #1433, opened 2016-07-25
Fixed crashing submit-queue when container restarts.

foxish:
While emptyDir presents an empty directory when a pod is killed and recreated by the deployment, it retains contents between container
restarts. This change guards against that scenario by not cloning the repository when a `PROJECT_NAME` directory already exists.

### DIFF
--- a/mungegithub/features/repo-updates.go
+++ b/mungegithub/features/repo-updates.go
@@ -155,15 +155,28 @@ func (o *RepoInfo) Initialize(config *github.Config) error {
 	if !finfo.IsDir() {
 		return fmt.Errorf("--repo-dir is not a directory")
 	}
+
+	// check if the cloned dir already exists, if yes, cleanup.
+	if _, err := os.Stat(o.projectDir); !os.IsNotExist(err) {
+		if err := o.cleanUp(o.projectDir); err != nil {
+			return fmt.Errorf("Unable to remove old clone directory at %v: %v", o.projectDir, err)
+		}
+	}
+
 	if cloneUrl, err := o.cloneRepo(); err != nil {
 		return fmt.Errorf("Unable to clone %v: %v", cloneUrl, err)
 	}
 	return o.updateRepoUsers()
 }
 
+func (o *RepoInfo) cleanUp(path string) error {
+	err := os.RemoveAll(path)
+	return err
+}
+
 func (o *RepoInfo) cloneRepo() (string, error) {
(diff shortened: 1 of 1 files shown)


foxish:
@apelisse PTAL

apelisse:
What if the directory already exists but is in an invalid state?

foxish:
I thought of that, and it could happen I guess if the container crashed right as the clone process was starting, before it created the .git directory. I think that one is a little harder to detect. This one should protect us from the sort of crash we saw this morning though. I'll follow this up with more robust checking, and there's perhaps some function that the go-github library could give us to see if the git state of a directory is valid.

apelisse:
Yes, that would probably happen in `UpdateRepo`. Are we sure this is safe enough? Maybe we should re-clone if the update fails.

foxish:
I think this should hold us for now in case there is a container going bad at some point during the day. Long-term solutions could be:

1. We could re-clone the directory under a different randomly chosen name, but if we never clean up, it is possible we may run out of disk space at some point (?), especially if it is in a crash loop. 
2. We could check that it's a valid directory + valid git repository and let `updateRepoUsers` take care of it. If only a partial clone was performed before the container crashed, `updateRepoUsers` should still be able to fix it.

I feel like option 2 is better. What do you think?

foxish:
I just did some more investigation with `git clone`. It appears that clone first copies files into a temp directory, and only when complete, it renames the directory to _project-name_. This protects us I think, and that would make the above fix sufficient. If a directory by that name exists, it implies that the clone was successful.

apelisse:
I'm not worried about `clone` lying to us. I'm worried about corruption in a repo because of a crash.

Instead of doing "is the directory there already?", I would do "is there directory there? Then run `git fsck` in it"

foxish:
Okay, that makes sense. I'll add a `git fsck` check to it.

foxish:
@apelisse From my experiments so far:
- Making the container crash at different points is leading to different behavior with subsequent git pulls. 
- It does not appear that git-fsck fixes errors, just appears to report corruption.

Hence, I found it easiest to simply try and delete the directory. Cloning to a different path is a problem because we rely on the path being `/gitrepos/kubernetes/...` in the arguments supplied to several mungers. This deletion operation usually succeeds (I haven't seen it fail so far), but if it fails, it returns an error, and that error will restart the container, which will try again to delete the directory when it runs again.

apelisse:
lgtm

PR #1434, opened 2016-07-25
Assign test owners to default as a fallback. Pick a random owner if there are multiple options.

rmmh:
See also: kubernetes/kubernetes#29564

### DIFF
--- a/mungegithub/mungers/testowner/owner.go
+++ b/mungegithub/mungers/testowner/owner.go
@@ -20,6 +20,7 @@ import (
 	"encoding/csv"
 	"errors"
 	"io"
+	"math/rand"
 	"os"
 	"regexp"
 	"strings"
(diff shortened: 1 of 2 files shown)


mungegithub/mungers/testowner/owner.go
```diff
@@ -41,18 +42,28 @@ func normalize(name string) string {
...
 }
 
-// TestOwner returns the owner for a test, or the empty string if none is found.
+// TestOwner returns the owner for a test, an owner from $default if present,
+// or else the empty string if none is found.
 func (o *OwnerList) TestOwner(testName string) string {
 	name := normalize(testName)
 	owner, _ := o.mapping[name]
+	if owner == "" {
+		owner, _ = o.mapping["$default"]
```
fejta: I'm not super wild about $default as this looks like a environment variable

fejta:
LGTM

mungegithub/mungers/testowner/owner.go
```diff
@@ -41,18 +42,28 @@ func normalize(name string) string {
...
+	}
+	if strings.Contains(owner, "/") {
```
fejta: Would a comma or tab make more sense here?

mungegithub/mungers/testowner/owner.go
rmmh: What would you prefer?

mungegithub/mungers/testowner/owner.go
rmmh: It's in a CSV, so commas are right out. The / is more visible/understandable than a tab.

mungegithub/mungers/testowner/owner.go
fejta: `DEFAULT` would be clear enough for me

mungegithub/mungers/testowner/owner.go
fejta: Makes sense about the comma... Maybe a [semi-colon](https://en.wikipedia.org/wiki/Comma-separated_values#Specification)?

rmmh: `/` is more commonly used to separate items than `;`.

mungegithub/mungers/testowner/owner.go
```diff
@@ -41,18 +42,28 @@ func normalize(name string) string {
 // OwnerList uses a map to get owners for a given test name.
 type OwnerList struct {
 	mapping map[string]string
+	rng     *rand.Rand
 }
 
-// TestOwner returns the owner for a test, or the empty string if none is found.
+// TestOwner returns the owner for a test, an owner from $default if present,
```
ixdy: comment needs updating

PR #1437, opened 2016-07-26
Add automatic issue labeler

### LINKED ISSUE #1432: [mungegithub]  Create Munger To Automatically Add Labels To Issues
Issues currently require manual labeling for routing.   Let's automate the process with some simple machine learning.

grodrigues3:
Resolves #1432

The simple_app.py file loads 2 pretrained machine learning models that have learned patterns on existing issues.  For new issues, the munger file will send a request to the webapp with the new issue's title and body, and the webapp will respond with the team and component labels that it thinks are appropriate.


grodrigues3:
@apelisse The simple app no longer depends on any of the other code that was used for model selection and validating the model.  It logs to a file (which maybe should be on the PV and it is currently storing the model on the container as well).  Let's change the configuration to point to the persistent volume.  Let me know if you see anything else that should be change (I can change it to a get request if you really want)

mungegithub/issue-labeler/simple_app.py
```diff
@@ -0,0 +1,125 @@
...
+myPenalty = 'l2'
+myHasher = FeatureHasher(input_type="string", n_features= numFeatures, non_negative=True)
+myStemmer = PorterStemmer()
+tokenizer = RegexpTokenizer(r'\w+')
+
+
+try:
+  stop_words = []
+  if not stopwords:
+    stop_fn = "./data/stopwords.txt"
```
apelisse: where is this file coming from?

mungegithub/issue-labeler/simple_app.py
```diff
@@ -0,0 +1,125 @@
...
+    comp_mod = joblib.load(component_fn)
+    vec = myHasher.transform([tokens])
+    tlabel = team_mod.predict(vec)[0]
+    clabel = comp_mod.predict(vec)[0]
+    return ",".join([tlabel, clabel])
+
+
+def tokenize_stem_stop(inputString):
+    curTitleBody = tokenizer.tokenize(inputString.decode('utf-8').lower())
+    return map(myStemmer.stem, filter(lambda x: x not in stopwords, curTitleBody))
```
apelisse: how many stopwords do you have? Maybe you should use a `set` instead

mungegithub/issue-labeler/simple_app.py
```diff
@@ -0,0 +1,125 @@
...
+  with open(stop_fn, 'r') as f:
+    stopwords = [word.strip() for word in f]
+except:
+  #don't remove any stopwords
+  stopwords = []
+
+@app.errorhandler(500)
+def internal_error(exception):
+  app.logger.error(exception)
+  return "Internal Server Error-> Check the Logs"
```
apelisse: I would have done: `return str(exception)`. We already know it's an internal error (because of 500). And it's much simpler than checking the logs.

mungegithub/mungers/issue-categorizer.go
```diff
@@ -0,0 +1,96 @@
+/*
```
apelisse: Please apply some sort of `gofmt` to this file.

mungegithub/mungers/issue-categorizer.go
```diff
@@ -0,0 +1,96 @@
...
+	if obj.IsPR() {
+		return
+	}
+
+
+
+	issue := obj.Issue
+
+	existingLabels := issue.Labels
+	for _, label := range existingLabels{
```
apelisse: There is a function named `GetLabelsWithPrefix` in `mungegithub/github`:

```
github.GetLabelsWithPrefix(issue.Labels, 'team/')
github.GetLabelsWithPrefix(issue.Labels, 'component/')
```

mungegithub/mungers/issue-categorizer.go
```diff
@@ -0,0 +1,96 @@
...
+		}
+	}
+
+
+
+	routingLabelsToApply, err := http.PostForm("http://localhost:5000",
+		url.Values{"title": {*issue.Title}, "body": {*issue.Body}})
+
+	if routingLabelsToApply.StatusCode != 200{
+		glog.Error("Not 200 status code: " + string(routingLabelsToApply.StatusCode))
```
apelisse: You could print the error received from the server (the body)

mungegithub/issue-labeler/simple_app.py
```diff
@@ -0,0 +1,125 @@
...
+def get_labels():
+    title = request.form.get('title', "")
+    body = request.form.get('body', "")
+    tokens = tokenize_stem_stop(" ".join([title, body]))
+    team_mod = joblib.load(team_fn)
+    comp_mod = joblib.load(component_fn)
+    vec = myHasher.transform([tokens])
+    tlabel = team_mod.predict(vec)[0]
+    clabel = comp_mod.predict(vec)[0]
+    return ",".join([tlabel, clabel])
```
apelisse: This is not consistent with the string split below (`;` vs `,`)

mungegithub/issue-labeler/simple_app.py
```diff
@@ -0,0 +1,125 @@
...
+    return map(myStemmer.stem, filter(lambda x: x not in stopwords, curTitleBody))
+
+
+@app.route("/update_models", methods = ["PUT"])
+def update_model():
+    """
+    data should contain three fields
+      titles: list of titles
+      bodies: list of bodies
+      labels: list of list of labels
```
apelisse: Not pretty.

More pretty:
```
data should be of the following form:
[
   {
      "title": "Issue Title",
      "body": "Issue body",
      "labels": ["team/something", "component/something"]
   },
   ...
]
```

mungegithub/issue-labeler/simple_app.py
```diff
@@ -0,0 +1,125 @@
...
+    titles = data.get('titles')
+    bodies = data.get('bodies')
+    labels = data.get('labels')
+
+    tTokens = []
+    cTokens = []
+    team_labels = []
+    component_labels = []
+    for (title, body, label_list) in zip(titles, bodies, labels):
+      tLabel = filter(lambda x: x.startswith('team'), label_list)
```
apelisse: I find it interesting that you use filter rather than list comprehension. Whatever =)

mungegithub/issue-labeler/simple_app.py
```diff
@@ -0,0 +1,125 @@
...
+      if tLabel:
+        team_labels += tLabel
+        tTokens += [tokens]
+      if cLabel:
+        component_labels += cLabel
+        cTokens += [tokens] 
+    tVec = myHasher.transform(tTokens)
+    cVec = myHasher.transform(cTokens)
+
+    if os.path.isfile(team_fn):
```
apelisse: You could simplify the logic by switching if statements here and below:
```
if team_labels:
    if os.path.isfile(team_fn):
        ...
    else:
        ...
```

mungegithub/issue-labeler/simple_app.py
```diff
@@ -0,0 +1,125 @@
...
+        component_model.partial_fit(cVec, np.array(component_labels))
+    else:
+      #create a new component model
+      if component_labels:
+        component_model = SGDClassifier(loss=myLoss, penalty=myPenalty, alpha=myAlpha) 
+        component_model.fit(cVec, np.array(component_labels))
+    
+    joblib.dump(team_model, team_fn)
+    joblib.dump(component_model, component_fn)
+    return "Nothing went wrong"
```
apelisse: You can return an empty string :-)

mungegithub/issue-labeler/simple_app.py
grodrigues3: modified the dockerfile to add it to the container.  there are only 100ish words but I don't want to copy them all into the source file.

mungegithub/mungers/issue-categorizer.go
grodrigues3: the line below this also checks if `err != nil`.   Is that redundant.  Does a non 200 status code automatically indicate there was an error and can I removed the second check.

mungegithub/issue-labeler/simple_app.py
grodrigues3: this is for the comment right.  if so, i changed it for you.

mungegithub/issue-labeler/simple_app.py
grodrigues3: interesting or bad? is it bad form for some reason?

mungegithub/submit-queue/deployment.yaml
```diff
@@ -68,3 +68,18 @@ spec:
       - name: cache-volume
         persistentVolumeClaim:
           claimName: submit-queue-cache
+      - name: machine-learning-container
```
apelisse: Please revert this change!

mungegithub/issue-labeler/README.md
```diff
@@ -0,0 +1,27 @@
...
+In order to run the webapp, build the container using the following command.
+
+$sudo docker build -no-cache -t <choose-image-name> . 
+
+Then, run the container making sure to forward traffic to port 5000.
+
+$docker run -p <port>:5000 <image name>
+
+
+
```
apelisse: It misses a couple of blank lines I think ;-)

apelisse:
Cool, can you please revert `deployment.yaml` change and squash these commits?

PR #1438, opened 2016-07-26
OWNERS files for each directory under contrib

foxish:
Used a modified version of https://github.com/coreos/findowner (at https://github.com/foxish/findowner/tree/foxish-edits) to generate the YAML files.
It populates the top 3 committers over the past 6 months as potential assignees. 

PTAL if the lists make sense and we can get blunderbuss running on contrib.
cc @lavalamp @bgrant0607 @apelisse @pwittrock

### DIFF
--- a/404-server/OWNERS
+++ b/404-server/OWNERS
@@ -1,0 +1,5 @@
+assignees:
+- bprashanth
+- luxas
+- mikedanese
+
--- a/ansible/OWNERS
+++ b/ansible/OWNERS
@@ -1,0 +1,7 @@
+assignees:
+- danehans
+- eparis
+- ingvagabund
+- rutsky
+- stephenrlouie
+
--- a/OWNERS
+++ b/OWNERS
@@ -1,0 +1,5 @@
+assignees:
+- aledbf
+- bprashanth
+- eparis
+- lavalamp
--- a/addon-resizer/OWNERS
+++ b/addon-resizer/OWNERS
@@ -1,0 +1,5 @@
+assignees:
+- Q-Lee
+- piosz
+- wojtek-t
+
--- a/cluster-autoscaler/OWNERS
+++ b/cluster-autoscaler/OWNERS
@@ -1,0 +1,5 @@
+assignees:
+- fgrzadkowski
+- mwielgus
+- piosz
+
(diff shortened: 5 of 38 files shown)


eparis:
Does it matter that some (many?) of these people don't have any permission to label things?

ansible/OWNERS
```diff
@@ -0,0 +1,5 @@
+assignees:
+- rutsky
+- stephenrlouie
+- danehans
```
eparis: Add myself and ingvagabund

eparis:
Is there any way to get these alphabetized? I know it doesn't actually matter...

ansible/OWNERS
ingvagabund: @foxish is there any way to detect reviewers? Reviewer = user adding labels && merging PR && commenting PR

bgrant0607:
@eparis We intentionally want to broaden the reviewer pool and we're going to automate application of the `lgtm` label (see #1428).

ansible/OWNERS
bgrant0607: @ingvagabund See #1428 and #1389.

404-server/OWNERS
```diff
@@ -0,0 +1,4 @@
+assignees:
+- wernight
```
bgrant0607: @foxish 
Replace with bprashanth. Why didn't findowners choose him?

He wrote the code:
https://github.com/kubernetes/contrib/blame/master/404-server/server.go

ansible/OWNERS
foxish: Ah, right. Will replace. I think it did not choose him because his changes are over 6 months old. We should probably alter the logic to use more than just the commits in the past 6 months if that is often the case.

404-server/OWNERS
foxish: Ah, right. Will replace. I think it did not choose him because his changes are over 6 months old. We should probably alter the logic to use more than just the commits in the past 6 months if that is often the case.

foxish:
Alphabetized the OWNERS lists, and generated it using entire commit history rather than just the last 6 months. PTAL.

lavalamp:
LGTM, we can merge and iterate.

Issue #762, opened 2016-04-12
Move code that should be maintained to dedicated repos


bgrant0607:
contrib isn't monitored adequately for maintained code. Issues aren't triaged. PRs aren't reviewed. We need to fix those problems, but it's not a good idea to mix maintained and unmaintained code. 

Examples:
- addon-resizer
- ingress and service-loadbalancer
- test infra/utils and mungeithub
- do we still need release-notes?
- ansible?

cc @bprashanth @eparis @david-mcmahon @mikedanese @ixdy @Q-Lee

bprashanth:
Yeah I've brought this up before. 
I think we decided (at the time) that we don't really get much from splitting Ingress out into another repo. 

Fwiw both serviceloadbalancer and ingress have e2es (https://github.com/kubernetes/kubernetes/blob/master/test/e2e/ingress.go) in the main repo that consume released images from: https://github.com/kubernetes/contrib/releases. Unittests will run as part of contrib pre-checking github hooks. I usually include a line in the release notes when I bump up the image in: https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/cluster-loadbalancing/glbc/glbc-controller.yaml#L21

It's a hand wavy process that I'd like to improve. 

The meta issue is: How do we manage cluster addons?    
* Where are they run? 
  - If they're on the node, then we need to be really careful about increasing request limits 
  - If they're on the master, we need to allow users to swap them out (eg: gce l7 for nginx)
* Where do they live? 
  - kube-dns is checked into main 
  - heaptster has it's own repo but not sure about e2es 
  - Ingress is described above
  - cadvisor runs as part of kubelet but still has its own repo/release.
* Are they all shipped by default, with every cluster? 
   - Increases resources required for smallest cluster
   - If we don't, Ux is worse, docs harder to write (eg: start an Ingress controller, then create this)
* How are they tested?
  - Ingress would benefit from a builder available to the community
  - Running e2es from contrib/-HEAD would also be a big simplification
  - But if each of these gets a seperate e2e-builder how do these block a release?

bgrant0607:
I would like to nuke contrib. It's no longer needed by the community.

davidopp:
> I would like to nuke contrib. It's no longer needed by the community.

Can you clarify? What would we do with the stuff that's currently in kubernetes/contrib?

bgrant0607:
@davidopp See the PR description.

davidopp:
[...]

PR #1440, opened 2016-07-27
Provide an Ingress Controller that use Citrix NetScaler

chiradeep:
This contribution provides an Ingress Controller that can configure a Citrix NetScaler.


bprashanth:
Firstly, thanks for the pull request. 

We are in the process of dismantling the Kubernetes contrib repo into several sub-repositories as discussed on https://github.com/kubernetes/contrib/issues/762. With each large pull request this gets harder to do. The Ingress model is intentionally losely coupled from the controllers so the maintenance of each one doesn't fall on the same spof (read: me). 

Instead of upstreaming such a large change, can I suggest an update to the ingress docs with a link to your repository? If you're up for the docs update, please add a section to https://github.com/kubernetes/contrib/tree/master/ingress/controllers. I'm closing this pr assuming you understand the problem as described.

thockin:
I concur with Prashanth.  I don't think you get much value out of putting
it in our org or repo, at least not more than us simply linking to you.
We're happy to advise and review the kubernetes-specific parts of your
code, but the name on the repo should be your own :)

I think we'll be doing more of this.

chiradeep:
Makes sense, thanks.

PR #1443, opened 2016-07-28
Update the README for the nginx-slim image

ide:
The NGINX image actually adds back SPDY support, which was removed when NGINX fully launched HTTP/2. There is also a patch for dynamic sizing of the TLS records that look at the TCP segment sizes.


bprashanth:
LGTM thanks

PR #1447, opened 2016-07-28
Expose pprof handlers from healthz

bprashanth:


girishkalele:
Looks correct - I have a similar patch for kube-dns.

Do we want this permanently in place or have an alternate debug container ?

bprashanth:
This exposes running goroutines etc, it needs to be part of the same process

girishkalele:
Ack, these hooks are process-private.

bprashanth:
LGTM?

PR #1453, opened 2016-07-29
Fix health check link

bprashanth:
Just convering health_check to health_checks


bprashanth:
one char change, merging

PR #1454, opened 2016-07-29
Fix pv definition to match running instance.

foxish:
The name of the PD mounted on the utility cluster is kubernetes-cache.

cc @lavalamp


lavalamp:
lgtm

PR #1455, opened 2016-07-30
Mention assignees and author when LGTM is removed

apelisse:
After a push, the LGTM flag is removed by the bot. It can be hard to
notice because it doesn't mention anybody.

Make it easier for people to notice by mentioning reviewers and author,
and keep the pull-request moving.

### DIFF
--- a/mungegithub/mungers/lgtm_after_commit.go
+++ b/mungegithub/mungers/lgtm_after_commit.go
@@ -17,6 +17,9 @@ limitations under the License.
 package mungers
 
 import (
+	"fmt"
+	"regexp"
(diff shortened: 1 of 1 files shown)


apelisse:
cc @bgrant0607

bgrant0607:
Thanks!

FYI, in general, I plan to point out examples of problems I come across. It doesn't necessarily mean that they are higher priority than other issues.

LGTM

apelisse:
When it's too fast to fix, it's not worth the overhead of
scheduling/keeping it somewhere in my head :-)

PR #1460, opened 2016-08-01
Mention code

apelisse:
Create generic code to mention people, so that we don't reimplement it in each munger. It comes with tests and interesting utility functions.

### DIFF
--- a/mungegithub/mungers/close-stale-pr.go
+++ b/mungegithub/mungers/close-stale-pr.go
@@ -19,11 +19,11 @@ package mungers
 import (
 	"fmt"
 	"regexp"
-	"strings"
 	"time"
 
 	"k8s.io/contrib/mungegithub/features"
 	"k8s.io/contrib/mungegithub/github"
+	"k8s.io/contrib/mungegithub/mungers/mungerutil"
 
 	"github.com/golang/glog"
 	githubapi "github.com/google/go-github/github"
@@ -222,7 +222,7 @@ func durationToDays(duration time.Duration) string {
 }
 
 func closePullRequest(obj *github.MungeObject, inactiveFor time.Duration) {
-	mention := mentionUsers(getInvolvedUsers(obj))
+	mention := mungerutil.GetIssueUsers(obj.Issue).AllUsers().Mention().Join()
 	if mention != "" {
 		mention = "cc " + mention + "\n"
 	}
@@ -236,36 +236,8 @@ func closePullRequest(obj *github.MungeObject, inactiveFor time.Duration) {
 	obj.ClosePR()
 }
 
-func getInvolvedUsers(obj *github.MungeObject) []string {
-	var users []string
-
-	var user string
-	if obj.Issue.User != nil && obj.Issue.User.Login != nil {
-		user = *obj.Issue.User.Login
-		users = append(users, user)
-	}
-
-	for _, assignee := range obj.Issue.Assignees {
-		if assignee.Login == nil || *assignee.Login == user {
-			continue
-		}
-		users = append(users, *assignee.Login)
-	}
-
-	return users
-}
-
-func mentionUsers(users []string) string {
-	var mentions []string
-	for _, user := range users {
-		mentions = append(mentions, "@"+user)
-	}
-
-	return strings.Join(mentions, " ")
-}
-
 func postWarningComment(obj *github.MungeObject, inactiveFor time.Duration, closeIn time.Duration) {
-	mention := mentionUsers(getInvolvedUsers(obj))
+	mention := mungerutil.GetIssueUsers(obj.Issue).AllUsers().Mention().Join()
 	if mention != "" {
 		mention = "cc " + mention + "\n"
 	}
--- a/mungegithub/mungers/mungerutil/util.go
+++ b/mungegithub/mungers/mungerutil/util.go
@@ -1,0 +1,98 @@
+/*
+Copyright 2016 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package mungerutil
+
+import (
+	"strings"
+
+	"k8s.io/kubernetes/pkg/util/sets"
+
+	"github.com/google/go-github/github"
+)
+
+// UserSet is a set a of users
+type UserSet sets.String
+
+// GetUsers returns a UserSet
+func GetUsers(users ...*github.User) UserSet {
+	allUsers := sets.String{}
+
+	for _, user := range users {
+		if user != nil && user.Login != nil {
+			allUsers.Insert(*user.Login)
+		}
+	}
+
+	return UserSet(allUsers)
+}
+
+// Has tells you if the users can be found in the set
+func (u UserSet) Has(user ...*github.User) bool {
+	return len(u.intersection(GetUsers(user...))) != 0
+}
+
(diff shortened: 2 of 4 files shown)


apelisse:
@hongchaodeng you probably want to have a look

apelisse:
@foxish Can I ask you to have a look and review that?

Thanks!

mungegithub/mungers/close-stale-pr.go
```diff
@@ -222,7 +222,7 @@ func durationToDays(duration time.Duration) string {
 }
 
 func closePullRequest(obj *github.MungeObject, inactiveFor time.Duration) {
-	mention := mentionUsers(getInvolvedUsers(obj))
+	mention := mungerutil.JoinMentions(mungerutil.CreateActiveUsersFromIssue(obj.Issue).AllUsers())
```
foxish: Would `GetActiveUsersFromIssue` be a better name? I found line 225 a little confusing when reading it. Although mungetutil is creating an object internally, because the function is exported, `GetSomething` may be more meaningful as a name.

I feel like mungerutil should export functions like GetUsersFromIssue, GetAssigneesFromIssue, and GetAuthorsFromIssue, which would make using it easier. Does that make sense?

mungegithub/mungers/mungerutil/util_test.go
```diff
@@ -0,0 +1,297 @@
...
+			users:     []string{"Jane", "John"},
+			mentioned: []string{"@Jane", "@John"},
+		},
+	}
+
+	for _, test := range tests {
+		actualList := sort.StringSlice(MentionUsers(test.users))
+		actualList.Sort()
+		if !reflect.DeepEqual([]string(actualList), test.mentioned) {
+			t.Error(actualList, "dones't match expected", test.mentioned)
```
foxish: s/dones't/doesn't/

mungegithub/mungers/mungerutil/util.go
```diff
@@ -16,14 +16,122 @@ limitations under the License.
 
 package mungerutil
 
-import "github.com/google/go-github/github"
+import (
+	"strings"
 
-// FilterValid filters a list of user data and returns only user data that contains valid username.
-func FilterValid(users []*github.User) (res []*github.User) {
```
foxish: Looks like getInvolvedUsers and FilterValid were doing overlapping things before. Good that we got rid of it. :)

mungegithub/mungers/mungerutil/util_test.go
```diff
@@ -0,0 +1,297 @@
...
+
+func TestActiveUserIsAssignee(t *testing.T) {
+	tests := []struct {
+		activeUsers       *ActiveUsers
+		potentialAssignee *github.User
+		isAssignee        bool
+	}{
+		{
+			activeUsers: createActiveUsers(makeUser("John"),
+				[]*github.User{makeUser("Jane")}, nil),
```
foxish: Is the indentation off here?

mungegithub/mungers/mungerutil/util.go
```diff
@@ -16,14 +16,137 @@ limitations under the License.
...
+		validUser := GetUser(user)
+		if validUser != nil {
+			allUsers = append(allUsers, *validUser)
+		}
+	}
+	return mergeUsers(allUsers)
+}
+
+// CreateActiveUsersFromIssue make a new object from a github.Issue
+func CreateActiveUsersFromIssue(issue *github.Issue) *ActiveUsers {
```
foxish: My very last nit. The name `GetActive..` would be easier to read from the outside in my opinion. Also, the comment above is a little ambiguous.

foxish:
LGTM

mungegithub/mungers/lgtm_handler.go
```diff
@@ -118,25 +117,21 @@ func (h LGTMHandler) addLGTMIfReviewerSaidSo(obj *github.MungeObject) {
...
 		}
 	}
 	if len(users) != 0 {
 		// if somebody that is NOT the reviewer says "LGTM", it would be nice to mention the reviewer, rather than just ignore.
 		// Note: assuming assignees are reviewers here. If not true anymore, also notify assignees.
-		obj.WriteComment(fmt.Sprintf("LGTM from users (%v). Can reviewers (%v) review this?", atUserFormat(users), atUserFormat(reviewers)))
+		obj.WriteComment(
+			fmt.Sprintf("LGTM from users (%v). Can reviewers (%v) review this?",
+				mungerutil.JoinMentions(mungerutil.GetUsersList(users)),
+				mungerutil.JoinMentions(activeUsers.Assignees())))
```
lavalamp: style nit: closing parens on lines than balance out their opening paren. (you can add extra commas if need be)

mungegithub/mungers/mungerutil/util.go
```diff
@@ -16,14 +16,137 @@ limitations under the License.
...
+)
+
+// ActiveUsers tracks Users involved in a github Issue
+type ActiveUsers struct {
+	assignees []string
+	author    string
+}
+
+func mergeUsers(userslists ...[]string) []string {
+	uniqueUsers := map[string]struct{}{}
```
lavalamp: we have a pkg/util/sets with a String class that does this for you.

lavalamp: This whole function becomes a one liner: `sets.NewString(userslists...).List()`

mungegithub/mungers/mungerutil/util.go
```diff
@@ -16,14 +16,137 @@ limitations under the License.
...
+	"github.com/google/go-github/github"
+)
+
+// ActiveUsers tracks Users involved in a github Issue
+type ActiveUsers struct {
+	assignees []string
+	author    string
+}
+
+func mergeUsers(userslists ...[]string) []string {
```
lavalamp: userslist? It's not multiple lists...

mungegithub/mungers/mungerutil/util.go
```diff
@@ -16,14 +16,137 @@ limitations under the License.
...
+}
+
+// IsAssignee checks if a user is an assignee
+func (u *ActiveUsers) IsAssignee(user *github.User) bool {
+	if GetUser(user) == nil {
+		return false
+	}
+	username := *GetUser(user)
+
+	for _, assignee := range u.Assignees() {
```
lavalamp: Maybe better to just store a sets.String here instead, then you can just do `return u.assignees.Has(username)`.

mungegithub/mungers/mungerutil/util.go
apelisse: Cool!

mungegithub/mungers/mungerutil/util.go
```diff
@@ -16,14 +16,137 @@ limitations under the License.
...
+	return &ActiveUsers{
+		assignees: mergeUsers(
+			GetUsersList(assignees),
+			GetUsersList([]*github.User{assignee})),
+		author: *authorName,
+	}
+}
+
+// MentionUsers adds a @ before each name in the list
+func MentionUsers(users []string) []string {
```
lavalamp: consider making a userlist type with methods instead of having a bunch of functions that take a list of users.

mungegithub/mungers/mungerutil/util.go
```diff
@@ -16,14 +16,137 @@ limitations under the License.
...
+func CreateActiveUsersFromIssue(issue *github.Issue) *ActiveUsers {
+	return createActiveUsers(
+		issue.Assignee,
+		issue.Assignees,
+		issue.User,
+	)
+}
+
+// GetIssueAuthor returns the author of the issue
+func GetIssueAuthor(issue *github.Issue) string {
```
lavalamp: Consider making an IssueUsers type instead of making a bunch of loose functions.

lavalamp: e.g. instead folks would write `NewIssueUsers(issue).Author()`.

apelisse:
PTAL @foxish @lavalamp

lavalamp:
I am satisfied, I will let @foxish make a final pass.

foxish:
Very clean and readable I think. Especially the `.AllUsers().Mention().Join()` style.

apelisse:
I don't think it makes a big difference, but I initially split because it made sense to me to keep it separated :-)

Issue #1415, opened 2016-07-21
[mungegithub] Modify mungegithub to use config files instead of command-line arguments.


foxish:
The number of command-line arguments is getting large and unwieldy. We should switch to reading config files instead.
It would also let us group these command-line options into logical sub-units which is currently not possible.

eparis:
I've not looked at it, but I know `github.com/spf13/viper` is supposed to help add config files on top of cobra/pflag...

foxish:
@eparis I will take a look at that. Thanks!

foxish:
This issue was moved to kubernetes/test-infra#1633

PR #1461, opened 2016-08-01
move kubemark to nonblocking

lavalamp:
@gmarek & @wojtek-t... I'm really sorry but when I told @quinton-hoole I wasn't taking any more post-submit tests that didn't have corresponding presubmit tests (e.g., federation tests), he was like, "Kubemark does that," and I was like, "Well, I'll kick it out next time it causes a prolonged blockage." Still trying to figure out what the current problem is, if it's not a particular PR then I'll reconsider.

### DIFF
--- a/mungegithub/submit-queue/deployment/kubernetes/configmap.yaml
+++ b/mungegithub/submit-queue/deployment/kubernetes/configmap.yaml
@@ -16,10 +16,10 @@ data:
 
   # submit-queue options.
   submit-queue.required-contexts: "Jenkins GCE Node e2e"
-  submit-queue.nonblocking-jenkins-jobs: kubernetes-e2e-gke-staging,kubernetes-e2e-gke-staging-parallel,kubernetes-e2e-gce-serial,kubernetes-e2e-gke-serial,kubernetes-e2e-gke-test,kubernetes-e2e-gce-examples,kubernetes-e2e-gce-federation,kubernetes-e2e-gce-scalability,kubernetes-soak-continuous-e2e-gce,kubernetes-soak-continuous-e2e-gke
-  submit-queue.jenkins-jobs: kubelet-gce-e2e-ci,kubernetes-build,kubernetes-test-go,kubernetes-verify-master,kubernetes-e2e-gce,kubernetes-e2e-gce-slow,kubernetes-e2e-gke,kubernetes-e2e-gke-slow,kubernetes-kubemark-5-gce
+  submit-queue.nonblocking-jenkins-jobs: kubernetes-e2e-gke-staging,kubernetes-e2e-gke-staging-parallel,kubernetes-e2e-gce-serial,kubernetes-e2e-gke-serial,kubernetes-e2e-gke-test,kubernetes-e2e-gce-examples,kubernetes-e2e-gce-federation,kubernetes-e2e-gce-scalability,kubernetes-soak-continuous-e2e-gce,kubernetes-soak-continuous-e2e-gke,kubernetes-kubemark-5-gce,kubernetes-kubemark-500-gce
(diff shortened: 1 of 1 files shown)


mungegithub/submit-queue/deployment/kubernetes/configmap.yaml
```diff
@@ -16,10 +16,10 @@ data:
...
-  submit-queue.jenkins-jobs: kubelet-gce-e2e-ci,kubernetes-build,kubernetes-test-go,kubernetes-verify-master,kubernetes-e2e-gce,kubernetes-e2e-gce-slow,kubernetes-e2e-gke,kubernetes-e2e-gke-slow,kubernetes-kubemark-5-gce
+  submit-queue.nonblocking-jenkins-jobs: kubernetes-e2e-gke-staging,kubernetes-e2e-gke-staging-parallel,kubernetes-e2e-gce-serial,kubernetes-e2e-gke-serial,kubernetes-e2e-gke-test,kubernetes-e2e-gce-examples,kubernetes-e2e-gce-federation,kubernetes-e2e-gce-scalability,kubernetes-soak-continuous-e2e-gce,kubernetes-soak-continuous-e2e-gke,kubernetes-kubemark-5-gce,kubernetes-kubemark-500-gce
+  submit-queue.jenkins-jobs: kubelet-gce-e2e-ci,kubernetes-build,kubernetes-test-go,kubernetes-verify-master,kubernetes-e2e-gce,kubernetes-e2e-gce-slow,kubernetes-e2e-gke,kubernetes-e2e-gke-slow
   submit-queue.presubmit-jobs: kubernetes-pull-build-test-e2e-gce
-  submit-queue.weak-stable-jobs: kubernetes-kubemark-500-gce
+  submit-queue.weak-stable-jobs:
```
foxish: Keeping it blank throws errors. Needs to be `"\"\""` because it's a field that accepts a stringSlice.

apelisse: Can we fix that?

foxish: Previously, when we weren't using the configmap, we'd just remove the argument entirely and that would work (because of the default set in code to `[]string{}`. We can't do that anymore since we're populating environment variables with these configmap values. We should be able to once we consume the configmap directly in the pod (https://github.com/kubernetes/contrib/issues/1415), instead of supplying commandline parameters.

lavalamp: Updated, PTAL

On Mon, Aug 1, 2016 at 5:12 PM, Anirudh Ramanathan <notifications@github.com

foxish:
LGTM

lavalamp:
I reverted https://github.com/kubernetes/kubernetes/pull/28955 locally and was able to start a kubemark cluster. However it's time to go home so I think I'm going to push this configmap change so the queue can make progress.

lavalamp:
I pushed this config and restarted. I still haven't confirmed whether that PR is the problem.

lavalamp:
(I will try to confirm tonight if I have a chance)

lavalamp:
Yeah, so that PR definitely was the culprit and a presubmit here would have caught this and saved me hours today, and kept the queue moving.

wojtek-t:
@gmarek - we should fix this ASAP and:
- add small-ish kubemark to presubmit (like kubemark-50 or kubemark-100 or sth like that - it shouldn't be expensive)
- readd kubemark-500 to blocking

wojtek-t:
Filed https://github.com/kubernetes/kubernetes/issues/29897 for this

PR #1462, opened 2016-08-02
GCE ingress http-only annotation naming fix

FourSigma:
```
GCE ingress http-only annotation naming fix
```


FourSigma:
I signed it!

bprashanth:
Thanks, LGTM

PR #1468, opened 2016-08-02
Adding submit-queue munger to the contrib deployment.

foxish:
The check-labels munger (https://github.com/kubernetes/contrib/pull/1306) needs to be merged before, so that it can create the necessary labels.

Also adding the submit-queue munger to the deployment to enable automatic merging of PRs with LGTM, after they pass CI tests.

### DIFF
--- a/mungegithub/submit-queue/deployment/contrib/configmap.yaml
+++ b/mungegithub/submit-queue/deployment/contrib/configmap.yaml
@@ -7,7 +7,7 @@ data:
   config.http-cache-dir: /cache/httpcache
   config.organization: kubernetes
   config.project: contrib
-  config.pr-mungers: blunderbuss,size
+  config.pr-mungers: blunderbuss,submit-queue
   config.state: open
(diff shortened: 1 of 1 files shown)


mungegithub/submit-queue/deployment/contrib/configmap.yaml
```diff
@@ -7,7 +7,7 @@ data:
   config.http-cache-dir: /cache/httpcache
   config.organization: kubernetes
   config.project: contrib
-  config.pr-mungers: blunderbuss,size
```
lavalamp: Why remove size?

foxish: The currently running deployment has size removed as well. We need the labels to be created using the check-labels munger before turning on size, or it complains about being unable to find various size labels.

lavalamp: ah ok

PR #1469, opened 2016-08-02
Removing non-org members

foxish:
@aledbf
@danehans
@simonswine
@rutsky
@itayariel
@stephenrlouie
@mattthias
@jdavis7257

Please add yourself to the kubernetes org, so that PR assignments can be made.

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,5 +1,4 @@
 assignees:
-- aledbf
 - bprashanth
 - eparis
 - lavalamp
--- a/ansible/OWNERS
+++ b/ansible/OWNERS
@@ -1,7 +1,4 @@
 assignees:
-- danehans
 - eparis
 - ingvagabund
-- rutsky
-- stephenrlouie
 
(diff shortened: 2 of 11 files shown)


lavalamp:
LGTM

lavalamp:
All mentioned folks--happy to add you back here, but merging this now to reduce the number of failed github API requests our bot is making.

eparis:
we need a better way to get people into the org. Doesn't seem like you can request, only someone can add you...

lavalamp:
I'm pretty sure you can request, I've gotten emails saying people have requested to be in-- maybe you have to request at the group rather than the org level? If that's the case, I think there's a "collaborators" group.

simonswine:
I think the request is only possible, if you are already a member of any other team in the organisation. Happy to join the org if someone with privileges to do hat is around...

lavalamp:
I see. I'll invite all of the mentioned people, then. :)

lavalamp:
Invites sent.

simonswine:
@lavalamp thanks for the invite. Probably by tomorrow we can revert the PR

aledbf:
@lavalamp thanks for the invite

PR #1471, opened 2016-08-03
Turning on blunderbuss on the docs submit-queue.

foxish:


mungegithub/submit-queue/deployment/docs/configmap.yaml
```diff
@@ -7,7 +7,7 @@ data:
   config.http-cache-dir: /cache/httpcache
   config.organization: kubernetes
   config.project: kubernetes.github.io
-  config.pr-mungers: size
```
apelisse: Do we not care about size?

foxish: We cannot turn it on before the check-labels munger creates the size labels for this repository. We'll turn it back on after that is done.

foxish:
@lavalamp What would be the right thing to do here? If we have no e2e tests, should we disable the `@k8s-bot test this` comment and merge as soon as we get an LGTM?

apelisse:
Wait, it must at least wait for travis-ci to pass.

foxish:
travis-ci is a [required context]((https://github.com/kubernetes/contrib/blob/master/mungegithub/submit-queue/deployment/contrib/configmap.yaml#L19) in the contrib deployment, so we do wait for it to pass before we merge.

apelisse:
I got scared by the "merge as soon as we get an LGTM" :-)

Cool!

PR #1473, opened 2016-08-03
Adding labels.yaml for check-labels munger.

foxish:
List of labels to be created in the contrib repository.
The list may need to be pruned, in case some labels do not make sense for contrib.

### DIFF
--- a/labels.yaml
+++ b/labels.yaml
@@ -1,0 +1,277 @@
+labels:
+  - name: 0 - Backlog
+    color: CCCCCC
+  - name: 0 - Triage
+    color: AAAAAA
+  - name: 1 - Ready
+    color: CCCCCC
+  - name: 2 - Working <= 5
+    color: CCCCCC
+  - name: 3 - Review
+    color: CCCCCC
+  - name: 4 - Done
+    color: CCCCCC
+  - name: approved
+    color: 0ffa16
+  - name: area/HA
+    color: bfd4f2
+  - name: area/admin
+    color: "009800"
+  - name: area/api
+    color: e1ed21
+  - name: area/app-config-deployment
+    color: fb6420
+  - name: area/auth
+    color: 00cc00
+  - name: area/autoscaling
+    color: "5319e7"
+  - name: area/batch
+    color: c7def8
+  - name: area/build-release
+    color: bfe5bf
+  - name: area/cadvisor
+    color: fef2c0
+  - name: area/cloudprovider
+    color: f1fd21
+  - name: area/cluster-federation
+    color: fad8c7
+  - name: area/cluster-lifecycle
+    color: d4c5f9
+  - name: area/devexp
+    color: "5319e7"
+  - name: area/downward-api
+    color: 8fffcf
+  - name: area/example/cassandra
+    color: d4c5f9
+  - name: area/extensibility
+    color: ffaa04
+  - name: area/images-registry
+    color: eba420
+  - name: area/introspection
+    color: bfe5bf
+  - name: area/isolation
+    color: aba420
+  - name: area/kubelet-api
+    color: fbca04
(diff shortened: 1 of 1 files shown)


eparis:
is alphabetical sort possible to help review? There are a bunch we don't need/want...

foxish:
@eparis Done. PTAL.

foxish:
cc @bgrant0607

labels.yaml
```diff
@@ -0,0 +1,285 @@
+labels:
+  - name: 0 - Backlog
```
bgrant0607: Not needed

labels.yaml
```diff
@@ -0,0 +1,285 @@
...
+    color: CCCCCC
+  - name: 0 - Triage
```
bgrant0607: Not needed

labels.yaml
```diff
@@ -0,0 +1,285 @@
...
+    color: AAAAAA
+  - name: 1 - Ready
```
bgrant0607: Not needed

labels.yaml
```diff
@@ -0,0 +1,285 @@
...
+    color: CCCCCC
+  - name: 2 - Working <= 5
```
bgrant0607: Not needed

labels.yaml
```diff
@@ -0,0 +1,285 @@
...
+    color: CCCCCC
+  - name: 3 - Review
+    color: CCCCCC
+  - name: 4 - Done
+    color: CCCCCC
+  - name: '[deprecated--DO NOT USE]old-team/master'
```
bgrant0607: Not needed.

labels.yaml
```diff
@@ -0,0 +1,285 @@
...
+    color: e11d21
+  - name: documentation/better-examples
+    color: f9d0c4
+  - name: documentation/confusing
+    color: f9d0c4
+  - name: documentation/content-gap
+    color: f9d0c4
+  - name: documentation/out-of-date
+    color: f9d0c4
+  - name: e2e-not-required [deprecated use retest-not-required]
```
bgrant0607: Not needed.

labels.yaml
```diff
@@ -0,0 +1,285 @@
...
+    color: c7def8
+  - name: kind/flake
+    color: f7c6c7
+  - name: kind/friction
+    color: c7def8
+  - name: kind/mesos-flake
+    color: f7c6c7
+  - name: kind/new-api
+    color: c7def8
+  - name: kind/old-docs
```
bgrant0607: Not needed.

labels.yaml
```diff
@@ -0,0 +1,285 @@
...
+    color: 0052cc
+  - name: component/kubelet
+    color: 0052cc
+  - name: component/node-e2e
+    color: 0052cc
+  - name: component/nodecontroller
+    color: 0052cc
+  - name: component/ui
+    color: 0052cc
+  - name: davidopp
```
bgrant0607: Not needed.

bgrant0607:
I don't think it's that important to prune this list. I commented on a few -- deprecated labels, etc. 

LGTM after those are removed.

eparis:
I wondered if we should come up with a 'standard' list. One without all of the `areas`. Maybe even without all of the `teams`. These don't hurt. But they don't help either...

bgrant0607:
@eparis 

It would be helpful to identify labels upon which the bot actually depends.

I do hope we can use the issue classifier on multiple repos, but I agree not all labels make sense for all repos.

I'd actually like to rename many of the labels (e.g., team and component labels) before we copy them to lots of repos.

foxish:
@bgrant0607 Done!

bgrant0607:
LGTM

PR #1474, opened 2016-08-03
Adding early exit condition for submit-queue merging

foxish:
When there are no e2e tests to be run, we can merge after LGTM. Pre-submit tests are already being performed as part of https://github.com/foxish/contrib/blob/4216d7608c9906edc5607c4092f4d8119f4511fc/mungegithub/mungers/submit-queue.go#L1133

PTAL @lavalamp @eparis

### DIFF
--- a/mungegithub/mungers/submit-queue.go
+++ b/mungegithub/mungers/submit-queue.go
@@ -1148,35 +1148,8 @@ func (sq *SubmitQueue) doGithubE2EAndMerge(obj *github.MungeObject) bool {
 		}
 		glog.Infof("Skipping retest since head and base sha match previous attempt!")
 		atomic.AddInt32(&sq.retestsAvoided, 1)
-	} else {
-		if err := obj.WriteComment(retestBody); err != nil {
-			glog.Errorf("%d: unknown err: %v", *obj.Issue.Number, err)
-			sq.SetMergeStatus(obj, unknown)
-			return true
-		}
-
-		// Wait for the retest to start
-		sq.SetMergeStatus(obj, ghE2EWaitingStart)
-		atomic.AddInt32(&sq.prsTested, 1)
-		err = obj.WaitForPending(sq.RequiredRetestContexts)
-		if err != nil {
-			sq.SetMergeStatus(obj, fmt.Sprintf("Failed waiting for PR to start testing: %v", err))
-			return true
-		}
-
-		// Wait for the status to go back to something other than pending
-		sq.SetMergeStatus(obj, ghE2ERunning)
-		err = obj.WaitForNotPending(sq.RequiredRetestContexts)
-		if err != nil {
-			sq.SetMergeStatus(obj, fmt.Sprintf("Failed waiting for PR to finish testing: %v", err))
-			return true
-		}
-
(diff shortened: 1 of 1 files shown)


mungegithub/mungers/submit-queue.go
```diff
@@ -1140,6 +1140,13 @@ func (sq *SubmitQueue) doGithubE2EAndMerge(obj *github.MungeObject) bool {
 		return true
 	}
 
+	if len(sq.RequiredRetestContexts) == 0 && len(sq.BlockingJobNames) == 0 && len(sq.WeakStableJobNames) == 0 {
```
eparis: I'd kinda rather we moved all of the 'retest' stuff around https://github.com/kubernetes/contrib/pull/1474/files#diff-2ab7cb9ccaa72dce30323c5555c00c66L1152 into its own function which just returned immediately if there was nothing to retest. And make sure that `sq.E2EStalbe()` returns quickly.

foxish: Done. PTAL.

foxish:
@eparis Ping. The submit-queue on contrib doesn't merge things unless [retest-not-needed] label is applied, due to this issue.

apelisse:
@foxish I think the point of re-running the pre-submit is that you may test against a different base commit.

foxish:
If I understand you correctly, you're saying that the presubmit job should be re-run on contrib prior to submission. That makes sense, and I'll make a PR such that we retest with the same presubmit job, and it should stop the stalling that is currently occurring because we supply an empty retest-contexts argument to the contrib submit-queue.

However, this PR still would need to go through before we can turn on the submit-queue on the docs repo, because it has no test, or re-test contexts.

foxish:
@apelisse In order to re-test in this repo, we would need to teach k8s-bot to test this repository (in a manner similar to [this](https://github.com/kubernetes/test-infra/blob/3cbf8d996ce6a7cb55ea3b27f295848e79d27de7/jenkins/job-configs/kubernetes-jenkins-pull/kubernetes-pull.yaml#L504)). Do we need that for this repository?

mungegithub/mungers/submit-queue.go
```diff
@@ -1149,6 +1149,26 @@ func (sq *SubmitQueue) doGithubE2EAndMerge(obj *github.MungeObject) bool {
...
+		return true
+	}
+
+	sq.mergePullRequest(obj)
+	return true
+}
+
+// Returns true if we can discard the PR from the queue, false if we must keep it for later.
+func (sq *SubmitQueue) retestPR(obj *github.MungeObject) bool {
+	if len(sq.RequiredRetestContexts) > 0 {
```
eparis: style nit: Invert this logic to `return false` if `len(sq.RequiredRetestContexts) == 0` and remove a level on indentation. Then I'll happy.

foxish: Fixed! PTAL. Thanks!

eparis:
oh yeah, this is going to sit for 45 minutes until the restest request times out...

foxish:
@eparis Hopefully for the last time... :)

PR #1475, opened 2016-08-03
Deployment flag to control owners embedded in md.

foxish:
The docs repository reads owners from the front-matter in markdown
files as well as dedicated OWNERS files.

cc @pwittrock @lavalamp


lavalamp:
In a followup, can you change the flag name? It's not at all clear what this does based on the flag name.

lavalamp:
Hm, seems like we need to tell it not to ask jenkins for a retest in the contrib repo.

foxish:
@lavalamp, working on it in https://github.com/kubernetes/contrib/pull/1474.

PR #1485, opened 2016-08-04
Cluster-autoscaler: max node total count flag for CIDR

mwielgus:
In some cases we cannot grow the cluster beyond a certain number of nodes. This PR adds a flag to pass this value + appropriate handling in scale up. The cluster won't be forcefully scaled down to meet the passed value.

cc: @piosz @fgrzadkowski

### DIFF
--- a/cluster-autoscaler/scale_up.go
+++ b/cluster-autoscaler/scale_up.go
@@ -36,9 +36,10 @@ type ExpansionOption struct {
 }
 
 // ScaleUp tries to scale the cluster up. Return true if it found a way to increase the size,
-// false if it didn't and error if an error occured.
+// false if it didn't and error if an error occured. Assumes that all nodes in the cluster are
+// ready and in sync with instance groups.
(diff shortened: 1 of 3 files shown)


cluster-autoscaler/scale_up.go
```diff
@@ -122,6 +123,15 @@ func ScaleUp(unschedulablePods []*kube_api.Pod, nodes []*kube_api.Node, cloudPro
 			glog.V(1).Infof("Capping size to MAX (%d)", bestOption.nodeGroup.MaxSize())
 			newSize = bestOption.nodeGroup.MaxSize()
 		}
+
+		if maxNodesTotal > 0 && len(nodes)+(newSize-currentSize) > maxNodesTotal {
+			glog.V(1).Infof("Capping size to max cluster total size (%d)", maxNodesTotal)
```
piosz: Shouldn't an event be exported also here?

piosz:
LGTM, just a nit

cluster-autoscaler/scale_up.go
mwielgus: Next pr. Same comment applies to node pool limits.

PR #1489, opened 2016-08-05
submit-queue: Display uncached request count to github API

apelisse:
We currently display number of total requests to github API, and number
of cached requests. Let's print number of uncached request as this is
the number that matters the most.


foxish:
LGTM

eparis:
You could change the default sort order, aka change `self.StatOrder = "-Count";` in `script.js` to show what's most important. But this is good/fine/lgtm.

foxish:
@eparis We should no longer need the retest-not-required label as https://github.com/kubernetes/contrib/pull/1474 is live on contrib.

PR #1494, opened 2016-08-05
Reverting commit to README.md.

foxish:
Edited using the web UI :(


lavalamp:
Web UI is dangerous for those of us with powers :)

foxish:
Duly noted. Bad habits from when I didn't have powers. :)

apelisse:
> Duly noted. Bad habits from when I didn't have powers. :)

That doesn't make any sense! Did you get the habit of clicking, changing the text and getting permissions denied? :-D

foxish:
Oh, without powers, it creates a branch and makes a PR when I try to edit something if I recall correctly.

PR #1497, opened 2016-08-07
Makes use of a config-file possible along with command-line options

### LINKED ISSUE #1415: [mungegithub] Modify mungegithub to use config files instead of command-line arguments.

foxish:
Fixes https://github.com/kubernetes/contrib/issues/1415.

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1497)


mungegithub/submit-queue/deployment/contrib/configmap.yaml
```diff
@@ -3,41 +3,42 @@ kind: ConfigMap
...
-  config.organization: kubernetes
-  config.project: contrib
-  config.pr-mungers: blunderbuss,lgtm-after-commit,submit-queue,check-labels
-  config.state: open
-  config.token-file: /etc/secret-volume/token
-  gitrepo.repo-dir: /gitrepos
-  gitrepo.enable-md-yaml: "false"
+  munge-config: |
+    # basic config options.
+    http-cache-dir: /cache/httpcache
```
eparis: Should we leave things like which, which are defined in the deployment template, as command line flags? So both defining the location of the volume mounts and telling the app where those volume mounts are can be changed together? (This is true for a couple of these flags)

eparis:
CI failed because you didn't vendor all of the deps (`godep save ./...`) 

I'm also unsure if we are allows to use `BurntSushi/toml`. I remember something about that from @bgrant0607 and it is a dep of viper.

foxish:
@eparis Tried again with `godep save` like you mentioned. Hopefully that should work. BurntSushi/toml is under the wtfpl, which might be why we cannot use it. I could make a fork of viper that does not use it, because we don't use the TOML parser anyway if necessary.

foxish:
I realize that the way I've fixed the issue right now, removing the toml parsing from the vendored import files is not the right way. Would creating a fork of viper without that package be the right step? I can see that there was an effort to get rid of it in https://github.com/spf13/viper/commit/abafbf243b9583ac1360ae4ef680d8acaa361089 but it was reverted.

eparis:
I'll wait for @bgrant0607 to say +1 -1 on the state of the toml parser.

mungegithub/submit-queue/deployment/contrib/configmap.yaml
apelisse: I think this is a good point

mungegithub/mungegithub.go
```diff
@@ -128,6 +129,22 @@ func main() {
 		r.AddFlags(root, &config.Config)
 	}
 
+	// set values from config file if available.
+	viper.SetConfigName("munge-config")
+	viper.SetConfigType("yaml")
+	viper.AddConfigPath("/etc/config/")
+	viper.AddConfigPath(".")
+	if err := viper.ReadInConfig(); err != nil {
+		glog.Warningf("Cannot read configuration file: %v", err)
```
apelisse: So the config is not an option itself, it's a mandatory thing?

foxish: It currently reads a config file of that name if present in one of those two locations, and sets the various option flags. If a config file is not present, it will use the flag values supplied via the commandline  

I guess we should make the configfile path an option, so that it is clearer _where_ we're reading values from. Good point.

mungegithub/submit-queue/deployment/contrib/configmap.yaml
foxish: Agreed. We don't want someone changing one without changing the other. I'll make this change.

mungegithub/submit-queue/deployment/contrib/configmap.yaml
```diff
@@ -3,41 +3,42 @@ kind: ConfigMap
 metadata:
   name: contrib-sq-config
 data:
-  # basic config options.
```
apelisse: Would it make sense to "create configmap --from-file" instead?

bgrant0607:
That wasn't resolved by https://github.com/spf13/viper/pull/217 ?

cc @caesarxuchao

bgrant0607:
Hmm. Ok, I guess not.

caesarxuchao:
Unfortunately https://github.com/spf13/viper/pull/217 got reverted.

mungegithub/mungegithub.go
foxish: Done.

mungegithub/submit-queue/deployment/contrib/configmap.yaml
foxish: I think that's a good idea. I don't want to change the Makefile as well in this PR, I'll check that approach in a subsequent one.

foxish:
@eparis, addressed comments and fixed the parser issue by forking viper as of https://github.com/spf13/viper/pull/217 and using that as our dependency instead. PTAL.

mungegithub/submit-queue/deployment/contrib/configmap.yaml
foxish: Done.

mungegithub/mungegithub.go
```diff
@@ -25,12 +25,13 @@ import (
 	"k8s.io/contrib/mungegithub/features"
 	github_util "k8s.io/contrib/mungegithub/github"
 	"k8s.io/contrib/mungegithub/mungers"
+	"k8s.io/contrib/mungegithub/mungers/mungerutil"
 	"k8s.io/contrib/mungegithub/reports"
 	utilflag "k8s.io/kubernetes/pkg/util/flag"
 
+	"github.com/foxish/viper"
```
lavalamp: Probably shouldn't point at your own fork of this. Are you waiting for changes to get merged upstream?

foxish: There is a discussion in the comments. There was a patch to replace the TOML library that was being used (as it is under the WTFPL), but it was reverted. @bgrant0607 suggested that we've waited long enough and should just fork it.

lavalamp:
wow that dependency tree is huge. Is viper really doing something that takes more than a few hundred lines of code?

mungegithub/mungegithub.go
eparis: Since the toml library which was reverted has now been fixed, I'm ok with the fork, but can you take the lead of trying to update upstream viper again?

foxish: @eparis Sure, I can do that. However, @lavalamp made a good point that the huge number of dependencies we're adding for relatively simple reading of a config YAML might not be worth it. So, I think I might move away to reading the yaml ourselves in this case, unless there is a good reason to stick with viper.

apelisse:
Reviewed 6 of 6 files at r1, 362 of 362 files at r2.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.kubernetes.io:443/reviews/kubernetes/contrib/1497)*

test-foxish:
recomputing cla status...

foxish:
Closing for now, as there is no real need for this, and spf13/viper can be used directly in a follow-up as toml library licensing issues have been resolved.

PR #1498, opened 2016-08-07
[nginx-ingress-controller] Refactoring of template handling

### LINKED ISSUE #1457: NGINX Ingress Controller - net.core.somaxconn=128
The setup works fine (btw thank you community :) ), however the logs is flooded with:
```
W0801 05:21:58.225140       1 utils.go:231] system net.core.somaxconn=128. Using NGINX default (511)
```

Is there any way to tune `net.core.somaxconn` or ignore the log?

aledbf:
This also watch for changes in the template

fixes #1457

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1498)


ingress/controllers/nginx/nginx/template/template.go
```diff
@@ -54,48 +56,86 @@ var (
...
+	if err != nil {
+		return nil, err
+	}
+
+	template := &Template{
+		tmpl:    tmpl,
+		watcher: watcher,
+	}
+
+	go func() {
```
bprashanth: I would vastly prefer if we did this in a compsable way. Perhaps we can have a sidecar that watches the config map, notices a change, and forces a reload on the nginx controller through a /reload http endpoint ? that would make for a nice reusable component. 
```
$ watch-configmaps --reload-url="url-in-same-netns" --file="/path/to/file"
```

When we upgrade to docker1.12 we will get shared pid namespaces and can consider replacing the --reload-url arg with just `pkill SIGHIUP nginx`. 

The *most* reusable way to do this right now would be to avoid even the http server, and do:
```
$ watch-configmaps --pod-name="foopod, defaults to self" \
    --container-name="nginx, defaults to first peer container in list of own pod" \
    --file="/path/to/file" \
    --script="/reload.sh"
```

That would kubectl exec in the give pod and exec the the signal to the given container/process. If you're interested you can download the kubectl client from: 
`wget https://storage.googleapis.com/kubernetes-release/release/${KUBERNETES_VERSION}/bin/linux/amd64/kubectl`

Or exec through a REST client like we do in our e2es: 
https://github.com/kubernetes/kubernetes/blob/4bb5fdc47f4c409f7d290b3af348fb135d0daf0f/test/e2e/framework/util.go#L3756

Quite a few people have asked for something like the last one. I would put this under ingress/ so it doesn't get lost when we split ingress out of contrib.

aledbf: > I would vastly prefer if we did this in a composable way

Agree. I'll start working on this.

aledbf:
@bprashanth the last commit contains a sidecar that watch changes in a file or a kubernetes resource (`configmap` and `secrets` for now).
Writing the [example](https://github.com/aledbf/contrib/blob/517275f35169a79c6851723a5d1bf0b6be3d5de4/ingress/controllers/nginx/examples/watch-template/watch.yaml) I realized that this requires the use of volumes (at least `emptyDir`, copying the template before the tstart of the controller) to be able to watch the file located in a different container.

Is this change acceptable just to be able to use a sidecar and not add this code in the controller?

aledbf:
@bprashanth ping

ingress/controllers/nginx/examples/watch-template/watch.yaml
```diff
@@ -0,0 +1,111 @@
...
+        k8s-app: nginx-ingress-lb
+        name: nginx-ingress-lb
+    spec:     
+      terminationGracePeriodSeconds: 60
+      containers:
+      - image: aledbf/watch-resource-cmdrunner:0.1
+        name: watch-template
+        command:
+        - /watch-resource-cmdrunner
+        - --command="echo 'nginx template updated'"
```
bprashanth: don't you want to reload nginx when this happens?

ingress/controllers/nginx/examples/watch-template/watch.yaml
```diff
@@ -0,0 +1,111 @@
...
+        - --file /etc/nginx/template/nginx.tmpl
```
bprashanth: Can you provide an example where this is actuallly a configmap mounted into both pods, and the watcher sends the reload signal to nginx?

ingress/watch-resource/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM alpine:3.4
+
+COPY watch-resource-cmdrunner /
```
bprashanth: Would be nice to make this part of the generic https example also: https://github.com/kubernetes/kubernetes/issues/30532, since it is a modular sidecar that watches and reloads

ingress/watch-resource/configmap_watcher.go
```diff
@@ -0,0 +1,109 @@
...
+	"log"
+
+	"k8s.io/kubernetes/pkg/api"
+	"k8s.io/kubernetes/pkg/client/unversioned"
+	"k8s.io/kubernetes/pkg/registry/generic"
+	"k8s.io/kubernetes/pkg/watch"
+)
+
+// ConfigmapWatcher ...
+type ConfigmapWatcher struct {
```
bprashanth: I'm not sure why I'd ever want to watch the config map directly, can you clarify the use case? the sidecar might see the config map change through watch before the app container, and the reload will be too soon. Or is the sidecar also responsible for writing the configmap into a shared volume (in which case it's duplicating kubelet functionality).

ingress/controllers/nginx/examples/watch-template/watch.yaml
aledbf: Right. The correct content should be `curl http://localhost:10254/reload-template`

bprashanth:
Suggest simplifying the pr to just watch configmaps mounted into 2 containers, and send a signal from the first to the second through kubeClient. Then write a good example that uses a configmap, modifies it , and has an nginx reload. We can iterate on growing that out as required.

ingress/controllers/nginx/examples/watch-template/watch.yaml
aledbf: Like this https://gist.github.com/aledbf/e9e669657ab00ae2f8562e89e47c3cb1

aledbf:
> Suggest simplifying the pr to just watch configmaps mounted into 2 containers, and send a signal from the first to the second through kubeClient. 

Just in case if an user just upgrades the current version (without a volume) the controller behaves like now (no reload if the template changes)

> Then write a good example that uses a configmap, modifies it , and has an nginx reload.

I think this  https://gist.github.com/aledbf/e9e669657ab00ae2f8562e89e47c3cb1 is that example

aledbf:
> Suggest simplifying the pr to just watch

done

[Here](https://github.com/kubernetes/contrib/pull/1498/commits/1385a049ee408b3be770d8fae344bcf518db77d8#diff-21d9bcf6a62d0c34b86a1ead0de11faaR83) is the important part

Basically I watch for the template file for changes. Is not important from where the file comes (emptyDir, another volume, configmap, etc).

aledbf:
This is an example of the change

```
kubectl exec nginx-ingress-controller-7g2rv -- echo "" >> /etc/nginx/template/nginx.tmpl
```

```
I0825 00:21:01.317244       1 main.go:94] new NGINX template loaded
I0825 00:21:04.832265       1 command.go:81] change in configuration detected. Reloading...
```


```
kubectl exec nginx-ingress-controller-7g2rv -- echo "{{ something }}" >> /etc/nginx/template/nginx.tmpl
```

```
W0825 00:21:59.913088       1 main.go:88] invalid NGINX template: template: nginx.tmpl:370: function "something" not defined
```

aledbf:
Also contains update to godeps required by #1623

ingress/controllers/nginx/main.go
```diff
@@ -195,3 +188,28 @@ func handleSigterm(lbc *loadBalancerController) {
...
+
+const (
+	defTmpl      = "/etc/nginx/template/nginx.tmpl"
+	fallbackTmpl = "/etc/nginx/nginx.tmpl"
+)
+
+// checkTemplate verifies the NGINX template exists (file /etc/nginx/template/nginx.tmpl)
+// If the file does not exists it means:
+// a. custom docker image
+// b. standard image using watch-resource sidecar with emptyDir volume
```
bprashanth: why emptyDir volume? isn't this just like the nginx example in main repo, except this sidecar is reusable with *any* existing pod that mounts a config map?

ingress/controllers/nginx/main.go
```diff
@@ -158,12 +151,12 @@ func registerHandlers(lbc *loadBalancerController) {
...
 	healthz.InstallHandler(mux, lbc.nginx)
 
-	http.HandleFunc("/build", func(w http.ResponseWriter, r *http.Request) {
+	mux.HandleFunc("/build", func(w http.ResponseWriter, r *http.Request) {
 		w.WriteHeader(http.StatusOK)
 		fmt.Fprintf(w, "build: %v - %v", gitRepo, version)
 	})
 
-	http.HandleFunc("/stop", func(w http.ResponseWriter, r *http.Request) {
+	mux.HandleFunc("/stop", func(w http.ResponseWriter, r *http.Request) {
```
bprashanth: maybe also add a handler for https://github.com/kubernetes/contrib/issues/1507 ?

ingress/controllers/nginx/nginx/template/file_watcher.go
```diff
@@ -0,0 +1,72 @@
...
+	err := fw.watch()
+	return fw, err
+}
+
+func (f fileWatcher) close() error {
+	return f.watcher.Close()
+}
+
+// watch creates a fsnotify watcher for a file and create of write events
+func (f *fileWatcher) watch() error {
```
bprashanth: I though the plan was to break this out? did we decide against that? or iterating?

ingress/controllers/nginx/main.go
aledbf: > maybe also add a handler for #1507 ?

Let's do this in a following PR

ingress/controllers/nginx/nginx/template/file_watcher.go
aledbf: > or iterating?

I think is better to iterate to see the behavior with this change.

bprashanth:
LGTM

tjliupeng:
@aledbf , I change the content of /etc/nginx/template/ngnix.tmpl in the running container, but no new nginx.conf is generated. What do I miss?

PR #1501, opened 2016-08-09
Correctly handle ingress.class in GCE controller 

bprashanth:
This bug would only get activated when a user has both `ingess.class=gce` and `ingress.class=nginx` ingresses active in the same GCE/GKE cluster, and would manifest as a set of cloud resources created wastefully for the `ingress.class=nginx` ingress as well.

We were previously only ignoring ingress.class (documented here: https://github.com/kubernetes/contrib/blob/master/ingress/controllers/gce/BETA_LIMITATIONS.md#disabling-glbc) when the ingress was created/deleted/modified. There's a chance another ingress with the correct class results in us entering the `sync` routine and listing all ingresses. The listing routine was not smart enough to ignore `ingress.class=nginx`, so we ended up creating resources for the nginx ingress anyway. 

The second commit fixes some of the nginx examples to include a `readiness` probe that is == liveness probe. 

Minhan or Girish, whichever one has spare cycles first.

### DIFF
--- a/ingress/controllers/gce/controller/utils.go
+++ b/ingress/controllers/gce/controller/utils.go
@@ -188,7 +188,10 @@ type StoreToIngressLister struct {
 // List lists all Ingress' in the store.
 func (s *StoreToIngressLister) List() (ing extensions.IngressList, err error) {
 	for _, m := range s.Store.List() {
-		ing.Items = append(ing.Items, *(m.(*extensions.Ingress)))
+		newIng := m.(*extensions.Ingress)
+		if isGCEIngress(newIng) {
+			ing.Items = append(ing.Items, *newIng)
+		}
 	}
 	return ing, nil
 }
--- a/ingress/controllers/gce/Makefile
+++ b/ingress/controllers/gce/Makefile
@@ -1,7 +1,7 @@
 all: push
 
 # 0.0 shouldn't clobber any released builds
-TAG = 0.7.0
+TAG = 0.7.1
 PREFIX = gcr.io/google_containers/glbc
 
 server:
--- a/ingress/controllers/gce/controller/controller.go
+++ b/ingress/controllers/gce/controller/controller.go
@@ -291,7 +291,10 @@ func (lbc *LoadBalancerController) sync(key string) (err error) {
 	}
 	nodePorts := lbc.tr.toNodePorts(&paths)
 	lbNames := lbc.ingLister.Store.ListKeys()
-	lbs, _ := lbc.ListRuntimeInfo()
+	lbs, err := lbc.ListRuntimeInfo()
+	if err != nil {
+		return err
+	}
 	nodeNames, err := lbc.getReadyNodeNames()
 	if err != nil {
 		return err
@@ -404,14 +407,17 @@ func (lbc *LoadBalancerController) updateIngressStatus(l7 *loadbalancers.L7, ing
 
 // ListRuntimeInfo lists L7RuntimeInfo as understood by the loadbalancer module.
 func (lbc *LoadBalancerController) ListRuntimeInfo() (lbs []*loadbalancers.L7RuntimeInfo, err error) {
-	for _, m := range lbc.ingLister.Store.List() {
-		ing := m.(*extensions.Ingress)
-		k, err := keyFunc(ing)
+	ingList, err := lbc.ingLister.List()
+	if err != nil {
+		return lbs, err
+	}
+	for _, ing := range ingList.Items {
+		k, err := keyFunc(&ing)
 		if err != nil {
 			glog.Warningf("Cannot get key for Ingress %v/%v: %v", ing.Namespace, ing.Name, err)
 			continue
 		}
-		tls, err := lbc.tlsLoader.load(ing)
+		tls, err := lbc.tlsLoader.load(&ing)
 		if err != nil {
 			glog.Warningf("Cannot get certs for Ingress %v/%v: %v", ing.Namespace, ing.Name, err)
 		}


ingress/controllers/nginx/examples/default/rc-default.yaml
```diff
@@ -14,11 +14,18 @@ spec:
...
-      terminationGracePeriodSeconds: 60      
+      terminationGracePeriodSeconds: 60
       containers:
       - image: gcr.io/google_containers/nginx-ingress-controller:0.8.2
         name: nginx-ingress-lb
         imagePullPolicy: Always
+        readinessProbe:
+          httpGet:
+            path: /healthz
+            port: 80
```
girishkalele: Are ports 80 and 10249 equivalent inside the container ?

ingress/controllers/nginx/examples/custom-configuration/rc-custom-configuration.yaml
```diff
@@ -14,11 +14,18 @@ spec:
...
-      terminationGracePeriodSeconds: 60      
+      terminationGracePeriodSeconds: 60
       containers:
       - image: gcr.io/google_containers/nginx-ingress-controller:0.8.2
         name: nginx-ingress-lb
         imagePullPolicy: Always
+        readinessProbe:
+          httpGet:
+            path: /healthz
+            port: 80
```
girishkalele: The existing liveness probe uses port 10249 instead of 80 ?

ingress/controllers/gce/controller/utils.go
```diff
@@ -188,7 +188,10 @@ type StoreToIngressLister struct {
 // List lists all Ingress' in the store.
 func (s *StoreToIngressLister) List() (ing extensions.IngressList, err error) {
 	for _, m := range s.Store.List() {
-		ing.Items = append(ing.Items, *(m.(*extensions.Ingress)))
+		newIng := m.(*extensions.Ingress)
+		if isGCEIngress(newIng) {
```
girishkalele: Would we need this check (to exclude GCE ingress controllers) in the nginx controller too ? Or there are no resources wasted so we don't need it there ?

ingress/controllers/nginx/examples/default/rc-default.yaml
aledbf: > Are ports 80 and 10249 equivalent inside the container ?

No. 
NGINX -> 80
Go (controller) -> 10249 -> [this](https://github.com/kubernetes/contrib/blob/master/ingress/controllers/nginx/nginx/command.go#L104) checks for nginx.
(Port 10249 should be used for both probes)

ingress/controllers/gce/controller/utils.go
aledbf: > Would we need this check (to exclude GCE ingress controllers) in the nginx controller too

the check already exists [here] (https://github.com/kubernetes/contrib/blob/master/ingress/controllers/nginx/controller.go#L156)

ingress/controllers/nginx/examples/default/rc-default.yaml
bprashanth: The nginx controller also serves a /healthz: https://github.com/kubernetes/contrib/blob/master/ingress/controllers/nginx/nginx.tmpl#L289
I want the readiness to be on :80, because that's the port that matches the nodeport on the service, so the GCE ingress controller will adopt that as a cloud health check (for this to happen the nodePort's target port and the readiness probe must be the same port).

aledbf: but in a different port `listen 18080`

bprashanth: what's happening here is the readiness probe is getting proxied to default-http-backend/healthz. At this point we have 2 options:     

1. Make the default rc.yamls work behind a cloud provider lb by documenting that the rc must serve a health check on :80, either on some un-claimable url (weird) or by proxying to a health check url on the default backend
2. Document a pointer to a special rc if people want to deploy the nginx lb behind a cloud lb

we can't use another port, because that needs another node port, but the health checks at the cloud provider layer are per node port. I'm doing 2 with this pr, somewhat inadvertently. 

Thoughts on just adding a comment saying: "this is the health check serviced by the default backend, remove it..." something like that?

aledbf: > Make the default rc.yamls work behind a cloud provider lb by documenting that the rc must serve a health check on :80, either on some un-claimable url (weird) or by proxying to a health check url on the default backend

This is not hard to change. I can add a `/healthz` location in the default server.

bprashanth: a default backend can always grab that url and start failing the check though right? eitherways we will have to set some convention and document it.

aledbf: > a default backend can always grab that url and start failing the check though right?

No. Remember that the order of the url are sorted in descending order which means `/` is the last Location. This means `/healthz` is reached before.

bprashanth: well some other ingress then :)

bprashanth: But I think we can just document that healthz/ is special till we have a need for a more rigid protocol. Or just make it more obvious, like `/ingress-controller-healthz`

aledbf: > well some other ingress then :)

good point :)

aledbf: >  Or just make it more obvious, like /ingress-controller-healthz

right.

bprashanth: I've removed the readiness probe changes for now, we can handle it in a follow up pr and discuss convention on a bug

bprashanth:
PTAL, dropped the nginx yaml commit

Issue #1334, opened 2016-07-08
pet/redis: redis-server won't run


simonswine:
I am not too sure what changed, but I am not able to run the redis PetSet.

The `install` and `bootstrap` container succeeds, but somehow the binary links against musl c library which is not available in the latest debian:jessie:

```
root@9e6ec95b7fab:/# ldd /opt/redis/redis-server
        linux-vdso.so.1 (0x00007ffed4774000)
        libc.musl-x86_64.so.1 => not found
```

after install musl everything works fine

``````
root@9e6ec95b7fab:/# apt-get install --reinstall musl
[...]

root@9e6ec95b7fab:/# /opt/redis/redis-server
215:C 08 Jul 16:54:28.225 # Warning: no config file specified, using the default config. In order to specify a config file use /opt/redis/redis-server /path/to/redis.conf
                _._
           _.-``__ ''-._
      _.-``    `.  `_.  ''-._           Redis 3.2.0 (00000000/0) 64 bit
  .-`` .-```.  ```\/    _.,_ ''-._
 (    '      ,       .-`  | `,    )     Running in standalone mode
 |`-._`-...-` __...-.``-._|'` _.-'|     Port: 6379
 |    `-._   `._    /     _.-'    |     PID: 215
  `-._    `-._  `-./  _.-'    _.-'
 |`-._`-._    `-.__.-'    _.-'_.-'|
 |    `-._`-._        _.-'_.-'    |           http://redis.io
  `-._    `-._`-.__.-'_.-'    _.-'
 |`-._`-._    `-.__.-'    _.-'_.-'|
 |    `-._`-._        _.-'_.-'    |
  `-._    `-._`-.__.-'_.-'    _.-'
      `-._    `-.__.-'    _.-'
          `-._        _.-'
              `-.__.-'

``````

@bprashanth

bprashanth:
Might be the image, please try: https://github.com/kubernetes/kubernetes/issues/28078#issuecomment-228629148
@aledbf

simonswine:
yes the e2e example is working for me

simonswine:
@bprashanth t y for the quick reply. Should I create a PR using the e2e image? 

Another solution would be to use jessie as base image like in `redis-install:0.1`

bprashanth:
I think there was a size difference that we tried to optimize. I'm fine with either, if we could add musl to the existing image without the overhead of jessie that would probably be best.

PR #1502, opened 2016-08-09
Revert redis image

bprashanth:
Temporarily revert the image till we can figure out https://github.com/kubernetes/contrib/issues/1334

### DIFF
--- a/pets/redis/redis.yaml
+++ b/pets/redis/redis.yaml
@@ -32,7 +32,7 @@ spec:
         pod.alpha.kubernetes.io/init-containers: '[
             {
                 "name": "install",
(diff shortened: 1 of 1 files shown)


aledbf:
LGTM

bprashanth:
I'll spend some time reducing the image size just like you did and figure out why it didn't work when i have some time, just going to get a working example for now

aledbf:
> figure out why it didn't work

The issue here is that using alpine to compile redis introduces musl as dependency. 
I think that replacing the debian:jessie image for alpine in the petset should fix the issue (but this requires the inspection of the scripts)

PR #1503, opened 2016-08-10
Don't clobber backends.

bprashanth:
If a parallel ingress controller updates the backend service with another instance group, don't clobber it, just make sure the instance group you put in is there.


bprashanth:
planning weekend test runs, can i get lgtm?

freehan:
LGTM. assuming IsSuperset works

PR #1508, opened 2016-08-10
Add kubelet-serial-gce-e2e-ci to non-blocking.

Random-Liu:
Add kubelet-serial-gce-e2e-ci to non-blocking.

/cc @kubernetes/sig-node


Random-Liu:
@dchen1107 Thanks! :) Will merge this after test passes.

PR #1510, opened 2016-08-11
Do not retest if retest-not-required is applied

apelisse:
When a PR hasn't been tested in a while, it is retested by the bot.
This behavior also happens when the PR has the label retest-not-required
or retest-not-required-docs-only, because the label is initially meant
for the submit-queue. Let's not retest if it's not required.


bgrant0607:
@apelisse Thanks!

foxish:
LGTM.  I think we should move this commenting logic into our state machine since it is actually just acting like a pinger activating the jenkins bot.

PR #1514, opened 2016-08-11
git-sync: Fix .git path for worktrees

gregory-lyons:
@mikedanese @huggsboson

Found a bug in my implementation of atomic pulling for git-sync in https://github.com/kubernetes/contrib/pull/1356.

When using "git worktree add" to create a new worktree directory, that worktree directory contains a .git file instead of a .git directory. The .git file is simply a reference to a path within the root .git directory:

```
gitdir: /git/.git/worktrees/rev-...
```

Because that reference path is absolute, any other container that attempts to run a git command inside the worktree will fail if the volume was mounted under a different name. For example, if another container mounts the volume at /repo and then tries to run a git command in one of the worktree directories within /repo, it will fail because the .git file in that worktree will point to a /git path that doesn't exist.

To avoid the bug, we overwrite the .git file in the worktree to use a relative path instead of an absolute path.

```
gitdir: ../.git/worktrees/rev-...
```

### DIFF
--- a/git-sync/main.go
+++ b/git-sync/main.go
@@ -23,6 +23,7 @@ import (
 	"flag"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"log"
 	"math/rand"
 	"os"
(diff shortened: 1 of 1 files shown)


gregory-lyons:
Covered under Box's CLA, have submitted PRs [here](https://github.com/kubernetes/contrib/pull/1356/) and [here](https://github.com/kubernetes/contrib/pull/1431).

gregory-lyons:
For some context, I came across a similar bug that I fixed in https://github.com/kubernetes/contrib/pull/1356/commits/52ad7d8db6697b93ed0977e9feea40621f3a56a2 as part of the initial PR.

That was related to other containers (with different volume mount names) not being able to access the symlinked repo at all, because the symlink was set initially to point to an absolute path instead of a relative path.

This bug was a bit more subtle, as the other containers could access the files in the worktree but just couldn't run any Git commands from inside it.

PR #1519, opened 2016-08-11
Added Total PR Count to Header

grodrigues3:


apelisse:
That looks like a very good idea to me.

cc @eparis what do you think?

foxish:
I agree. It seems like a good idea for consistency with the queue tab. 
I'd expect the number to match the number of open PRs in the repo.

mungegithub/submit-queue/www/index.html
```diff
@@ -34,7 +34,7 @@ <h2 class="md-toolbar-tools">E2E Tests Failing. Entire submit queue blocked. Las
           <md-tab label="PRs" md-on-select="cntl.selectTab('prs')">
             <md-content class="md-whiteframe-z2">
               <md-toolbar>
-                <h2 class="md-toolbar-tools">Pull Requests</h2>
+                <h2 class="md-toolbar-tools">Pull Requests ({{cntl.prsCount}})</h2>
```
eparis: I'm pretty stupid here, but does `{{cntl.prs.length}}` work without the .js change?  Also, do we care that the 'search' will change the pr list but not this count?

grodrigues3: cntrl.prs.length didn't work because prs is an object and not an array.  I initially tried to get the count without modifying the js, and this is the only solution that worked for me.  Regarding search, you're right that it's strange not to show the total count of the "displayed items".

foxish: Could update that count also [here](https://github.com/grodrigues3/contrib/blob/2c918d7e60403f7a14bf28521adf500f78376289/mungegithub/submit-queue/www/script.js#L354) to give us the length of the returned list when we filter.

eparis: yeah, can/should we do the same thing on the 'history' tab?

eparis:
anyway, i'll go ahead and merge this. it's not worse.

PR #1520, opened 2016-08-11
track flakes in more presubmit jobs

lavalamp:
This turns on flake filing for presubmit jobs that are or soon will be tolerant of flakes.


lavalamp:
@pwittrock fyi I'm going to send a PR to you to actually turn on the tolerance for the node pull test, because I hit a flake on it earlier.

foxish:
LGTM

PR #1526, opened 2016-08-12
[rescheduler] Implemented main control loop

piosz:
ref https://github.com/kubernetes/kubernetes/issues/29023

cc @davidopp @wojtek-t

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1526)


rescheduler/rescheduler.go
```diff
@@ -17,18 +17,28 @@ limitations under the License.
...
+	"k8s.io/kubernetes/pkg/fields"
 	kubectl_util "k8s.io/kubernetes/pkg/kubectl/cmd/util"
+	"k8s.io/kubernetes/plugin/pkg/scheduler/schedulercache"
 
 	"github.com/golang/glog"
 	flag "github.com/spf13/pflag"
 )
 
+const (
+	criticalAnnotation = "scheduler.alpha.kubernetes.io/critical-pod"
```
davidopp: maybe criticalPodAnnotation

rescheduler/rescheduler.go
```diff
@@ -17,18 +17,28 @@ limitations under the License.
...
+	criticalAddonsOnly = "CriticalAddonsOnly"
```
davidopp: A more descriptive variable name would be `criticalAddonsOnlyTaintKey`

davidopp: Unrelated (but Github won't let me comment on a line you didn't change...):

`inCluster = flags.Bool("running-in-cluster", true, ...`

Where would the controller be running if not in a cluster? (One someone's laptop?) When do you set this to false?

rescheduler/utils.go
```diff
@@ -0,0 +1,69 @@
...
+	return &podSet{
+		set:   make(map[string]struct{}),
+		mutex: sync.Mutex{},
+	}
+}
+
+// Add the pod to the set.
+func (s *podSet) Add(pod *kube_api.Pod) {
+	s.mutex.Lock()
+	go s.mutex.Unlock()
```
davidopp: did you mean "defer" all the places you wrote "go" in this file?

rescheduler/utils.go
```diff
@@ -0,0 +1,69 @@
...
+	delete(s.set, podId(pod))
+}
+
+// Has checks whether the pod is in the set.
+func (s *podSet) Has(pod *kube_api.Pod) bool {
+	return s.HasId(podId(pod))
+}
+
+// Has checks whether the pod is in the set.
```
davidopp: HasId

rescheduler/rescheduler.go
```diff
@@ -77,8 +88,247 @@ func main() {
...
 				}
-				glog.Infof("%+v", allUnschedulablePods)
+
+				criticalPods := filterCriticalPods(allUnschedulablePods)
+
+				if len(criticalPods) > 0 {
+					for _, pod := range criticalPods {
+						glog.Infof("Critical pod %s is unschedulable. Trying to find a spot for it.", podId(pod))
+
+						nodes, err := nodeLister.List()
```
davidopp: you could move this List() outside the for loop

rescheduler/rescheduler.go
```diff
@@ -77,8 +88,247 @@ func main() {
...
+	for _, node := range nodes {
+		taints, err := kube_api.GetTaintsFromNodeAnnotations(node.Annotations)
+		if err != nil {
+			glog.Warningf("Error while getting Taints for node %v: %v", node.Name, err)
+			continue
+		}
+
+		newTaints := make([]kube_api.Taint, 0)
+		for _, taint := range taints {
+			if taint.Key == criticalAddonsOnly && podsBeingProcessed.HasId(taint.Value) {
```
davidopp: shouldn't this be `&& !podsBeingProcessed.HasId(taint.Value)` ?

rescheduler/rescheduler.go
```diff
@@ -77,8 +88,247 @@ func main() {
...
+				glog.Warningf("Error while releasing taints on node %v: %v", node.Name, err)
+			} else {
+				glog.Infof("Successfully released all taints on node %v", node.Name)
+			}
+		}
+	}
+}
+
+func prepareNodeForPod(client *kube_client.Client, predicateChecker *ca_simulator.PredicateChecker, node *kube_api.Node, criticalPod *kube_api.Pod) error {
+	// This taint will be removed if this function will return an error.
```
davidopp: I think you mean "The caller of this function must remove the taint if this function returns error" ?
(There's no code that I can see in this function that removes it if this function returns error.)

rescheduler/rescheduler.go
```diff
@@ -77,8 +88,247 @@ func main() {
...
+	}
+	requiredPods = append(requiredPods, criticalPod)
+	nodeInfo = schedulercache.NewNodeInfo(requiredPods...)
+	nodeInfo.SetNode(node)
+
+	for _, p := range otherPods {
+		if err := predicateChecker.CheckPredicates(p, nodeInfo); err != nil {
+			glog.Infof("Pod %s will be deleted in order to schedule critical pods %s.", podId(p), podId(criticalPod))
+			// TODO(piosz): add better support of graceful deletion
+			delErr := client.Pods(p.Namespace).Delete(p.Name, kube_api.NewDeleteOptions(0))
```
davidopp: 0 seems to trigger various kinds of problems. If you're going to hardcode a value here for now, how about hard-coding 30s which is the default anyway? If not, maybe 10s?

rescheduler/rescheduler.go
piosz: For development purposes mostly. I follow the pattern from Ingress project.

rescheduler/rescheduler.go

rescheduler/rescheduler.go
piosz: Resolved first comment

rescheduler/utils.go
piosz: Nice catch! Thanks. Fixed in all places

rescheduler/utils.go
piosz: Done

rescheduler/rescheduler.go
piosz: Right. Fixed

rescheduler/rescheduler.go
piosz: That's what I meant. Changed comment to what you wrote.

rescheduler/rescheduler.go
piosz: In the first version I wanted to not care about grace period at all, so changing to 10s which is reasonable small.

piosz:
Thanks @davidopp for the review. Addressed your comments.

rescheduler/rescheduler.go
```diff
@@ -77,8 +88,247 @@ func main() {
 					glog.Errorf("Failed to list unscheduled pods: %v", err)
 					continue
 				}
-				glog.Infof("%+v", allUnschedulablePods)
+
+				criticalPods := filterCriticalPods(allUnschedulablePods)
```
mwielgus: you should probably include podsbeingprocessed

rescheduler/rescheduler.go
```diff
@@ -77,8 +88,247 @@ func main() {
...
+				}
+
+				releaseAllTaints(kubeClient, nodeLister, podsBeingProcessed)
 			}
 		}
 	}
 }
+
+func waitForScheduled(podsBeingProcessed *podSet, pod *kube_api.Pod) {
+	// TODO(piosz): improve this
```
mwielgus: How do you want to improve it?

rescheduler/rescheduler.go
```diff
@@ -77,8 +88,247 @@ func main() {
...
+	otherPods := make([]*kube_api.Pod, 0)
+	for _, p := range podsOnNode.Items {
+		pod := &p
+
+		creatorRef, err := ca_simulator.CreatorRefKind(pod)
+		if err != nil {
+			return []*kube_api.Pod{}, []*kube_api.Pod{}, err
+		}
+
+		if ca_simulator.IsMirrorPod(pod) || creatorRef == "DaemonSet" || isCriticalPod(pod) {
```
mwielgus: How does this differ from drain?

rescheduler/rescheduler.go
```diff
@@ -77,8 +88,247 @@ func main() {
...
+	}
+	node.Annotations[kube_api.TaintsAnnotationKey] = string(taintsJson)
+	_, err = client.Nodes().Update(node)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func findNodeForPod(client *kube_client.Client, predicateChecker *ca_simulator.PredicateChecker, nodes []*kube_api.Node, pod *kube_api.Pod) *kube_api.Node {
```
mwielgus: How are the nodes prioritiezed?

rescheduler/rescheduler.go
```diff
@@ -77,8 +88,247 @@ func main() {
...
 					continue
 				}
-				glog.Infof("%+v", allUnschedulablePods)
+
+				criticalPods := filterCriticalPods(allUnschedulablePods)
+
+				if len(criticalPods) > 0 {
+					for _, pod := range criticalPods {
+						glog.Infof("Critical pod %s is unschedulable. Trying to find a spot for it.", podId(pod))
+
```
mwielgus: empty line

rescheduler/rescheduler.go
```diff
@@ -77,8 +88,247 @@ func main() {
...
+	glog.Infof("Waiting for pod %s to be scheduled", podId(pod))
+	time.Sleep(10 * time.Minute)
+	podsBeingProcessed.Remove(pod)
+}
+
+func createKubeClient(flags *flag.FlagSet, inCluster bool) (*kube_client.Client, error) {
+	if inCluster {
+		return kube_client.NewInCluster()
+	}
+
```
mwielgus: empty lie

rescheduler/rescheduler.go
```diff
@@ -77,8 +88,247 @@ func main() {
...
+				return fmt.Errorf("Failed to delete pod %s: %v", podId(p), delErr)
+			}
+		} else {
+			newPods := append(nodeInfo.Pods(), p)
+			nodeInfo = schedulercache.NewNodeInfo(newPods...)
+			nodeInfo.SetNode(node)
+		}
+	}
+
+	// TODO(piosz): how to reset scheduler backoff?
```
mwielgus: empty lines

rescheduler/rescheduler.go
davidopp: @mwielgus I assume you mean "filterCriticalPods() should exclude pods in podsBeingProcessed"? If so, I agree -- good catch.

rescheduler/rescheduler.go

rescheduler/rescheduler.go
piosz: Drain removes only pods which will be automatically recreated with the same state, which means being managed by higher level abstraction, without local disk, etc.

Here we only want to avoid removing pods that has a higher or equal 'priority' than the critical pod, which means manifest pods, daemon set pods and other critical addons.

rescheduler/rescheduler.go
piosz: Randomly. Added comment and todo to improve this

rescheduler/rescheduler.go

rescheduler/rescheduler.go

rescheduler/rescheduler.go

davidopp:
LGTM

I would not suggest squashing the commits.

PR #1528, opened 2016-08-12
Make IsMergeable retry up to 5 times with exponential delays. (2-32s)

rmmh:
This should fix the bouncing in the submit queue where the
mergeability of a PR is indeterminate after each round of merging.
I think Github is just taking longer than 2 seconds to respond.

### DIFF
--- a/mungegithub/github/github.go
+++ b/mungegithub/github/github.go
@@ -1626,10 +1626,13 @@ func (obj *MungeObject) IsMergeable() (bool, error) {
 		return false, err
 	}
 	prNum := *pr.Number
-	if pr.Mergeable == nil {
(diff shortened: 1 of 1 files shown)


spxtr:
Idea seems good.

mungegithub/github/github.go
```diff
@@ -1626,10 +1626,13 @@ func (obj *MungeObject) IsMergeable() (bool, error) {
 		return false, err
 	}
 	prNum := *pr.Number
-	if pr.Mergeable == nil {
+	// Github might not have computed mergeability yet. Try again a few times.
+	for try := 1; try <= 5 && pr.Mergeable == nil; try++ {
```
lavalamp: @eparis thoughts on this?

PR #1531, opened 2016-08-12
Add exclusion for HA.yaml files as they are part of core tests that need to be editable.

timothysc:
My regex is rusty, hopefully I got it right. 
/cc @eparis

### DIFF
--- a/mungegithub/block-path.yaml
+++ b/mungegithub/block-path.yaml
@@ -10,5 +10,6 @@ doNotBlockRegexp:
   - ^docs/admin/kubelet.md$
   - ^docs/admin/federation-apiserver.md$
   - ^docs/admin/federation-controller-manager.md$
+  - ^docs/admin/high-availability/.*\.(yaml)$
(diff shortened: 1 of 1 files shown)


timothysc:
xref: https://github.com/kubernetes/kubernetes/pull/30514

apelisse:
Why do you have extra parenthesis around `yaml`?

timothysc:
> Why do you have extra parenthesis around yaml?

I put it there in case I needed to ```|``` extensions but it's not needed. 
Happy to remove if desired.

apelisse:
Yeah it doesn't make a difference,

timothysc:
Hmm it does not look like the bot is respecting this rule?  Does the bot need to be reset?

apelisse:
Yes, we would have to redeploy the bot, which may not have happened yet. 
Let me check ...

Image:	docker.io/foxish/submit-queue:2016-07-29-68fd252

It hasn't been deployed in quite some time. I'll try to let you know when it happens.

PR #1532, opened 2016-08-12
Updated readme with instructions to run locally.

foxish:
Added instructions to run/test locally as several people seem to be having trouble with testing their changes.


apelisse:
Perfect!

PR #1537, opened 2016-08-14
Update the scale-demo readme for Kubernetes 1.2.0.

roberthbailey:
A bit out of date, now that 1.3 has been out for a while, but had this change sitting around so I thought I'd send it.

### DIFF
--- a/scale-demo/README.md
+++ b/scale-demo/README.md
@@ -1,9 +1,9 @@
 # Kubernetes One Million Requests Per Second Demo
 
 ## Prerequisites
-### Kubernetes 1.1.x
-This demo requires a Kubernetes 1.1.x cluster, it uses iptables proxy support that was released
-in Kubernetes 1.1.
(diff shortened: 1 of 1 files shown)


scale-demo/README.md
```diff
@@ -1,8 +1,8 @@
 # Kubernetes One Million Requests Per Second Demo
 
 ## Prerequisites
-### Kubernetes 1.1.x
-This demo requires a Kubernetes 1.1.x cluster, it uses iptables proxy support that was released
+### Kubernetes 1.1.x or 1.2.x
```
brendandburns: Perhaps Kubernetes > 1.1?  Since 1.3 is clearly ok too.

brendandburns:
May as well make it clear that the requirement is 1.1+

Feel free to self LGTM after that.

scale-demo/README.md
roberthbailey: fixed.

PR #1538, opened 2016-08-14
Add a simple script to scale up a cluster

roberthbailey:
Also change the html headings to "Scale Demo" since we now go well beyond 1M QPS.

### DIFF
--- a/scale-demo/scale-up.sh
+++ b/scale-demo/scale-up.sh
@@ -1,0 +1,39 @@
+#!/bin/bash
+
+# Copyright 2016 The Kubernetes Authors All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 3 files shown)


scale-demo/scale-up.sh
```diff
@@ -0,0 +1,38 @@
...
+declare -ir target_size="${3}"
+
+echo "Scaling from ${initial_size} to ${target_size} by ${step_size}"
+
+for ((i=${initial_size}; i<=${target_size}; i+=${step_size})); do
+  current_size="${i}"
+  echo "Scaling to ${current_size} loadbots"
+  echo "kubectl scale rc vegeta --replicas=${current_size}"
+  kubectl scale rc vegeta --replicas="${current_size}"
+  sleep 7
```
brendandburns: May as well add an environment variable for sleep time?

brendandburns:
LGTM, one nit, feel free to self-LGTM.

Thanks!

scale-demo/scale-up.sh
roberthbailey: done.

PR #1543, opened 2016-08-16
Fix relative paths to init directory and k8s build

ingvagabund:
When deploying locally built kubernetes, systemd service files and binaries are one level up.
Due to changes of top directory layout another level needs to be added.

### DIFF
--- a/ansible/roles/node/defaults/main.yml
+++ b/ansible/roles/node/defaults/main.yml
@@ -1,3 +1,3 @@
 kubelet_working_dir: /var/lib/kubelet
-localBuildOutput: ../../_output/local/go/bin
+localBuildOutput: ../../../_output/local/go/bin
 open_cadvisor_port: false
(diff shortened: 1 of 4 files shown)


ingvagabund:
@redhat-k8s-bot test please

ansible/roles/node/defaults/main.yml
```diff
@@ -1,3 +1,3 @@
 kubelet_working_dir: /var/lib/kubelet
-localBuildOutput: ../../_output/local/go/bin
+localBuildOutput: ../../../_output/local/go/bin
```
sttts: This is intentionally? Does not match the description of the PR.

ingvagabund: It is. Description and subject updated to reflect that.

sttts:
lgtm

PR #1550, opened 2016-08-17
Fixes race between lgtm-handler and lgtm-after-commit & adds tests.

### LINKED ISSUE #1546: Potential race: lgtm_handler and lgtm_after_commit
Before deploying https://github.com/kubernetes/contrib/pull/1539, I want to ensure that we have no races with the bot fighting against itself and exhausting quota.

It appears that it could happen in the following situation:
* Someone comments `/lgtm`.
* The bot adds the `lgtm` label.
* The author pushes new commits.

What happens:
* The lgtm_handler munger continues to see the comment and tries to apply the label. 
* The lgtm_after_commit munger tries to remove the lgtm label.

cc @eparis @apelisse @lavalamp

foxish:
Fixes #1546
- Adds tests for lgtm_handler.
- Change the message posted by lgtm-after-commit when removing LGTM.

cc @lavalamp @eparis

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1550)

### DIFF
--- a/mungegithub/mungers/lgtm_handler.go
+++ b/mungegithub/mungers/lgtm_handler.go
@@ -81,9 +81,17 @@ func (h *LGTMHandler) addLGTMIfCommented(obj *github.MungeObject, comments []*gi
 	// Assumption: The comments should be sorted (by default from github api) from oldest to latest
 	for i := len(comments) - 1; i >= 0; i-- {
 		comment := comments[i]
-		fields := strings.Fields(*comment.Body)
+		if !mungerutil.IsValidUser(comment.User) {
+			continue
+		}
+
+		if !(mungerutil.IsMungeBot(comment.User) || isReviewer(comment.User, reviewers)) {
+			continue
+		}
+
+		fields := getFields(*comment.Body)
 		if isCancelComment(fields) {
-			// "/lgtm cancell" if commented more recently than "/lgtm"
+			// "/lgtm cancel" if commented more recently than "/lgtm"
 			return
 		}
 		if isLGTMComment(fields) {
@@ -91,11 +99,9 @@ func (h *LGTMHandler) addLGTMIfCommented(obj *github.MungeObject, comments []*gi
 			// See https://github.com/kubernetes/contrib/issues/1389#issuecomment-235161164
 			// TODO: An approver should be acceptable.
 			// See https://github.com/kubernetes/contrib/pull/1428#discussion_r72563935
-			if mungerutil.IsValidUser(comment.User) && isReviewer(comment.User, reviewers) {
-				glog.Infof("Adding lgtm label. Reviewer (%s) LGTM", *comment.User.Login)
-				obj.AddLabel(lgtmLabel)
-				return
-			}
+			glog.Infof("Adding lgtm label. Reviewer (%s) LGTM", *comment.User.Login)
+			obj.AddLabel(lgtmLabel)
+			return
 		}
 	}
 }
@@ -103,17 +109,23 @@ func (h *LGTMHandler) addLGTMIfCommented(obj *github.MungeObject, comments []*gi
 func (h *LGTMHandler) removeLGTMIfCancelled(obj *github.MungeObject, comments []*githubapi.IssueComment, reviewers mungerutil.UserSet) {
 	for i := len(comments) - 1; i >= 0; i-- {
 		comment := comments[i]
-		fields := strings.Fields(*comment.Body)
+		if !mungerutil.IsValidUser(comment.User) {
+			continue
+		}
+
+		if !(mungerutil.IsMungeBot(comment.User) || isReviewer(comment.User, reviewers)) {
+			continue
+		}
+
+		fields := getFields(*comment.Body)
 		if isLGTMComment(fields) {
 			// "/lgtm" if commented more recently than "/lgtm cancel"
 			return
 		}
 		if isCancelComment(fields) {
-			if mungerutil.IsValidUser(comment.User) && isReviewer(comment.User, reviewers) {
-				glog.Infof("Removing lgtm label. Reviewer (%s) cancelled", *comment.User.Login)
-				obj.RemoveLabel(lgtmLabel)
-				return
-			}
+			glog.Infof("Removing lgtm label. Reviewer (%s) cancelled", *comment.User.Login)
+			obj.RemoveLabel(lgtmLabel)
+			return
 		}
 	}
 }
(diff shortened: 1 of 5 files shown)


mungegithub/mungers/lgtm_handler.go
```diff
@@ -81,39 +81,51 @@ func (h *LGTMHandler) addLGTMIfCommented(obj *github.MungeObject, comments []*gi
...
 
 func (h *LGTMHandler) removeLGTMIfCancelled(obj *github.MungeObject, comments []*githubapi.IssueComment, reviewers mungerutil.UserSet) {
 	for i := len(comments) - 1; i >= 0; i-- {
 		comment := comments[i]
-		fields := strings.Fields(*comment.Body)
+		if !mungerutil.IsValidUser(comment.User) {
+			continue
+		}
+
+		if !(mungerutil.IsMungeBot(comment.User) || isReviewer(comment.User, reviewers)) {
```
eparis: So any reviewer (or bot) can cancel any other reviewers `/lgtm`? Just making sure I understand.

foxish: Yes, that is correct.

mungegithub/mungers/lgtm_handler.go
```diff
@@ -81,39 +81,51 @@ func (h *LGTMHandler) addLGTMIfCommented(obj *github.MungeObject, comments []*gi
...
+			continue
+		}
+
+		fields := getFields(*comment.Body)
 		if isCancelComment(fields) {
-			// "/lgtm cancell" if commented more recently than "/lgtm"
+			// "/lgtm cancel" if commented more recently than "/lgtm"
 			return
 		}
 		if isLGTMComment(fields) {
```
eparis: Should we invert this conditional and return early? (nit)

Should we then check the push time vs comment time before even bothering to add lgtm? Thus fixing the race a second way?

mungegithub/mungers/lgtm_handler.go
hongchaodeng: Can you change it to
```
if !IsMungeBot(...) && !isReviewer(...) {
}
```
Outer "!" is confusing.

hongchaodeng:
lgtm

mungegithub/mungers/lgtm_handler.go
foxish: > Should we invert this conditional and return early? (nit)

I tried inverting the conditional, but it became a little harder for me to read with the early return before it. It might just be me, so, if you feel strongly about it, I'll change it. 

```
for ... {
    ...
    if isLGTMComment(fields) {
	  return
    }

    if !isCancelComment(fields) {
	  continue
    }

    glog.Infof("Removing lgtm label. Reviewer (%s) cancelled", *comment.User.Login)
    obj.RemoveLabel(lgtmLabel)
    return
}
```

> Should we then check the push time vs comment time before even bothering to add lgtm? Thus fixing the race a second way?

I think that needs more thought because in case we don't honor a `/lgtm` command, we need to post a comment to that effect as well, along with the reason. Perhaps we should do the same thing as the lgtm-after-commit handler in that case, and post that message. I can follow up in a later PR if this causes issues.

mungegithub/mungers/lgtm_handler.go
foxish: Done

mungegithub/mungers/lgtm_handler.go
eparis: Actually what if we made the lgtm_after_commit handler do it, rather than do it here. He if change /lgtm into something else (with a comment why it changed it)

foxish: @eparis Wouldn't the order in which the mungers run start to impact what we see then? If the lgtm_handler runs first, it will probably see `/lgtm` and not the edited version.

apelisse:
Reviewed 4 of 5 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[mungegithub/mungers/lgtm_after_commit.go, line 33 \[r1\]](https://reviewable.kubernetes.io:443/reviews/kubernetes/contrib/1550#-KPKpDE7uu50Pt8QReu6:-KPKpDE8gTB7yZXEF9U_:b5gl92i) ([raw file](https://github.com/kubernetes/contrib/blob/50e64d64816d3d76ca3d125d0d2f3182b4ac36a9/mungegithub/mungers/lgtm_after_commit.go#L33)):*
> ```Go
> 
> const (
> 	lgtmRemovedBody = "/lgtm cancel //PR changed after LGTM, removing LGTM. %s"
> ```

I think we should make the comment on a different line rather than make it harder than necessary to parse

---

*[mungegithub/mungers/lgtm_handler_test.go, line 119 \[r2\]](https://reviewable.kubernetes.io:443/reviews/kubernetes/contrib/1550#-KPPEDEaC4294FbQHK2A:-KPPEDEaC4294FbQHK2B:bjg9nty) ([raw file](https://github.com/kubernetes/contrib/blob/5285f5f4779b18fcbc66e6096b6c74926c90f346/mungegithub/mungers/lgtm_handler_test.go#L119)):*
> ```Go
> 			issue: prWithoutLGTM,
> 			comments: []*github.IssueComment{
> 				github_test.IssueComment(1, "/lgtm //this is a comment", "user 1", 0),
> ```

I think we should really agree on that (how to comment "code") quickly

---


*Comments from [Reviewable](https://reviewable.kubernetes.io:443/reviews/kubernetes/contrib/1550)*

foxish:
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[mungegithub/mungers/lgtm_after_commit.go, line 33 \[r1\]](https://reviewable.kubernetes.io:443/reviews/kubernetes/contrib/1550#-KPKpDE7uu50Pt8QReu6:-KPPNin_uZi7vFh882DS:b-fwgowj) ([raw file](https://github.com/kubernetes/contrib/blob/50e64d64816d3d76ca3d125d0d2f3182b4ac36a9/mungegithub/mungers/lgtm_after_commit.go#L33)):*
<details><summary><i>Previously, apelisse (Antoine Pelisse) wrote…</i></summary>
> I think we should make the comment on a different line rather than make it harder than necessary to parse
</details>
Makes sense. Once we have the comment handler, we can make it into the description field and update the tests. Do you think we should decide on it right away?

---


*Comments from [Reviewable](https://reviewable.kubernetes.io:443/reviews/kubernetes/contrib/1550)*

foxish:
@eparis, could you apply the LGTM label here this as well? 
I've talked to @apelisse about his comment.

foxish:
Thanks! Starting it up shortly. I'll run a couple of experiments on a PR and see how it goes.

PR #1377, opened 2016-07-15
Cluster-autoscaler: cloud provider interface implementation for AWS

osxi:
Part of https://github.com/kubernetes/contrib/issues/1311

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1377)

PR #1552, opened 2016-08-17
AWS Cluster Autoscaler README

andrewsykim:
under https://github.com/kubernetes/contrib/issues/1311

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1552)

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/README.md
+++ b/cluster-autoscaler/cloudprovider/aws/README.md
@@ -1,0 +1,79 @@
+# Cluster Autoscaler on AWS
+The cluster autoscaler on AWS scales worker nodes within an autoscaling group. It will run as a `Deployment` in your cluster. This README will go over some of the necessary steps required to get the cluster autoscaler up and running.
+
+## Kubernetes Version
+Cluster autoscaler must run on v1.3.0 or greater.
+
+## Permissions
+The worker running the cluster autoscaler will need access to certain resources and actions:
+```json
+{
+    "Version": "2012-10-17",
+    "Statement": [
+        {
+            "Effect": "Allow",
+            "Action": [
+                "autoscaling:DescribeAutoScalingGroups",
+                "autoscaling:DescribeAutoScalingInstances",
+                "autoscaling:SetDesiredCapacity",
+                "autoscaling:TerminateInstanceInAutoScalingGroup"
+            ],
+            "Resource": "*"
+        }
+    ]
+}
+```
+Unfortunately AWS does not support ARNs for autoscaling groups yet so you must use "*" as the resource. More information [here](http://docs.aws.amazon.com/autoscaling/latest/userguide/IAM.html#UsingWithAutoScaling_Actions).
+
+## Deployment Specification
+Your deployment configuration should look something like this:
+```yaml
+---
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: cluster-autoscaler
+  labels:
+    app: cluster-autoscaler
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: cluster-autoscaler
+  template:
+    metadata:
+      labels:
+        app: cluster-autoscaler
+    spec:
+      containers:
+        - image: {{ YOUR IMAGE HERE }}
+          name: cluster-autoscaler
+          resources:
+            limits:
+              cpu: 100m
+              memory: 300Mi
+            requests:
+              cpu: 100m
+              memory: 300Mi
+          command:
+            - ./cluster-autoscaler
+            - -v=4
+            - --cloud-provider=aws
+            - --skip-nodes-with-local-storage=false
+            - --nodes={{ ASG MIN e.g. 1 }}:{{ASG MAX e.g. 5}}:{{ASG NAME e.g. k8s-worker-asg}}
+          env:
+            - name: AWS_REGION
+              value: us-east-1
+          volumeMounts:
+            - name: ssl-certs
+              mountPath: /etc/ssl/certs/ca-certificates.crt
+              readOnly: true
+          imagePullPolicy: "Always"
+      volumes:
+        - name: ssl-certs
+          hostPath:
+            path: "/etc/ssl/certs/ca-certificates.crt"
+```
+Note:
+- The `/etc/ssl/certs/ca-certificates.crt` should exist by default on your ec2 instance.
+- The autoscaling group should span 1 availability zone for the cluster autoscaler to work. If you want to distribute workloads evenly across zones, set up multiple ASGs, with a cluster autoscaler for each ASG. At the time of writing this, cluster autoscaler is unaware of availability zones and although autoscaling groups can contain instances in multiple availability zones when configured so, the cluster autoscaler can't reliably add nodes to desired zones. That's because AWS AutoScaling determines which zone to add nodes which is out of the control of the cluster autoscaler. For more information, see https://github.com/kubernetes/contrib/pull/1552#discussion_r75533090.


andrewsykim:
@mwielgus @pbitty @osxi @iterion r?

cluster-autoscaler/cloudprovider/aws/README.md
```diff
@@ -0,0 +1,79 @@
...
+            "Resource": "*"
+        }
+    ]
+}
+```
+Unfortunately AWS does not support ARNs for autoscaling groups yet so you must use "*" as the resource.
+
+## Deployment Specification
+Your deployment configuration should look something like this:
+```
```
iterion: You can hint to Github that this should be formmated as yaml with: "```yaml". Same with the json above.

cluster-autoscaler/cloudprovider/aws/README.md
```diff
@@ -0,0 +1,79 @@
...
+            - --nodes={{ ASG MIN e.g. 1 }}:{{ASG MAX e.g. 5}}:{{ASG NAME e.g. k8s-worker-asg}}
+          env:
+            - name: AWS_REGION
+              value: us-east-1
+          volumeMounts:
+            - name: ssl-certs
+              mountPath: /etc/ssl/certs/ca-certificates.crt
+              readOnly: true
+          imagePullPolicy: "Always"
+      volumes:
```
iterion: the GCE example in this repo had this:
```yaml
volumes:
  - name: ssl-certs
    hostPath:
      path: "/etc/ssl/certs"
```

I'm not sure if this difference would cause issues in other setups?

cluster-autoscaler/cloudprovider/aws/README.md
andrewsykim: cool didn't know that

cluster-autoscaler/cloudprovider/aws/README.md
pbitty: @iterion, where is this example?  I can't find it.

andrewsykim: @pbitty think he was referring to https://github.com/kubernetes/contrib/blob/master/cluster-autoscaler/deploy/ca-controller.yaml

iterion: Correct.

andrewsykim: @iterion there's a lot of other certs in that directory in most ec2 instances, not sure if we want all of them mounted so for aws it could be problematic?

cluster-autoscaler/cloudprovider/aws/README.md
```diff
@@ -0,0 +1,79 @@
...
+        - name: ssl-certs
+          hostPath:
+            path: "/etc/ssl/certs/ca-certificates.crt"
+```
+Note: 
+- the `/etc/ssl/certs/ca-certificates.crt` should exist by default on your ec2 instance.
+- at the time of writing this, cluster autoscaler is unaware of availability zones, the availability zone of the instance should be configured by the autoscaling group. Although autoscaling groups can contain instances in multiple availability zones, the autoscaling group should span 1 availability zone for the cluster autoscaler to work.
```
mumoshu: Hi, really excited to see this PR!!

Execute me if I'm misreading but I couldn't figure out why `the autoscaling group should span 1 availability zone for the cluster autoscaler to work`.

As noted, an cluster autoscaler is unaware of AZs, but an ASG is aware of them. If configured, spreading EC2 instances over multiple AZs will be done completely out of band in AWS AutosCaling. So I believe you can just assign 2 or more AZs to an ASG to eventually make cluster autoscaler multi-AZ aware as far as the AWS impl for Cluster Autoscaler delegates adding/removing instances to AWS AutoScaling.

And it seems so(according to https://github.com/kubernetes/contrib/pull/1377/files#diff-ade7b95627ea0dd6b6f4deee7f24fa7eR124 it is calling `SetDesiredCapacity` to delegate adding/removing instances)

osxi: Hey @mumoshu, I can't really speak to multi-AZ setups since our use case was a single-AZ ASG but that sounds right. You're correct regarding `SetDesiredCapacity` for instance creation, but [this](https://github.com/kubernetes/contrib/pull/1377/files#diff-ade7b95627ea0dd6b6f4deee7f24fa7eR132) is where the instances are deleted.

pbitty: @mumoshu, it will likely work with a multi-AZ ASG, but there may be some caveats if Kubernetes is configured to be zone-aware.  Pardon the long-winded explanation.  Here is the reasoning:

The cluster-autoscaler asks the AWS CloudProvider for a sample Node from the NodeGroup (backed by an ASG), and uses it to make scaling decisions.  It assumes that the sample Node is equivalent to all other nodes in the ASG - eg. same instance type, storage, etc.  When it needs to scale up, for example, it will know with certainty that new Nodes will have the same capacity and will be able to accommodate the pending pods.

The cluster-autoscaler has logic that simulates the Scheduler's decisions to see if a new Node will be able to accommodate pending workloads.  If the Scheduler is zone-aware, it may specifically want to distribute workloads across AZs.

Consider this scenario:

* The Scheduler wants to run some Pods in Zone A because there are already equivalent pods in Zone B and it wants a multi-AZ Pod distribution.
* There are not enough Nodes in Zone A
* In this situation the Pods will be pending, waiting for a Node in Zone A
* The cluster autoscaler sees pending pods and starts looking for a NodeGroup (an ASG) that can accommodate these Pods
* Say we have a NodeGroup whose ASG spans Zones A and B
* This NodeGroup returns a random sampleNode that is in Zone A
* The cluster-autoscaler says "Great, this NodeGroup has an appropriate node, let me scale it up", and it increases the DesiredCapacity by 1
* The ASG launches a Node in Zone B (because it is trying to keep the Zones balanced)
* The new Node appears in Kubernetes and the Scheduler sees that is in Zone B
* The Scheduler continues to wait a Node in Zone A in order to schedule the pending Pods
* The new Node did not help accommodate the pending Pods

Eventually the cluster-autoscaler will try to launch another Node because there are still Pods pending.  By chance if a new Node lands in Zone A, the Pods will get scheduled and things will work, but this would be non-deterministic and would make the process less reliable

The crux of the matter is the contract between NodeGroup that provides a sampleNode, and the scaling logic, which expects that a sampleNode is equivalent to all other nodes in that NodeGroup.

These are theoretical, I haven't tested these scenarios, but from my understanding of the Node sampling logic, something like this would happen.

pbitty: On that note, maybe we should say in the README that if one wants the Scheduler to be Zone-aware and distribute workloads evenly across Zones, then the 1-zone-per-ASG rule must be followed.  Otherwise, it doesn't matter.

Thoughts?

andrewsykim:
```
- At the time of writing this, cluster autoscaler is unaware of availability zones, the availability zone of the instance should be configured by the autoscaling group. Although autoscaling groups can contain instances in multiple availability zones, the cluster autoscaler will not evenly distribute pods across zones. Use mutli-AZ ASG at your own risk. 
```

mumoshu: @osxi @pbitty @andrewsykim Your explanation really helped me understand what's under the hood. Thanks!

So, to be clear, 

* As long as we keep single AZ per ASG, cluster-autoscaling works reliably. That's because cluster autoscaler isn't AZ aware. Multiple AZs per ASG can result in a node added to an unneeded AZ because cluster scaler doesn't know/is unable to control which AZ to add a node.
* If you are O.K. with undeterministic behavior @pbitty described, you can assign multiple AZs to an ASG but it isn't necessary because as @pbitty pointed out in https://github.com/kubernetes/contrib/pull/1552#discussion_r75533090, keeping 1-zone-per-ASG fixes the reliability gotcha without any unwanted side-efffect

?

Then, IMHO, something like the below makes sense to me:

> The autoscaling group should span exactly 1 availability zone for the cluster autoscaler to work. If you want to distribute workloads evenly across zones, set up multiple ASGs, each in a distinct availability zone.
> 
> At the time of writing this, cluster autoscaler is unaware of availability zones. Although autoscaling groups can contain instances in multiple availability zones when configured so, cluster autoscaler can't reliably add nodes to desired zones then. That's because AWS AutoScaling determines which zone to add node completely out-of-band from cluster autoscaler. For more information, see https://github.com/kubernetes/contrib/pull/1552#discussion_r75532949

mwielgus:
Is the current state of the doc acceptable?

cluster-autoscaler/cloudprovider/aws/README.md
```diff
@@ -0,0 +1,79 @@
...
+                "autoscaling:DescribeAutoScalingInstances",
+                "autoscaling:SetDesiredCapacity",
+                "autoscaling:TerminateInstanceInAutoScalingGroup"
+            ],
+            "Resource": "*"
+        }
+    ]
+}
+```
+Unfortunately AWS does not support ARNs for autoscaling groups yet so you must use "*" as the resource.
```
osxi: We could probably add this link: http://docs.aws.amazon.com/autoscaling/latest/userguide/IAM.html#UsingWithAutoScaling_Actions

osxi:
LGTM after adding a link to the autoscaling permissions docs!

cluster-autoscaler/cloudprovider/aws/README.md
andrewsykim: looks good to me, I'll make the changes

andrewsykim:
Comments addressed. @mwielgus I was wondering if there's an official gcr image for the CA that supports AWS as a cloud provider we can put [here](https://github.com/kubernetes/contrib/pull/1552/files#diff-1af4eaf7829acf6bfc72755c74d6b9e2R49)?

mwielgus:
There is one: gcr.io/google_containers/cluster-autoscaler:v0.3.0-beta2 but it is a beta image. I will will update the doc once we get the final release.

mwielgus:
lgtm

PR #1553, opened 2016-08-17
Add mungegithub to kubernetes/test-infra

apelisse:
Make sure the mungegithub is running for test-infra repository.

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1553)

### DIFF
--- a/mungegithub/submit-queue/deployment/test-infra/configmap.yaml
+++ b/mungegithub/submit-queue/deployment/test-infra/configmap.yaml
@@ -1,0 +1,44 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: test-infra-sq-config
+data:
+  # basic config options.
+  config.http-cache-dir: /cache/httpcache
+  config.organization: kubernetes
+  config.project: test-infra
+  config.pr-mungers: lgtm-after-commit,submit-queue,needs-rebase
+  config.state: open
+  config.token-file: /etc/secret-volume/token
+  gitrepo.repo-dir: /gitrepos
+  gitrepo.enable-md-yaml: "false"
+
+  # test-options feature options.
+  test-options.required-retest-contexts: "\"\""
+
+  # submit-queue options.
+  submit-queue.required-contexts: "continuous-integration/travis-ci/pr"
+  submit-queue.nonblocking-jenkins-jobs: "\"\""
+  submit-queue.jenkins-jobs: "\"\""
+  submit-queue.presubmit-jobs: "\"\""
+  submit-queue.weak-stable-jobs: "\"\""
+  submit-queue.do-not-merge-milestones: "\"\""
+  submit-queue.admin-port: "9999"
+  submit-queue.chart-url: ""
+  submit-queue.history-url: ""
+
+  # options used by the GCS feature.
+  gcs.gcs-bucket: ""
+  gcs.gcs-logs-dir: ""
+  gcs.pull-logs-dir: ""
+  gcs.pull-key: ""
+
+  # munger specific options.
+  path-label.path-label-config: ""
+  block-path.block-path-config: ""
+  flake-manager.test-owners-csv: ""
+  old-test-getter.number-of-old-test-results: "0"
+  assign-fixes.fixes-issue-reassign: "false"
+  blunderbuss.blunderbuss-reassign: "false"
+  size.generated-files-config: ""
+  check-labels.label-file: "/gitrepos/test-infra/labels.yaml"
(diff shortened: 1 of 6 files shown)


apelisse:
@foxish Can you please make sure I didn't forget anything? Also, we would need to create the test-infra.submit-queue.k8s.io host

mungegithub/submit-queue/ingress.yaml
```diff
@@ -22,3 +22,9 @@ spec:
         - backend:
             serviceName: docs-sq-status
             servicePort: 80
+  - host: test-infra.submit-queue.k8s.io
+    http:
+      paths:
+        - backend:
+            serviceName: contrib-sq-status
```
foxish: service-name should be test-infra-sq-status

foxish:
It should work mostly. Do we need the submit-queue itself to run? https://github.com/kubernetes/contrib/issues/762#issuecomment-235930596 stated that we may not want the submit-queue on each repo. 
We should first get some consensus on which mungers make sense to run on test-infra. 
To run blunderbuss for example, we need OWNERS files to be generated first.

 cc @fejta

fejta:
Yes I want the submit-queue. I want a consistent experience in every repo in the kubernetes org.

mungegithub/submit-queue/deployment/test-infra/service.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: submit-queue
```
fejta: Can we add the grapher now that it exists?

fejta:
LGTM

foxish:
Makes sense. @apelisse, can we prune the other mungers from that list for test-infra? We can start with the submit-queue, lgtm-after-commit, and needs-rebase mungers perhaps. Blunderbuss and so on can be started up later.

mungegithub/submit-queue/deployment/test-infra/service.yaml
foxish: Do you mean the queue-health grapher?

mungegithub/submit-queue/deployment/test-infra/configmap.yaml
```diff
@@ -0,0 +1,44 @@
...
+  submit-queue.nonblocking-jenkins-jobs: "\"\""
+  submit-queue.jenkins-jobs: "\"\""
+  submit-queue.presubmit-jobs: "\"\""
+  submit-queue.weak-stable-jobs: "\"\""
+  submit-queue.do-not-merge-milestones: "\"\""
+  submit-queue.admin-port: "9999"
+  submit-queue.chart-url: ""
+  submit-queue.history-url: ""
+
+  # options used by the GCS feature.
```
fejta: Shouldn't these be non-empty?

apelisse: @foxish I can't remember what these are?
@fejta They are not set for contrib, so I assumed we don't need it ... (your talking about GCS variables below, right?)

foxish: These are useful only if we use the jenkins instances. They can be blank here since we run only the travis presubmit and no other tests.

mungegithub/submit-queue/deployment/test-infra/configmap.yaml
```diff
@@ -0,0 +1,44 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: test-infra-sq-config
+data:
+  # basic config options.
+  config.http-cache-dir: /cache/httpcache
+  config.organization: kubernetes
+  config.project: test-infra
+  config.pr-mungers: blunderbuss,lgtm-after-commit,submit-queue,check-labels
```
foxish: s/blunderbuss//

apelisse: I'll address that now

apelisse:
@foxish Can you please add LGTM label if it's ok with you. I've removed the blunderbuss munger.

foxish:
@apelisse I think we need to remove the check-labels munger as well since the labels.yaml file isn't present in test-infra repo.

fejta:
We definitely want the check-labels munger.

fejta:
It is very important the experience is the same in every repo... then I don't have the friction of trying to learn how each and every repo behaves.

apelisse:
Yeah, but maybe we also want to do it incrementally

fejta:
Sure, doesn't have to be in this PR, but any reason not to create a labels.yaml in the test-infra now?

foxish:
Yes, I agree. I think the next step would be to create the label file and turn up that munger as well.

apelisse:
I'm not sure we want to create lots of unused labels?

PR #1563, opened 2016-08-18
switch queue tie-break to first LGTM time

### LINKED ISSUE #1286: PRs in the queue should sort by first LGTM date
...instead of by PR number. I think this would be fairer to people.

lavalamp:
Fixes #1286

### DIFF
--- a/mungegithub/github/github.go
+++ b/mungegithub/github/github.go
@@ -274,6 +274,11 @@ type MungeObject struct {
 	Annotations map[string]string //annotations are things you can set yourself.
 }
 
+// Number is short for *obj.Issue.Number.
+func (obj *MungeObject) Number() int {
+	return *obj.Issue.Number
+}
+
 // DebugStats is a structure that tells information about how we have interacted
 // with github
 type DebugStats struct {
@@ -595,8 +600,41 @@ func (obj *MungeObject) LastModifiedTime() *time.Time {
 	return lastModified
 }
 
-// labelEvent returns the most recent event where the given label was added to an issue
-func (obj *MungeObject) labelEvent(label string) *github.IssueEvent {
+// FirstLabelTime returns the first time the request label was added to an issue.
+// If the label was never added you will get a nil time.
+func (obj *MungeObject) FirstLabelTime(label string) *time.Time {
+	event := obj.labelEvent(label, firstTime)
+	if event == nil {
+		return nil
+	}
+	return event.CreatedAt
+}
+
+// Return true if 'a' is preferable to 'b'. Handle nil times!
+type timePred func(a, b *time.Time) bool
+
+func firstTime(a, b *time.Time) bool {
+	if a == nil {
+		return false
+	}
+	if b == nil {
+		return true
+	}
+	return !a.After(*b)
+}
+func lastTime(a, b *time.Time) bool {
+	if a == nil {
+		return false
+	}
+	if b == nil {
+		return true
+	}
+	return a.After(*b)
+}
+
+// labelEvent returns the event where the given label was added to an issue.
+// 'pred' is used to select which label event is chosen if there are multiple.
+func (obj *MungeObject) labelEvent(label string, pred timePred) *github.IssueEvent {
 	var labelTime *time.Time
 	var out *github.IssueEvent
 	events, err := obj.GetEvents()
@@ -606,7 +644,7 @@ func (obj *MungeObject) labelEvent(label string) *github.IssueEvent {
 	index := 0
 	for i, event := range events {
 		if *event.Event == "labeled" && *event.Label.Name == label {
-			if labelTime == nil || event.CreatedAt.After(*labelTime) {
+			if pred(event.CreatedAt, labelTime) {
 				labelTime = event.CreatedAt
 				out = event
 				index = i
@@ -619,9 +657,9 @@ func (obj *MungeObject) labelEvent(label string) *github.IssueEvent {
 }
 
 // LabelTime returns the last time the request label was added to an issue.
-// If the label was never added you will get the 0 time.
+// If the label was never added you will get a nil time.
 func (obj *MungeObject) LabelTime(label string) *time.Time {
-	event := obj.labelEvent(label)
+	event := obj.labelEvent(label, lastTime)
 	if event == nil {
 		return nil
 	}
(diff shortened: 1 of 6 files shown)


lavalamp:
Also fyi @krousey, current build cop.

mungegithub/github/github.go
```diff
@@ -595,8 +600,41 @@ func (obj *MungeObject) LastModifiedTime() *time.Time {
...
+	if a == nil {
+		return false
+	}
+	if b == nil {
+		return true
+	}
+	return !a.After(*b)
+}
+func lastTime(a, b *time.Time) bool {
+	if a == nil {
```
apelisse: nit: this is the opposite of lastTime (removes 5 lines of code yay!)

mungegithub/github/github.go
```diff
@@ -595,8 +600,41 @@ func (obj *MungeObject) LastModifiedTime() *time.Time {
...
+func firstTime(a, b *time.Time) bool {
+	if a == nil {
+		return false
+	}
+	if b == nil {
+		return true
+	}
+	return !a.After(*b)
+}
+func lastTime(a, b *time.Time) bool {
```
apelisse: nit: this is the opposite of firstTime (removes 5 lines of code yay!)

mungegithub/github/testing/github.go
```diff
@@ -104,6 +104,27 @@ func Issue(user string, number int, labels []string, isPR bool) *github.Issue {
 	return issue
 }
 
+// MultiIssueEvents packages up events for when you have multiple issues in the
+// test server.
+func MultiIssueEvents(issueToEvents map[int][]LabelTime) (out []*github.IssueEvent) {
```
apelisse: what is the point of the named return?

mungegithub/mungers/submit-queue.go
```diff
@@ -1305,7 +1324,11 @@ func (sq *SubmitQueue) servePriorityInfo(res http.ResponseWriter, req *http.Requ
       <li>PR with no release milestone will be considered after any PR with a milestone</li>
     </ul>
   </li>
-  <li>PR number</li>
+  <li>First time at which the LGTM label was applied.
```
krousey: So I should just create my PRs with LGTM and DO_NOT_MERGE and then remove them, right?

lavalamp: That's effectively how it works for everyone today, since it goes off of
number... If people abuse this in that way we can switch to something else.

mungegithub/github/github.go
```diff
@@ -274,6 +274,11 @@ type MungeObject struct {
 	Annotations map[string]string //annotations are things you can set yourself.
 }
 
+// Number is short for *obj.Issue.Number.
+func (obj *MungeObject) Number() int {
```
eparis: I feel like an idiot for never making this...

mungegithub/github/testing/github.go
eparis: he doesn't have to declare the empty array as the first line of the function?

mungegithub/github/github.go
lavalamp: I only did it so I could make an interface. But yeah, it'd be a convenient
place to check for nil ;)

mungegithub/github/github.go
eparis: It's not quite the opposite because of him using the same `if a== nil { return false } ...`   But I have not idea if those are correct. Haven't worked through it all.

mungegithub/github/github.go
```diff
@@ -595,8 +600,41 @@ func (obj *MungeObject) LastModifiedTime() *time.Time {
 	return lastModified
 }
 
-// labelEvent returns the most recent event where the given label was added to an issue
-func (obj *MungeObject) labelEvent(label string) *github.IssueEvent {
+// FirstLabelTime returns the first time the request label was added to an issue.
+// If the label was never added you will get the 0 time.
```
eparis: If the label was never added you will get nil, not the 0 time (although similar)

mungegithub/mungers/mungerutil/time_cache_test.go
```diff
@@ -0,0 +1,62 @@
...
+		{testLTG{1, "lgtm", timeA}, timeA},
+		{testLTG{2, "blah", timeA}, time.Time{}},
+		{testLTG{1, "lgtm", timeB}, timeA}, // because it will be cached
+		{testLTG{2, "lgtm", timeB}, timeB}, // the empty time was not cached.
+	}
+
+	cache := NewLabelTimeCache("lgtm")
+	for i, tt := range table {
+		got, ok := cache.FirstLabelTime(&tt.obj)
+		if e, a := (tt.expect != time.Time{}), ok; e != a {
```
eparis: what's the point of `a` in all this? why not just use `ok` ? I feel like I'm missing something...

lavalamp: oh, it's just a common pattern to avoid typing long variables multiple
times. not needed here, I can remove if you like

mungegithub/github/github.go
lavalamp: will fix.

mungegithub/github/github.go
apelisse: Yeah, I feel dumb right now :-)

mungegithub/github/testing/github.go
lavalamp: Right. Saves a line.

mungegithub/github/github.go
lavalamp: It probably would work with one being the opposite of the other, but I think it's OK as is.

lavalamp:
Changes pushed, PTAL. Would like to get this in asap.

eparis:
```
!!! 'gofmt -s' needs to be run on the following files: 
./mungegithub/mungers/submit-queue_test.go
!!! 'golint' problems: 
/home/travis/gopath/src/k8s.io/contrib/mungegithub/mungers/mungerutil/time_cache.go:39:1: exported function NewLabelTimeCache should have comment or be unexported
/home/travis/gopath/src/k8s.io/contrib/mungegithub/mungers/mungerutil/time_cache.go:46:1: exported method LabelTimeCache.FirstLabelTime should have comment or be unexported
```

Fix that and feel free to self LGTM

lavalamp:
applying label. Will push when this merges.

PR #1564, opened 2016-08-18
autoscaler: Remove minimum size requirement

philk:
There are use cases (data processing nodes) where an instance group
would need to be zero most of the time and only scale up on demand.

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1564)

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
+++ b/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
@@ -208,9 +208,6 @@ func buildAsg(value string, awsManager *AwsManager) (*Asg, error) {
 		awsManager: awsManager,
 	}
 	if size, err := strconv.Atoi(tokens[0]); err == nil {
-		if size <= 0 {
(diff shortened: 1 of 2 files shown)


philk:
I signed it!

eparis:
@DirectXMan12

philk:
Ah, in testing this I see why it's not as easy as that. Since there are zero nodes the autoscaler can't determine the resource capacity.

I maintain that a zero size autoscale group is a useful thing to have (since c4.8xlarge instances are $1.76/hour) but I don't have any immediate good ideas on how to handle this. Should I open an issue for discussion or keep this here? Happy to work on it just don't see a reasonable solution to getting node capacity when there aren't any nodes (somehow mapping from the MIG/ASG but that seems prone to failure)

fgrzadkowski:
@mwielgus

fgrzadkowski:
This change will not work.

For our simulations we need to know how each "node type" looks like. Internally we build a "node template", which is based on the existing nodes. If we scale a group to 0, then we'll not have any nodes to base the template on.

PR #1565, opened 2016-08-18
fix bug in fluentd-es-logging sidecar

harryge00:
`filename=$(basename $filepath)` will cause problem if multiple files are of the same base name, 
e.g. 
`"/mnt/log1/synthetic-dates.log /mnt/log2/synthetic-dates.log"` are two different files, but there will be only one `/etc/td-agent/files/synthetic-dates.log` indicating fluentd to collect logs from  `/mnt/log2/synthetic-dates.log`.

In my case, the config under /etc/td-agent/files become like `/etc/td-agent/files/_mnt_log1_synthetic-dates.log`

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1565)

### DIFF
--- a/logging/fluentd-sidecar-es/config_generator.sh
+++ b/logging/fluentd-sidecar-es/config_generator.sh
@@ -19,9 +19,24 @@ if [ -z "$FILES_TO_COLLECT" ]; then
   exit 0
 fi
 
+declare -A names_count
+
+function get_uncollided_name {
+  next_name=$1
+  base_name=$1
+  while [ ${names_count[$next_name]} ];
+  do
+    ((names_count[$base_name]++))
+    next_name=${base_name}_${names_count[$base_name]}
+  done
+  names_count+=([$next_name]=1)
+}
+
 for filepath in $FILES_TO_COLLECT
 do
   filename=$(basename $filepath)
+  get_uncollided_name $filename
+  filename=$next_name
   cat > "/etc/td-agent/files/${filename}" << EndOfMessage
 <source>
   type tail


logging/fluentd-sidecar-es/config_generator.sh
```diff
@@ -21,7 +21,7 @@ fi
 
 for filepath in $FILES_TO_COLLECT
 do
-  filename=$(basename $filepath)
+  filename=`echo $filepath|sed -e "s/\//_/g"`
```
jimmidyson: This could still lead to collisions if files have underscore in the name. Unlikely, but possible.

Can you check if the target file exists & add perhaps an incrementing suffix if it does?

harryge00:
@jimmidyson I have updated. Now an associate array is used to record the count of filename. If a filename has been used before, then add "_nextIndex" as suffix. 

If "FILES_TO_COLLECT=/data/base /mnt/base /test/base /home/root/base_1 /var/log/base_2", then
```
/etc/td-agent/files/base is for /data/base
/etc/td-agent/files/base_2 is for /mnt/base
/etc/td-agent/files/base_3 is for /test/base
/etc/td-agent/files/base_1 is for /home/root/base_1
/etc/td-agent/files/base_2_2 is for /var/log/base_2
```

test-foxish:
recomputing cla status...

harryge00:
I signed it!

grodrigues3:

- **logging**

cc @eparis 

PR #1568, opened 2016-08-19
Consider .png files as doc and don't rerun tests

### LINKED ISSUE #1566: retest-not-required-docs-only should count images as docs
E.g. the label was removed from [this PR](https://github.com/kubernetes/kubernetes/pull/30840) since it includes several png's referenced in the docs.

apelisse:
Fixes #1566

### DIFF
--- a/mungegithub/mungers/docs-no-retest.go
+++ b/mungegithub/mungers/docs-no-retest.go
@@ -31,7 +31,7 @@ const (
 )
 
 var (
-	ignoreFilesRegex = regexp.MustCompile(".*\\.md$")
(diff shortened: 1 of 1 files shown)


eparis:
does it have to be caps?
/LGTM

apelisse:
It's case insensitive. But the munger is not set-up yet in contrib

foxish:
@apelisse @eparis It would've worked if you were an assignee. I think after the approvers mechanism is in place, the approvers would also be able to lgtm, even if they're not assigned. The bot should have posted a message that it wasn't honoring the `/lgtm` as you weren't an assignee.

PR #1571, opened 2016-08-19
ingress: use POD_NAMESPACE as a namespace in cli parameters

kayrus:
When you deploy ingress not into `default` namespace, ingress RC fails with the `no service with name default/default-http-backend found: services "default-http-backend" not found` error message.

This fix uses `POD_NAMESPACE` which we already pass into the pod ENV.

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1571)

### DIFF
--- a/ingress/controllers/gce/rc.yaml
+++ b/ingress/controllers/gce/rc.yaml
@@ -78,5 +78,5 @@ spec:
             cpu: 100m
             memory: 50Mi
         args:
-        - --default-backend-service=default/default-http-backend
+        - --default-backend-service=$(POD_NAMESPACE)/default-http-backend
         - --sync-period=300s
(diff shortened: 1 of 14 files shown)


bprashanth:
makes sense to me, will wait for Manuel to lgtm

aledbf:
@kayrus just a comment: in all the examples the default backend service is created in the default namespace ([default-backend.yaml](https://github.com/kubernetes/contrib/tree/master/ingress/controllers/nginx/examples/default-backend.yaml)). Is up to the user to specify the correct service. 
From the docs: `--default-backend-service: Service used to serve a 404 page for the default backend. Takes the form namespace/name. The controller uses the first node port of this Service for the default backend.`
If @bprashanth is ok with this change then is lgtm.

aledbf:
@kayrus please do not change the content of the `ingress/vendor/` directory. The content of that directory is generated by `godep`.

kayrus:
@aledbf made this vendor change by script. restored.

as for the default backend, I know that.

bprashanth:
LGTM thanks

PR #1578, opened 2016-08-19
[rescheduler] Fixed few issues in rescheduler

piosz:
1. Wait 2 minutes after rescheduler is created in order to
   make sure that all addons has a chance to start.
2. Operate on copy of a node (without taint) while doing simulations.
3. Correctly handle iteration over PodList.

### DIFF
--- a/rescheduler/rescheduler.go
+++ b/rescheduler/rescheduler.go
@@ -53,12 +53,19 @@ var (
 
 	systemNamespace = flags.String("system-namespace", kube_api.NamespaceSystem,
 		`Namespace to watch for critical addons.`)
+
+	initialDelay = flags.Duration("initial-delay", 2*time.Minute,
+		`How long should rescheduler wait after start to make sure
+		 all critical addons had a chance to start.`)
 )
 
 func main() {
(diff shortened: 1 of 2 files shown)


rescheduler/rescheduler.go
```diff
@@ -58,6 +58,10 @@ var (
 func main() {
 	glog.Infof("Running Rescheduler")
 
+	// Wait for cluster stabilization.
+	// TODO(piosz): figure our a better way of verifying cluster stabilization here.
+	time.Sleep(2 * time.Minute)
```
mwielgus: can you make it a flag?

rescheduler/rescheduler.go
```diff
@@ -58,6 +58,10 @@ var (
...
+
 	flags.Parse(os.Args)
```
mwielgus: Flag parsing should be the first thing. It may affect glog.

rescheduler/rescheduler.go

rescheduler/rescheduler.go

mwielgus:
lgtm

PR #1591, opened 2016-08-22
[nginx-ingress-controller] Always listen on port 443, even without ingress rules

aledbf:
Currently is required at least one ingress rule to listen on port 443


skinny:
as (the) one having issues with this I can confirm this solves the problem. the default (_) server now listens on port 443 as well. Thanks!

bprashanth:
LGTM

PR #1593, opened 2016-08-22
Only show 512 characters of comments in the logs

eparis:
When creating new issues or adding comments to existing issue we can
create VERY long comments which make the logs really hard to read. We
don't need all of that in the logs.

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1593)

### DIFF
--- a/mungegithub/github/github.go
+++ b/mungegithub/github/github.go
@@ -1611,9 +1611,13 @@ func (obj *MungeObject) ListComments(withListOpts ...WithListOpt) ([]*github.Iss
 // WriteComment will send the `msg` as a comment to the specified PR
 func (obj *MungeObject) WriteComment(msg string) error {
 	config := obj.config
-	prNum := *obj.Issue.Number
+	prNum := obj.Number()
 	config.analytics.CreateComment.Call(config, nil)
-	glog.Infof("Commenting %q in %d", msg, prNum)
+	comment := msg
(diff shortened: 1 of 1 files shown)


mungegithub/github/github.go
```diff
@@ -1611,9 +1611,13 @@ func (obj *MungeObject) ListComments(withListOpts ...WithListOpt) ([]*github.Iss
 // WriteComment will send the `msg` as a comment to the specified PR
 func (obj *MungeObject) WriteComment(msg string) error {
 	config := obj.config
-	prNum := *obj.Issue.Number
+	prNum := obj.Number()
 	config.analytics.CreateComment.Call(config, nil)
-	glog.Infof("Commenting %q in %d", msg, prNum)
+	comment := msg
+	if len(comment) > 512 {
+		comment = comment[:512]
```
lavalamp: I would prefer to truncate error messages within the comment. This is likely to leave out important stuff like "happened in presubmit on #blah"

eparis: I'm not sure exactly what you think would be ok to truncate. This is only truncating what goes into the `kubectl logs`, (obviously not what goes in the issue itself)

lavalamp: oh! sorry. I totally missed that this was just the log message.

/lgtm

PR #1596, opened 2016-08-22
[nginx-ingress-controller]: Adapt nginx hash sizes to the number of ingress

### LINKED ISSUE #1487: [nginx-ingress-controller] Improve nginx default
Default values of `types_hash_max_size`, `server_names_hash_max_size` and `server_names_hash_bucket_size` should take in account to the number of Ingress rules and adjust the default value accordingly

aledbf:
This change allows the tuning of 2 important NGINX variables:
- server_names_hash_max_size
- server_names_hash_bucket_size

The default values should be enough for most of the users but after +300 Ingress rules or long hostnames as FQDN NGINX requires tuning of this values or it will not start.

The introduced change allows the self-tuning using the Ingress information
Using `--v=3` it's possible to see the changes:

```
...
I0822 21:42:10.517778       1 template.go:84] adjusting ServerNameHashMaxSize variable from 4096 to 16384
...
```

fixes #1487

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1596)

### DIFF
--- a/ingress/controllers/nginx/nginx/config/config.go
+++ b/ingress/controllers/nginx/nginx/config/config.go
@@ -167,7 +167,7 @@ type Configuration struct {
 	// http://nginx.org/en/docs/http/ngx_http_core_module.html#server_names_hash_max_size
 	ServerNameHashMaxSize int `structs:"server-name-hash-max-size,omitempty"`
 
-	// Size of the bucker for the server names hash tables
+	// Size of the bucket for the server names hash tables
 	// http://nginx.org/en/docs/hash.html
 	// http://nginx.org/en/docs/http/ngx_http_core_module.html#server_names_hash_bucket_size
 	ServerNameHashBucketSize int `structs:"server-name-hash-bucket-size,omitempty"`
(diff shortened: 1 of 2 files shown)


ingress/controllers/nginx/nginx/template.go
```diff
@@ -64,6 +64,27 @@ func (ngx *Manager) loadTemplate() error {
 }
 
 func (ngx *Manager) writeCfg(cfg config.Configuration, ingressCfg IngressConfig) (bool, error) {
+	var longestName int
+	var serverNames int
+	for _, srv := range ingressCfg.Servers {
```
bprashanth: what's the tradeoff in being so precise vs just defaulting it to something high, documenting the limit, and not writing this code?

aledbf: > what's the tradeoff in being so precise

Using a small value means nginx cannot start or be restarted
```
[emerg] 17617#0: could not build the server_names_hash, you should increase server_names_hash_bucket_size: 32
```

> just defaulting it to something high

This depends on the amount of hostnames and FQDN length

bprashanth: yeah i mean just start off with it set to 16384, what do we lose?

aledbf: > what do we lose?

Nothing, but for instance I need at least 49152 :)

aledbf:
@bprashanth ping

bprashanth:
LGTM

PR #1597, opened 2016-08-23
Stop processing the pull-request if Github fails

apelisse:
Today we keep going because we don't properly handle errors, and we may
end-up posting a comment everytime we can't read the comments.
Let's check if we can fetch the comment and fail if it's broken.

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1597)

### DIFF
--- a/mungegithub/mungers/close-stale-pr.go
+++ b/mungegithub/mungers/close-stale-pr.go
@@ -48,7 +48,7 @@ You can add 'keep-open' label to prevent this from happening, or add a comment t
 
 var (
 	closingCommentRE = regexp.MustCompile(`This PR hasn't been active in \d+ days?\..*label to prevent this from happening again`)
-	warningCommentRE = regexp.MustCompile(`This PR hasn't been active in \d+ days?\..*be closed in \d+ days`)
+	warningCommentRE = regexp.MustCompile(`This PR hasn't been active in \d+ days?\..*be closed in \d+ days?`)
 )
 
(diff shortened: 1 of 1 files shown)


apelisse:
@lavalamp this is an attempt to fix the bug on https://github.com/kubernetes/kubernetes/pull/27193.

I'm not sure if this is what went wrong, but that can't hurt anyway.

mungegithub/mungers/close-stale-pr.go
```diff
@@ -295,7 +303,7 @@ func (CloseStalePR) Munge(obj *github.MungeObject) {
 		checkAndWarn(obj, inactiveFor, closeIn)
 	} else {
 		// Pull-request is active. Remove previous potential warning
-		comment := findLatestWarningComment(obj)
+		comment, _ := findLatestWarningComment(obj)
```
lavalamp: handle the error.

mungegithub/mungers/close-stale-pr.go
```diff
@@ -180,12 +180,12 @@ func findLastModificationTime(obj *github.MungeObject) (*time.Time, error) {
 	return lastModif, nil
 }
 
-func findLatestWarningComment(obj *github.MungeObject) *githubapi.IssueComment {
+func findLatestWarningComment(obj *github.MungeObject) (*githubapi.IssueComment, error) {
```
lavalamp: nit: document, include that `nil, nil` is a possible return value.

PR #1606, opened 2016-08-24
copy inventory group_vars to vagrant inventory directory

ingvagabund:
Vagrant creates its own inventory file without predefined ansible group_vars.
Thus, it is necessary to copy the directory to vagrant inventory as well.

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1606)


ingvagabund:
@jasonbrooks, @stephenrlouie @danehans can you verify the fix works for you? I have run it locally with and without group_vars directory created under .vagrant directory. In both cases the cluster got deployed.

jasonbrooks:
This works for me. +1

PR #1610, opened 2016-08-24
Add docker source type

### LINKED ISSUE #455: Ansible: Allow for use of Docker yum repositories

ingvagabund:
Continuing with effort in https://github.com/kubernetes/contrib/pull/456 and https://github.com/kubernetes/contrib/pull/291. All credits to @rsmitty and @furikake for bringing the topic up. I have just put crumbs together.

With this PR docker can be installed in addtion to the usual way (via distribution package manager) from custom docker repository or from distribution rpm.

Introducing separate playbook and script so it is easier to test the docker role separately.

Fixes: #455

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1610)

### DIFF
--- a/ansible/scripts/deploy-docker.sh
+++ b/ansible/scripts/deploy-docker.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+# Copyright 2016 The Kubernetes Authors All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+. ./init.sh
+
+inventory=${INVENTORY:-${INVENTORY_DIR}/inventory}
+ansible-playbook -i ${inventory} ${PLAYBOOKS_DIR}/deploy-docker.yml "$@"
--- a/ansible/playbooks/deploy-cluster.yml
+++ b/ansible/playbooks/deploy-cluster.yml
@@ -14,13 +14,7 @@
 - include: deploy-etcd.yml
 
 # Install docker
-- hosts: all
-  become: yes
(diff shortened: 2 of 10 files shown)


ansible/scripts/deploy-docker.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+# Copyright 2015 The Kubernetes Authors All rights reserved.
```
eparis: 2016

ansible/scripts/deploy-docker.sh
```diff
@@ -0,0 +1,20 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+. ./init.sh
+
+inventory=${INVENTORY:-${INVENTORY_DIR}/inventory}
+ansible-playbook -i ${inventory} ${PLAYBOOKS_DIR}/deploy-docker.yml "$@"
```
eparis: I get how this is good for testing, but do we need this actually in the repo? would we expect people to use kuberntes-ansible to deploy docker?

ingvagabund: Someone can use it to update docker in kubernetes cluster at most. Otherwise, no.

ansible/scripts/deploy-docker.sh
ingvagabund: updated

ansible/scripts/deploy-docker.sh
ingvagabund: Later, it can be used to diagnose docker state in the cluster. For instance adding a task that will do ``systemctl status docker`` (and other checks). Then, it would make more sense.

ingvagabund:
@redhat-k8s-bot test please

ingvagabund:
Running deployment with both ``distribution-rpm`` and ``custom-repository`` source types. As long as the installed docker rpm is ``docker``, the cluster gets deployed. For the case of ``docker-engine``, kubernetes distribution spec files needs to be update to allow both ``docker`` and ``docker-engine``. Meantime, ``kube_source_type=github-release``.

PR #1612, opened 2016-08-24
iptables: adapt for new iptables mode

jlebon:
In the new "iptables" mode of kube-proxy, packets addressed for a node
port will get DNAT'ed to one of the endpoints. If that endpoint does not
reside on the node, then flannel needs to be able to do the proxying to
the endpoint. We generalize and add flannel rules for this.

If the endpoint does reside on the node, then the packet will be
addressed to the docker bridge, in which case we need to allow it. We
modify the node rules for this. The previous node rules there are no
longer needed (they were only relevant in the "userspace" mode).

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1612)

### DIFF
--- a/ansible/roles/flannel/tasks/firewalld.yml
+++ b/ansible/roles/flannel/tasks/firewalld.yml
@@ -13,14 +13,24 @@
   # we have to manually invoke firewall-cmd
 - name: Open flanneld subnet traffic
   command: /bin/firewall-cmd --direct --add-rule ipv4 filter FORWARD 1
-           -i flannel.1 -o docker0 -j ACCEPT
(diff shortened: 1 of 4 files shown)


jlebon:
Modified to keep the previous rules as well so that we support both userspace and iptables proxy modes.

In the bigger picture, this might be something that should be offloaded to kube-proxy during startup.

jasonbrooks:
I just tested this w/ centos atomic & it works for me. Thanks!

ingvagabund:
@redhat-k8s-bot test please

PR #1613, opened 2016-08-24
Add kubernetes-cross-build to list of non-blocking builds

ixdy:
I'd prefer this to be a blocking build, but we aren't doing a full crossbuild on PRs, so I guess we don't want to make it blocking?

Broken crossbuild does already block cutting releases, though.

cc @luxas @david-mcmahon @spxtr 

x-ref https://github.com/kubernetes/test-infra/issues/393

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1613)

### DIFF
--- a/mungegithub/submit-queue/deployment/kubernetes/configmap.yaml
+++ b/mungegithub/submit-queue/deployment/kubernetes/configmap.yaml
@@ -16,7 +16,7 @@ data:
 
   # submit-queue options.
   submit-queue.required-contexts: "Jenkins GCE Node e2e"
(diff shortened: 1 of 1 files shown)


lavalamp:
/lgtm

Yup, can't be blocking if there's no presubmit :)

luxas:
We want it blocking, but we should take one step at a time.
Eventually, a `make release` job should run in parallel with the other jobs that run on every PR.
It clearly has shown that it just is worth it to run `make release` before PRs are merged...
The alpha.3 release is now _delayed_ about two weeks just because we don't have it as presubmit

Issue #1325, opened 2016-07-07
systemd defaults: remove ServiceAccount from admissions


ingvagabund:
As a starting user of kubernetes I would like to use the default account as much as possible and discover concept of admissions later. Currently, when the kubernetes is installed and the ``kube-apiserver`` started, I can not start playing with k8s right away. I have to tune the configuration first.

Let's discuss what are the disadvantages of disabling the ``ServiceAccount`` admission control [1].

[1] https://github.com/kubernetes/contrib/blob/master/init/systemd/environ/apiserver#L23

sjenning:
I think it should be removed from the default list and a comment added saying it can be added back after the controller-manager is configured with a ```--service-account-private-key```.

ingvagabund:
http://stackoverflow.com/questions/33528398/why-dont-i-have-a-default-serviceaccount-on-kubernetes

ingvagabund:
Maybe, it would make more sense to write down an article/tutorial how to setup the service account. Or at least collect a list of already existing articles. With ServiceAccount on a user is forced to google out the issue and learn more about (at least to remove the ServiceAccount from the admissions). However, yes, it is not install and run.

sjenning:
IIRC the minimum procedure for enabling services accounts is:
- create a self-signed apiserver key/cert
- modify the apiserver ``` --service-account-key-file``` and controller-manager ```- -service-account-private-key``` configs to use the apiserver key for service account secret creation and authorization
- restart the apiserver and controller-manager

This is at least a 4 or 5 step process

ingvagabund:
What about to extend the kubernetes rpms with ``deploy.sh`` script that will deploy the kubernetes cluster on the host? It would run a subset of playbooks and take care of all the steps. Something like:

```
$ cat deploy.sh
sudo dnf install ansible kubernetes-ansible -y
cd /usr/share/ansible/scripts; ./deploy-one-node-cluster.sh
```

ingvagabund:
At the same time, it make sense to install kubernetes-addons as well so a user have running DNS and other extensions.

ingvagabund:
With #1689 the kubernetes is deployed on localhost with addons on.

PR #1621, opened 2016-08-25
WIP: replace some resources get via get_url with local copy

ingvagabund:
Minimize reliance on remote resources, use local ones instead.

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1621)

### DIFF
--- a/ansible/roles/master/tasks/retrieve-systemd-files.yml
+++ b/ansible/roles/master/tasks/retrieve-systemd-files.yml
@@ -1,9 +1,6 @@
 ---
-- name: Get Systemd unit files from Kubernetes repository
-  get_url:
-    url=https://raw.githubusercontent.com/kubernetes/contrib/master/init/systemd/{{ item }}.service
-    dest=/etc/systemd/system/{{ item }}.service
-    force=yes
+- name: Install Systemd unit files
+  copy: src={{ item }}.service dest=/etc/systemd/system/{{ item }}.service
   register: "{{ item }}_service"
   notify:
     - reload systemd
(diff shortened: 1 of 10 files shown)


ingvagabund:
@redhat-k8s-bot test please

ingvagabund:
Tested within the same CI with "kube_source_type=github-release kube_version=1.2.4". Kubernetes got deployed.

eparis:
I hate having two places for these things to live and to get out of date...
/lgtm

ingvagabund:
@eparis I don't like it either. What about to completely remove init/systemd and move its content under corresponding files directories of master/node/kubernetes under ansible? There is no difference if people get the service and config files from ``init/systemd`` or from ``ansible/roles/[master|node|kubernetes]/files``.

ansible/roles/master/tasks/retrieve-systemd-files.yml
```diff
@@ -1,9 +1,6 @@
 ---
-- name: Get Systemd unit files from Kubernetes repository
-  get_url:
-    url=https://raw.githubusercontent.com/kubernetes/contrib/master/init/systemd/{{ item }}.service
```
sttts: Are those files now duplicated in the contrib repo?

ingvagabund: They are. However, the contrib repository is planned to be retired. So we need to find new place for all the systemd and config files.

ingvagabund:
The config files are meant for ``1``-node cluster anyway [1]. Though, it does not make much sense to use the ansible to deploy the ``1``-node cluster, it generates tokens and launches kubernetes-addons. Which is the missing piece when you install kubernetes from rpms. Still not a plug&play solution [2].

In short:
**environ & config files**
* `config` and `kubernetes.conf` are under kubernetes role as it is common for all components
* `apiserver`, `controller-manager` and `scheduler` belong to master components only (plus with `kubelet` in some cases)
* `kubelet` and `proxy` belong to node components only

**service files**
* `kube-apiserver.service`, `kube-controller-manager.service`, `kube-scheduler.service` belongs to master components (plus with `kubelet.service` in some cases)
* `kubelet.service` and `kube-proxy.service` belongs to node components

Service files are not templated so far. Only configs are. And afaik, all the non-templated config files are used in distribution builds only.

But still, we will have at duplicates. Until resolved, let's not merge the PR.

[1] https://github.com/kubernetes/contrib/tree/master/init/systemd
[2] https://github.com/kubernetes/contrib/issues/1325

ingvagabund:
Or another option is to

1. move init/systemd under ansible directory
1. in every occurrence of the systemd service of config file check if "../init/systemd/[NAME.service|environ/NAME] file exists. If it does, copy it, if it does not, download it.

It will not change behaviour of the current PR. At the same time when each role gets publish in Ansible galaxy, it will work. And, there will be no (obvious) duplicates. But still the files get downloaded unless the roles are used directly from the contrib repository.

There is already some duplication between templated and non-templated config files. However, it should not matter as the non-templated configs are used purely for `1`-node cluster and can be independent of templated ones (unless some fundamental thing needs to change).

sttts:
If the other files outside of ansible have a different purpose and life-cycle, I don't care about duplicates. But a self contained ansible directory without unversioned files downloaded with http(s) is a good goal.

test-foxish:
recomputing cla status...

grodrigues3:

- **ansible**

cc @bprashanth 

PR #1629, opened 2016-08-26
Fixed #1628: kube-down should not delete multi-user.target.wants dir

### LINKED ISSUE #1628: kube-down should not delete multi-user.target.wants directory on centos
```
- name: Remove remaining files
      file: path={{ item }} state=absent
      with_items:
          ...
          - /etc/systemd/system/multi-user.target.wants
          ...
```

Because in multi-user.target.wants directory, there are some other useful services.
For example sshd.service, once deleted this direcotry, we will fail to access the machine by ssh.

keontang:
On Centos, there are some other service in multi-user.target.wants directory, so we should not just delete it.

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1629)

### DIFF
--- a/ansible/playbooks/adhoc/uninstall.yml
+++ b/ansible/playbooks/adhoc/uninstall.yml
@@ -138,7 +138,14 @@
         - /etc/systemd/system/kube-scheduler.service
         - /etc/systemd/system/flanneld.service
         - /etc/systemd/system/flanneld.service.wants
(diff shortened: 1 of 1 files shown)


keontang:
[https://github.com/kubernetes/contrib/issues/1628](https://github.com/kubernetes/contrib/issues/1628)

PR #1630, opened 2016-08-26
[rescheduler] actually wait for pod to be scheduled

piosz:
https://github.com/kubernetes/kubernetes/issues/29023

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1630)

### DIFF
--- a/rescheduler/rescheduler.go
+++ b/rescheduler/rescheduler.go
@@ -28,6 +28,7 @@ import (
 	kube_client "k8s.io/kubernetes/pkg/client/unversioned"
 	"k8s.io/kubernetes/pkg/fields"
 	kubectl_util "k8s.io/kubernetes/pkg/kubectl/cmd/util"
+	"k8s.io/kubernetes/pkg/util/wait"
 	"k8s.io/kubernetes/plugin/pkg/scheduler/schedulercache"
 
 	"github.com/golang/glog"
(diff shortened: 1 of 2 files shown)


piosz:
cc @mwielgus

rescheduler/rescheduler.go
```diff
@@ -130,10 +134,22 @@ func main() {
 	}
 }
 
-func waitForScheduled(podsBeingProcessed *podSet, pod *kube_api.Pod) {
-	// TODO(piosz): periodically check whether pod is actually scheduled
-	glog.Infof("Waiting for pod %s to be scheduled", podId(pod))
-	time.Sleep(10 * time.Minute)
+func waitForScheduled(client *kube_client.Client, podsBeingProcessed *podSet, pod *kube_api.Pod) {
+	for start := time.Now(); time.Since(start) < *podScheduledTimeout; time.Sleep(10 * time.Second) {
```
wojtek-t: Not sure if you already have this dependency in contrib (and if not how easy it is to import it), but it would be great if you could use:
wait.Poll(...)
from: https://github.com/kubernetes/kubernetes/blob/master/pkg/util/wait/wait.go#L158
function instead.

Also, please make the sleep smaller (like 500ms or sth like that).

wojtek-t:
LGTM

PR #1636, opened 2016-08-26
ansible: added etcd cert auth support

kayrus:
subj

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1636)

### DIFF
--- a/ansible/roles/etcd/defaults/main.yaml
+++ b/ansible/roles/etcd/defaults/main.yaml
@@ -9,8 +9,13 @@ etcd_peer_port: 2380
 etcd_client_legacy_port: 4001
 etcd_peer_legacy_port: 7001
 etcd_peers_group: etcd
+
 etcd_url_scheme: http
 etcd_peer_url_scheme: http
+# Set correspondingly to true when use https values above to enable cert auth
+etcd_client_cert_auth: false
+etcd_peer_client_cert_auth: false
+
 etcd_conf_dir: /etc/etcd
 etcd_script_dir: /usr/libexec/etcd
 etcd_ca_file: "{{ etcd_conf_dir }}/ca.crt"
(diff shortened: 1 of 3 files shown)


ansible/roles/etcd/defaults/main.yaml
```diff
@@ -9,8 +9,13 @@ etcd_peer_port: 2380
 etcd_client_legacy_port: 4001
 etcd_peer_legacy_port: 7001
 etcd_peers_group: etcd
+
 etcd_url_scheme: http
 etcd_peer_url_scheme: http
+# Set correspondingly to true when use https values above to enable cert auth
+etcd_client_cert_auth: true
+etcd_peer_client_cert_auth: true
```
ingvagabund: Both ``ETCD_CLIENT_CERT_AUTH`` and ``ETCD_PEER_CLIENT_CERT_AUTH`` are false by default.

Setting both to ``true`` changes the default behaviour of deployed etcd. I would rather set it to false and leave the final decision to operator. What do you think?

[1] https://github.com/coreos/etcd/blob/master/Documentation/op-guide/configuration.md#--client-cert-auth
[2] https://github.com/coreos/etcd/blob/master/Documentation/op-guide/configuration.md#--peer-client-cert-auth

kayrus: May be. But there should be a note somewhere that TLS cert doesn't imply the auth by cert.

ingvagabund: Can you write that down into README.md?

kayrus: Ok, little bit later.

kayrus:
@ingvagabund Updated README.md

ingvagabund:
@redhat-k8s-bot test please

ingvagabund:
@redhat-k8s-bot test please

PR #1626, opened 2016-08-25
Ping people when flaky issues don't get attention

apelisse:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1626)

PR #1638, opened 2016-08-26
First step for FSM Munger. [WIP]

foxish:
Author:    Anirudh Ramanathana ramanathana@google.com

@apelisse PTAL, this is still a WIP but I'd like to know what's good and bad about the approach. Just wanted to check in before everything gets rethought. Will revert changes to mungegithub/mungers/matchers/comment/*.go after your https://github.com/kubernetes/contrib/pull/1626 goes through.

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1638)

### DIFF
--- a/mungegithub/github/github.go
+++ b/mungegithub/github/github.go
@@ -720,6 +720,22 @@ func (obj *MungeObject) AddLabel(label string) error {
 	return obj.AddLabels([]string{label})
 }
 
+// AddLabelIfAbsent adds a single `label` to the issue, if it isn't present already.
+func (obj *MungeObject) AddLabelIfAbsent(label string) error {
+	if !obj.HasLabel(label) {
+		return obj.AddLabels([]string{label})
+	}
+	return nil
+}
+
+// RemoveLabelIfPresent removes a single `label` from the issue, if it is present.
+func (obj *MungeObject) RemoveLabelIfPresent(label string) error {
+	if obj.HasLabel(label) {
+		return obj.RemoveLabel(label)
+	}
+	return nil
+}
+
 // AddLabels will add all of the named `labels` to the issue
 func (obj *MungeObject) AddLabels(labels []string) error {
 	config := obj.config
--- a/hack/verify-flags/known-flags.txt
+++ b/hack/verify-flags/known-flags.txt
@@ -97,6 +97,7 @@ source-file
 ssl-ca-cert
 ssl-cert
 start-from
+state-machine-enabled
 stats-port
 sync-period
 test-owners-csv
--- a/mungegithub/github/testing/github.go
+++ b/mungegithub/github/testing/github.go
@@ -125,6 +125,14 @@ func MultiIssueEvents(issueToEvents map[int][]LabelTime) (out []*github.IssueEve
 	return out
 }
 
+// User returns a github user with the necessary details filled in.
+func User(name string) *github.User {
+	return &github.User{
+		Login:     stringPtr(name),
+		AvatarURL: stringPtr("MyAvatarURL"),
+	}
+}
+
 // LabelTime is a struct which can be used to call Events()
 // It expresses what label the event should be about and what time
 // the event took place.
--- a/mungegithub/mungegithub.go
+++ b/mungegithub/mungegithub.go
@@ -30,6 +30,8 @@ import (
 
 	"github.com/golang/glog"
 	"github.com/spf13/cobra"
+	"k8s.io/contrib/mungegithub/mungers/fsm"
+	"k8s.io/contrib/mungegithub/mungers/fsm/states"
 	"k8s.io/contrib/mungegithub/mungers/mungerutil"
 )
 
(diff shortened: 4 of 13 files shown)


mungegithub/github/github.go
```diff
@@ -720,6 +720,22 @@ func (obj *MungeObject) AddLabel(label string) error {
 	return obj.AddLabels([]string{label})
 }
 
+// AddLabelIfAbsent adds a single `label` to the issue, if it isn't present already.
+func (obj *MungeObject) AddLabelIfAbsent(label string) error {
```
apelisse: would it make sense to add this code directly to AddLabel(s) and RemoveLabel?

apelisse:
Reviewed 13 of 13 files at r1.
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks broke.

---

*[mungegithub/mungegithub.go, line 74 \[r1\]](https://reviewable.kubernetes.io:443/reviews/kubernetes/contrib/1638#-KQ6XElNog42C655vFTq:-KQ6XElNog42C655vFTr:b-kmtffn) ([raw file](https://github.com/kubernetes/contrib/blob/404e0f4ae459994b8a78bf75faaa1fc5d546eb39/mungegithub/mungegithub.go#L74)):*
> ```Go
> 		}
> 
> 		if config.StateMachineEnabled {
> ```

Should we do it after or before the munger pass?

---

*[mungegithub/github/github.go, line 723 \[r1\]](https://reviewable.kubernetes.io:443/reviews/kubernetes/contrib/1638#-KQ6X0qJ0vrHOAIxOKNf:-KQ6X0qJ0vrHOAIxOKNg:b-a1ghpe) ([raw file](https://github.com/kubernetes/contrib/blob/404e0f4ae459994b8a78bf75faaa1fc5d546eb39/mungegithub/github/github.go#L723)):*
> ```Go
> }
> 
> // AddLabelIfAbsent adds a single `label` to the issue, if it isn't present already.
> ```

would it make sense to add this code directly to AddLabel(s) and RemoveLabel?

---

*[mungegithub/mungers/fsm/states/changes-needed.go, line 39 \[r1\]](https://reviewable.kubernetes.io:443/reviews/kubernetes/contrib/1638#-KQ7P9lzSelbdjKRLstX:-KQ7P9lzSelbdjKRLstY:bfq7tbp) ([raw file](https://github.com/kubernetes/contrib/blob/404e0f4ae459994b8a78bf75faaa1fc5d546eb39/mungegithub/mungers/fsm/states/changes-needed.go#L39)):*
> ```Go
> 	// There is no proceeding from this state.
> 	obj.AddLabelIfAbsent(labelChangesNeeded)
> 	return &End{}, nil
> ```

I guess it makes sense to stay in "ChangesNeeded", or we would never have come in the state in the first place?

---

*[mungegithub/mungers/fsm/states/needs-review.go, line 41 \[r1\]](https://reviewable.kubernetes.io:443/reviews/kubernetes/contrib/1638#-KQ7TA2rkXm5S0uacy5m:-KQ7TA2rkXm5S0uacy5n:bi0xfn2) ([raw file](https://github.com/kubernetes/contrib/blob/404e0f4ae459994b8a78bf75faaa1fc5d546eb39/mungegithub/mungers/fsm/states/needs-review.go#L41)):*
> ```Go
> 	var result bool
> 	var err error
> 	if nr.checkLGTM(obj) {
> ```

If we have LGTM, then we don't needreview anymore? So we shouldn't return End, should we? I would have expected a move to NeedApproval (or however it's named)

---

*[mungegithub/mungers/fsm/states/needs-review.go, line 69 \[r1\]](https://reviewable.kubernetes.io:443/reviews/kubernetes/contrib/1638#-KQ7uAb1ZeHz6_ScQO7_:-KQ7uAb1ZeHz6_ScQO7a:b-xta6ef) ([raw file](https://github.com/kubernetes/contrib/blob/404e0f4ae459994b8a78bf75faaa1fc5d546eb39/mungegithub/mungers/fsm/states/needs-review.go#L69)):*
> ```Go
> 	}
> 
> 	lastAuthorComment := comment.FilterComments(comments, comment.AuthorLogin(*obj.Issue.User.Login)).GetLast()
> ```

I would maybe make something different than `AuthorLogin`, just `AuthorUser` take a `...*github.User` (it would also be safer than all the "implicit" pointer de-references you have here)

---

*[mungegithub/mungers/fsm/states/prereview_test.go, line 76 \[r1\]](https://reviewable.kubernetes.io:443/reviews/kubernetes/contrib/1638#-KQ7vjTXKiEe3jZdHIxH:-KQ7vjTXKiEe3jZdHIxI:byroiv6) ([raw file](https://github.com/kubernetes/contrib/blob/404e0f4ae459994b8a78bf75faaa1fc5d546eb39/mungegithub/mungers/fsm/states/prereview_test.go#L76)):*
> ```Go
> 		test.prIssue.Assignees = test.assignees
> 		client, server, mux := github_test.InitServer(t, test.prIssue, test.pr, nil, nil, nil, nil, nil)
> 		path := fmt.Sprintf("/repos/o/r/issue/%d/labels", *test.prIssue.Number)
> ```

Shouldn't this code be moved into the testing framework instead?

---

*[mungegithub/mungers/matchers/comment/comment.go, line 74 \[r1\]](https://reviewable.kubernetes.io:443/reviews/kubernetes/contrib/1638#-KQ7vwarsU4kMUgTR6jR:-KQ7vwarsU4kMUgTR6jS:b-l2rjhg) ([raw file](https://github.com/kubernetes/contrib/blob/404e0f4ae459994b8a78bf75faaa1fc5d546eb39/mungegithub/mungers/matchers/comment/comment.go#L74)):*
> ```Go
> 
> // AuthorLogins matches comment made by these Authors
> type AuthorLogins []string
> ```

As discussed somewhere else, could be simplified.

---


*Comments from [Reviewable](https://reviewable.kubernetes.io:443/reviews/kubernetes/contrib/1638)*

grodrigues3:
why did you close this?

PR #1644, opened 2016-08-26
added documentation for commands that the bot accepts

grodrigues3:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1644)

### DIFF
--- a/mungegithub/README.md
+++ b/mungegithub/README.md
@@ -106,3 +106,12 @@ make secret APP=submit-queue TARGET=<TARGET>
 * Create the service which is of type NodePort.
 * Finally, update the ingress.yaml with the new URL and the new service to point to.
 * Apply changes to the running ingress instance.
+
+## Communicating with the Bot
+
+Github contributors and reviewers can communicate with the mungebot by commenting on a PR with the following commands entered alone in the text field. All commands require the following syntax: /<COMMAND> [OPTIONAL ARGS].  Note the forward slashing preceding the command name.
+
+### List of Commands
+
+1. **/lgtm** : applies the lgtm label
+2. **/lgtm cancel** : removes a previously applied lgtm label
--- a/mungegithub/submit-queue/www/index.html
+++ b/mungegithub/submit-queue/www/index.html
@@ -248,6 +248,21 @@
                     <span><img ng-src="{{cntl.metadata.ChartUrl}}" style="max-width: 100%"/></span>
                   </md-content>
                 </md-tab>
+
+                <md-tab label="Bot Commands">
+                    <h1>Communicating with the Bot</h1>
+
+                Github contributors and reviewers can communicate with the mungebot by commenting on a PR with the
+                  following commands entered alone in the text field. All commands require the following syntax:
+                  <br><br> /Command [Optional Args]. <br><br> Note the forward slashing preceding the command name.
+                    <h2>List of Commands</h2>
+
+                    <ol>
+                      <li>lgtm: applies the lgtm label </li>
+                      <li> lgtm cancel: removes a previously applied lgtm label </li>
+                    </ol>
+
+                </md-tab>
               </md-tabs>
             </md-content>
           </md-tab>


mungegithub/README.md
```diff
@@ -105,4 +105,11 @@ make secret APP=submit-queue TARGET=<TARGET>
...
-* Apply changes to the running ingress instance.
\ No newline at end of file
+* Apply changes to the running ingress instance.
+
+### Commands for Communicating with the Bot
+
+Github contributors and reviewers can communicate with the mungebot by commenting on a PR with the following commands entered alone in the text field. Note the forward slash preceding the text.
+
+1. **/lgtm** : applies the lgtm label
+2. **/lgtm-cancel** : removes a previously applied lgtm label
```
foxish: It is `/lgtm cancel` without the hyphen.

eparis:
and chance you'd be willing to put this information instead in the submit-queue web interface? if we add/remove things here people runing the bot in the wild will have no idea what their version supports...

grodrigues3:
I don't really have a strong opinion about where the documentation lives.  I just want it to be easily accessible and only in one place (so we only have to update it one place).   @foxish, what do you think?  also, @eparis I think the commands are going to be more than just submit-queue related (e.g /ack - i accept this review and will get to it /nack - reassign to someone else) if that changes anything.

grodrigues3:
we might also, try documenting docs/devel in the main repo:

https://github.com/kubernetes/kubernetes/blob/master/docs/devel/automation.md

eparis:
Honestly I think that https://github.com/kubernetes/kubernetes/blob/master/docs/devel/automation.md should largely be killed in favor of pointing people to the web page...

foxish:
I agree with @eparis that these commands within the submit-queue webpage help text makes sense, as folks are more likely to see that, rather than the README here. Quite a few users don't know where the bot's source lives.

Also, we might also want to include that we can add comments to the command, like so:
```
/lgtm //some-comment-here
/lgtm cancel //some-comment-here
```

grodrigues3:
Sounds good.  Will update later today.

grodrigues3:
@eparis @foxish I added a bot commands tab to the info page.  I know nothing about angular, so if there's a cleaner way then just editing raw (possibly outdated) html, let me know.

mungegithub/submit-queue/www/index.html
```diff
@@ -248,6 +248,21 @@ <h2 class="md-toolbar-tools">Health chart for the past week</h2>
...
+                  following commands entered alone in the text field. All commands require the following syntax:
+                  <br><br> /Command [Optional Args]. <br><br> Note the forward slashing preceding the command name.
+                    <h2>List of Commands</h2>
+
+                    <ol>
+                      <li>lgtm: applies the lgtm label </li>
+                      <li> lgtm cancel: removes a previously applied lgtm label </li>
+                    </ol>
+
+                </md-tab>
```
foxish: I have seen that some parts of the submit-queue help text is populated from within the submit-queue munger. I'm not sure what the advantages are, but if we kept the help text in one place, that might help.

foxish: On second thought, making it a separate tab page this way also makes sense.

eparis: I know that I did `queue order` and `merge requirements` from inside submit-queue.go because it was the single place the information made sense.

If we expect all (or most) of the commands to be implement/used in submit-queue.go it might make sense to do that. It would likely be best to have all of the (running) mungers register commands which would be outputted.

But I'll merge this patch if others are happy and think we can iterate rather than do it up front...

eparis: Agreed that the tab like this makes sense in the html, only question is if it should be implemented like the `merge requirements` or `queue order` tab.

foxish: Most commands are expected to be implemented outside of submit-queue.go, so perhaps this is as good a place as any. We can iterate later; SGTM.

foxish:
LGTM

PR #1645, opened 2016-08-27
Consistent bool returns on exposed methods for err handling

eparis:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1645)

### DIFF
--- a/mungegithub/github/github.go
+++ b/mungegithub/github/github.go
@@ -624,26 +624,27 @@ func (config *Config) SetBranchProtection(name string, contexts []string) error 
 
 // Refresh will refresh the Issue (and PR if this is a PR)
 // (not the commits or events)
-func (obj *MungeObject) Refresh() error {
+func (obj *MungeObject) Refresh() bool {
 	num := *obj.Issue.Number
 	issue, err := obj.config.getIssue(num)
 	if err != nil {
-		return err
+		glog.Errorf("Error in Refresh: %v", err)
+		return false
 	}
 	obj.Issue = issue
 	if !obj.IsPR() {
-		return nil
+		return true
 	}
 	pr, err := obj.config.getPR(*obj.Issue.Number)
 	if err != nil {
-		return err
+		return false
 	}
 	obj.pr = pr
-	return nil
+	return true
 }
 
 // ListMilestones will return all milestones of the given `state`
-func (config *Config) ListMilestones(state string) []*github.Milestone {
+func (config *Config) ListMilestones(state string) ([]*github.Milestone, bool) {
 	listopts := github.MilestoneListOptions{
 		State: state,
 	}
@@ -652,7 +653,7 @@ func (config *Config) ListMilestones(state string) []*github.Milestone {
 	if err != nil {
 		glog.Errorf("Error getting milestones of state %q: %v", state, err)
 	}
-	return milestones
+	return milestones, true
 }
 
 // GetObject will return an object (with only the issue filled in)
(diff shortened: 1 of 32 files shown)


lavalamp:
I recall debating this in another PR, did you convince past me that this was better than returning errors?

eparis:
'its how everything else is' was my argument and 1/2 1/2 is worse than all one or all the other.  Errorf() at the bottom your munger code just bails.

We can switch everything to errors in another go if we have some consistency. (and this fixes real bugs).

lavalamp:
OK, sure. @foxish can you do a careful review? I only skimmed :)

foxish:
@lavalamp ack. I'll get to it later today.

mungegithub/github/github.go
```diff
@@ -558,35 +559,39 @@ func (config *Config) NewIssue(title, body string, labels []string, owner string
...
-	return ""
+	return "", ok
 }
 
 // IsForBranch return true if the object is a PR for a branch with the given
 // name. It return false if it is not a pr, it isn't against the given branch,
 // or we can't tell
-func (obj *MungeObject) IsForBranch(branch string) bool {
-	objBranch := obj.Branch()
+func (obj *MungeObject) IsForBranch(branch string) (bool, bool) {
```
foxish: I feel that to return (bool, bool) is confusing, in that one needs to check the result of the second return statement to know if the first is meaningful.

foxish:
Review status: 0 of 30 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[mungegithub/github/github.go, line 511 \[r1\]](https://reviewable.kubernetes.io:443/reviews/kubernetes/contrib/1645#-KQT0tYG4L4EREz4SnOV:-KQT0tYG4L4EREz4SnOW:b-alza3n) ([raw file](https://github.com/kubernetes/contrib/blob/232f7654dc5a2fab0851e21235aad0775af14e48/mungegithub/github/github.go#L511)):*
> ```Go
> 	config.analytics.ListMilestones.Call(config, resp)
> 	if err != nil {
> 		glog.Errorf("Error getting milestones of state %q: %v", state, err)
> ```

Return (nil, false) here?

---

*[mungegithub/github/github.go, line 872 \[r1\]](https://reviewable.kubernetes.io:443/reviews/kubernetes/contrib/1645#-KQT1jbmEk9KTsjGJZnF:-KQT1jbmEk9KTsjGJZnG:bkkshzb) ([raw file](https://github.com/kubernetes/contrib/blob/232f7654dc5a2fab0851e21235aad0775af14e48/mungegithub/github/github.go#L872)):*
> ```Go
> 		return "", false
> 	}
> 	if !releaseMilestoneRE.MatchString(*title) {
> ```

We return true because this isn't an error condition?

---


*Comments from [Reviewable](https://reviewable.kubernetes.io:443/reviews/kubernetes/contrib/1645)*

foxish:
@eparis I made one pass over it. I am a little worried about the fact that it will no longer be possible for any munger to see the error and act differently based on what kind of error it is, because we will always return a boolean for "something went wrong".

grodrigues3:

- **mungegithub**

cc @bprashanth 

eparis:
After about 5 large rebases I'm thinking about just self-LGTM. We know it fixes bugs. If we want to convert to an error that's fine, but no doubt this is 100% better than it is today.

Also I intentionally used my personal e-mail on the commit to make sure the merge robot can handle cla: human-approved after we turned on protected branches.

foxish:
@eparis Let's go ahead and merge this :)

PR #1655, opened 2016-08-29
matchers: Fix matching on multi-lines comments

apelisse:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1655)

### DIFF
--- a/mungegithub/mungers/matchers/comment/command_test.go
+++ b/mungegithub/mungers/matchers/comment/command_test.go
@@ -37,12 +37,24 @@ func TestParseCommand(t *testing.T) {
 			comment:         " /COMMAND Line can't start with spaces",
 		},
 		{
+			expectedCommand: nil,
+			comment:         "Command not at the beginning:\n/COMMAND\nAnd something else...",
+		},
+		{
 			expectedCommand: &Command{Name: "COMMAND"},
 			comment:         "/COMMAND",
 		},
 		{
 			expectedCommand: &Command{Name: "COMMAND", Arguments: "Valid command"},
 			comment:         "/COMMAND Valid command",
+		},
+		{
+			expectedCommand: &Command{Name: "COMMAND", Arguments: "Multiple Lines"},
+			comment:         "/COMMAND Multiple Lines\nAnd something else...",
+		},
+		{
+			expectedCommand: &Command{Name: "COMMAND", Arguments: "Args"},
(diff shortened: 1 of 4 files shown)


foxish:
@apelisse  Could you explain what was wrong with the previous regexes and what the issue was? 
I'm trying to understand what happened, and it'll help review this :)

mungegithub/mungers/matchers/comment/notification.go
```diff
@@ -33,7 +33,7 @@ type Notification struct {
 
 var (
 	// Matches a notification: [NOTIFNAME] Arguments
-	notificationRegex = regexp.MustCompile(`^\[([^\]\s]+)\]\s?(.*)$`)
+	notificationRegex = regexp.MustCompile(`(?m:^\[([^\]\s]+)\] *?(.*)$)`)
```
foxish:
```
(?m:^\[([^\]\s]+)\] *?(.*)$)
-------^ do we need this capturing group?
```

foxish: Also why do we explicitly include `\s`? Wouldn't `[^\]]` include spaces anyway?

mungegithub/mungers/matchers/comment/command_test.go
```diff
@@ -45,6 +45,18 @@ func TestParseCommand(t *testing.T) {
 			comment:         "/COMMAND Valid command",
 		},
 		{
+			expectedCommand: &Command{Name: "COMMAND", Arguments: "Multiple Lines"},
```
foxish: Do we want to support this syntax of the command not being at the beginning of a comment?

mungegithub/mungers/matchers/comment/notification.go
apelisse: - Do we need to capture this group: Yes, it's the name of the notification
- Why do we include \s twice: No, we are actually excluding \s (confusing with all those squared brackets ...)

mungegithub/mungers/matchers/comment/command_test.go
foxish: @grodrigues3 WDYT?

grodrigues3: I'd like commands to be at the beginning of comments.  Easier for humans to parse and figure out where bot commands came from.

foxish:
LGTM

grodrigues3:
LGTM

PR #1658, opened 2016-08-30
AppArmor loader DaemonSet

timstclair:
This PR adds a proof-of-concept DaemonSet for loading AppArmor profiles from a ConfigMap onto a cluster.

The DaemonSet runs a small go program that is mostly a wrapper around `apparmor_parser`, which actually loads the profiles. The additional features the loader provides are:
- Polling
- Sourcing from multiple directories
- Better handling of symlinks (necessary for ConfigMap volumes)
- Better logging and messaging

This is meant to provide an example of using AppArmor, and be a starting point for users trying out AppArmor or building there own solutions.

/cc @jfrazelle @thockin

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1658)

### DIFF
--- a/apparmor/loader/example-configmap.yaml
+++ b/apparmor/loader/example-configmap.yaml
@@ -1,0 +1,76 @@
+# An example ConfigMap demonstrating how profiles can be stored as Kubernetes objects, and loaded by
+# the apparmor-loader DaemonSet.
+
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: apparmor-profiles
+  namespace: apparmor
+data:
+  # Filename k8s-nginx maps to the definition of the nginx profile.
+  k8s-nginx: |-
+    #include <tunables/global>
+
+    # From https://github.com/jfrazelle/bane/blob/master/docker-nginx-sample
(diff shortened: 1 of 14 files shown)


apparmor/loader/example-configmap.yaml
```diff
@@ -0,0 +1,76 @@
+# An example ConfigMap demonstrating how profiles can be stored as Kubernetes objects, and loaded by
+# the DaemonSet.
```
dchen1107: s/DaemonSet/AppArmor Profile Loader DaemonSet?

timstclair: Done. I went with "apparmor-loader DaemonSet" since that's the actual object name.

dchen1107:
LGTM to me overall. But not sure why travis complained the following:

```
This job is running on container-based infrastructure, which does not allow use of 'sudo', setuid and setguid executables.
If you require sudo, add 'sudo: required' to your .travis.yml
```

timstclair:
I think the failure was just a flake. It shows the sudo message for every build.

timstclair:
Squashed commits.

dchen1107:
LGTM

PR #1661, opened 2016-08-30
Add kubemark-scale to non-blocking SQ suites

### LINKED ISSUE #31105: [k8s.io] Density [Feature:Performance] should allow starting 30 pods per node {Kubernetes e2e suite}
https://k8s-gubernator.appspot.com/build/kubernetes-jenkins/logs/kubernetes-e2e-gce-scalability/10290/

Failed: [k8s.io] Density [Feature:Performance] should allow starting 30 pods per node {Kubernetes e2e suite}

```
/go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/test/e2e/framework/framework.go:129
Expected
    <[]string | len:1, cap:1>: [
        "Container kube-apiserver-e2e-scalability-master/kube-apiserver is using 2000.9453125/2000 MB of memory",
    ]
to be empty
/go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/test/e2e/framework/resource_usage_gatherer.go:376
```

Previous issues for this test: #26185 #27142 #30275 #31101

wojtek-t:
Fix kubernetes/kubernetes#31105

@gmarek

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1661)

### DIFF
--- a/mungegithub/submit-queue/deployment/kubernetes/configmap.yaml
+++ b/mungegithub/submit-queue/deployment/kubernetes/configmap.yaml
@@ -16,7 +16,7 @@ data:
 
   # submit-queue options.
   submit-queue.required-contexts: "Jenkins GCE Node e2e"
-  submit-queue.nonblocking-jenkins-jobs: kubernetes-cross-build,kubernetes-e2e-gke-staging,kubernetes-e2e-gke-staging-parallel,kubernetes-e2e-gce-serial,kubernetes-e2e-gke-serial,kubernetes-e2e-gke-test,kubernetes-e2e-gce-examples,kubernetes-e2e-gce-federation,kubernetes-e2e-gce-scalability,kubernetes-soak-continuous-e2e-gce,kubernetes-soak-continuous-e2e-gke,kubernetes-kubemark-5-gce,kubernetes-kubemark-500-gce,kubelet-serial-gce-e2e-ci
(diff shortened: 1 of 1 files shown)


lavalamp:
@wojtek-t why "do not merge"? it lgtm if the test is ready to go.

apelisse:
It's syntactically correct, can't really validate if it's something we want to do :-)

lavalamp:
It is something we want to do, yes, as soon as @wojtek-t removes the
do-not-merge label.

gmarek:
Our (well - mostly mine) concern is that we said that we'll block the release on flakes and IIUC adding the suite to non-blocking set will generate flake issues for failures. I just don't want to make the release process any harder than it needs to be.

wojtek-t:
My motivation for this PR is so that we have issues automatically filed for kubemark-scale. However, we seem to have some issues with kubemark-scale now (they seem to be related to overloaded Jenkins, but we are still trying to fix them), so it may not be the best option to have it now.

Personally I would wait for the first one to happen:
- we fix kubemark-scale (from constantly failing)
- we launch the release 1.4
If any of those happen, we should merge it.

If you don't agree with the above - feel free to remove "do-not-merge" label.

wojtek-t:
Actually, now I really think that we fixed kubemark-scale. So its failures actually should be debugged and i think we are ready to merge it.
Removing "do-not-merge" label I added.

PR #1670, opened 2016-08-31
Cluster-autoscaler: pick one of the matching node groups at random

mwielgus:
Currently the first matching node group is picked for scale up. CA should select it at random so that all node groups grow in a similar way.

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1670)

### DIFF
--- a/cluster-autoscaler/utils.go
+++ b/cluster-autoscaler/utils.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"fmt"
+	"math/rand"
 	"reflect"
 	"time"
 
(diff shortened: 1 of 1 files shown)


piosz:
LGTM

Do we have a plan to choose node group in a more sophisticated way?

piosz:
```
# k8s.io/contrib/cluster-autoscaler
./utils.go:185: cannot use len(expansionOptions) (type int) as type int32 in argument to "math/rand".Int31n
```

mwielgus:
Fixed.

mwielgus:
k8s-merge-boot seems broken. Merging manually

PR #1676, opened 2016-09-01
Fix fetch comments

apelisse:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1676)

### DIFF
--- a/mungegithub/github/github.go
+++ b/mungegithub/github/github.go
@@ -1558,17 +1558,34 @@ func (obj *MungeObject) ListReviewComments() ([]*github.PullRequestComment, erro
 
 	config := obj.config
 	page := 1
+	// Try to work around not finding comments--suspect some cache invalidation bug when the number of pages changes.
+	tryNextPageAnyway := false
 	for {
 		listOpts.ListOptions.Page = page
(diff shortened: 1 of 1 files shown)


mungegithub/github/github.go
```diff
@@ -1589,23 +1606,40 @@ func (obj *MungeObject) ListComments(withListOpts ...WithListOpt) ([]*github.Iss
...
+				// Cached last page was actually truthful -- expected error.
+				break
+			}
 			return nil, err
 		}
+		if tryNextPageAnyway {
+			if len(comments) == 0 {
+				break
+			}
+			glog.Infof("For %v: supposedly there weren't more comments, but we asked anyway and found %v more.", issueNum, len(comments))
```
eparis: can you distinguish these glog's so we know if it it only 1 or both of the APIs that are broken?

apelisse: True

eparis:
@lavalamp it lgtm but you also helped run down this bug in the Events API.

apelisse:
Fixed the nit,

lavalamp:
/lgtm

but maybe you can scan the logs and make sure this is actually necessary.

PR #1684, opened 2016-09-02
Fix #1678 by speeding up IsMergeable retries when testing.

### LINKED ISSUE #1678: mungegithub/github: Tests are very slow
Tests are much slower since 06d9dd190. One of the test checks what happens when mergeability flag isn't there. The new code is blocking for up to 62 seconds (2 + 4 + 8 + 16 + 32), this may be a problem on different levels.

Test time basically went from 5 seconds to 65 seconds.

cc @foxish

rmmh:
before:
ok      k8s.io/contrib/mungegithub/mungers      65.728s
after:
ok      k8s.io/contrib/mungegithub/mungers      3.886s

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1684)

### DIFF
--- a/mungegithub/github/github.go
+++ b/mungegithub/github/github.go
@@ -163,8 +163,8 @@ type Config struct {
 	// If true, don't make any mutating API calls
 	DryRun bool
 
-	// Defaults to 30 seconds.
-	PendingWaitTime *time.Duration
+	// Base sleep time for retry loops. Defaults to 1 second.
+	BaseWaitTime time.Duration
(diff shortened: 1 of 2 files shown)


apelisse:
@eparis it should, but you must be assignee. And it was supposed to tell you why it doesn't work when you're not assignee...

apelisse:
/lgtm 

I certainly like the result! Thanks for taking care of that promptly

PR #1689, opened 2016-09-05
Add playbook to install 1-node cluster from within running host machine

### LINKED ISSUE #1325: systemd defaults: remove ServiceAccount from admissions

ingvagabund:
The playbook is useful for starting users of kubernetes. It just runs through etcd, docker and kubernetes roles. So user can run&play without need to generate certificates or tokens.

It skips configuration of etcd and docker in case a user has already any of them installed.
On the other hand, kubernetes configuration is always carried as tokens/certs require config files update.

**TODO**
- [x] refactor `docker` role (at least introduce `configure` tag) https://github.com/kubernetes/contrib/pull/1690

Fixes: #1325

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1689)

### DIFF
--- a/ansible/scripts/deploy-local-cluster.sh
+++ b/ansible/scripts/deploy-local-cluster.sh
@@ -1,0 +1,37 @@
+#!/bin/bash
+
+# Copyright 2016 The Kubernetes Authors All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+. ./init.sh
+
+inventory=${INVENTORY_DIR}/localhost.ini
+
+# use localhost inventory
+# run etcd playbook
+# run docker playbook
+# run kubernetes-master playbook
+# run add-ons playbook
+# run kubernetes-node playbook
+
+# kubernetes roles takes care of token and certs generating
+
+# skipping configure tasks as we don't want to override default configuration
+# of etcd and docker.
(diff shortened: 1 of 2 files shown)


ingvagabund:
Tested locally, both etcd and docker configuration tasks are skipped.

ingvagabund:
I want more people to take a look at it. Though etcd and docker could be skipped as kubernetes is installed via package manager, one can decide to install them from different sources. That will need some modification of the script and make the ``--skip-tags="configure"`` optional.

ingvagabund:
@mdshuai @aveshagarwal @sttts @sjenning this is more like to discuss how much it make sense to provide the ``deploy-local-cluster.sh`` and if it is senseful enough what should be inside.

ingvagabund:
I have just used it to deploy scratch build of Kubernetes in RHEL before I do the raw build. I guess this can go through. It's a new script, new playbook, nothing affected.

ansible/scripts/deploy-local-cluster.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+
+# Copyright 2016 The Kubernetes Authors All rights reserved.
```
sttts: nit: the "All rights reserved." is gone in the default kube boilerplate.

sttts:
One nit. Otherwise lgtm

ansible/scripts/deploy-local-cluster.sh
ingvagabund: Since when? Running the ./hack/verify-all.sh with "All rights reserved." ends with en error.

ingvagabund: I got feeling that is valid on kubernetes/kubernetes repository, not in the contrib one.

ingvagabund: Unless https://github.com/kubernetes/contrib/issues/1707 gets resolved,  "All rights reserved." has to be there.

Issue #1674, opened 2016-09-01
[ansible] uninstall playbook doesn't work


mkumatag:
Modified the inventory file and installed the cluster. Later tried uninstalling using `adhoc/uninstall.yml ` playbook but failed with below error:

```
[root@jupiter-vm536 ansible]# pwd
/root/contrib/ansible
[root@jupiter-vm536 ansible]# ansible-playbook -i inventory/inventory playbooks/adhoc/uninstall.yml 

```
fails with following error:

```
TASK [Remove remaining files] **************************************************
[DEPRECATION WARNING]: Skipping task due to undefined Error, in the future this will be a fatal error.: 'ansible_ssh_user' is undefined.
This feature will be removed 
in a future release. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
[DEPRECATION WARNING]: Skipping task due to undefined Error, in the future this will be a fatal error.: 'ansible_ssh_user' is undefined.
This feature will be removed 
in a future release. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
fatal: [jupiter-vm1055.pok.stglabs.ibm.com]: FAILED! => {"failed": true, "msg": "the field 'args' has an invalid value, which appears to include a variable that is undefined. The error was: 'item' is undefined\n\nThe error appears to have been in '/root/contrib/ansible/playbooks/adhoc/uninstall.yml': line 114, column 7, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n    - name: Remove remaining files\n      ^ here\n"}
fatal: [jupiter-vm1194.pok.stglabs.ibm.com]: FAILED! => {"failed": true, "msg": "the field 'args' has an invalid value, which appears to include a variable that is undefined. The error was: 'item' is undefined\n\nThe error appears to have been in '/root/contrib/ansible/playbooks/adhoc/uninstall.yml': line 114, column 7, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n    - name: Remove remaining files\n      ^ here\n"}

NO MORE HOSTS LEFT *************************************************************
	to retry, use: --limit @playbooks/adhoc/uninstall.retry

PLAY RECAP *********************************************************************
jupiter-vm1055.pok.stglabs.ibm.com : ok=15   changed=2    unreachable=0    failed=1   
[...]

PR #1694, opened 2016-09-06
Fix uninstall playbook

mkumatag:
Fixes - https://github.com/kubernetes/contrib/issues/1674

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1694)

### DIFF
--- a/ansible/playbooks/adhoc/uninstall.yml
+++ b/ansible/playbooks/adhoc/uninstall.yml
@@ -114,7 +114,7 @@
     - name: Remove remaining files
       file: path={{ item }} state=absent
       with_items:
-        - "~{{ ansible_ssh_user }}/.kube"
+        - "{{ ansible_env.HOME }}/.kube"
         - /root/.kube
         - "{{ bin_dir }}/kube-apiserver"
         - "{{ bin_dir }}/kube-controller-manager"


mkumatag:
@ingvagabund

ansible/playbooks/adhoc/uninstall.yml
```diff
@@ -114,7 +114,7 @@
     - name: Remove remaining files
       file: path={{ item }} state=absent
       with_items:
-        - "~{{ ansible_ssh_user }}/.kube"
+        - "{{ ansible_env.HOME }}/.kube"
```
ingvagabund: What is the difference between ``~/.kube`` and ``{{ ansible_env.HOME }}/.kube`` then?

mkumatag: I feel both are same. :)

ingvagabund: Got the same feeling :). Reading through stackoverflow articles and ansible docs and can't find anything particular and usable for the ``ansible_env.HOME`` case. So far have not found anything saying it is incorrect usage. With the ``ansible_env.HOME`` it is explicitly mentioned I want remote host directory which corresponds to a user under which the ansible is ssh'ed.

ingvagabund:
Looks good, thanks.

PR #1714, opened 2016-09-07
Open up submit queue

pwittrock:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1714)


mungegithub/submit-queue/deployment/kubernetes/configmap.yaml
```diff
@@ -20,7 +20,7 @@ data:
   submit-queue.jenkins-jobs: kubelet-gce-e2e-ci,kubernetes-build,kubernetes-test-go,kubernetes-verify-master,kubernetes-e2e-gce,kubernetes-e2e-gce-slow,kubernetes-e2e-gke,kubernetes-e2e-gke-slow
   submit-queue.presubmit-jobs: kubernetes-pull-build-test-e2e-gce,kubernetes-pull-build-test-e2e-gke,kubernetes-pull-build-test-federation-e2e-gce,node-pull-build-e2e-test
   submit-queue.weak-stable-jobs: "\"\""
-  submit-queue.do-not-merge-milestones: "next-candidate,v1.5,"
+  submit-queue.do-not-merge-milestones: ""
```
lavalamp: @foxish do we need to copy the `"\"\""` from the line above if we delete everything?

pwittrock: Good catch

foxish: Yes, it is a string slice, so we will need the `"\"\""` because of the way the cmd-line parsing library.

PR #1727, opened 2016-09-10
implement approval handler

hongchaodeng:
ref: https://github.com/kubernetes/contrib/issues/1389

What?
This PR adds ApprovalHandler that will try to add "approved" label once all files of change has been approved by approvers.

```
// The algorithm goes as:
// - Initially, we set up approverMap
//   - Go through all comments after latest commit. If any approver said "/approve", add him to approverMap.
// - For each file, we see if any approver of this file is in approverMap.
//   - An approver of a file is defined as:
//     - It's known that each dir has a list of approvers. (This might not hold true. For usability, current situation is enough.)
//     - Approver of a dir is also the approver of child dirs.
//   - We look at only top 3 dir level's approvers .
// - Iff all files has been approved, the bot will add "approved" label.
```

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1727)

### DIFF
--- a/mungegithub/mungers/approval-handler.go
+++ b/mungegithub/mungers/approval-handler.go
@@ -20,11 +20,13 @@ import (
 	"path"
 	"strings"
 
-	"github.com/golang/glog"
-	"github.com/spf13/cobra"
 	"k8s.io/contrib/mungegithub/features"
 	"k8s.io/contrib/mungegithub/github"
 	"k8s.io/contrib/mungegithub/mungers/mungerutil"
+	"k8s.io/kubernetes/pkg/util/sets"
+
+	"github.com/golang/glog"
+	"github.com/spf13/cobra"
 )
 
 // ApprovalHandler will try to add "approved" label once
@@ -84,7 +86,7 @@ func (h *ApprovalHandler) Munge(obj *github.MungeObject) {
 		return
 	}
 
-	approverMap := map[string]struct{}{}
+	approverMap := sets.String{}
 
 	// from oldest to latest
 	for i := len(comments) - 1; i >= 0; i-- {
@@ -97,12 +99,12 @@ func (h *ApprovalHandler) Munge(obj *github.MungeObject) {
 		fields := strings.Fields(strings.TrimSpace(*c.Body))
 
 		if len(fields) == 1 && strings.ToLower(fields[0]) == "/approve" {
-			approverMap[*c.User.Login] = struct{}{}
+			approverMap.Insert(*c.User.Login)
 			continue
 		}
 
 		if len(fields) == 2 && strings.ToLower(fields[0]) == "/approve" && strings.ToLower(fields[1]) == "cancel" {
-			delete(approverMap, *c.User.Login)
+			approverMap.Delete(*c.User.Login)
 		}
 	}
 
@@ -114,7 +116,7 @@ func (h *ApprovalHandler) Munge(obj *github.MungeObject) {
 	obj.AddLabel(approvedLabel)
 }
 
-func (h *ApprovalHandler) hasApproval(filename string, approverMap map[string]struct{}) bool {
+func (h *ApprovalHandler) hasApproval(filename string, approverMap sets.String) bool {
 	paths := strings.Split(filename, "/")
 	var p string
 	for i := 0; i < len(paths) && i < 3; i++ {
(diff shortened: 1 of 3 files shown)


hongchaodeng:
cc @bgrant0607 @philips @xiang90

mungegithub/mungers/approval-handler.go
```diff
@@ -0,0 +1,144 @@
...
+limitations under the License.
+*/
+
+package mungers
+
+import (
+	"path"
+	"strings"
+
+	"github.com/golang/glog"
```
foxish: The imports need to be reordered.

mungegithub/mungers/approval-handler.go
```diff
@@ -0,0 +1,144 @@
...
+		return
+	}
+
+	comments, err := getCommentsAfterLastModified(obj)
+	if err != nil {
+		glog.Errorf("failed to get comments in this PR: %v", err)
+		return
+	}
+
+	approverMap := map[string]struct{}{}
```
foxish: If the intent is to use it like a set, we could use sets.String (`k8s.io/kubernetes/pkg/util/sets`) instead?

mungegithub/mungers/approval-handler.go
```diff
@@ -0,0 +1,146 @@
...
+			return
+		}
+	}
+	obj.AddLabel(approvedLabel)
+}
+
+func (h *ApprovalHandler) hasApproval(filename string, approverMap sets.String) bool {
+	paths := strings.Split(filename, "/")
+	var p string
+	for i := 0; i < len(paths) && i < 3; i++ {
```
grodrigues3: can we make 3 a parameter so it's clear that it represents the maxDepth of the search?

mungegithub/mungers/approval-handler.go
```diff
@@ -0,0 +1,146 @@
...
+		return
+	}
+
+	comments, err := getCommentsAfterLastModified(obj)
+	if err != nil {
+		glog.Errorf("failed to get comments in this PR: %v", err)
+		return
+	}
+
+	approverMap := sets.String{}
```
grodrigues3: nit: change the variable name from approverMap to approverSet if it's a set.

mungegithub/mungers/approval-handler.go
```diff
@@ -0,0 +1,144 @@
...
+	if !obj.IsPR() {
+		return
+	}
+	files, err := obj.ListFiles()
+	if err != nil {
+		glog.Errorf("failed to list files in this PR: %v", err)
+		return
+	}
+
+	comments, err := getCommentsAfterLastModified(obj)
```
grodrigues3: @apelisse in the event that file A is modified, only file A should require re-approval right?  I think this implementation will require all approvers write ```/approve``` after every new commit, if if the file or files they previously approved have not changed at all.  is that right?

hongchaodeng: That's a good point. I think this is one of the optimization that we can do in the future.
I want to keep make this PR simple and usable due to lack of testing. Once it's working, we can incrementally add optimizations.

hongchaodeng:
All comments addressed.
PTAL.

apelisse:
/lgtm

---

Reviewed 2 of 3 files at r1, 1 of 1 files at r4.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[mungegithub/mungers/approval-handler.go, line 86 \[r3\]](https://reviewable.kubernetes.io:443/reviews/kubernetes/contrib/1727#-KRH799tXp1kBxHfpYSt-r3-86:-KRdnz8xw20YraeJuVEA:b-4au5xj) ([raw file](https://github.com/kubernetes/contrib/blob/f43f3b8fc5565415e777960b3d5b6cf1fc0010d1/mungegithub/mungers/approval-handler.go#L86)):*
<details><summary><i>Previously, hongchaodeng (Hongchao Deng) wrote…</i></summary>
> That's a good point. I think this is one of the optimization that we can do in the future.
> 
> I want to keep make this PR simple and usable due to lack of testing. Once it's working, we can incrementally add optimizations.
</details>
Yeah, that's a good point. Pretty useful optimization, but let's go for incremental :-)

---


*Comments from [Reviewable](https://reviewable.kubernetes.io:443/reviews/kubernetes/contrib/1727#-:-KRdpa4XjSPeghNRbcJO:b-8ks5qa)*

PR #1739, opened 2016-09-13
Support multiple commands on multiple URLs in exechealthz

### LINKED ISSUE #1735: Check multiple commands on multiple URLs in Exechealthz
Refer kubernetes/kubernetes#30633.

Would it be possible to support checking multiple commands on multiple URLs in one Exechealthz process? I am going to run two Exechealthz processes within one pod but consider it would be better if Exechealthz supports this intuitively.

@luxas

MrHohn:
Fixes #1735.

Adding support for multiple commands on URLs for [exec-healthz](https://github.com/kubernetes/contrib/tree/master/exec-healthz).

Example as below:

```
$ ./exechealthz --cmd="ls /tmp/test1" --url="/healthz1" --cmd="ls /tmp/test2" --url="/healthz2"
```

The --url flag indicates the path healthz server needs to serve on.
Notes: Number of cmds and URLs have to be the same (if more than 1). URL need to start with "/". URLs and cmds match up based on their orders (first URL to first cmd).

The previous behavior remains the same (as the previous examples in README).

@thockin @bprashanth @luxas

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1739)


thockin:
I don't know this sub-repo - any opportunity for tests?

MrHohn:
Seems like the only way is to add unit tests under the same folder.

Is it Okay to run the main function in a go routine while using another go routine to verify?

thockin:
no really.  I would try to test this by injecting a pkg/util/exec.Interface
which is a fake, and verifying that it gets called.  I don't want to derail
tooo much - it's all about confidence in the code.  You should not be stuck
doing tests for everything you touch, but if you CAN that is great.

bprashanth:
Tests sound like a good idea. I didn't write them initially because I thought it was borderline complexity, but adding more probably pushes it over the edge. 

You can give the `execworker` an interface to exec and a fakeClock, then trigger the clock and confirm the exec hits. Both those packages are available in main kuberepo (https://github.com/kubernetes/kubernetes/tree/master/pkg/util), but require godeps, which this project has thus far steered clear from. I leave it to you, maybe it's easier to add a godep, or just define your own fake.

MrHohn:
@thockin @bprashanth Thanks for the detailed instructions. I tried `exec` and `clock` packages and added tests mainly for the execWorker.

Not sure to also test the web server or not. I consider holding a real server that occupy a port may introduce uncertainty and flaky result.

cc @matchstick

MrHohn:
Sorry, there is a mistake in exechealthz_test.go. I misunderstood the usage of wait.Poll(). Will fix it soon.

MrHohn:
Fixed.

test-foxish:
recomputing cla status...

exec-healthz/exechealthz.go
```diff
@@ -37,15 +39,47 @@ import (
 // 3. 404 for unknown endpoints
 
 var (
-	port       = flag.Int("port", 8080, "Port number to serve /healthz.")
-	cmd        = flag.String("cmd", "echo healthz", "Command to run in response to a GET on /healthz. If the given command exits with 0, /healthz will respond with a 200.")
+	cmdDefault = "echo healthz"
```
thockin: having a default cmd is not that useful.  I won't block this PR on it, but if anything else needs to be fixed, I would just remove this and demand at least one --cmd

exec-healthz/exechealthz.go
```diff
@@ -88,16 +126,16 @@ func logf(format string, args ...interface{}) {
 // start attemtps to run the probeCmd every `period` seconds.
 // Meant to be called as a goroutine.
 func (h *execWorker) start() {
```
thockin: rename `h` to `ew` or `worker`

thockin (approved):
LGTM.  This is a good PR.

thockin:
Now you need to tag this and push it.  I might also try to convince you to move this to github.com/thockin/go-build-template

PR #1693, opened 2016-09-06
Allow customisation of the nginx proxy_buffer_size directive via ConfigMap

vyshane:
When using nginx as a proxy we can run into the following error:

```
upstream sent too big header while reading response header from upstream
```

In order to fix this, we need to be able to configure the `proxy_buffer_size` nginx directive to increase its value. This PR updates the nginx-ingress-controller to allow that.

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1693)

PR #1743, opened 2016-09-15
Allow customisation of the nginx proxy_buffer_size directive via ConfigMap

vyshane:
I'm opening a new PR with the same changes as #1693 because I pushed the latter with an email address that can't be used to sign the CLA. Description from the previous PR:

When using nginx as a proxy we can run into the following error:

```
upstream sent too big header while reading response header from upstream
```

In order to fix this, we need to be able to configure the proxy_buffer_size nginx directive to increase its value. This PR updates the nginx-ingress-controller to allow that.

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1743)


vyshane:
@aledbf, this PR replaces #1693.

bprashanth:
probably won't get time to look at this before monday because of the 1.4 release, hopefully by then though. Manuel might have time to review.

aledbf:
LGTM

bprashanth:
good enough :)

PR #1749, opened 2016-09-16
[nginx-ingress-controller] Readiness probe that works behind a CP lb

### LINKED ISSUE #1507: Figure out readiness probe for nginx ingress that works behind a CP lb
Flesh out the details discussed in: https://github.com/kubernetes/contrib/pull/1501#discussion-diff-74324489 and update the probes accordingly.

aledbf:
fixes #1507

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1749)

### DIFF
--- a/ingress/controllers/nginx/nginx.tmpl
+++ b/ingress/controllers/nginx/nginx.tmpl
@@ -265,6 +265,12 @@ http {
         {{ end }}
 
         {{ if eq $server.Name "_" }}
+        # health checks in cloud providers require the use of port 80
+        location {{ $cfg.healthzUrl }} {
+            access_log off;
+            return 200;
+        }
+
         # this is required to avoid error if nginx is being monitored
         # with an external software (like sysdig)
         location /nginx_status {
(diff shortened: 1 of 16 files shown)


ingress/controllers/nginx/nginx.tmpl
```diff
@@ -265,6 +265,12 @@ http {
         {{ end }}
 
         {{ if eq $server.Name "_" }}
+        # health checks in cloud providers require the use of port 80
+        location /ingress-controller-healthz {
```
bprashanth: can you make this the default, and allow people to config it through yaml?

bprashanth (commented):
awesome! thanks just the one nit

ingress/controllers/nginx/nginx.tmpl
aledbf: Using a flag with `/ingress-controller-healthz` as default is ok?

bprashanth: any reason it shouldn't be a flag? if i have an ingress controller behind an ingress controller how would i modify it :)

bprashanth:
Thanks LGTM, modifying title to note have wip

bprashanth:
oh please add a todo to scrub incoming ingresses for the url passed through the arg (can come laters, or just file a bug so we can track it)

aledbf:
> file a bug so we can track it)

done #1754

test-foxish:
recomputing cla status...

PR #1751, opened 2016-09-19
[nginx-ingress-controller] Add the name of the upstream in the log

### LINKED ISSUE #1750: [Ingress] Service/Upstream name in the log
Is there a way to get the service (or upstream) name logged in the NGINX log_format ?

We are trying to export the NGINX logs into Elasticsearch, but we are only interested by the logs of particular applications. Right now, it's all mixed, there's no way to differentiate which pod/service is targeted. Sure, the targeted upstream address gets logged, but that does not help to process the log.

aledbf:
fixes #1750

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1751)


bprashanth:
Oh nice, it would be good to document the format of output logs so people can parse them and use in eg: feedback loops/stats collection etc. LGTM.

PR #1756, opened 2016-09-20
Redirect STOPSIGNAL for kube-dns graceful termination

### LINKED ISSUE #33050: kube-dns's graceful termination is not acting as expected.
After PR #31894, I did a bit experiments on kube-dns graceful termination and got some unintended result as below:
```
# scenario: scale down kube-dns from 4 to 1 with graceful termination (250ms period lookup with udp) 
2016/09/16 18:31:58 Resolution took 1.383041ms
2016/09/16 18:31:58 Resolution took 3.750672ms
2016/09/16 18:31:58 err: read udp 10.244.2.6:53645->10.0.0.10:53: read: connection refused
2016/09/16 18:31:59 Resolution took 927.836µs
2016/09/16 18:31:59 Resolution took 2.034077ms
2016/09/16 18:31:59 Resolution took 2.549052ms

# scenario: scale down kube-dns from 4 to 1 with graceful termination (250ms period lookup with tcp)
2016/09/19 18:28:59 Resolution took 672.151µs
2016/09/19 18:28:59 Resolution took 806.637µs
2016/09/19 18:28:59 Resolution took 585.271µs
2016/09/19 18:28:59 err: read tcp 10.244.0.12:37854->10.0.0.10:53: read: connection reset by peer
2016/09/19 18:29:00 Resolution took 639.391µs
2016/09/19 18:29:00 Resolution took 603.815µs
2016/09/19 18:29:00 Resolution took 576.174µs
```

I guess it is because the other container `dnsmasq` was terminated before `kubedns`. DNS query from clients failed because `dnsmasq` is the actual connection point.

After a bit discussion with @girishkalele and @freehan, two solutions for this are:
1. Create another program wrap up the dnsmasq and capture the SIGTERM signal.
2. Figure out a way to modify the sig mask of process without touching the original codes.

@thockin

MrHohn:
Fixes kubernetes/kubernetes#33050.

Current `dnsmasq` container exit immediately after docker sends STOPSIGNAL. In order to support graceful termination as kubernetes/kubernetes#31807 indicated, we need to delay this operation.

Similar to that we do in `kubedns` --- ignoring the SIGTERM and exit when SIGKILL from docker occurs, @thockin's solution is to redirect docker's STOPSIGNAL to SIGCONT. I did a couple rounds of test and it works as expected.

The tag should be bumped up when we need to get the graceful termination feature in.

@girishkalele @ArtfulCoder @luxas

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1756)

### DIFF
--- a/dnsmasq/Dockerfile
+++ b/dnsmasq/Dockerfile
@@ -16,4 +16,6 @@ FROM alpine:3.4
 MAINTAINER Girish Kalele <gkalele@google.com>
 RUN apk --no-cache add dnsmasq
 COPY dnsmasq.conf /etc/dnsmasq.conf
+# Replace SIGTERM with SIGCONT as stop signal to support graceful termination
+STOPSIGNAL SIGCONT
 ENTRYPOINT ["/usr/sbin/dnsmasq", "--keep-in-foreground"]
--- a/dnsmasq/Dockerfile.cross
+++ b/dnsmasq/Dockerfile.cross
@@ -19,5 +19,7 @@ CROSS_BUILD_COPY qemu-ARCH-static /usr/bin/
 MAINTAINER Girish Kalele <gkalele@google.com>
 COPY dnsmasq.conf /etc/dnsmasq.conf
 COPY dnsmasq /usr/sbin/dnsmasq
+# Replace SIGTERM with SIGCONT as stop signal to support graceful termination
+STOPSIGNAL SIGCONT
 RUN mkdir /var/run/
 ENTRYPOINT ["/usr/sbin/dnsmasq", "--keep-in-foreground"]


thockin:
Did you actually roll this out to a cluster and test it?  Sometimes I have
really bad ideas :)

MrHohn:
Yes I did. I think this solution works just well because we are using [docker stop](https://docs.docker.com/engine/reference/commandline/stop/) in pod deletion. As long as the we have no special usage with SIGCONT, it should be safe to replacing SIGTERM with SIGCONT for this image.

thockin:
I'm sort of disgusted with myself for thinking of it :)

dnsmasq/Dockerfile
```diff
@@ -16,4 +16,5 @@ FROM alpine:3.4
 MAINTAINER Girish Kalele <gkalele@google.com>
 RUN apk --no-cache add dnsmasq
 COPY dnsmasq.conf /etc/dnsmasq.conf
+STOPSIGNAL SIGCONT
```
thockin: Can you comment this in both files?  LGTM otherwise.

MrHohn: Sure thing. Done.

bprashanth: Hai guys, 
don't we have a feature for exactly this called pre-stop hook? I mean: 
```yaml
lifecycle:
  preStop:
    exec:
      command:
      - /bin/sh
      - -ec
      - |
        echo "sleeping 60s"
        sleep 60
```

The pre-stop duration is taken from deletionGrace, meaning if you set `terminationGrace` to 60s and `sleep 60s` in your pre-stop, you will receive a sigkill in 60s. Prestop should be called *before* you get the `docker stop`, ie before the SIGTERM. 

This should work cross container runtime, maybe I misunderstood what you're trying to do?

MrHohn: I think you are correct. By using this preStop hook, we don't need to separately configure kubedns and dnsmasq to support graceful termination...

PR #1771, opened 2016-09-22
Bump glbc to 0.8.0

### LINKED ISSUE #1776: gce ingress is too eager in converting probes to health checks
the gce ingress controller goes looking in `AllNamespaces` for pods matching selectors of a given service, to convert their readiness probes into http health checks. It should only adopt readiness probes from pods in the same namespace as the service.

bprashanth:
Don't think this hits the bar for 1.4.0, but hopefully it can make 1.4.1. The version bump is for the godep update that fixes an issue with the throttling workqueue (https://github.com/kubernetes/kubernetes/pull/31396). I should've done this sooner, dropped it.

Also fixes https://github.com/kubernetes/contrib/issues/1776 and https://github.com/kubernetes/contrib/issues/1783

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1771)


freehan:
LGTM

PR #1778, opened 2016-09-22
Adopt go-build-template and push images as 1.2

MrHohn:
The latest version(1.2) exechealthz docker images have been pushed to gcr.io. Listed as below:
- gcr.io/google_containers/exechealthz:1.2
- gcr.io/google_containers/exechealthz-amd64:1.2
- gcr.io/google_containers/exechealthz-arm:1.2
- gcr.io/google_containers/exechealthz-arm64:1.2
- gcr.io/google_containers/exechealthz-ppc64le:1.2

Previous Makefile is broken due to the new appear vendor folder. It is fixed now.

I tried [go-build-template](https://github.com/thockin/go-build-template) but found it not very friendly to set multiple BASEIMAGEs for different ARCHs. So I switch back.

@thockin @bprashanth

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1778)


exec-healthz/Makefile
```diff
@@ -53,11 +53,11 @@ endif
 	cd $(TEMP_DIR) && sed -i "s|BASEIMAGE|$(BASEIMAGE)|g" Dockerfile
 
 	# Compile the binary inside a container for reliable builds
-	docker run --rm -it -v $(TEMP_DIR):/build golang:$(GOLANG_VERSION) /bin/bash -c "make -C /build server ARCH=$(ARCH)"
+	docker run --rm -it -v $(TEMP_DIR):/build golang:$(GOLANG_VERSION) /bin/bash -c "mv /build/vendor/* /go/src && make -C /build server ARCH=$(ARCH) TAG=$(TAG)"
```
thockin: Can you explain why this is needed?  It should not be needed (the mv part)

MrHohn: The previous command tried to build under `/build`, which is outside $GOPATH and $GOROOT. So go will not look in the vendor folder(golang/go#12511).

What about this command?
`docker run --rm -it -v $(TEMP_DIR):/go/src/build golang:$(GOLANG_VERSION) /bin/bash -c "make -C /go/src/build server ARCH=$(ARCH) TAG=$(TAG)"`

MrHohn: By the way, for golang:1.6 image:
$GOROOT -> /usr/local/go/src
$GOPATH -> /go/src

thockin:
yeah /go/src is the right place.  Take another look at the build template
repo I linked before.  I fixed multi-arch.

MrHohn:
@thockin PTAL, the new commit adopts go-build-template.

thockin:
Once we commit this we should apply a tag.  I'd like to advocate for semver (3-field) versions everywhere.  Arguments?

thockin:
I will send you a PR with a couple tweaks - I an constantly refining :)

MrHohn:
Yeah, `MAJOR.MINOR.PATCH` seems good to me.

One more question is: should the tagging happen locally? It does not make much sense applying an union tag for the contrib repo. There are some tags on the contrib repo like `cluster-autoscaler-0.3.0`, which adds the sub-repo name as prefix. Though I don't think that's good neither.

PR #1784, opened 2016-09-23
Add more 1.4 and head GCI jobs to non-blocking submit queue

mtaufen:
/cc @vishh

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1784)


vishh:
LGTM since these are non-blocking. We need to add support for auto filing issues for all test by default to avoid having to these kind of changes in the future.

mtaufen:
The jobs we care about for GCI and v1.4 have already made it into the submit queue or are pending in https://github.com/kubernetes/contrib/pull/1794

PR #1793, opened 2016-09-27
Add more gci jobs to the non-blocking list

mtaufen:
So issues will be auto-filed for test failures in these jobs.

/cc @vishh

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1793)


apelisse:
Yeah, LGTM. We'll need a way to do it automatically, this is becoming pretty gross...

mtaufen:
I knowwwwwwwwww.

vishh:
LGTM. I don't think we want to display all these jobs in the UI as well.

PR #1794, opened 2016-09-27
Add non-blocking jobs for "hidden" GCI tests

mtaufen:
These are test jobs that were switched to GCI but don't have gci in the job name.

I think this should be it for today.

/cc @vishh

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1794)


vishh:
LGTM

mtaufen:
@apelisse PTAL

PR #1796, opened 2016-09-28
[nginx-ingress-controller] Add external authentication using auth_request

### LINKED ISSUE #1492: nginx-ingress-controller: support for oauth2_proxy
I'd like to have an nginx-ingress-controller that that works with https://github.com/bitly/oauth2_proxy,

It looks like this will require more than surface level changes to nginx-ingress-controller. I'm curious if this is something that is worth accommodating in this "upstream" nginx-ingress-controller or if it would be invasive enough to not be desired?

I'm just starting to look into this but it seems like it would requiring:
* augmenting the nginx template to accept other chunks of config for configuring oauth2_proxy
* alternatively, modifying it to optionally accept another ConfigMap that contains the nginx template so users could do arbitrary other config changes while still benefiting from the rest of the nginx-ingress-controller machinery
* needs to consume a Secret to set relevant env vars or flags for oauth2_proxy (the oauth client_secret)

The idea here is to be able to deploy Ingress along with kube-lego (for SSL to protect auth info/cookies) and then have oauth2_proxy configured  to accept users for a certain Azure Active Directory tenant or to a GitHub organization so that I can automatically give my coworkers access to cluster service while keeping others out.

I suppose this could also be done as an extra hop, with another nginx with oauth2_proxy sitting infront of the ingress nginx Pods but that seems a bit less clean.

I'm curious if anyone else is interested in this, and if it makes sense to be part of the upstream nginx-ingress-controller or if its more appropriate as a separate project?

aledbf:
fixes #1492

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1796)

### DIFF
--- a/ingress/controllers/nginx/controller.go
+++ b/ingress/controllers/nginx/controller.go
@@ -42,6 +42,7 @@ import (
 
 	"k8s.io/contrib/ingress/controllers/nginx/nginx"
 	"k8s.io/contrib/ingress/controllers/nginx/nginx/auth"
+	"k8s.io/contrib/ingress/controllers/nginx/nginx/authreq"
 	"k8s.io/contrib/ingress/controllers/nginx/nginx/config"
 	"k8s.io/contrib/ingress/controllers/nginx/nginx/cors"
 	"k8s.io/contrib/ingress/controllers/nginx/nginx/healthcheck"
@@ -723,6 +724,12 @@ func (lbc *loadBalancerController) getUpstreamServers(ngxCfg config.Configuratio
 				glog.V(3).Infof("error reading CORS annotation in Ingress %v/%v: %v", ing.GetNamespace(), ing.GetName(), err)
 			}
 
+			ra, err := authreq.ParseAnnotations(ing)
+			glog.V(3).Infof("nginx auth request %v", ra)
+			if err != nil {
+				glog.V(3).Infof("error reading auth request annotation in Ingress %v/%v: %v", ing.GetNamespace(), ing.GetName(), err)
+			}
+
 			host := rule.Host
 			if host == "" {
 				host = defServerName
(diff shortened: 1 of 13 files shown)


ingress/controllers/nginx/nginx/authreq/main.go
```diff
@@ -0,0 +1,82 @@
...
+)
+
+const (
+	// external URL that provides the authentication
+	authURL = "ingress.kubernetes.io/auth-url"
+)
+
+var (
+	// ErrMissingAnnotations is returned when the ingress rule
+	// does not contains annotations related with authentication
```
bprashanth: s/does not contains/does not contain/

ingress/controllers/nginx/controller.go
```diff
@@ -723,6 +724,12 @@ func (lbc *loadBalancerController) getUpstreamServers(ngxCfg config.Configuratio
 				glog.V(3).Infof("error reading CORS annotation in Ingress %v/%v: %v", ing.GetNamespace(), ing.GetName(), err)
 			}
 
+			ra, err := authreq.ParseAnnotations(ing)
```
bprashanth: where are we using ra?

ingress/controllers/nginx/nginx/authreq/main.go
aledbf: done

ingress/controllers/nginx/controller.go
aledbf: controller.go (last commit)

bprashanth:
Saw the related discussion on the bug, can you add an example of how to use this, and how that enables further integration with things like the oauth_proxy? (fine if you want to do this in a follow up)

bprashanth:
I think your failure is legit 
```
# k8s.io/contrib/ingress/controllers/nginx/nginx/authreq
controllers/nginx/nginx/authreq/main_test.go:86: t.Run undefined (type *testing.T has no field or method Run)
```

aledbf:
> controllers/nginx/nginx/authreq/main_test.go:86: t.Run undefined (type *testing.T has no field or method Run)

oops I forgot to use go 1.6 features (t.Run is a new thing in 1.7)

PR #1808, opened 2016-09-29
ingress/controllers/nginx: WebSocket documentation

whitlockjc:
For those that do not understand the default way in which nginx proxies
requests not containing a "Connection" header, the approach for enabling
WebSocket support might not make sense.  This commit adds documentation
that explains why things are done this way.

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1808)

### DIFF
--- a/ingress/controllers/nginx/nginx.tmpl
+++ b/ingress/controllers/nginx/nginx.tmpl
@@ -92,6 +92,13 @@ http {
     resolver {{ .defResolver }} valid=30s;
     {{ end }}
 
+    {{/* Whenever nginx proxies a request without a "Connection" header, the "Connection" header is set to "close" */}}
+    {{/* when making the target request.  This means that you cannot simply use */}}
+    {{/* "proxy_set_header Connection $http_connection" for WebSocket support because in this case, the */}}
+    {{/* "Connection" header would be set to "" whenever the original request did not have a "Connection" header, */}}
+    {{/* which would mean no "Connection" header would be in the target request.  Since this would deviate from */}}
+    {{/* normal nginx behavior we have to use this approach. */}}
+    # Retain the default nginx handling of requests without a "Connection" header
     map $http_upgrade $connection_upgrade {
         default upgrade;
         ''      close;


ingress/controllers/nginx/nginx.tmpl
```diff
@@ -92,6 +92,13 @@ http {
     resolver {{ .defResolver }} valid=30s;
     {{ end }}
 
+    {{/* Whenever nginx proxies a request without a "Connection" header, the "Connection" header is set to "close" */}}
+    {{/* when making the target request.  This means that you cannot simply use */}}
+    {{/* "proxy_set_header Connection $http_connection" for WebSocket support because in this case, the */}}
+    {{/* "Connection" header would be set to "" whenever the original request did not have a "Connection" header, */}}
+    {{/* which would mean no "Connection" header would be in the target request.  Since this would deviate from */}}
+    {{/* normal nginx behavior we have to use this approach. */}}
```
bprashanth: why add this here instead of a section in the readme? or better yet, add an example under https://github.com/kubernetes/contrib/tree/master/ingress/controllers/nginx/examples and document it there?

bprashanth:
Thanks for the pr, you probably need to auth linuxfoundation access to your github account so it verifies CLA

ingress/controllers/nginx/nginx.tmpl
aledbf: @bprashanth I think this a better place because the next map changes the http behavior and the comment show the why and consequences of changing the values

whitlockjc: That was my thought as well.  It might not make sense why we're going through all of this effort just to set the `Connection` header and documenting this as close to where it's happening made most sense to me.  I'm up for whatever you decide.

whitlockjc:
I've signed both CLAs.

ingress/controllers/nginx/nginx.tmpl
bprashanth: SG, a websockets example would still be good if you have time.

bprashanth:
LGTM

aledbf:
@bprashanth ping

Issue #1809, opened 2016-09-29
HAProxy Ingress Controller


odacremolbap:
**FEATURE REQUEST /  CONTRIBUTION**

We want to contribute our **HAProxy** Ingress controller.

It is simple enough to get going easily, only Ingress spec are supported.
That limits the features of the Ingress (no configuration management through config maps, no advanced configuration using annotations, not a lot of parameters).

Although backends (kubernetes watchers, stores, ...) are mostly the same for every ingress controller, unifying all ingresses is a WIP that will require an agreement on arguments, annotations, config maps ...

That won't probably happen very soonish, so we would like to contribute this HAProxy controller for anyone who wants to use it.

PR #1810, opened 2016-09-29
HAProxy Ingress Controller

odacremolbap:
Related to #1809 

Simple Ingress spec only HAProxy ingress controller

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1810)


bprashanth:
Can I ask that you integrate with the nginx controller?

odacremolbap:
Sounds like merging common functionality is the best path to allow anyone build a new Ingress.

* Getting the kubernetes Ingress and other related objects informers to a common *lib*
* Having an entry **cmd** to extend
* Taking care of healthz, logs, ...

**Caveats**
Depending on the functionality exposed some informers might be necessary or not (for now we don't allow a lot of configuration customisation, and no ConfigMap informer is needed)

Command line arguments are also tied to functionality. I guess we can minimise the number of arguments (which should be common to all ingress controllers, be it nginx, haproxy, vulcand, traefik, ...) and use config maps managed by each implementation


So, I'd say yes, we will help integrating ingress controllers to reduce maintenance and allow anyone to easily write her own, but for now we will be maintaining this one, since it's being already used by our clients.

I'd propose to contribute this one, and in the meantime, we will open a thread or continue with [this one](https://github.com/kubernetes/contrib/pull/223) to get both ingress controllers merged. As soon as the new ingress will be ready, we will be glad to drop the current HAProxy.

mqliang:
cc @ddysher @Pendoragon

bprashanth:
@aledbf has started work on modularizing the nginx controller, maybe you can take a look when you have time? It's the first step, not the only, IIUC the next step after we release nginx-controller:0.9 is to expose an interface: https://github.com/kubernetes/contrib/pull/1847. 

I think we need to be consistent about what more we accept into contrib at this point, see: https://github.com/kubernetes/contrib/pull/1440, though it breaks my heart to turn down an haproxy ingress controller :(

I'm hoping we can find some middle ground

baldwinSPC:
@bprashanth what would be the middle ground you'd propose? We're not in a position where we can take the nginx stuff and refactor that then refactor haproxy into it. Our team size is quite quite small. 

Is this being actively discussed in SIG Network? If so, we'd like to join that conversation. Happy to come back in and add our code to whatever the go forward solution for load balancing is. Ideally, we'd get a sense of timing on that -- like what's the ETA for the changes so that others can contribute load balancers or are *all* load balancers blocked in the project at this point until someone re-factors NGINX? Could we drop NGINX out of the mix and simply build something from scratch and allow contribs from folks for nginx, haproxy, traefik, etc?

bprashanth:
> @bprashanth what would be the middle ground you'd propose? We're not in a position where we can take the nginx stuff and refactor that then refactor haproxy into it. Our team size is quite quite small.

Manuel is already doing that in the mentioned pr, maybe help with a review pass? might be a good way to get familiar with the code

> Is this being actively discussed in SIG Network? If so, we'd like to join that conversation. 

We could, if we need fleshing out, but I think we understand what's required at this point? 

> Ideally, we'd get a sense of timing on that -- like what's the ETA for the changes so that others can contribute load balancers or are all load balancers blocked in the project at this point until someone re-factors NGINX? 

This is a little hard to pin down, because it depends on whenever contributors have time to refactor. As mentioned in https://github.com/kubernetes/contrib/pull/1440#issuecomment-236072020, it's not clear that there's too much value in maintaining a completely independent controller here in contrib vs linking to an implementation in your repo. 

> Could we drop NGINX out of the mix and simply build something from scratch and allow contribs from folks for nginx, haproxy, traefik, etc?

Sure. If you think that's the better approach please file a bug with some reasoning why, and we can take it from there. Personally I don't think it'll be too hard to just refactor.

baldwinSPC:
@odacremolbap we should refactor into the new way ingress controllers are working.

sandys:
@odacremolbap we are very keen to use this with features like proxy protocol, etc. is this going to be merged with k8s?

jefflaplante:
Any status update on the PR?

ddysher:
@jefflaplante I think most of the development effort happens in https://github.com/kubernetes/ingress, but I didn't see haproxy backend yet.

aledbf:
@jefflaplante @ddysher here is the link to the haproxy ingress controller https://github.com/kubernetes/ingress/blob/master/docs/catalog.md

ddysher:
@aledbf thanks for the pointer.  Do we plan to add those into kubernetes/ingress?

PR #1813, opened 2016-09-30
Added domain name and server port into nginx logs

kayrus:
I replaced dashes so it shouldn't brake log parsers.
In addition here are the [fluentd rules](https://github.com/kayrus/elk-kubernetes/blob/a3727fdf1411114da7e2fe921cf693afa97ea478/docker/fluentd/td-agent.conf#L262..L269) which parse these logs (requires https://github.com/tagomoris/fluent-plugin-parser plugin)

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1813)

### DIFF
--- a/ingress/controllers/nginx/nginx.tmpl
+++ b/ingress/controllers/nginx/nginx.tmpl
@@ -73,8 +73,8 @@ http {
 
     client_max_body_size "{{ $cfg.bodySize }}";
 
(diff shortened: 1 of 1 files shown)


kayrus:
I signed it!

kayrus:
ping

bprashanth:
this lgtm, allowing @aledbf to take a look

Issue #1406, opened 2016-07-20
Optimize github rate-limit quota usage of bot


bgrant0607:
Forked from #1390

Github says we should be using webhooks instead of polling.

cc @lavalamp @kubernetes/contributor-experience

bgrant0607:
@rmmh @spxtr This is the issue I mentioned earlier today. Please explain what are you working on.

rmmh:
Right now each loop takes ~10 minutes to apply every munger to every open issue, so improving this will help responsiveness for everything. (Issue assignment, flake handling, etc)

The general idea is to change the core loop from `config.ForEachIssueDo(mungers.MungeIssue)` to `config.ForEachChangedIssueDo(mungers.MungeIssue)`. 

Changed issues could be determined by (from easiest to hardest swag estimate):
- searching for open issues and comparing their updated_at times to the last loop
- polling the [events api](https://developer.github.com/v3/activity/events/)
- having a webhook receive events and note issues that have been updated

This should reduce the number of API calls for each loop from O(number of open issues) to O(number of issues updated since last loop).

There may be some extra challenges with status updates (CLA, CI pass/fail information from Jenkins), since those don't appear to show up in the event-polling API. Maybe the issue updated_at field is enough to determine that?

bgrant0607:
Some info:

According to http://submit-queue.k8s.io/#/info, the bot is making 10 API calls per second.

Rate limit is 5k per hour, which is 1 per second:
https://developer.github.com/v3/#rate-limiting

So we use quota until we run out and then stall?

rmmh:
We cheat by using ETag headers, which doesn't consume our quota, but it
takes just as much time.

eparis:
github support suggested we use the `since` argument to https://developer.github.com/v3/issues/#list-issues-for-a-repository instead of checking every issue every time. That might be a couple of lines of code changed and be a big win.

web-sockets would get us a much better place, but last I looked, there wasn't a good library for web sockets and I didn't feel like writing it for go-github.

apelisse:
Yes, using `since` seems like the easy fix right now. It's not going to work for anything "time-based" like close-stale-pr though. Not something we can't work-around ...

rmmh:
Since also doesn't won't work for statuses, since they don't modify the updated_at timestamp. :(

We can categorize mungers based on whether they might have work whether or not updated_at has changed-- ones that look at statuses will always need to run.

[...]

PR #1819, opened 2016-09-30
Munge pr since last loop

apelisse:
I've tentatively explain what this does in the commit messages, please read those instead.

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1819)

### DIFF
--- a/mungegithub/github/github.go
+++ b/mungegithub/github/github.go
@@ -184,6 +184,9 @@ type Config struct {
 	// When we clear analytics we store the last values here
 	lastAnalytics analytics
 	analytics     analytics
+
+	// Last fetch
+	since time.Time
 }
 
 type analytic struct {
@@ -2080,11 +2083,39 @@ func (obj *MungeObject) MergedAt() (*time.Time, bool) {
 	return pr.MergedAt, true
 }
 
+func (config *Config) runMungeFunction(obj *MungeObject, fn MungeFunction) error {
+	if obj.Issue.Number == nil {
+		glog.Infof("Skipping issue with no number, very strange")
+		return nil
+	}
+	if obj.Issue.User == nil || obj.Issue.User.Login == nil {
+		glog.V(2).Infof("Skipping PR %d with no user info %#v.", *obj.Issue.Number, obj.Issue.User)
+		return nil
+	}
+	if *obj.Issue.Number < config.MinPRNumber {
+		glog.V(6).Infof("Dropping %d < %d", *obj.Issue.Number, config.MinPRNumber)
+		return nil
+	}
+	if *obj.Issue.Number > config.MaxPRNumber {
+		glog.V(6).Infof("Dropping %d > %d", *obj.Issue.Number, config.MaxPRNumber)
+		return nil
+	}
+
+	glog.V(2).Infof("----==== %d ====----", *obj.Issue.Number)
+	glog.V(8).Infof("Issue %d labels: %v isPR: %v", *obj.Issue.Number, obj.Issue.Labels, obj.Issue.PullRequestLinks != nil)
+	if err := fn(obj); err != nil {
+		return err
+	}
+	return nil
+}
+
 // ForEachIssueDo will run for each Issue in the project that matches:
 //   * pr.Number >= minPRNumber
 //   * pr.Number <= maxPRNumber
 func (config *Config) ForEachIssueDo(fn MungeFunction) error {
 	page := 1
+	since := time.Time{}
+
 	for {
 		glog.V(4).Infof("Fetching page %d of issues", page)
 		listOpts := &github.IssueListByRepoOptions{
@@ -2093,6 +2124,7 @@ func (config *Config) ForEachIssueDo(fn MungeFunction) error {
 			Labels:      config.Labels,
 			Direction:   "asc",
 			ListOptions: github.ListOptions{PerPage: 100, Page: page},
+			Since:       config.since,
 		}
 		issues, response, err := config.client.Issues.ListByRepo(config.Org, config.Project, listOpts)
 		config.analytics.ListIssues.Call(config, response)
@@ -2100,32 +2132,17 @@ func (config *Config) ForEachIssueDo(fn MungeFunction) error {
 			return err
 		}
 		for i := range issues {
-			issue := issues[i]
-			if issue.Number == nil {
-				glog.Infof("Skipping issue with no number, very strange")
-				continue
-			}
-			if issue.User == nil || issue.User.Login == nil {
-				glog.V(2).Infof("Skipping PR %d with no user info %#v.", *issue.Number, issue.User)
-				continue
-			}
-			if *issue.Number < config.MinPRNumber {
-				glog.V(6).Infof("Dropping %d < %d", *issue.Number, config.MinPRNumber)
-				continue
-			}
-			if *issue.Number > config.MaxPRNumber {
-				glog.V(6).Infof("Dropping %d > %d", *issue.Number, config.MaxPRNumber)
-				continue
-			}
-			glog.V(2).Infof("----==== %d ====----", *issue.Number)
-			glog.V(8).Infof("Issue %d labels: %v isPR: %v", *issue.Number, issue.Labels, issue.PullRequestLinks != nil)
-			obj := MungeObject{
+			obj := &MungeObject{
 				config:      config,
-				Issue:       issue,
+				Issue:       issues[i],
 				Annotations: map[string]string{},
 			}
-			if err := fn(&obj); err != nil {
-				continue
+			err := config.runMungeFunction(obj, fn)
+			if err != nil {
+				return err
+			}
+			if obj.Issue.UpdatedAt != nil && obj.Issue.UpdatedAt.After(since) {
+				since = *obj.Issue.UpdatedAt
 			}
 		}
 		if response.LastPage == 0 || response.LastPage <= page {
@@ -2133,6 +2150,7 @@ func (config *Config) ForEachIssueDo(fn MungeFunction) error {
 		}
 		page++
 	}
+	config.since = since
 	return nil
 }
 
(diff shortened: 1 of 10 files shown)


apelisse:
This is an attempt to improve the situation described by #1406

mungegithub/github/timer.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+}
+
+// GetReached returns the list of issues whose timer has been recched
+func (timer *Timer) GetReached() []int {
+	reached := []int{}
+
+	for id, date := range timer.ids {
+		if timer.now().After(date) {
+			delete(timer.ids, id)
```
eparis: This is likely fine (I just started reviewing) but it makes me feel icky that `Get` has in implicit `Remove`.

mungegithub/github/state_change.go
```diff
@@ -0,0 +1,77 @@
...
+limitations under the License.
+*/
+
+package github
+
+import "sync"
+
+// StateChange keeps track of issue/commit for state changes
+type StateChange struct {
+	issues  map[string]map[int]struct{} // Issues for each commit
```
eparis: Should we be using https://github.com/kubernetes/kubernetes/tree/master/pkg/util/sets

mungegithub/github/state_change.go
```diff
@@ -0,0 +1,77 @@
...
+	commits map[int]string              // Commit for each issue
+	changed map[int]struct{}            // Changed issues since last call
```
eparis: When I see these I think of sets...

mungegithub/github/state_change.go
```diff
@@ -0,0 +1,77 @@
...
+*/
+
+package github
+
+import "sync"
+
+// StateChange keeps track of issue/commit for state changes
+type StateChange struct {
+	issues  map[string]map[int]struct{} // Issues for each commit
+	commits map[int]string              // Commit for each issue
```
eparis: Is this just tracking the last commit? Not all of the commits?

eparis (commented):
I don't get the whole idea. My not understanding the github APIs in question. How does a 'commit' change? If I have a PR with HEAD=abcdef this will get registered in `commits` and `issues`. Great. Now lets say I edit my PR such that HEAD=123456. What is going to happen here?

commits will be updated to 123456.
issues will still have abcdef pointing to my PR, but it will not also have 123456 pointing to my PR.

When does `Change` get called? 

I'm going to go read the 3rd PR now, so this might all be non-sense.

mungegithub/github/state_change.go
apelisse: Yeah, agreed, I should be using set, will replace

apelisse:
@eparis I'll try to clarify the idea

It looks like a Status change doesn't update the UpdateAt field of an issue/pr, which means that using the "Since" parameter, we won't get issues whose status as changed (but nothing else). And I assume this is a problem.

Listening to Status change webhooks event helps, but the problem is that it doesn't say what pull-requests are impacted by the status-change, just the commit sha. That's why I'm keeping track of "what issues are associated with what commit sha" so that I can say "Ah! This sha status has changed, it means these PRs have changed, we should probably also munge them."

apelisse:
Also, I only care about the HEAD because this is the status we are interested in.

apelisse:
@eparis I've updated the PR, and re-ordered the commits to maybe clarify the idea. I've also dropped the timer for now.

What this is doing:
- Use "Since" so we only munge issues that have changed since last loop
- Use webhooks and track StateChange so that we can munge issues whose HEAD commit status has changed.

Does it make more sense?

mungegithub/github/state_change.go
```diff
@@ -0,0 +1,71 @@
...
+	s.issues[commit].Insert(issue)
+	s.commits[issue] = commit
+	s.mutex.Unlock()
+}
+
+// Change should be called when the status for this commit has changed
+func (s *StateChange) Change(commit string) {
+	s.mutex.Lock()
+	s.changed.Insert(s.issues[commit].List()...)
+	s.mutex.Unlock()
```
eparis: Please use
```
lock()
defer unlock()
```
throughout.

mungegithub/mungegithub.go
```diff
@@ -107,6 +107,8 @@ func main() {
 			if err := mungers.InitializeMungers(&config.Config, &config.Features); err != nil {
 				glog.Fatalf("unable to initialize mungers: %v", err)
 			}
+			config.HookHandler.Initialize()
+			go config.HookHandler.Listen()
```
eparis: Didn't you have code to make it all work without turning the web hook on? Do we need such code? Is there a reason not to put the `Listen()` inside the `Initialize()`?

mungegithub/github/webhook.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+// Initialize the webhook. Returns false if it's missing parameters/shouldn't run.
+func (webhook *WebHook) Initialize() {
+	webhook.States = NewStateChange()
+}
+
+// Listen receives webhooks.
+func (webhook *WebHook) Listen() {
+	http.Handle("/", webhook)
```
eparis: Isn't re-using the `http.DefaultServeMux` here going to interfere with the usage for the submit queue UI on :80 ?

mungegithub/github/state_change.go
```diff
@@ -0,0 +1,71 @@
...
+func NewStateChange() *StateChange {
+	return &StateChange{
+		issues:  map[string]sets.Int{},
+		commits: map[int]string{},
+		changed: sets.NewInt(),
+	}
+}
+
+// UpdateCommit must be called when we know how commits and issues are associated
+func (s *StateChange) UpdateCommit(issue int, commit string) {
```
eparis: Is this dead code now?

mungegithub/github/state_change.go
```diff
@@ -0,0 +1,71 @@
...
+import (
+	"sync"
+
+	"k8s.io/kubernetes/pkg/util/sets"
+)
+
+// StateChange keeps track of issue/commit for state changes
+type StateChange struct {
+	issues  map[string]sets.Int // Issues for each commit
+	commits map[int]string      // Commit for each issue
```
eparis: If `UpdateCommit()` is not dead code, so is `commits` I think.

mungegithub/github/github.go
```diff
@@ -1851,13 +1860,30 @@ func (config *Config) ForEachIssueDo(fn MungeFunction) error {
...
 		}
 		if response.LastPage == 0 || response.LastPage <= page {
 			break
 		}
 		page++
 	}
 	config.since = since
+
+	// Handle additional issues
+	for id := range extraIssues {
```
eparis: It'd be pretty awesome if we could do these async instead of waiting for the next loop, but this is certainly safest. I wonder how many of the mungers can handle getting called twice at the same time...

mungegithub/github/state_change.go
apelisse: No, I just forgot to call it ..

mungegithub/github/state_change.go
apelisse: Sure, why not.

mungegithub/github/webhook.go
apelisse: Very good catch

mungegithub/mungegithub.go
apelisse: No, we need the webhooks to listen to status (unfortunately). I'll move the listen.

apelisse:
@eparis I tried to address your comment and give methods of StatusChange (that I renamed)  a better name.

Hopefully this goes in the right direction

apelisse:
@eparis I tried it with ngrok against kubernetes/kubernetes and fixed a couple of small bugs. This seems to be working now and definitely improves the token situation by orders of magnitude. Can we get that in?

Thanks

mungegithub/github/github.go
```diff
@@ -1800,46 +1834,30 @@ func (config *Config) ForEachIssueDo(fn MungeFunction) error {
 			Labels:      config.Labels,
 			Direction:   "asc",
 			ListOptions: github.ListOptions{PerPage: 100, Page: page},
+			Since:       config.since,
```
rmmh: The JSON for this is `omitempty` -- why wouldn't zero time work?

mungegithub/github/github.go
```diff
@@ -1787,11 +1790,42 @@ func (obj *MungeObject) MergedAt() *time.Time {
...
 //   * pr.Number >= minPRNumber
 //   * pr.Number <= maxPRNumber
 func (config *Config) ForEachIssueDo(fn MungeFunction) error {
 	page := 1
+
+	// Using Zero time doesn't work with github
+	if config.since.IsZero() {
+		config.since = time.Date(1900, 1, 1, 0, 0, 0, 0, time.UTC)
+	}
+	since := time.Now()
```
rmmh: We may get some weirdness with clock skew here. Can we use the updated time of the last issue in the loop instead?

mungegithub/github/github.go
```diff
@@ -1800,46 +1852,47 @@ func (config *Config) ForEachIssueDo(fn MungeFunction) error {
...
-			obj := MungeObject{
+			obj := &MungeObject{
 				config:      config,
-				Issue:       issue,
+				Issue:       issues[i],
 				Annotations: map[string]string{},
 			}
-			if err := fn(&obj); err != nil {
-				continue
+			err := config.runMungeFunction(obj, fn)
```
rmmh: there are mungers that need to run not just when things change -- like the stale issue nagging. How will this be handled?

apelisse:
@eparis @rmmh Updated the PR to consider the following problem:

Some PR assume that they will receive the full list of open PRs (submit-queue, cherrypick-queue). I think I fixed this behavior.

apelisse:
Also, I realize this is currently being migrated to test-infra, I don't want to delay the review though. I'll diff/patch as soon as possible. But please feel free to start reviewing :-)

PR #1820, opened 2016-10-01
Don't index ingress.status.ip if empty.

bprashanth:
This can happen in some odd situations where the user has asked for invalid ingress/secrets. It results in a nil pointer deref.

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1820)


freehan:
LGTM

PR #1759, opened 2016-09-20
[nginx-ingress-controller] Add support for default backend in Ingress rule

aledbf:
fixes #1590

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1759)

PR #1830, opened 2016-10-03
[nginx-ingress-controller] Add support for default backend in Ingress rule

aledbf:
replaces #1759

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1830)

### DIFF
--- a/ingress/controllers/nginx/controller.go
+++ b/ingress/controllers/nginx/controller.go
@@ -648,6 +648,9 @@ func (lbc *loadBalancerController) getStreamServices(data map[string]string, pro
 	return svcs
 }
 
+// getDefaultUpstream returns an NGINX upstream associated with the
+// default backend service. In case of error retrieving information
+// configure the upstream to return http code 503.
 func (lbc *loadBalancerController) getDefaultUpstream() *ingress.Upstream {
 	upstream := &ingress.Upstream{
 		Name: defUpstreamName,
@@ -671,7 +674,7 @@ func (lbc *loadBalancerController) getDefaultUpstream() *ingress.Upstream {
 	endps := lbc.getEndpoints(svc, svc.Spec.Ports[0].TargetPort, api.ProtocolTCP, &healthcheck.Upstream{})
 	if len(endps) == 0 {
 		glog.Warningf("service %v does not have any active endpoints", svcKey)
-		upstream.Backends = append(upstream.Backends, nginx.NewDefaultServer())
+		endps = []ingress.UpstreamServer{nginx.NewDefaultServer()}
 	} else {
 		upstream.Backends = append(upstream.Backends, endps...)
 	}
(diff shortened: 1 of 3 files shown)


aledbf:
@bprashanth ready for a first review

ingress/controllers/nginx/controller.go
```diff
@@ -739,19 +740,38 @@ func (lbc *loadBalancerController) getUpstreamServers(ngxCfg config.Configuratio
 				server = servers[defServerName]
 			}
 
+			if rule.HTTP == nil && host != defServerName {
+				glog.Info("multiple root ISSUE")
```
bprashanth: maybe send an event, and turn the message into a WARNING or ERROR? Please either extend the message itself or add a comment.

ingress/controllers/nginx/controller.go
```diff
@@ -739,19 +740,38 @@ func (lbc *loadBalancerController) getUpstreamServers(ngxCfg config.Configuratio
...
+						ups = defUps
+					}
+				}
+
 				nginxPath := path.Path
 				// if there's no path defined we assume /
 				if nginxPath == "" {
 					lbc.recorder.Eventf(ing, api.EventTypeWarning, "MAPPING",
-						"Ingress rule '%v/%v' contains no path definition. Assuming /", ing.GetNamespace(), ing.GetName())
+						"Ingress rule '%v/%v' contains no path definition. Assuming /",
```
bprashanth: / is the same as /* right? do you lose anything by just making it the latter (more explicit)

ingress/controllers/nginx/controller.go
```diff
@@ -739,19 +740,38 @@ func (lbc *loadBalancerController) getUpstreamServers(ngxCfg config.Configuratio
...
+						ing.GetNamespace(), ing.GetName())
 					nginxPath = rootLocation
 				}
 
-				// Validate that there is no another previuous
+				// Validate that there is no another previous
```
bprashanth: s/no another/no/

ingress/controllers/nginx/controller.go
```diff
@@ -844,41 +880,54 @@ func (lbc *loadBalancerController) createUpstreams(ngxCfg config.Configuration,
...
-						break
-					}
-				}
+	var upstreams []ingress.UpstreamServer
+	if err != nil {
+		return upstreams, fmt.Errorf("error getting service %v from the cache: %v", svcKey, err)
+	}
+
+	if !svcExists {
+		return upstreams, fmt.Errorf("service %v does not exists", svcKey)
```
bprashanth: do these translate into events? (the gce controller has the same problem of not sending an event for this case, pretty confusing to debug)

ingress/controllers/nginx/nginx/command.go
```diff
@@ -128,7 +128,10 @@ func (ngx Manager) testTemplate(cfg []byte) error {
 		return err
 	}
 	defer tmpfile.Close()
-	defer os.Remove(tmpfile.Name())
+	// do not remove temporal nginx.conf file if log level is >= 3
+	if !glog.V(3) {
```
bprashanth: hmm, why? this looks like a weird lateral dependency. are we just leaking the file to debug later?

ingress/controllers/nginx/controller.go
bprashanth: This is actually worth documenting. We just enforce that if there're multiple ingresses asking for a default backend, they must have the same backend? do we pick this backed on fcfs basis?

aledbf: done.

ingress/controllers/nginx/controller.go
aledbf: done, comment added

ingress/controllers/nginx/controller.go

ingress/controllers/nginx/controller.go
aledbf: no but this appears in the log of the pod

ingress/controllers/nginx/nginx/command.go
aledbf: change the behavior to remove the file only if there's no error running the check

bprashanth:
LGTM sorry for the delay

Issue #1826, opened 2016-10-03
[nginx-ingress-controller] metrics on reload


micheleorsi:
Hello,
we are adding more and more services/ingresses/pods to our bare metal cluster. 
Since the reload operation is _heavy_ and affects the balancing of our requests we would like to keep track of this event.

I developed a simple endpoint with [expvar](https://golang.org/pkg/expvar/) library to expose such information. 

Are you interested in this feature? If you want I can submit a merge request and discuss about it. 

Thanks

aledbf:
@micheleorsi sure, please share information about the endpoint (gist or a comment)

micheleorsi:
As you prefer @aledbf!

In [nginx/command.go](https://github.com/kubernetes/contrib/blob/master/ingress/controllers/nginx/nginx/command.go) I added 

```

import (
        ...

	"expvar"
)
```

```

var (
	reload = expvar.NewInt("reload")

)

```

.. and ..

```
if changed {
	reload.Add(1)

```

.. then I registered the [endpoint ](https://github.com/kubernetes/contrib/blob/master/ingress/controllers/nginx/main.go#L164)
```
mux.Handle("/debug/vars", http.DefaultServeMux)
```

Let me know if you like it, so I can submit a merge request!

aledbf:
@micheleorsi please open a PR

micheleorsi:
added in the new version outside kubernetes/contrib (look into PR comments for more details)

PR #1835, opened 2016-10-04
[nginx-ingress-controller] Added reload metrics

micheleorsi:
relates to #1826

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1835)

### DIFF
--- a/ingress/controllers/nginx/nginx/command.go
+++ b/ingress/controllers/nginx/nginx/command.go
@@ -29,7 +29,21 @@ import (
 
 	"k8s.io/contrib/ingress/controllers/nginx/nginx/config"
 	"k8s.io/contrib/ingress/controllers/nginx/nginx/ingress"
+
+	"expvar"
 )
+
+// number of reloads performed
+var (
+	reload      = new(expvar.Int)
+	reloadError = new(expvar.Int)
+)
+
+// publish metrics in order to access from path specified in registerHandlers function
+func init() {
+	expvar.Publish("reload", reload)
+	expvar.Publish("reloaderror", reloadError)
+}
 
 // Start starts a nginx (master process) and waits. If the process ends
 // we need to kill the controller process and return the reason.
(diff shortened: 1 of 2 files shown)


aledbf:
@micheleorsi can you add a second variable to differentiate reload that finish ok or with errors?

micheleorsi:
@aledbf here it is!

aledbf:
@micheleorsi thanks!

ingress/controllers/nginx/nginx/command.go
```diff
@@ -29,6 +29,14 @@ import (
 
 	"k8s.io/contrib/ingress/controllers/nginx/nginx/config"
 	"k8s.io/contrib/ingress/controllers/nginx/nginx/ingress"
+
+	"expvar"
+)
+
+// number of reload performed
```
bprashanth: nit: number of reloads performed

ingress/controllers/nginx/nginx/command.go
```diff
@@ -29,6 +29,14 @@ import (
...
+var (
+	reload = expvar.NewInt("reload")
```
micheleorsi: well, this is a way to create a variable for expvar library.
It will then be exposed to this [path](https://github.com/kubernetes/contrib/pull/1835/files#diff-c5f50015a7af6b5dfb050056b4d9338fR165) though the http.DefaultServeMux

bprashanth: Ah so /debug/vars/ dumps all such variables as json or some k:v format I'm guessing. Thanks for the clarification. Please add a comment so someone doesn't refactor it out. It actually might make more sense to setup an explicit handler that just returns the required variables so the relationship is explicit, but I leave that to you since it's pretty minor.

bprashanth:
Please also squash commits

micheleorsi:
@bprashanth: just refactored [here](https://github.com/kubernetes/contrib/pull/1835/files#diff-c5f50015a7af6b5dfb050056b4d9338fR166) and [here](https://github.com/kubernetes/contrib/pull/1835/files#diff-cf81a40001b0ad858babe4e5d4878a55R37) and [here](https://github.com/kubernetes/contrib/pull/1835/files#diff-cf81a40001b0ad858babe4e5d4878a55R42).

micheleorsi:
Sorry guys, is there something missing on my side?

micheleorsi:
re-opened #1955 here, so that I can continue to work on my master!

PR #1836, opened 2016-10-04
AWS: aws and kops-aws-updown to non-blocking list

zmerlynn:
So we start getting flakiness and testrun failures.


lavalamp:
LGTM

Wow, we need a better format for this, it's impossible to review right now.

zmerlynn:
@lavalamp: Yeah. I was wondering if there was a way to do raw strings broken over several lines and have the string pulled in correctly, but I don't know YAML well enough.

lavalamp:
note you'll need a restart for the queue to pick this up.

PR #1840, opened 2016-10-05
Docker Micro Benchmark: Add dockerd and containerd support.

Random-Liu:
@timstclair Can you help me review this? I remember you made similar change locally before.

I used this to get the result https://github.com/kubernetes/kubernetes/issues/28698#issuecomment-244514829, but forgot to submit the PR at that time.

### DIFF
--- a/docker-micro-benchmark/benchmark.sh
+++ b/docker-micro-benchmark/benchmark.sh
@@ -37,14 +37,23 @@ doBenchmark() {
   $DOCKER_MICRO_BENCHMARK $1 > $RDIR/result_benchmark.dat &
   BENCHMARK_PID=$!
   $STAT_TOOL -p $BENCHMARK_PID 1 > $RDIR/cpu_benchmark.dat &
-  DOCKER_PID=`ps -ef | awk '$8=="/usr/bin/docker" {print $2}'`
+  DOCKER_PID=`pidof dockerd`
+  if [ $? -ne 0 ]; then
+    # Be compatible with old version.
+    DOCKER_PID=`pidof docker`
+  fi
   $STAT_TOOL -p $DOCKER_PID 1 > $RDIR/cpu_docker_daemon.dat &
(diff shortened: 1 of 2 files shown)


docker-micro-benchmark/benchmark.sh
```diff
@@ -37,14 +37,23 @@ doBenchmark() {
...
+  if [ $? -ne 0 ]; then
+    # Be compatible with old version.
+    DOCKER_PID=`pidof docker`
+  fi
   $STAT_TOOL -p $DOCKER_PID 1 > $RDIR/cpu_docker_daemon.dat &
   DOCKER_PIDSTAT=$!
+  # For old version without containerd, there is no data for containerd,
+  # and the corresponding data won't be plotted.
+  CONTAINERD_PID=`pidof docker-containerd`
+  $STAT_TOOL -p $CONTAINERD_PID 1 > $RDIR/cpu_containerd_daemon.dat &
```
timstclair: Maybe you should either check than `CONTAINERD_PID != ""` or `$? == 0`?

Random-Liu: It seems that we still need an empty file for older version, or else gnuplot will report error https://github.com/kubernetes/contrib/pull/1840/files/32950d2e568ae09879d392287358529682fadc13#diff-b59587c4198e373e90b2198c5b137a41R9.

I remember so, but let me double check.

Random-Liu: * If we add check here, `cpu_containerd_daemon.dat` will not be generated, and we'll get a warning`warning: Skipping unreadable file "cpu_containerd_daemon.dat"`
* If we don't check here, an empty `cpu_containerd_daemon.dat` will be generated, and we'll get a warning `warning: Skipping data file with no valid points`.

None of them will affect the result, and they will only happen for docker <= 1.10, which should be acceptable. :)

PR #1847, opened 2016-10-06
[nginx-ingress-controller] Change structure of packages

### LINKED ISSUE #1665: [nginx-ingress-controller] Add cloud detection
In order to remove invalid settings, like `use-proxy-protocol` in GCE/GKE

### LINKED ISSUE #1799: [nginx-ingress-controller] Ingress address counts don't match ingress-controller replica count
I'm running into an anomaly where the Ingress resource for my nginx-ingress-controller has a count of addresses for a host that **do not** actively match the number of replicas of the nginx-ingress-controller itself.

I'm running nginx-ingress-controller v0.8.3 in a Deployment with no hostPort, but instead a NodePort-typed Service fronting the Deployment, and the Ingress resource references this Service for a single host on a single path, in its rules.

This nginx-ingress-controller + Ingress + NodePort Service setup was chosen based on docs read, and mainly done to allow each namespace to run its own nginx-ingress-controller. Each controller is configured in its own namespace, with its own default backend as well, and a `--watch-namespace=$(POD_NAMESPACE)` to only watch for Ingresses in the namespace it operates in.

By default, the ingress-controller Deployment only has 1 replica.

The nginx-ingress-controller is working and I have external access to my app, but in a spurt of curiosity, I scaled the ingress-controller Deployment from 1 replica to 4 (I have 3 Nodes in my cluster), and the address count reported by the ingress resource shows that I have 3 addresses for the Ingress host (one for each of the 3 Nodes). It seemed logical to only have 3 addresses for the Ingress after scaling up as I only had 3 Nodes in the cluster. I also figured that by using a NodePort-typed Service, iptables was already handling the forwarding of the packets for the 4th replica, wherever it ended up living in the set of 3 Nodes, given that I can't physically have a 4th address.

But when I scaled the Deployment back down to 1 replica, the ingress resource address count changed to 2 addresses, instead of 1.

So questions:
1. In this setup, how many addresses _should_ the Ingress resource be reporting? Is it supposed to match the number of replicas of the ingress-controller itself?
   - Better said, how come when I scale down to 1 replica, down from 4, do I have any number of addresses other than a single one being reported by the Ingress? Is this logically incorrect?
2. Given that the ingress-controller's Service is a NodePort type, which opens the same random port on **ALL** nodes, shouldn't the addresses listed for the ingress **always** be the address for each Node, or is there something strange going on that I haven't caught onto given the nature of the nginx-ingress-controller's role?
   - I primarily ask this because even in cases where only one single address is reported for the Ingress, the host:NodePort for **all** Nodes still works in an external curl command to the app, proxying to the app just fine, no matter the Node IP used - the Service used for the Ingress is a NodePort Service after all, so this should be the case, regardless. Is it possible that something to do with the logic of which addresses the Ingress reports to satisfy the host, is off?

Thanks!

/cc @aledbf @bprashanth

aledbf:
Goals of this PR:
- change the layout of the packages to be agnostic to the backend (first step)
- remove most of the references to nginx
- remove unnecessary function exports
- add test 
- refactor how the status of the Ingress rules are keep up to date (using leader election)
- move main inside cmd package

My wish is that to implement a new controller (like caddy) you just need to write a template for the tool, calling the generic code and a Dockerfile

replaces #1762

fixes #1665
fixes #1799

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1847)


mqliang:
cc @ddysher @Pendoragon

aledbf:
@bprashanth ready for a first review

kayrus:
@aledbf why this PR was closed?

aledbf:
@kayrus because I'm working in the next version [here](https://github.com/aledbf/ingress-controller). When ready I will open a PR in this repo.

PR #1741, opened 2016-09-13
ClusterAutoscaler: first fit decreasing binpacking-based estimate algorithm

mwielgus:
This should help with #1722.

The goal of the PR is to provide a better, immediate estimation of how many nodes should be added to a cluster to handle a given set of pods. The idea is to run an approximate bin-packing of pending pods to the cluster using FFD heuristic. See 

https://en.wikipedia.org/wiki/Bin_packing_problem

for details.

cc: @piosz @fgrzadkowski

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1741)

PR #1850, opened 2016-10-07
Cluster-autoscaler: Add a flag to enable more aggressive new node count nt estimation via binpacking estimator

mwielgus:
Binpacking estimator was added here: #1741

cc: @piosz @fgrzadkowski

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1850)

### DIFF
--- a/cluster-autoscaler/cluster_autoscaler.go
+++ b/cluster-autoscaler/cluster_autoscaler.go
@@ -55,6 +55,13 @@ func (flag *MultiStringFlag) Set(value string) error {
 	return nil
 }
 
+const (
+	//BasicEstimatorName is the name of basic estimator.
+	BasicEstimatorName = "basic"
+	// BinpackingEstimatorName is the name of binpacking estimator.
+	BinpackingEstimatorName = "binpacking"
+)
+
 var (
 	nodeGroupsFlag          MultiStringFlag
 	address                 = flag.String("address", ":8085", "The address to expose prometheus metrics.")
@@ -72,10 +79,14 @@ var (
 		"Node utilization level, defined as sum of requested resources divided by capacity, below which a node can be considered for scale down")
 	scaleDownTrialInterval = flag.Duration("scale-down-trial-interval", 1*time.Minute,
 		"How often scale down possiblity is check")
-	scanInterval  = flag.Duration("scan-interval", 10*time.Second, "How often cluster is reevaluated for scale up or down")
-	maxNodesTotal = flag.Int("max-nodes-total", 0, "Maximum number of nodes in all node groups. Cluster autoscaler will not grow the cluster beyond this number.")
-
+	scanInterval      = flag.Duration("scan-interval", 10*time.Second, "How often cluster is reevaluated for scale up or down")
+	maxNodesTotal     = flag.Int("max-nodes-total", 0, "Maximum number of nodes in all node groups. Cluster autoscaler will not grow the cluster beyond this number.")
 	cloudProviderFlag = flag.String("cloud-provider", "gce", "Cloud provider type. Allowed values: gce, aws")
+
+	// AvailableEstimators is a list of available estimators.
+	AvailableEstimators = []string{BasicEstimatorName, BinpackingEstimatorName}
+	estimatorFlag       = flag.String("estimator", BasicEstimatorName,
+		"Type of resource estimator to be used in scale up. Available values: ["+strings.Join(AvailableEstimators, ",")+"]")
 )
 
 func createKubeClient() *kube_client.Client {
@@ -241,7 +252,7 @@ func run(_ <-chan struct{}) {
 					scaleUpStart := time.Now()
 					updateLastTime("scaleup")
 					scaledUp, err := ScaleUp(unschedulablePodsToHelp, nodes, cloudProvider, kubeClient, predicateChecker, recorder,
-						*maxNodesTotal)
+						*maxNodesTotal, *estimatorFlag)
 
 					updateDuration("scaleup", scaleUpStart)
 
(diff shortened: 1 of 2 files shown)


fgrzadkowski:
Where you able to run performance tests to see what's the impact of this
new algorithm? Complexity is higher than earlier, so CPU usage could be
potentially much higher.

-- sent from a mobile device

07.10.2016 5:46 PM "Marcin Wielgus" <notifications@github.com> napisał(a):

mwielgus:
Total CPU won't change. I agree that the CPU consumption in the estimate phase would be higher but the global change would be marginal. Much more CPU is used in the scale down phase where we re-evaluate then nodes and pods every 10 seconds. Anyway, the current default is to use the old (faster) algorightm.

cluster-autoscaler/cluster_autoscaler.go
```diff
@@ -72,8 +72,9 @@ var (
 		"Node utilization level, defined as sum of requested resources divided by capacity, below which a node can be considered for scale down")
 	scaleDownTrialInterval = flag.Duration("scale-down-trial-interval", 1*time.Minute,
 		"How often scale down possiblity is check")
-	scanInterval  = flag.Duration("scan-interval", 10*time.Second, "How often cluster is reevaluated for scale up or down")
-	maxNodesTotal = flag.Int("max-nodes-total", 0, "Maximum number of nodes in all node groups. Cluster autoscaler will not grow the cluster beyond this number.")
+	scanInterval         = flag.Duration("scan-interval", 10*time.Second, "How often cluster is reevaluated for scale up or down")
+	maxNodesTotal        = flag.Int("max-nodes-total", 0, "Maximum number of nodes in all node groups. Cluster autoscaler will not grow the cluster beyond this number.")
+	aggressiveEstimation = flag.Bool("aggressive-estimation", false, "More aggressive estimation of how many extra nodes are needed")
```
fgrzadkowski: Is it actually aggressive? Maybe this flag name should refer to the fact that we are imitating scheduling or bin-packing? Maybe we can add a string flag `estimator` which will have two values `basic` and `binpacking`. WDYT?

fgrzadkowski (commented):
From your comment I assume you have verified that the total CPU usage does not grow?

Apart from the above only small comment about flag name.

cluster-autoscaler/cluster_autoscaler.go

mwielgus:
@fgrzadkowski 
The complexity of the binpacking estimation is O(new_nodes * pending_pods + pending_pods log pending_pods). Even if we had to add 1000 nodes to handle 10000 pending pods then the number of operations would be just ~10M, doable in ~0.1 sec on a modern computer. And this is just a one-time cost. 
Situations like that are heavy unlikely, even after we move out of beta. And if we had cluster of 1000-2000 nodes with such big amount of pods I would be much more worried about scale down performance that evaluates this amount of pods/nodes every 10 sec, not just one-time.

mwielgus:
Moreover we already have a worse thing just nearby. The verifier that a pod is really unschedulable (by @piosz) consumes O(current_nodes * pending_pods).

See: https://github.com/kubernetes/contrib/blob/master/cluster-autoscaler/utils.go#L90

cluster-autoscaler/cluster_autoscaler.go
```diff
@@ -72,10 +77,13 @@ var (
...
-	scanInterval  = flag.Duration("scan-interval", 10*time.Second, "How often cluster is reevaluated for scale up or down")
-	maxNodesTotal = flag.Int("max-nodes-total", 0, "Maximum number of nodes in all node groups. Cluster autoscaler will not grow the cluster beyond this number.")
-
+	scanInterval      = flag.Duration("scan-interval", 10*time.Second, "How often cluster is reevaluated for scale up or down")
+	maxNodesTotal     = flag.Int("max-nodes-total", 0, "Maximum number of nodes in all node groups. Cluster autoscaler will not grow the cluster beyond this number.")
 	cloudProviderFlag = flag.String("cloud-provider", "gce", "Cloud provider type. Allowed values: gce, aws")
+
+	AvailableEstimators = []string{BasicEstimatorName, BinpackingEstimatorName}
+	estimatorFlag       = flag.String("estimator", BasicEstimatorName,
+		"Type of resource estimator to be used in scale up. Avaliable values: ["+strings.Join(AvailableEstimators, ",")+"]")
```
fgrzadkowski: Available

cluster-autoscaler/scale_up.go
```diff
@@ -87,38 +89,43 @@ func ScaleUp(unschedulablePods []*kube_api.Pod, nodes []*kube_api.Node, cloudPro
...
+				option.pods = append(option.pods, pod)
 			} else {
 				glog.V(2).Infof("Scale-up predicate failed: %v", err)
 				podsRemainUnshedulable[pod] = struct{}{}
 			}
 		}
-		if groupHelpsSomePods {
+		if len(option.pods) > 0 {
+			if estimatorName == BinpackingEstimatorName {
+				aggresiveEstimator := estimator.NewBinpackingNodeEstimator(predicateChecker)
```
fgrzadkowski: s/aggresiveEsitmator/binpackingEstimator

cluster-autoscaler/scale_up.go
```diff
@@ -87,38 +89,43 @@ func ScaleUp(unschedulablePods []*kube_api.Pod, nodes []*kube_api.Node, cloudPro
...
+				option.nodeCount = aggresiveEstimator.Estimate(option.pods, nodeInfo)
+			} else if estimatorName == BasicEstimatorName {
+				safeEstimator := estimator.NewBasicNodeEstimator()
```
fgrzadkowski: s/safeEstimator/BasicEstimator

fgrzadkowski (commented):
LGTM. Only minor things. Feel free to add lgtm label once you fix them

cluster-autoscaler/cluster_autoscaler.go

cluster-autoscaler/scale_up.go
mwielgus: Done.

cluster-autoscaler/scale_up.go
mwielgus: Done.

mwielgus:
all comments addressed, adding lgtm

PR #1865, opened 2016-10-11
state a custom inventory file must be at the same directory as group_vars

### LINKED ISSUE #1864: AnsibleUndefinedVariable for networking when running deploy-cluster.sh
Attempting to run the deploy-cluster.sh script like this:

cd scripts
INVENTORY=../inventory.kube ./deploy-cluster.sh 

Results in an error:
fatal: [192.10.0.4] => {'msg': "AnsibleUndefinedVariable: One or more undefined variables: 'networking' is undefined", 'failed': True}

ingvagabund:
SSIA

Fixes: #1864

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1865)

### DIFF
--- a/ansible/scripts/init.sh
+++ b/ansible/scripts/init.sh
@@ -18,3 +18,17 @@ CUR_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 
 INVENTORY_DIR="${CUR_DIR}/../inventory"
 PLAYBOOKS_DIR="${CUR_DIR}/../playbooks"
+
+check_inventory() {
+	inv_dir=$(dirname $1)
+	if [ ! -e "${inv_dir}/group_vars" ]; then
+		echo "Inventory directory ${inv_dir} is missing group_vars directory"
+		exit 1
+	fi
+}
+
+ansible_playbook() {
+	inventory=$1
+	check_inventory ${inventory}
+	ansible-playbook -i $@
+}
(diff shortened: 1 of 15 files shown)


rrati:
lgtm

ingvagabund:
Thanks for the review

eparis:
Should we do any checks in `init.sh`

ingvagabund:
@eparis what could work too

ingvagabund:
@eparis PTAL. I have rephrased the statement in README to be more open. The ``group_vars/all.yml`` is not the only solution. Individual rules can have its own group_vars files. Extending ``init.sh`` and underlying script files with check as well.

ansible/scripts/init.sh
```diff
@@ -18,3 +18,11 @@ CUR_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
...
 INVENTORY_DIR="${CUR_DIR}/../inventory"
 PLAYBOOKS_DIR="${CUR_DIR}/../playbooks"
+
+check_inventory() {
+	inv_dir=$(dirname $1)
+	if [ ! -e "${inv_dir}/group_vars" ]; then
+		echo "Inventory directory ${inv_dir} is missing group_vars directory"
+		exit 1
+	fi
+}
```
eparis: Is there ever a time you wouldn't want to just call 'check_inventory' here? instead of doing it in every script file?

ingvagabund: No. However, the ``inventory`` is known after you set it. And that is after you run the ``./init.sh``. I can create another function, e.g. ``ansible_playbook``, that will call ``check_inventory`` and then ``ansible-playbook``.

ingvagabund:
@eparis what about now?

PR #1871, opened 2016-10-11
Dnsmasq metrics

bowei:
Posting the dnsmasq-metrics daemon. This still needs to be done:

This commit is towards resolving issue https://github.com/kubernetes/kubernetes/issues/29060
- e2e test with a dnsmasq server running in a container
- readme.md instructions

I'm posting this early to get some feedback.

Note: for the build, I used thockin's template.

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1871)


freehan:
cc: @bprashanth @MrHohn

dnsmasq-metrics/pkg/server/options.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+// NewOptions creates a new options struct.
+func NewOptions() *Options {
+	// Defaults should be set by ConfigureFlags() below.
+	return &Options{}
+}
+
+// ConfigureFlags adds command line options for the daemon
+func (opt *Options) ConfigureFlags(flagSet *pflag.FlagSet) {
```
thockin: It's minor in this case, so maybe not worth doing, but a bit of thought on the pattern of flags in libs.  This couples the lib to a flag impl and menas that any app that uses this lib uses these flag names or else defines them all.  In this case we don't have another user, so not a huge deal.

A better pattern is to make a Config or Options struct in this package, with a NewConfig() func that fills in default values, and make the `main` package define the flags that populate the config.  Then pass the config down to Run().  You already do most of this.

dnsmasq-metrics/pkg/server/metrics.go
```diff
@@ -0,0 +1,98 @@
...
+			Subsystem: options.PrometheusSubsystem,
+			Name:      "insertions",
+			Help:      "Counter of DNS cache insertions (from start of process)",
+		})
+	gauges[dnsmasq.CacheSize] = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Namespace: options.PrometheusNamespace,
+			Subsystem: options.PrometheusSubsystem,
+			Name:      "max_size",
+			Help:      "Maximum size of the dns cache",
```
thockin: s/dns/DNS/g ?

dnsmasq-metrics/pkg/dnsmasq/dnsmasq.go
```diff
@@ -0,0 +1,130 @@
...
+var MetricToString = map[int]string{
+	CacheHits:       "hits",
+	CacheMisses:     "misses",
+	CacheEvictions:  "evictions",
+	CacheInsertions: "insertions",
+	CacheSize:       "cachesize",
+}
+
+// Metrics exported by dnsmasq via *.bind CHAOS queries.
+type Metrics [MetricCount]int64
```
thockin: Optional: fixed-size arrays and *ToString() methods are a very C-ish approach.  If I were doing this, I would a) define `type MetricName string`; b) define the consts as MetricName; c) make this a `map[MetricName]int64` Strings are pretty cheap in Go.

dnsmasq-metrics/build/Dockerfile.e2e
```diff
@@ -0,0 +1,10 @@
+# Builds the docker container for the e2e test.
```
thockin: I would put all the e2e stuff in a dir `test` or `e2e` or something, rather than `build`

thockin (commented):
minor points, lgtm otherwise

dnsmasq-metrics/pkg/server/options.go
bowei: done

dnsmasq-metrics/build/Dockerfile.e2e
bowei: moved to test/e2e/

dnsmasq-metrics/README.md
```diff
@@ -0,0 +1,16 @@
+# dnsmasq-metrics
+
+dnsmasq-metrics is a daemon that exports metrics tracked by dnsmasq as
+prometheus metrics. It is meant to be run in the same pod as a dnsmasq instance.
+
+## Building
+
+TODO
```
thockin: still not seeing this updated

thockin:
Another great example of the utility of Pods.

brancz:
Is there already an image of this on gcr and I can't find it or is it not there yet?

bowei:
@brancz : sorry about that -- it should be up now

````
$ docker pull gcr.io/google_containers/dnsmasq-metrics-amd64:1.0
1.0: Pulling from google_containers/dnsmasq-metrics-amd64
Digest: sha256:4063e37fd9b2fd91b7cc5392ed32b30b9c8162c4c7ad2787624306fc133e80a9
Status: Image is up to date for gcr.io/google_containers/dnsmasq-metrics-amd64:1.0
````

PR #1874, opened 2016-10-12
Remove "ok to test" munger.

spxtr:
https://github.com/kubernetes/test-infra/issues/812

You don't need to say "ok to test" anymore, "@k8s-bot test this" works. The only case we don't hit anymore is if you apply the lgtm label without saying anything, which we can fix elsewhere. In the meantime, lets fix the world.

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1874)


apelisse:
Deleted code never crashes!

eparis:
But it does break the system. As you said: `The only case we don't hit anymore is if you apply the lgtm label without saying anything, which we can fix elsewhere.`

But we didn't fix it. That was the entire point of this munger. I think we need to revert this change (although we can clearly remove the `ok to test` portion of the comment message)

https://github.com/kubernetes/kubernetes/pull/35042#issuecomment-255864327

spxtr:
Rather than add this one back, I'll try to get the equivalent correct behavior into prow in the next few days. This munger spams us with messages if we get behind on testing.

apelisse:
Can we make it into a slash-command though for consistency?

eparis:
This one shouldn't have spammed it behind...  All it did was approve PRs for test if they got LGTM...  did you remove the wrong munger?

spxtr:
No, it tested if a PR needed "ok to test" by looking at when the github status was last updated. We had some problems where Jenkins got behind and didn't update statuses fast enough, then this munger started saying "@k8s-bot test this" every 10 minutes, exacerbating the problem.

spxtr:
@apelisse when you're back lets discuss that. I don't like that we have "@k8s-bot test this"-type commands and also "/lgtm". Lets pick one or the other :)

PR #1876, opened 2016-10-13
Add --verbose to boilerplate check to show what does not match

bowei:
--verbose will print out detailed information about why a given
file did not match the boilerplate match

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1876)

### DIFF
--- a/hack/boilerplate/boilerplate.py
+++ b/hack/boilerplate/boilerplate.py
@@ -17,16 +17,30 @@
 from __future__ import print_function
 
 import argparse
+import difflib
 import glob
 import json
+import logging
 import mmap
 import os
 import re
 import sys
 
 parser = argparse.ArgumentParser()
-parser.add_argument("filenames", help="list of files to check, all files if unspecified", nargs='*')
+parser.add_argument(
+    "filenames",
+    help="list of files to check, all files if unspecified",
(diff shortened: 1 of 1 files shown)


thockin:
Do we have an equiv in the main repo, too?

hack/boilerplate/boilerplate.py
```diff
@@ -88,6 +105,9 @@ def file_passes(filename, refs, regexs):
 
     # if we don't match the reference at this point, fail
     if ref != data:
+        logger.info('Header does not match reference, diff:')
+        for line in difflib.context_diff(ref, data, 'reference', 'actual'):
```
eparis: I personally prefer unified diffs, as I'd guess do most people who use github.

Can you use the actual filename instead of 'actual'?
Can you make the out put directly usable in `patch` to fix the problem?

eparis (changes requested):
Really cool! I'd rather do this in the kubernetes/kubernetes repo first a copy it over here.

Thank you!

bowei:
I made the changes to main kubernetes instead, see here:

https://github.com/kubernetes/kubernetes/pull/34760

PR #1879, opened 2016-10-14
Make map_hash_bucket_size configurable

### LINKED ISSUE #1817: ingress: nginx - could not build map_hash
Same issue I faced months ago when I provided new nginx config in this PR https://github.com/kubernetes/contrib/pull/1310.

```
E0930 16:07:51.073733       1 command.go:87] failed to execute nginx -s reload: 2016/09/30 16:07:51 [emerg] 16#16: could not build map_hash, you should increase map_hash_bucket_size: 32
nginx: [emerg] could not build map_hash, you should increase map_hash_bucket_size: 32
W0930 16:07:51.073971       1 utils.go:91] requeuing prod-uu/icp-media, err error reloading nginx: exit status 1
```

Today I faced it again. But the only difference - I use docker 1.11.2 and HDD instead of 1.12 and SSD. Same Ubuntu Xenial VMs with the same amount of RAM.

I successfully reproduced this error message on pure docker container with this config:
https://gist.github.com/kayrus/dcc092fe2b80b268d01dc591cfcec89a

for some reason increasing `map_hash_bucket_size` didn't help me months ago. But today it helped (`map_hash_bucket_size 64;`). Most probably it's worth to make `map_hash_bucket_size` option configurable in ingress.

r2d4:
I was getting an error while trying to run the nginx controller in minikube.  This allows this nginx configuration option to be passed in through a configmap.  

The default value depends on the processor's cache line size (32 | 64 | 128), however ServerNameHashBucketSize is determined similarly, so I've set it to the same default (64).

Fixes #1817

ref https://github.com/kubernetes/minikube/issues/611

cc @bprashanth

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1879)

### DIFF
--- a/ingress/controllers/nginx/nginx.tmpl
+++ b/ingress/controllers/nginx/nginx.tmpl
@@ -59,6 +59,7 @@ http {
     types_hash_max_size 2048;
     server_names_hash_max_size {{ $cfg.serverNameHashMaxSize }};
     server_names_hash_bucket_size {{ $cfg.serverNameHashBucketSize }};
+    map_hash_bucket_size {{ $cfg.mapHashBucketSize }};
 
     include /etc/nginx/mime.types;
     default_type text/html;
--- a/ingress/controllers/nginx/nginx/config/config.go
+++ b/ingress/controllers/nginx/nginx/config/config.go
@@ -141,6 +141,11 @@ type Configuration struct {
 	// http://nginx.org/en/docs/ngx_core_module.html#worker_connections
 	MaxWorkerConnections int `structs:"max-worker-connections,omitempty"`
 
+	// Sets the bucket size for the map variables hash tables.
+	// Default value depends on the processor’s cache line size.
+	// http://nginx.org/en/docs/http/ngx_http_map_module.html#map_hash_bucket_size
+	MapHashBucketSize int `structs:"map-hash-bucket-size,omitempty"`
+
 	// Defines a timeout for establishing a connection with a proxied server.
 	// It should be noted that this timeout cannot usually exceed 75 seconds.
 	// http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_connect_timeout
@@ -279,6 +284,7 @@ func NewDefault() Configuration {
 		GzipTypes:                gzipTypes,
 		KeepAlive:                75,
 		MaxWorkerConnections:     16384,
+		MapHashBucketSize:        64,
 		ProxyConnectTimeout:      5,
 		ProxyRealIPCIDR:          defIPCIDR,
 		ProxyReadTimeout:         60,


aledbf:
@r2d4  the next version contains a [fix](https://github.com/kubernetes/contrib/pull/1596) to adjust the size of the hash tables without user intervention

r2d4:
As far as my understanding of the nginx docs, these are distinct variables.  I was still encountering this bug building from master. 

Seems like this affects "The ngx_http_map_module module creates variables whose values depend on values of other variables." from the docs  

While the variables you've added are only configuration for the server name hash map

aledbf:
@r2d4 where are you running the ingress controller? Can you get information about the CPU?
(I'm just trying to replicate the error)

aledbf:
@r2d4 the default value is selected in at [compile](https://github.com/nginx/nginx/blob/master/auto/cc/clang#L25) and [runtime](
https://github.com/nginx/nginx/blob/master/src/core/ngx_cpuinfo.c#L94) time

r2d4:
@aledbf Thanks for the links.  I'm running it in https://github.com/kubernetes/minikube using KVM

r2d4:
just a friendly ping - anything else I can do to help this get merged?

r2d4:
@bprashanth @aledbf would it be possible to merge this and push a new image?  We would really like to have ingress working properly in minikube

bprashanth:
Sure sorry for the delay got drowned in usual 1.5 release stuff

r2d4:
No problem thanks!

timactive:
Hi
Have you a workaround solution. How to define this setting?

r2d4:
@timactive You can set custom configuration through a ConfigMap.  See here for an example

https://github.com/kubernetes/contrib/tree/master/ingress/controllers/nginx/examples/custom-configuration

timactive:
@r2d4 Thank but i already use a ConfigMap
```
apiVersion: v1
data:
  server-name-hash-bucket-size: "64"
  server-name-hash-max-size: "64"
kind: ConfigMap
metadata:
  namespace: kube-system
  name: nginx
```

kubectl exec nginx-ingress-controller-zei9g -- cat /etc/nginx/nginx.conf | grep -i "server"
    server_names_hash_max_size 64;
    server_names_hash_bucket_size 64;

aledbf:
@timactive until the next version you need to use a custom template adding the content of this PR
`map_hash_bucket_size 32;`

timactive:
@alebf thank you, i will try ;-)
https://github.com/kubernetes/contrib/blob/master/ingress/controllers/nginx/examples/custom-template/custom-template.yaml

timactive:
@aledbf it is working now thank, i have not view the custom template solution before.

Workaround solution

**1 Nginx tmpl**
wget https://raw.githubusercontent.com/kubernetes/contrib/2842513918f8fde3e2e0d0f442b658cbc371c0a9/ingress/controllers/nginx/nginx.tmpl

**2 Edit nginx.tmpl**

Add line 

> map_hash_bucket_size 64;

after

> server_names_hash_bucket_size

 ****

** 3 See readme **

https://github.com/kubernetes/contrib/blob/master/ingress/controllers/nginx/examples/custom-template/

PR #1880, opened 2016-10-14
nginx-ingress: configurable ingress class

roboll:
Add --ingress-class flag for configurable ingress class. Default
behavior is preserved, but adds flexibility to support, for example,
internal and external ELBs attached to nginx ingress controllers.

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1880)

### DIFF
--- a/hack/verify-flags/known-flags.txt
+++ b/hack/verify-flags/known-flags.txt
@@ -47,6 +47,7 @@ history-url
 http-cache-dir
 http-cache-size
 http-port
+ingress-class
 insecure-registry
 insecure-skip-verify
 issue-mungers
--- a/ingress/controllers/nginx/controller.go
+++ b/ingress/controllers/nginx/controller.go
@@ -462,7 +462,14 @@ func (lbc *loadBalancerController) sync(key string) error {
 	ngxConfig := lbc.nginx.ReadConfig(cfg)
 	ngxConfig.HealthzURL = lbc.defHealthzURL
 
-	ings := lbc.ingLister.Store.List()
+	allIngs := lbc.ingLister.Store.List()
+	ings := []interface{}{}
+	for _, ing := range allIngs {
+		curIng := ing.(*extensions.Ingress)
+		if isNGINXIngress(curIng) {
+			ings = append(ings, ing)
+		}
+	}
 	upstreams, servers := lbc.getUpstreamServers(ngxConfig, ings)
 
 	return lbc.nginx.CheckAndReload(ngxConfig, ingress.Configuration{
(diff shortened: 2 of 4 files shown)


roboll:
I signed it!

pieterlange:
Thanks for the contribution @roboll! There's some recent developments regarding this topic in https://github.com/kubernetes/kubernetes/issues/30151 and https://github.com/kubernetes/kubernetes/pull/34013 which seem to go in a different direction, but i think would still warrant the configurable class (i could have nginx'es in "gold" and "silver" QoS classes with this patch).

roboll:
Yeah I see the different direction there. 

This change preserves existing functionality and opens up a fair amount of flexibility to the current implementation, I'd still like to see it merged even if only for the short term. Thoughts?

pieterlange:
Pinging @aledbf @bprashanth 

Needless to say i'm happy to see this PR as it provides an immediate solution to a number of problems while we patiently wait for ingress claims.

aledbf:
@pieterlange I think this is a good idea until the definition of Ingress claims

aledbf:
@pieterlange @roboll right now I'm working to split the nginx ingress controller in two in order to allow the implementation of custom controllers using different products.
I already added the configurable classes feature (behind a flag)

roboll:
@aledbf thats great. Is that a long term effort? Do you think this warrants merging beforehand, or should we just wait for your work?

aledbf:
> Is that a long term effort?

No, next week I will have the first version ready.

emmanuel:
We are interested in running multiple nginx ingress controllers on our cluster. Our use-case is to segregate internal and external traffic flows (as described). 

@alebdf — any update on your efforts to split the nginx controller up? If that is still forthcoming, how about merging this (rebased) as a short-term fix for this use-case?

pieterlange:
@emmanuel See https://github.com/kubernetes/ingress

There's no official nginx release yet and we're still running into a few bugs, but this is mostly working now.

whereisaaron:
This is the main PR's for which we've been holding out for an 0.8.4 release. The good news is it does look like an equivalent change has made it into the refactored version: 

https://github.com/kubernetes/ingress/blob/master/core/pkg/ingress/controller/launch.go
```
		ingressClass = flags.String("ingress-class", "",
			`Name of the ingress class to route through this controller.`)
```

And thanks to @aledbf there is an in-progress build available where you should be able to test this feature.
> Here's is where I put the temporal images when I close a PR
> https://quay.io/repository/aledbf/nginx-ingress-controller?tab=tags
> 0.55 contains the refactoring from the PR kubernetes/ingress#95

alfirin:
Any news about this one guys ?

aledbf:
@alfirin this is already implemented [here](https://github.com/kubernetes/ingress/blob/master/core/pkg/ingress/controller/launch.go#L43)
Please check https://github.com/kubernetes/ingress/pull/181

PR #1888, opened 2016-10-18
allow both approvers and assignees in owners files

grodrigues3:
In preparation for the upcoming owners change, let's allow both assignees (old way) and approvers (new way) to lgtm code.  This does not affect the lgtm functionality.

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1888)

### DIFF
--- a/mungegithub/features/repo-updates.go
+++ b/mungegithub/features/repo-updates.go
@@ -43,6 +43,8 @@ const (
 
 type assignmentConfig struct {
 	Assignees []string `json:assignees yaml:assignees`
+	Approvers []string `json:approvers yaml:approvers`
+	//Reviewers []string `json:reviewers yaml:reviewers`
 	//Owners []string `json:owners`
 }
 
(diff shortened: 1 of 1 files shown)


mungegithub/features/repo-updates.go
```diff
@@ -287,14 +291,14 @@ func peopleForPath(path string, people map[string]sets.String, leafOnly bool, en
 	return out
 }
 
-// LeafAssignees returns a set of users who are the closest assginees to the
+// LeafAssignees returns a set of users who are the closest assignees to the
 // requested file. If pkg/OWNERS has user1 and pkg/util/OWNERS has user2 this
 // will only return user2 for the path pkg/util/sets/file.go
 func (o *RepoInfo) LeafAssignees(path string) sets.String {
 	return peopleForPath(path, o.assignees, true, o.EnableMdYaml)
```
foxish: For now maybe we should merge the two sets together? That would ensure that things continue to work as we transition and afterwards, we can separate them out again.

foxish: Also, aren't we supposed to rename assignees to approvers, as they mean the same thing?

eparis: They are merged together. Although under the "old" name `o.assignees`.  This patch LGTM, it is backwards compatible...

eparis (approved):

eparis:
although I would like to see a test...

grodrigues3:
suggestions for how to test?  It doesn't seem like something that I can write a unit test for easily

eparis:
/lgtm
uggh, we don't have any blunderbuss unit tests, do we.

grodrigues3:
@eparis  I think /lgtm works if it's the **only** text in the comment, to avoid cases like "/lgtm after you fix..."

foxish:
@eparis @grodrigues3 /lgtm //some comment here (this works)

PR #1889, opened 2016-10-18
mungegithub: Make approval label mimick lgtm label

apelisse:
Our goal is to move from **LGTM** to **approved** as the gating label for the submit-queue.
In order to transition smoothly, we will first begin by having both LGTM and Approved
labels that behave identically.  Initially, only LGTM will be needed for merge, though the approved label will be present, and eventually only the approved label will be needed.  

Notes: this implementation always adds and removes the two labels synchronously

The plan is to start warning about LGTM use later, indicating that its use will be phased out in favor of approved.  Then, it will be downgraded to be the prerequisite for approved.  

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1889)

### DIFF
--- a/mungegithub/github/github.go
+++ b/mungegithub/github/github.go
@@ -726,12 +726,16 @@ func (obj *MungeObject) AddLabel(label string) error {
 func (obj *MungeObject) AddLabels(labels []string) error {
 	config := obj.config
 	prNum := *obj.Issue.Number
-	config.analytics.AddLabels.Call(config, nil)
-	glog.Infof("Adding labels %v to PR %d", labels, prNum)
 	if len(labels) == 0 {
-		glog.Info("No labels to add: quitting")
+		glog.Info("Empty label list to add: ", prNum)
 		return nil
 	}
+	if obj.LabelSet().IsSuperset(sets.NewString(labels...)) {
+		return nil
+	}
+
+	config.analytics.AddLabels.Call(config, nil)
+	glog.Infof("Adding labels %v to PR %d", labels, prNum)
 
 	if config.DryRun {
 		return nil
@@ -753,6 +757,10 @@ func (obj *MungeObject) AddLabels(labels []string) error {
 func (obj *MungeObject) RemoveLabel(label string) error {
 	config := obj.config
 	prNum := *obj.Issue.Number
+
+	if !obj.LabelSet().Has(label) {
+		return nil
+	}
 
 	which := -1
 	for i, l := range obj.Issue.Labels {
--- a/mungegithub/mungers/approve-lgtm.go
+++ b/mungegithub/mungers/approve-lgtm.go
@@ -1,0 +1,93 @@
+/*
+Copyright 2015 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package mungers
+
+import (
+	"k8s.io/contrib/mungegithub/features"
+	"k8s.io/contrib/mungegithub/github"
+	"k8s.io/contrib/mungegithub/mungers/matchers/event"
+
+	"github.com/golang/glog"
+	"github.com/spf13/cobra"
+)
+
+// ApproveFollowsLGTM will try to follow the LGTM label closely. If it doesn't have LGTM, then removes approve.
+// If LGTM is there, then print a warning to let people know that this is not going to work forever.
+// No warning yet.
+type ApproveFollowsLGTM struct {
+}
+
+func init() {
+	a := &ApproveFollowsLGTM{}
+	RegisterMungerOrDie(a)
+}
+
+// Name is the name usable in --pr-mungers
+func (ApproveFollowsLGTM) Name() string { return "approve-lgtm" }
+
+// RequiredFeatures is a slice of 'features' that must be provided
+func (ApproveFollowsLGTM) RequiredFeatures() []string { return []string{} }
+
+// Initialize will initialize the munger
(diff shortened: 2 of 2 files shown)


foxish:
I'm concerned that a new munger which deals with `lgtm` introduces potential new races. Are we confident that this won't introduce any?

calebamiles:
that's at least something to think about, @foxish. maybe I take a look this evening or tomorrow morning (sorry my meeting chain continued unabated today) about adding in changes to the existing `lgtm*` mungers. How does that sound @apelisse?

apelisse:
@foxish, let's think about it. Here's what it does (or should do):
- Remove `lgtm` if `approved` was removed
- Remove `approved` if `lgtm` was removed
- Add `lgtm` if `approved` was added
- Add `approved` if `lgtm` was added

Today `approved` label is not used in any munger so no process can automatically remove or add it. The bot only add/remove the `lgtm` label so I don't see how we could end-up in a live-lock (but I agree that we must be careful how we proceed)

mungegithub/mungers/approve-lgtm.go
```diff
@@ -0,0 +1,89 @@
...
+
+	lastLabel := event.FilterEvents(events, lgtmOrApproved).GetLast()
+	// There is no such event, or the Event field is missing
+	if lastLabel == nil || lastLabel.Event == nil {
+		return
+	}
+
+	if *lastLabel.Event == "labeled" {
+		obj.AddLabel("approved")
+		obj.AddLabel("lgtm")
```
foxish: Maybe I misunderstand this. If the last event was `labeled`, and the bot added the `approved` and `lgtm` labels, we would come here and find it and try to add the label again?

foxish:
@apelisse I agree, there is no immediate race that could arise here. LGTM after the above comment is addressed.

mungegithub/mungers/approve-lgtm.go
foxish: As we discussed, [this is not a NOP](https://github.com/kubernetes/contrib/blob/f07b936a34401c5fb1af98573c69a0a392598639/mungegithub/github/github.go#L720-L750).

apelisse:
Hi @eparis, can you please verify that the "Do nothing if we add/remove existing/missing label" commit is OK?

mungegithub/mungers/approve-lgtm.go
apelisse: Should be addressed with the new commit

mungegithub/github/github.go
```diff
@@ -726,12 +726,17 @@ func (obj *MungeObject) AddLabel(label string) error {
...
 	config := obj.config
 	prNum := *obj.Issue.Number
-	config.analytics.AddLabels.Call(config, nil)
 	glog.Infof("Adding labels %v to PR %d", labels, prNum)
 	if len(labels) == 0 {
 		glog.Info("No labels to add: quitting")
 		return nil
 	}
+	if obj.LabelSet().IsSuperset(sets.NewString(labels...)) {
+		glog.Info("Labels are already set: quitting")
```
eparis: This seems too verbose for something common, no?

mungegithub/github/github.go
```diff
@@ -754,6 +759,11 @@ func (obj *MungeObject) RemoveLabel(label string) error {
 	config := obj.config
 	prNum := *obj.Issue.Number
 
+	if !obj.LabelSet().Has(label) {
+		glog.Info("Removing label that is not set: quitting")
```
eparis: Also this.

mungegithub/github/github.go
```diff
@@ -726,12 +726,17 @@ func (obj *MungeObject) AddLabel(label string) error {
 func (obj *MungeObject) AddLabels(labels []string) error {
 	config := obj.config
 	prNum := *obj.Issue.Number
-	config.analytics.AddLabels.Call(config, nil)
 	glog.Infof("Adding labels %v to PR %d", labels, prNum)
 	if len(labels) == 0 {
 		glog.Info("No labels to add: quitting")
```
eparis: Not your fault, but you are here. This is a pretty bad error message. At least say the obj.Number and maybe the obj.Issue.Labels...

mungegithub/github/github.go
```diff
@@ -726,12 +726,17 @@ func (obj *MungeObject) AddLabel(label string) error {
 func (obj *MungeObject) AddLabels(labels []string) error {
 	config := obj.config
 	prNum := *obj.Issue.Number
-	config.analytics.AddLabels.Call(config, nil)
 	glog.Infof("Adding labels %v to PR %d", labels, prNum)
```
eparis: This needs to be after the `IsSuperset()` early return (like you already did with the analytics collection)

mungegithub/mungers/approve-lgtm.go
```diff
@@ -0,0 +1,89 @@
...
+	} else if *lastLabel.Event == "unlabeled" {
+		obj.RemoveLabel("approved")
+		obj.RemoveLabel("lgtm")
+	} else {
+		glog.Fatal("Received unexpected event: ", *lastLabel.Event)
```
eparis: This will crash the entire mungegithub.  Is that necessary? Seems like a `glog.Errorf()` and return would be sufficient instead of constantly crashing if github api does something unexpected.

mungegithub/mungers/approve-lgtm.go
eparis: I think every munger previously has does the `if !obj.HasLabel; obj.AddLabel`, but I have no problem changing the github.go code like you did to handle it in there.

It may be ever so slightly faster to make a slice and use `AddLabels()`, I know we'll only hit the api one time, but we'll do less label->set conversions and stuff...    That's just bikeshedding.

Also, no hard coded label strings allowed in mungers. Use lgtmLabel (and make one for approvedLabel)

mungegithub/mungers/approve-lgtm.go
apelisse: Yes, we shouldn't crash if we receive an unexpected answer from github (I tend to like assertion-crash for programming errors though).

In this instance, you are correct, it will crash in the following scenario (amongst others): If github returns lgtm as a label, and event is neither label nor unlabeled, then it would crash.
I'll fix it.

apelisse:
@eparis I think I've addressed your comments. PTAL

apelisse:
We have decided to not do that anymore.

Closing.

PR #1890, opened 2016-10-18
allow either lgtm or approved for merge

grodrigues3:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1890)

### DIFF
--- a/mungegithub/mungers/submit-queue.go
+++ b/mungegithub/mungers/submit-queue.go
@@ -783,7 +783,7 @@ func (sq *SubmitQueue) getMetaData() []byte {
 const (
 	unknown                 = "unknown failure"
 	noCLA                   = "PR is missing CLA label; needs one of " + claYesLabel + ", " + cncfClaYesLabel + " or " + claHumanLabel
-	noLGTM                  = "PR does not have LGTM."
+	noLGTM                  = "PR does not have " + lgtmLabel + " label or " + approvedLabel + " label (needs at least one)."
 	lgtmEarly               = "The PR was changed after the LGTM label was added."
 	unmergeable             = "PR is unable to be automatically merged. Needs rebase."
 	undeterminedMergability = "Unable to determine is PR is mergeable. Will try again later."
@@ -799,6 +799,20 @@ const (
 	ghE2EFailed             = "Second github e2e run failed."
 	unmergeableMilestone    = "Milestone is for a future release and cannot be merged"
 )
+
+func getEarliestApprovedTime(obj *github.MungeObject) *time.Time {
+	lgtmTime := obj.LabelTime(lgtmLabel)
+	approvedTime := obj.LabelTime(approvedLabel)
+	// if both lgtmTime and approvedTime are nil, this func will return nil pointer
+	if lgtmTime == nil {
+		return approvedTime
+	} else if approvedTime == nil {
+		return lgtmTime
+	} else if lgtmTime.Before(*approvedTime) {
+		return lgtmTime
+	}
+	return approvedTime
+}
 
 // validForMerge is the base logic about what PR can be automatically merged.
 // PRs must pass this logic to be placed on the queue and they must pass this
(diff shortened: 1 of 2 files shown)


apelisse:
cc @lavalamp. We're trying to allow `approved` or `lgtm` to merge. Would that work?

mungegithub/mungers/submit-queue.go
```diff
@@ -866,22 +866,34 @@ func (sq *SubmitQueue) validForMerge(obj *github.MungeObject) bool {
...
+	if lastModifiedTime == nil || (lgtmTime == nil && approvedTime == nil) {
 		glog.Errorf("PR %d was unable to determine when LGTM was added or when last modified", *obj.Issue.Number)
 		sq.SetMergeStatus(obj, unknown)
 		return false
 	}
 
-	if lastModifiedTime.After(*lgtmTime) {
+	var timeToCompare *time.Time
+	if lgtmTime == nil {
+		timeToCompare = approvedTime
```
foxish: What if the approvedTime is nil here? It wouldn't enter the `else if...`. Maybe the next condition shouldn't be `else if..`.

foxish: We could probably encapsulate this in a function and add a comment there to make the intent clear.

mungegithub/submit-queue/deployment/contrib/configmap.yaml
```diff
@@ -7,7 +7,7 @@ data:
   config.http-cache-dir: /cache/httpcache
   config.organization: kubernetes
   config.project: contrib
-  config.pr-mungers: blunderbuss,lgtm-after-commit,submit-queue,check-labels,needs-rebase,lgtm-handler,cla
+  config.pr-mungers: blunderbuss,lgtm-after-commit,submit-queue,check-labels,needs-rebase,lgtm-handler,cla,approval-handler
```
foxish: Just activating on contrib for now, seems like a good idea :)

mungegithub/mungers/submit-queue.go
```diff
@@ -866,22 +866,34 @@ func (sq *SubmitQueue) validForMerge(obj *github.MungeObject) bool {
 	}
 
 	// Clearly
-	if !obj.HasLabel(lgtmLabel) {
+	if !(obj.HasLabel(lgtmLabel) || obj.HasLabel(approvedLabel)) {
```
foxish: In future we would break this down into two separate conditions? Because I imagine we want to set the status differently in each case.

mungegithub/mungers/submit-queue.go
apelisse: Yes, let's simplify the logic here.

eparis:
This basically LGTM. I do think getting the 'latest lgtm/approval time' should be in its own function. I also think we need to update the tests to check how these two things interact.

mungegithub/mungers/submit-queue.go
```diff
@@ -800,6 +800,18 @@ const (
 	unmergeableMilestone    = "Milestone is for a future release and cannot be merged"
 )
 
+func getEarliestApprovedTime(lgtmTime, approvedTime *time.Time) *time.Time {
+	if lgtmTime == nil {
+		return approvedTime
+	} else if approvedTime == nil {
+		return lgtmTime
+	} else if lgtmTime < approvedTime {
```
foxish: We return nil if both `lgtmTime` and `approvedTime` are nil, and we dereference the pointer later in the code.

mungegithub/mungers/submit-queue.go
```diff
@@ -800,6 +800,20 @@ const (
 	unmergeableMilestone    = "Milestone is for a future release and cannot be merged"
 )
 
+// getEarliestApprovedTime expects that lgtmTime and approvedTime are not both
+// nil.  If both of them ARE nil, it will return a nil pointer
+func getEarliestApprovedTime(lgtmTime, approvedTime *time.Time) *time.Time {
```
eparis: I feel like we should get the lgtmTime and the approvedTime here, in this function...

eparis:
so close!
```
# k8s.io/contrib/mungegithub/mungers
mungers/submit-queue.go:811: invalid operation: lgtmTime < approvedTime (operator < not defined on pointer)
godep: go exit status 2
```

mungegithub/mungers/submit-queue.go
```diff
@@ -783,7 +783,7 @@ func (sq *SubmitQueue) getMetaData() []byte {
 const (
 	unknown                 = "unknown failure"
 	noCLA                   = "PR is missing CLA label; needs one of " + claYesLabel + ", " + cncfClaYesLabel + " or " + claHumanLabel
-	noLGTM                  = "PR does not have LGTM."
+	noLGTM                  = "PR does not have LGTM or Approved Label (needs at least one)."
```
eparis: any chance while you are here you'd be willing to use `lgtmLabel` and `approvedLabel` in the string?

mungegithub/mungers/submit-queue.go
```diff
@@ -800,6 +800,21 @@ const (
...
 
+func getEarliestApprovedTime(obj *github.MungeObject) *time.Time {
+	lgtmTime := obj.LabelTime(lgtmLabel)
+	approvedTime := obj.LabelTime(approvedLabel)
+	// if both lgtmTime and approvedTime are nil, this func will return nil pointer
+	if lgtmTime == nil {
+		return approvedTime
+	} else if approvedTime == nil {
+		return lgtmTime
+	} else if lgtmTime < approvedTime {
```
eparis: I also wonder if we are supposed to use < or if we are supposed to use *lgtmTime.Before(*approvedTime)

eparis:
squash and feel free to self LGTM

PR #1903, opened 2016-10-20
Fix size check

### LINKED ISSUE #1892: Incorrect PR Size label
In PR https://github.com/kubernetes/kubernetes/pull/33944 the bot correctly applied the ` size/XXL` label to the PR that touches 72 files and adds 6936 lines.

It later removed that label and replaced it with `size/XS` <-- Buggy behavior

eparis:
fixes #1892 

The github api lies about the size of diffs if they are large. If we
then subtract out the size of generated files that it is not lieing
about we end up with negative numbers. So huge PRs can get XS size.

Instead of believing the API and deleting those things we don't care
about just sum the things we do care about and use that.

### DIFF
--- a/mungegithub/mungers/size.go
+++ b/mungegithub/mungers/size.go
@@ -151,43 +151,37 @@ func (s *SizeMunger) Munge(obj *github.MungeObject) {
 	}
 
 	issue := obj.Issue
-	pr, err := obj.GetPR()
-	if err != nil {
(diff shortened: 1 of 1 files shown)


foxish:
LGTM

foxish:
Hopefully there are no more surprises in the API :) Thanks for investigating and fixing!

PR #1909, opened 2016-10-21
Cluster-autoscaler: own drain

mwielgus:
This PR adds an own drain logic to ClusterAutoscaler. Previously we were using the code from kubectl drain command but:
- due to refactorings it was hard to update the the CA dependencies. 
- we had to add some extra logic on top of it in fast drain evaluation
- a recent feature request to check the number of replicas in rs/rc was impossible to fulfill. 

cc: @fgrzadkowski @davidopp @piosz

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1909)

### DIFF
--- a/cluster-autoscaler/simulator/drain.go
+++ b/cluster-autoscaler/simulator/drain.go
@@ -17,12 +17,10 @@ limitations under the License.
 package simulator
 
 import (
-	"fmt"
+	"k8s.io/contrib/cluster-autoscaler/utils/drain"
 
 	"k8s.io/kubernetes/pkg/api"
-	"k8s.io/kubernetes/pkg/controller"
-	"k8s.io/kubernetes/pkg/kubelet/types"
-	"k8s.io/kubernetes/pkg/runtime"
+	unversionedclient "k8s.io/kubernetes/pkg/client/unversioned"
 	"k8s.io/kubernetes/plugin/pkg/scheduler/schedulercache"
 )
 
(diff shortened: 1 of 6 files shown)


mwielgus:
@piosz @fgrzadkowski Ping

cluster-autoscaler/utils/drain/drain.go
```diff
@@ -0,0 +1,192 @@
...
+		}
+
+		if refKind == "ReplicationController" {
+			if checkReferences {
+				rc, err := client.ReplicationControllers(sr.Reference.Namespace).Get(sr.Reference.Name)
+				// Assume a reason for an error is because the RC is either
+				// gone/missing or that the rc has too few replicas configured.
+				// TODO: replace the minReplica check with pod disruption budget.
+				if err == nil && rc != nil {
+					if rc.Spec.Replicas < minReplica {
```
piosz: As @fgrzadkowski suggested how about checking here that we are not removing all pods from the given ReplicationController rather than whether the pod is just replicated?

cluster-autoscaler/utils/drain/drain.go
```diff
@@ -0,0 +1,192 @@
...
+			}
+		} else if refKind == "ReplicaSet" {
+			if checkReferences {
+				rs, err := client.ExtensionsClient.ReplicaSets(sr.Reference.Namespace).Get(sr.Reference.Name)
+
+				// Assume the only reason for an error is because the RS is
+				// gone/missing, not for any other cause.  TODO(mml): something more
+				// sophisticated than this
+				if err == nil && rs != nil {
+					if rs.Spec.Replicas < minReplica {
```
piosz: Same here

cluster-autoscaler/utils/drain/drain.go
```diff
@@ -0,0 +1,192 @@
...
+						replicated = true
+					}
+				} else {
+					return []*api.Pod{}, fmt.Errorf("replication controller for %s/%s is not available, err: %v", pod.Namespace, pod.Name, err)
+				}
+			} else {
+				replicated = true
+			}
+		}
+		if !daemonsetPod && !replicated {
```
piosz: You are checking in each case whether the pod is not created by daemon controller. How about just checking it once and continue?

cluster-autoscaler/utils/drain/drain.go
```diff
@@ -0,0 +1,192 @@
...
+)
+
+// GetPodsForDeletionOnNodeDrain returns pods that should be deleted on node drain as well as some extra information
+// about possibly problematic pods (unreplicated and deamon sets).
+func GetPodsForDeletionOnNodeDrain(
+	podList []*api.Pod,
+	decoder runtime.Decoder,
+	skipNodesWithSystemPods bool,
+	skipNodesWithLocalStorage bool,
+	checkReferences bool,
```
piosz: How about just passing client and when it's set to nil skip checking references? Having two correlated parameters is error prone.

cluster-autoscaler/simulator/drain.go
```diff
@@ -31,86 +29,25 @@ import (
...
-func isLocalVolume(volume *api.Volume) bool {
-	return volume.HostPath != nil || volume.EmptyDir != nil
+func DetailedGetPodsForMove(nodeInfo *schedulercache.NodeInfo, skipNodesWithSystemPods bool,
+	skipNodesWithLocalStorage bool, client *unversionedclient.Client, minReplicaCount int32) ([]*api.Pod, error) {
+	return drain.GetPodsForDeletionOnNodeDrain(
+		nodeInfo.Pods(),
+		api.Codecs.UniversalDecoder(),
+		skipNodesWithSystemPods,
+		skipNodesWithLocalStorage,
+		false,
```
piosz: Shouldn't be true?

piosz (changes requested):

cluster-autoscaler/utils/drain/drain.go
mwielgus: Same as above.

cluster-autoscaler/utils/drain/drain.go
mwielgus: Done.

cluster-autoscaler/simulator/drain.go
mwielgus: no, we don't want to check references in FAST get pods.

cluster-autoscaler/utils/drain/drain.go
piosz: So please add at least a comment that `checkReferences` set to `true` requires not-nil client.

cluster-autoscaler/simulator/drain.go
piosz: Something went wrong - there was false in both cases. Anyway now it's fixed.

cluster-autoscaler/utils/drain/drain.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+	if sr == nil {
+		return "", nil
+	}
+	return sr.Reference.Kind, nil
+}
+
+// CreatorRefKind returns the kind of the creator reference of the pod.
+func CreatorRef(pod *api.Pod) (*api.SerializedReference, error) {
+	creatorRef, found := pod.ObjectMeta.Annotations[controller.CreatedByAnnotation]
```
mwielgus: For the purpose of this PR we want to be compatible with kubectl drain which doesn't use ownerreference.

cluster-autoscaler/simulator/drain.go
mwielgus: Done.

piosz:
LGTM

piosz (approved):

PR #1937, opened 2016-10-27
Cluster-autoscaler: publish CA status as events in kube-system namespace

mwielgus:
More insight into what CA is doing was requested. Currently there is no good place to publish events about CA, especially in GKE where master node and CA pod is not visible. So we decided to publish them under kube-system namespace object as a temporary solution.

cc: @fgrzadkowski @piosz

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1937)

### DIFF
--- a/cluster-autoscaler/cluster_autoscaler.go
+++ b/cluster-autoscaler/cluster_autoscaler.go
@@ -78,6 +78,10 @@ var (
 	cloudProviderFlag = flag.String("cloud-provider", "gce", "Cloud provider type. Allowed values: gce, aws")
 )
 
+const (
+	clusterAutoscalerEventReason = "ClusterAutoscaler"
+)
+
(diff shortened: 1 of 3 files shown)


fgrzadkowski:
Please create a corresponding issue so that it's easier to track what's happening with cluster autoscaler.

fgrzadkowski:
As discussed offline please get a consensus what is proper way to create such global/cluster level events. I guess that starting with an issue or email thread should do the thing. Once you have agreement that using  `kube-system` namespace for such events is a good solution please assign this PR back to me.

mwielgus:
OK, created a bug for Eric Tune to decide what is the best place to do that from the UX perspective.

piosz:
Please reassign if this will be ready for the review.

PR #1940, opened 2016-10-27
Increase kubernetes security in ansible scripts

kayrus:
Inspired by https://github.com/kubernetes/kubernetes/issues/11816
By default the following locations are available through the internet:
- http://host:10251/configz
- http://host:10251/debug/pprof/heap
- http://host:10252/configz
- http://host:10252/debug/pprof/heap
- http://host:10255/metrics
- http://host:4194
- https://host:10250
- http://host:10255

Let's reduce probability to access them.

Forcing kubelet to listen the localhost interface results in inaccessible `kubectl exec` and `kubectl logs`. I'm trying to resolve this issue within https://github.com/kubernetes/kubernetes/pull/35693

Anyway it is recommended to set the following custom variables:
- `private_interface: "eth0"`
- `kubelet_bind_address: "{{ hostvars[inventory_hostname]['ansible_'+private_interface]['ipv4']['address'] }}"`
- `has_iptables: true`

And then `kubectl exec` and `kubectl logs` will work flawlessly.

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1940)

### DIFF
--- a/ansible/inventory/group_vars/all.yml
+++ b/ansible/inventory/group_vars/all.yml
@@ -99,3 +99,5 @@ dns_replicas: 1
 
 # See kube documentation for apiserver runtime config options.  Example below enables HPA, deployments features.
 # apiserver_extra_args: "--runtime-config=extensions/v1beta1/deployments=true"
+
+#private_interface: eth0
--- a/ansible/roles/etcd/tasks/install.yml
+++ b/ansible/roles/etcd/tasks/install.yml
@@ -26,7 +26,7 @@
   user:
     name=etcd
     comment="Etcd user"
-    shell=/sbin/nologin
+    shell=/usr/sbin/nologin
     state=present
     system=yes
     groups=etcd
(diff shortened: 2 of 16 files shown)


liggitt:
I'm not sure tunneling is the long-term recommendation for the kubelet secure port, though I would like to see the other ports disabled by default..

kayrus:
@liggitt which "other" ports?

liggitt:
10255, 4194

kayrus:
np. done.

liggitt:
> np. done.

I actually meant disabling in the default binaries... not here. More an aspirational goal than a request to change this PR.

I mostly wanted to make sure this wasn't moving toward tunneling as the primary solution for securing the kubelet API port if that wasn't our long-term goal

kayrus:
@liggitt 1.3.x branch is still used along with 1.4.x. kubelet security will be implemented only in 1.5.x. so why don't use at least this solution?

liggitt:
I'm not sure the expectation is that the ansible scripts install/run with non-default settings

kayrus:
@liggitt this PR will at least allow users to control these options.

kayrus:
ping

PR #1942, opened 2016-10-27
Swap submit queue blockers for GCI equivalents

mtaufen:
Ref: https://github.com/kubernetes/kubernetes/issues/34335

/cc @vishh

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1942)


apelisse:
I trust you a lot ...

mtaufen:
@apelisse does the queue need to be restarted to take up this change?

apelisse:
Yes, I should do it ...

mtaufen:
@apelisse http://submit-queue.k8s.io/#/e2e is showing the old CVM jobs disappeared from the list of blockers but the new GCI jobs still aren't there. My current theory is that this is because I missed removing them from the non-blocking list. Does our queue setup remove from the blocking list any jobs also present in the non-blocking list?

mtaufen:
https://github.com/kubernetes/contrib/pull/1965

PR #1944, opened 2016-10-27
ingress nginx: Should not interpret path as regex.

calebcase:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1944)

### DIFF
--- a/ingress/controllers/nginx/nginx/template/template.go
+++ b/ingress/controllers/nginx/nginx/template/template.go
@@ -187,12 +187,7 @@ func buildLocation(input interface{}) string {
 		return slash
 	}
 
-	path := location.Path
-	if len(location.Redirect.Target) > 0 && location.Redirect.Target != path {
-		return fmt.Sprintf("~* %s", path)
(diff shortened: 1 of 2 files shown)


calebcase:
I signed it!

aledbf:
@calebcase can you elaborate why this change is required?

calebcase:
Initially this resulted from one user having an ingress for `/a/b/c` while another user had `/myproject/a/b/c`. Because of the way the location lines are unanchored regex traffic that was expected to go to `/myproject/a/b/c` ended up going to `/a/b/c`.

The entire removal of the regex itself is motivated by two observations:
* You can't put a regex in the path spec for ingress - it will only accept a "plain" string.
* The regex chosen is case-insensitive meaning it will surprisingly match `/A/b/C` for `/a/b/c`.

I had a different patch for just anchoring the regex, but decided to propose removal first as that seems the better option.

calebcase:
Tests cleaned up.

PR #1467, opened 2016-08-02
[nginx-ingress-controller]: Use ClientConfig to configure connection

aledbf:
fixes #1459

Running with `docker run`:

```
core@localhost ~ $ docker run -it aledbf/nginx-third-party:0.31 bash
root@f6a96f46eab0:/# export KUBERNETES_MASTER=http://172.17.4.99:8080
root@f6a96f46eab0:/# /nginx-ingress-controller --default-backend-service=default/nginx-errors
I0802 14:44:58.604384       7 main.go:94] Using build: https://github.com/aledbf/contrib - git-5b9146a
W0802 14:44:58.605282       7 main.go:118] unexpected error getting runtime information: unable to get POD information (missing POD_NAME or POD_NAMESPACE environment variable)
I0802 14:44:58.607270       7 main.go:123] Validated default/nginx-errors as the default backend
W0802 14:44:58.611322       7 ssl.go:132] no file dhparam.pem found in secrets
I0802 14:44:58.615637       7 controller.go:1128] starting NGINX loadbalancer controller
I0802 14:44:58.615902       7 command.go:35] Starting NGINX process...
```

Running inside in a cluster:

```
I0802 14:47:50.254736       1 main.go:94] Using build: https://github.com/aledbf/contrib - git-5b9146a
I0802 14:47:50.254920       1 merged_client_builder.go:103] No kubeconfig could be created, falling back to service account.
I0802 14:47:50.343440       1 main.go:123] Validated default/nginx-errors as the default backend
W0802 14:47:50.343677       1 ssl.go:132] no file dhparam.pem found in secrets
I0802 14:47:50.347322       1 controller.go:1128] starting NGINX loadbalancer controller
I0802 14:47:50.347870       1 command.go:35] Starting NGINX process...
```

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1467)

PR #1951, opened 2016-10-29
Allow custom configuration for apiserver host

ghost:
Currently your system uses the kubectl_util.DefaultClientConfig() to retrieve the IP for the api-server, which may not work outside GCE in customized setups.

This PR provides optional flags to use a non-default api-server. It uses the same flags and borrows code from the kubernetes/dashboard project:
https://github.com/kubernetes/dashboard/blob/master/src/app/backend/client/apiserverclient.go
https://github.com/kubernetes/dashboard/blob/master/src/app/backend/dashboard.go#L31

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1951)


doug-clarifai:
I signed it!

doug-clarifai:
I signed it!

ingress/controllers/nginx/apiserverclient.go
```diff
@@ -0,0 +1,63 @@
...
+
+import (
+	"github.com/golang/glog"
+
+	client "k8s.io/kubernetes/pkg/client/unversioned"
+	"k8s.io/kubernetes/pkg/client/unversioned/clientcmd"
+	clientcmdapi "k8s.io/kubernetes/pkg/client/unversioned/clientcmd/api"
+)
+
+// Dashboard UI default values for client configs.
```
therc: Reference to Dashboard

bprashanth:
Is your goal to point at an apiserver that kubectl doesn't know about? because even outside a kubecluster, we should be looking up the default kubeconfig context, and pulling out the apiserver from there, at least the last time i checked

aledbf:
@doug-clarifai this is not required. Please check this comment https://github.com/kubernetes/contrib/pull/1467#issue-168905136

doug-clarifai:
@aledbf The comment you reference only allows a simple host name as an environment variable for the api server.

The goal of this change is to allow for more flexible configurations. kubernetes/dashboard added these flags to match kubectl and proxy: https://github.com/kubernetes/dashboard/pull/1256 https://github.com/kubernetes/dashboard/issues/905

PR #1952, opened 2016-10-30
Remove service should remove the deployments

hectorj2f:
I was trying to use this package with any newer Kubernetes version. But it always fails to cleanup the services cause it tries to remove the replica controllers.  It should remove the deployments instead when using the current version.

I wanted to add some other improvements but I want to start with something smaller :).

ping @mikedanese

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1952)


hectorj2f:
I signed it!

PR #1955, opened 2016-10-31
[nginx-ingress-controller] Added reload metrics

micheleorsi:
relates to #1826

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1955)

### DIFF
--- a/ingress/controllers/nginx/main.go
+++ b/ingress/controllers/nginx/main.go
@@ -29,6 +29,7 @@ import (
 	"github.com/golang/glog"
 	"github.com/spf13/pflag"
 
+	"expvar"
 	"k8s.io/kubernetes/pkg/api"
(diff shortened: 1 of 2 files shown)


micheleorsi:
@aledbf so also this PR now is superseded, since I see doc about metrics [here](https://github.com/aledbf/ingress-controller/tree/master/backends/nginx#metrics)

aledbf:
@micheleorsi same thing the previous PR, sorry. 
Please check [here](https://github.com/aledbf/ingress-controller/blob/master/pkg/ingress/controller/metrics.go), it uses prometheus (like the rest of k8s) so it provides more information (like memory and cpu). This is an example of the output

captura de pantalla 2016-10-31 a las 4 07 56 p m

micheleorsi:
ok thanks! So I'll close this pull request since it's not anymore useful!

PR #1959, opened 2016-10-31
ansible: Don't upgrade key cluster components on every playbook run

kayrus:
Latest software is good, but it should be under control.

Otherwise docker/etcd/flannel will be updated simultaneously on all nodes.

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1959)


kayrus:
ping

ingvagabund:
Is not purpose of the installation part to install the latest? If you want to keep the current versions, the installation part should be skipped instead.

PR #1965, opened 2016-11-01
Remove gci jobs that are now blocking from the list of non-blocking jobs

mtaufen:
/cc @apelisse

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1965)


apelisse:
Let's see what it does :D

PR #1973, opened 2016-11-02
Cluster Autoscaler Value for Node Not Needed Time (10 minutes) is Not Configurable

atomantic:
According to the docs page, "If a node is not needed for more than 10 min (configurable) then it can be deleted."
Where would I configure that 10 minute value. Since it will wait for 10 minutes per node it needs to delete, it could take a while to remove all the cruft if I have a lot of overage.

"I believe the value is not configurable." - David Oppenheimer from GCP.

If this value is configurable, please close this PR and document how to configure it. If it really isn't configurable, this is a missing feature.

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1973)

### DIFF
--- a/cluster-autoscaler/README.md
+++ b/cluster-autoscaler/README.md
@@ -70,7 +70,7 @@ manifest-run pods or pods created by daemonsets).
 * There are no pods with local storage. Applications with local storage would lose their 
 data if a node is deleted, even if they are replicated.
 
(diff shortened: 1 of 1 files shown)


mwielgus:
This is configurable:

```
scaleDownUnneededTime = flag.Duration("scale-down-unneeded-time", 10*time.Minute,
		"How long the node should be unneeded before it is eligible for scale down")
```

mwielgus:
cc: @davidopp

PR #1980, opened 2016-11-03
[WIP] OpenStack Heat cluster autoscaler

taimir:
Hi there, I've been working on an autoscaler for OpenStack for a while now and I thought I should share it. 

It  is entirely based on Heat. It utilizes Heat ResourceGroups (or AutoScalingGroups) to resize groups of instances that form a cluster of minions.

The rough idea is that the user specifies the name of the resource group (and Heat stack it is contained in) that should be auto-scaled when the autoscaler is started, like this:

```
min:max:resource-group:stack-name:size-param-name
```

The OpenStack-heat cloudprovider implementation then takes care of updating the parameters of the Heat stack in order to resize the ResourceGroup, based on the already existing auto-scaler policies (unschedulable pods, etc...).

I'll be happy to contribute this if you are interested :)

The commit contains no vendored dependencies, to ease up the review.

P.S.: I realize the new [Senlin](https://wiki.openstack.org/wiki/Senlin) service is better suited to autoscaling, but it is quite new and there are still no golang SDKs for it. I'd be happy to introduce a Senlin autoscaler later on, once it becomes more wide-spread (as there are currently many OpenStack installations that support Heat but not Senlin).

@kenan435

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1980)

### DIFF
--- a/cluster-autoscaler/cloudprovider/openstack_heat/heat_cloud_provider.go
+++ b/cluster-autoscaler/cloudprovider/openstack_heat/heat_cloud_provider.go
@@ -1,0 +1,100 @@
+package openstack_heat
+
+import (
+	"errors"
+	"fmt"
+	"strconv"
+	"strings"
+
+	"k8s.io/contrib/cluster-autoscaler/cloudprovider"
+	kube_api "k8s.io/kubernetes/pkg/api"
+)
+
+type OpenstackHeatProvider struct {
+	heatManager    *HeatManager
+	resourceGroups []*HeatResourceGroup
+}
+
+func BuildOpenstackHeatCloudProvider(heatManager *HeatManager, specs []string) (*OpenstackHeatProvider, error) {
+	provider := &OpenstackHeatProvider{
+		heatManager:    heatManager,
+		resourceGroups: make([]*HeatResourceGroup, 0),
+	}
+	for _, spec := range specs {
+		if err := provider.addResourceGroup(spec); err != nil {
+			return nil, fmt.Errorf("could not register resource group with spec %s: %v", spec, err)
+		}
+	}
+	return provider, nil
+}
+
+func (provider *OpenstackHeatProvider) addResourceGroup(spec string) error {
+	rscGroup, err := provider.buildGroupFromSpec(spec)
+	if err != nil {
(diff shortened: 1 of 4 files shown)


davidopp:
@kubernetes/sig-openstack @kubernetes/autoscaling

mwielgus:
@taimir What is the status of this PR? Is it ready for the review or can it be updated in the near future? I'm not an expert in OpenStatck so I will have to look around for some person to actually review the PR.

taimir:
@mwielgus It's been a while since I opened this PR, at the time I thought it would be a good idea to put it out there and if someone considers it useful, maybe merge it. The code can be reviewed in its current state (that's also why I didn't push anything else like vendor, etc.) so that someone experienced with OpenStack could quickly skip over it and grasp the general idea.

Since the OpenStack [clustering](http://developer.openstack.org/api-ref/clustering/) service `Senlin` is now a thing, I'm not too sure if doing autoscaling with `Heat` (as in this PR) is desirable any more. Maybe it would still be an appropriate solution for some use cases, where `Senlin` is not available as a service? I would appreciate the opinion of someone experienced with OpenStack on this, that would decide whether this PR is still viable or whether it should be replaced by one for `Senlin`.

If there's some interest in the `Heat` approach I'll be happy to rebase it and prep it for merging.

xsgordon:
@taimir my understanding based on the available survey data from the OpenStack Foundation as of October 2016 ( https://www.openstack.org/analytics/ ) is that while Heat is now seeing broader deployment (54% of production OpenStack clouds, 61% of Testing/POC OpenStack Clouds) Senlin is not yet registering as very widely deployed.

This seems like a true reflection of what I see in the community but may also be a chicken and egg thing (e.g. was Senlin a valid response for this edition of the survey, it does not appear on the graphs at all).

taimir:
OK, I thought that might well be the case. I'll try to rebase this PR when I have some time later today. I remember there was one problem in the `Gophercloud` OpenStack client, which I filed as an [issue](https://github.com/gophercloud/gophercloud/issues/185). So that will need to be resolved eventually. Feel free to ping me if you need anything else to review.

ironhouzi:
Any recent progress with these efforts?

Issue #1956, opened 2016-10-31
Autoscaler on AWS should discover automatically the name of the ASG


Raffo:
In the current implementation of the AutoScaler for the AWS CloudProvider, we need to pass the name of the autoscaling group (ASG). There's no need for this as it is possible to determine the ASG name from the instance itself via the AWS API.

PR #1982, opened 2016-11-03
AWS Autoscaler autodiscover ASG names and sizes

Raffo:
This is just a possible idea on how to implement #1956. It's not yet enough generic and I would love to take the chance to discuss it. 

I implemented automatic discovery of autoscaling groups in case the CLI flag is not set.

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1982)


Raffo:
I signed it!

mwielgus:
cc: @andrewsykim @osxi

cluster-autoscaler/cloudprovider/aws/aws_cloud_provider_test.go
```diff
@@ -276,3 +300,8 @@ func TestBuildAsg(t *testing.T) {
 	assert.Equal(t, 222, asg.MaxSize())
 	assert.Equal(t, "test-name", asg.Name)
 }
+
+func TestAutoDiscoverASG(t *testing.T) {
+	provider := testProvider(t, testAwsManager)
+	provider.autoDiscoverASG()
```
andrewsykim: I don't see how this is much of a test, we should be asserting that `autoDiscoverASG` returns nil and that the appropriate ASGs were registered

cluster-autoscaler/cloudprovider/aws/aws_manager.go
```diff
@@ -48,12 +50,17 @@ type autoScaling interface {
 	TerminateInstanceInAutoScalingGroup(input *autoscaling.TerminateInstanceInAutoScalingGroupInput) (*autoscaling.TerminateInstanceInAutoScalingGroupOutput, error)
 }
 
+type kubeEC2 interface {
```
andrewsykim: Seems this interface is not directly relevant to kube and should be named `ec2`

cluster-autoscaler/cloudprovider/aws/aws_manager.go
```diff
@@ -48,12 +50,17 @@ type autoScaling interface {
...
+	DescribeInstances(input *ec2.DescribeInstancesInput) (*ec2.DescribeInstancesOutput, error)
+}
+
 // AwsManager is handles aws communication and data caching.
 type AwsManager struct {
 	asgs     []*asgInformation
 	asgCache map[AwsRef]*Asg
 
 	service    autoScaling
+	ec2Service kubeEC2
```
andrewsykim: Seems like a bit of a code smell that we named this service `ec2Service` but the autoscaling service is just named `service`

andrewsykim (changes requested):
Auto discovering ASGs in the cluster is probably a bit more involved than this. The source of truth is now defined via AWS and not the cluster autoscaler. With this setup, any changes to the ASG would mean the cluster autoscaler would get out of sync until the cluster autoscaler pod is restarted which can lead to confusion. If we do want the cluster autoscaler discovering ASGs automatically we would probably need to run some sync loop to make sure that the cluster autoscaler's configurations match that of the cloud provider. If this is a feature we do want, I think all cloud providers should support it eventually, @mwielgus probably has a better say in that.

cluster-autoscaler/cloudprovider/aws/aws_manager.go
Raffo: Agree.

cluster-autoscaler/cloudprovider/aws/aws_manager.go
Raffo: You're right and I though the same. I guess renaming `service` into something that makes more sense is the right thing to do.

Raffo:
@andrewsykim Sorry for the hacky code, I just wanted to show something to get early feedback, thanks for giving me it. You're right saying that AWS is the source of truth and it probably makes sense to have some sort of sync loop. The current implementation that only works with an argument has the same strong limitation, which is something we can address here. 
I think this feature should be implemented for all cloud providers, I just focused on AWS first. If you think this makes sense I will try to implement some sort of sync loop for this case.

andrewsykim:
@mwielgus do we want other cloud providers to auto discover node groups eventually?

Raffo:
@mwielgus Any feedback on this?

osxi:
The idea of automatically discovering the min/max size for an ASG got brought up, which makes sense since we probably want AWS to define that and not an rc/deployment definition (see https://github.com/kubernetes/contrib/blob/master/cluster-autoscaler/cloudprovider/aws/README.md#deployment-specification).

I'm thinking out loud but maybe automatically discovering the ASGs could possibly be an issue if you had an ASG that you didn't want autoscaled by cluster_autoscaler.

mwielgus:
Sorry for the delay. The PR completely fall of my radar.

I agree with @andrewsykim that at this moment CA can easily get out of sync with the reality as the config is set only at the beginning. So it needs to be updated more often, possibly with every loop.

Auto-discovery should be enabled with an additional flag so that the users don't have a surprise. No-config = no action. 

With the current CA implementation min > 0. Is this requirement also present in AWS?

It seems that min and max are obligatory on AWS. In GCE/GKE migs don't have min/max, only the target size so making it more generic may be problematic.

Raffo:
@mwielgus Thanks for your comments. Yes, min and max are required and min must be > 1. I will update the PR with a fix to periodically auto update, but I think this is a further enhancement as with the string flag we have today it could already be out of sync. 
I totally agree on the flag to configure the autoconfiguration, will work on that and update the PR.

cluster-autoscaler/Dockerfile
```diff
@@ -15,6 +15,9 @@
 FROM gcr.io/google_containers/ubuntu-slim:0.1
 MAINTAINER Marcin Wielgus "mwielgus@google.com"
 
+RUN apt-get update  
+RUN apt-get install -y ca-certificates
```
mumoshu: nit: To save dozens of MB from the resulting docker image size, I'd suggest replacing these 2 lines with something like:
```
RUN apt-get update -y && \
  apt-get install -y --no-install-recommends ca-certificates && \
  apt-get clean -y && \
  rm -rf /var/cache/apt/archives/* /var/lib/apt/lists/*
```

Raffo: Thanks, updated.

cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
```diff
@@ -21,38 +21,130 @@ import (
...
+				}
+
+				var nodesList []*string
+
+				for _, n := range nl.Items {
+					glog.V(4).Infof("Considering node: %s", n.Name)
+					nodesList = append(nodesList, &n.Name)
+				}
+				AutoDiscoverNodeGroup(aws, nodesList)
+				time.Sleep(30 * time.Second)
```
andrewsykim: Use ticker? http://stackoverflow.com/a/16466581

andrewsykim: Could it run in the same loop [here](https://github.com/kubernetes/contrib/blob/master/cluster-autoscaler/cluster_autoscaler.go#L227)

mwielgus:
@mumoshu @andrewsykim Does this PR look good for you now?

mwielgus:
@andrewsykim ping

andrewsykim:
Sorry I missed your first message, I'll take a look later today.

cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
```diff
@@ -21,38 +21,134 @@ import (
...
-		if err := aws.addNodeGroup(spec); err != nil {
-			return nil, err
+	if autoDiscoverNodeGroup {
+		go func() {
+			for {
+				select {
+				case <-time.After(scanInterval):
+					{
+						selector := "master!=true"
+						listAll := apiv1.ListOptions{LabelSelector: selector}
```
andrewsykim: I think this is misleading, this var is not actually a `listAll` option. Maybe name it `notMaster`?

cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
```diff
@@ -21,38 +21,134 @@ import (
...
+							glog.Errorf("err: %s\n", err)
+						}
+
+						var nodesList []*string
+
+						for _, n := range nl.Items {
+							glog.V(4).Infof("Considering node: %s", n.Name)
+							nodesList = append(nodesList, &n.Name)
+						}
+						AutoDiscoverNodeGroup(aws, nodesList)
```
andrewsykim: This go routine doesn't do any error handling which concerns me a bit. What happens if AutoDiscoverNodeGroup returns an error? Do we just log the error and hope that on the next syncloop it would pass? If we continually fail to discover new ASGs I feel like we have to fail loudly in some way. What do you think?

cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
```diff
@@ -21,38 +21,134 @@ import (
...
+		go func() {
+			for {
+				select {
+				case <-time.After(scanInterval):
+					{
+						selector := "master!=true"
+						listAll := apiv1.ListOptions{LabelSelector: selector}
+						nl, err := kubeClient.Core().Nodes().List(listAll)
+						if err != nil {
+							glog.Errorf("err: %s\n", err)
```
andrewsykim: No point in adding `err:` when we're logging an error, maybe something like this might be better:
```
`glog.Errorf("Failed to list nodes from api server: %v", err)`
```

cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
```diff
@@ -21,38 +21,134 @@ import (
...
+		Filters: []*ec2.Filter{
+			{
+				Name:   aws_util.String("private-dns-name"),
+				Values: nodesList,
+			},
+		},
+	}
+	instances, err := aws.awsManager.ec2Service.DescribeInstances(params)
+	if err != nil {
+		glog.Errorf("Error describing EC2 instances: %s\n", err)
```
andrewsykim: `glog.Errorf("Error describing EC2 instances: %v", err)`

cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
```diff
@@ -21,38 +21,134 @@ import (
...
+		return err
+	}
+
+	//make a "uniq" on the names so that we don't consider autoscaling groups twice
+	asgs := make(map[string]bool, 0)
+	names := make([]*string, 0)
```
andrewsykim: Can't we construct `names` from the keys from `asgs`?

cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
```diff
@@ -21,38 +21,134 @@ import (
...
+		},
+	}
+	instances, err := aws.awsManager.ec2Service.DescribeInstances(params)
+	if err != nil {
+		glog.Errorf("Error describing EC2 instances: %s\n", err)
+		return err
+	}
+
+	//make a "uniq" on the names so that we don't consider autoscaling groups twice
+	asgs := make(map[string]bool, 0)
```
andrewsykim: Can we name this var something more meaningful?

cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
```diff
@@ -21,38 +21,134 @@ import (
...
+	names := make([]*string, 0)
+	for idx := range instances.Reservations {
```
andrewsykim: Seems cleaner to do this:
```go
for _, r := range instances.Reservations {
    for _, inst := range r.Instances {
```

cluster-autoscaler/cluster_autoscaler.go
```diff
@@ -113,6 +113,7 @@ var (
 	AvailableExpanders = []string{RandomExpanderName, MostPodsExpanderName, LeastWasteExpanderName}
 	expanderFlag       = flag.String("expander", RandomExpanderName,
 		"Type of node group expander to be used in scale up. Available values: ["+strings.Join(AvailableExpanders, ",")+"]")
+	autoDiscoverNodeGroup = flag.Bool("auto-discover-node-group", false, "Enables auto discovery of node groups. Currently supported only in AWS.")
```
andrewsykim: We should fail fast or a log some sort of warning/error if someone has this flag enabled but not on AWS.

cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
```diff
@@ -21,38 +21,134 @@ import (
...
+	}
+	asgGroups, err := aws.awsManager.autoscalingService.DescribeAutoScalingGroups(request)
+	if err != nil {
+		glog.Errorf("Error describing autoscaling groups: %s\n", err)
+		return err
+	}
+
+	aws.asgMutex.Lock()
+	defer aws.asgMutex.Unlock()
+	aws.asgs = aws.asgs[:0]
```
andrewsykim: Why do we need this if `ClearAsgs()` is already doing that?

cluster-autoscaler/Dockerfile
```diff
@@ -15,6 +15,11 @@
 FROM gcr.io/google_containers/ubuntu-slim:0.1
 MAINTAINER Marcin Wielgus "mwielgus@google.com"
 
+RUN apt-get update -y && \
```
andrewsykim: Can we remove the ca cert we have in this [example](https://github.com/kubernetes/contrib/blob/master/cluster-autoscaler/cloudprovider/aws/README.md#deployment-specification) since we no longer need it.

andrewsykim (changes requested):

mwielgus:
@Raffo could you please respond to the unanswered comments from @andrewsykim?

Raffo:
@mwielgus awesome, I had the comment pending and didn't submit them, I'll still unfamiliar with this  "review" functionality from GitHub, I'm deeply sorry.

mwielgus:
@Raffo The code doesn't compile now. We updated the dependencies and there is no release_1_5 client in Kubernetes codebase anymore. Sorry for inconvenience.

Raffo:
@mwielgus No problem at all! I update it again now, but there are still some comments that are unanswered. Maybe @andrewsykim can have a look?

Raffo:
@andrewsykim can you have a look once again? It would be great to get this done.

andrewsykim:
@Raffo sorry I missed your last message, I'll try to get a review in today.

andrewsykim:
@Raffo sorry this fell out of my radar last week, haven't forgotten about it, trying to get to it!

cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
andrewsykim: @Raffo the approach I'm leaning towards is running this sync code in the main cluster autoscaler loop (somewhere in [here](https://github.com/kubernetes/contrib/blob/master/cluster-autoscaler/cluster_autoscaler.go#L277-L301)) and skipping autoscaling operations if the sync fails, what do you think?

andrewsykim: That way we can do proper error handling of the ASG discovery. We're also using the same scanInterval so maybe goroutine doesn't make sense anymore (sorry, I know I suggested it in the first place)

cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
andrewsykim: Ahh I see, that is odd, I would ask if you could possibly fix that but it seems like the other cloud providers do that too. 

GCE: https://github.com/kubernetes/contrib/blob/master/cluster-autoscaler/cloudprovider/gce/gce_manager.go#L48 and https://github.com/kubernetes/contrib/blob/master/cluster-autoscaler/cloudprovider/gce/gce_cloud_provider.go#L31

andrewsykim: but you could still combine into the same function right? i.e. aws.ClearAsgs() can access aws.awsManager

cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
```diff
@@ -255,7 +354,7 @@ func buildAsg(value string, awsManager *AwsManager) (*Asg, error) {
 	}
 
 	if tokens[2] == "" {
-		return nil, fmt.Errorf("asg name must not be blank: %s got error: %v", tokens[2])
```
andrewsykim: 👍

cluster-autoscaler/cloudprovider/aws/aws_cloud_provider_test.go
```diff
@@ -30,11 +32,36 @@ type AutoScalingMock struct {
...
+}
+
+func (e *EC2Mock) DescribeInstances(input *ec2.DescribeInstancesInput) (*ec2.DescribeInstancesOutput, error) {
+	return &ec2.DescribeInstancesOutput{
+		Reservations: []*ec2.Reservation{
+			{Instances: []*ec2.Instance{
+				{
+					InstanceId: aws.String("test-instance-id"),
+					Tags: []*ec2.Tag{{
+						Key:   aws.String("aws:autoscaling:groupNam"),
```
andrewsykim: `s/aws:autoscaling:groupNam/aws:autoscaling:groupName`

cluster-autoscaler/cloudprovider/aws/aws_cloud_provider_test.go
```diff
@@ -276,3 +306,15 @@ func TestBuildAsg(t *testing.T) {
 	assert.Equal(t, 222, asg.MaxSize())
 	assert.Equal(t, "test-name", asg.Name)
 }
+
+func TestAutoDiscoverASG(t *testing.T) {
+	provider := testProvider(t, testAwsManager)
+	node := "aws:///us-east-1a/test-instance-i"
+	nl := []*string{&node}
+
+	err := AutoDiscoverNodeGroup(provider, nl)
```
andrewsykim: Can we add a test case where we run `AutoDiscoverNodeGroup` with the mock autoscaling API returning 2 ASGs on the second time?

cluster-autoscaler/cluster_autoscaler.go
```diff
@@ -472,6 +474,10 @@ func main() {
 		"Can be used multiple times. Format: <min>:<max>:<other...>")
 	kube_flag.InitFlags()
 
+	if *autoDiscoverNodeGroup == true && *cloudProviderFlag != "aws" {
```
andrewsykim: `s/*autoDiscoverNodeGroup == true/`*autoDiscoverNodeGroup`

cluster-autoscaler/cluster_autoscaler.go
```diff
@@ -472,6 +474,10 @@ func main() {
...
+		glog.Fatalf("Cannot turn on node group autodiscovery for cloud provider: %s", cloudProviderFlag)
```
andrewsykim: `s/cloudProviderFlag/*cloudProviderFlag`

cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
Raffo: Well this is exactly what I planned to do in the beginning, so I agree :D I also missed the comment and implemented a test for the goroutine now, but I can drop it of course.

cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
Raffo: Yes, I can fix that.

Raffo:
@andrewsykim I lost your comments above and in the meantime I already added another test with reference to the one you asked for which is slightly more complicated and I had to touch some other part of the tests. It looks better to me know as it tests that the goroutine actually works, but now I'm not sure that we need/want this goroutine anymore. 
Can we do a recap and say how we want this to work and have a list of all the changes that are needed. The PR has been open since the 3rd of Nov and it would be great to summarize the actions and do them or close the PR instead.

cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
andrewsykim: The plan in the beginning if I recall correctly was to discover ASGs only once. The go routine suggestion was mainly to ensure that the cluster-autoscaler state was in sync with the cloud provider state. Given that we absolutely need to escalate errors when we fail to sync cloud provider state we should probably put this back into the main sync loop. Again, sorry that I'm stepping back from my original suggestion, I didn't think implementing a go routine for this would be this complicated.

andrewsykim:
@Raffo I left a comment re: goroutine for ASG state [here](https://github.com/kubernetes/contrib/pull/1982#discussion_r102071891). Once that's addressed I think this PR is good to go. You'll need approval from @mwielgus to get this merged though.

cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
Raffo: No problem at all! I will implement the change then, thanks for the quick answer (and sorry for my late reply).

Raffo: The only problem of putting the code in the main sync loop and why I agreed originally in having the goroutine is that the sync loop is totally not specific for the single cloud provider while this functionality is only implemented for aws. An `if` condition will thus be very ugly, but I don't see other options. I could probably just call a function that is no-op for other cloud providers.

cw-kuoka: @Raffo The `Autoscaler` interface is recently introduced to the code-base. Would it be possible to utilize it here so that we can achieve it without goroutines?
https://github.com/kubernetes/contrib/blob/master/cluster-autoscaler/core/autoscaler.go#L38

Just a rough idea but add an `Autoscaler` implementation. The implementation, when `currentTime > lastFetchTime + delay`, fetches autoscaling groups from AWS and update node groups like we do for [DynamicAutoscaler](https://github.com/kubernetes/contrib/blob/master/cluster-autoscaler/core/dynamic_autoscaler.go#L125).

mumoshu: @Raffo The `Autoscaler` interface is recently introduced to the code-base. Would it be possible to utilize it here so that we can achieve it without goroutines?
https://github.com/kubernetes/contrib/blob/master/cluster-autoscaler/core/autoscaler.go#L38

Just a rough idea but add an `Autoscaler` implementation. The implementation, when `currentTime > lastFetchTime + delay`, fetches autoscaling groups from AWS and update node groups like we do for [DynamicAutoscaler](https://github.com/kubernetes/contrib/blob/master/cluster-autoscaler/core/dynamic_autoscaler.go#L125).

mumoshu:
Hi @Raffo @andrewsykim, thanks for your great work here!
I'm really looking forward to this.
Btw my two cents https://github.com/kubernetes/contrib/pull/1982#discussion_r103338220

cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
Raffo: @mumoshu Sorry for the late reply. I'm not sure this is the best idea here. I saw the changes in code, but why not just fetching the updated information somewhere in this function? https://github.com/kubernetes/contrib/blob/master/cluster-autoscaler/core/static_autoscaler.go#L116 
The problem is that this will be pretty much specific for AWS once again and that it would be mostly a no-op for other cloud providers.

mumoshu: Hi @Raffo, thanks for the reply and good question!
Yes, it will be no-op for cloud providers other than AWS when implemented in static_autoscaler.go.

That's why I'd like to add a brand-new Autoscaler implementation specific to autodiscover ASGs. Doing so would enable cluster-autoscaler to run AWS-specific code(=ASG auto-discovery in this case) if and only if it is required(=cloud-provider is AWS and the autodiscovery is enabled).

Actually, we're already doing the same thing for static vs dynamic config.

Here:

```golang
func NewAutoscaler(opts AutoscalerOptions, predicateChecker *simulator.PredicateChecker, kubeClient kube_client.Interface, kubeEventRecorder kube_record.EventRecorder, listerRegistry kube_util.ListerRegistry) Autoscaler {
	autoscalerBuilder := NewAutoscalerBuilder(opts.AutoscalingOptions, predicateChecker, kubeClient, kubeEventRecorder, listerRegistry)
	if opts.ConfigMapName != "" {
		configFetcher := dynamic.NewConfigFetcher(opts.ConfigFetcherOptions, kubeClient, kubeEventRecorder)
		return NewDynamicAutoscaler(autoscalerBuilder, configFetcher)
	}
	return autoscalerBuilder.Build()
}
```

If `--configmap` is omitted, `DynamicAutoscaler` is not even instantiated hence code specific to dynamic reconfiguration(#2181) is not executed run at all.
I thought this is preferred over adding a mostly no-op if-else condition to the core of cluster-autoscaler(which is now in static_autoscaler.go).

Raffo: Ok, I get it now, it makes totally sense. I'll have a look at the code later next week (it's KubeCon week, let's see how much time I'll have) and update the PR!

Raffo:
Closing this in favor of #2549

PR #1989, opened 2016-11-04
Cluster-autoscaler: restrict one-time cluster scale up to 100

mwielgus:
So that a small mistake doesn't result in requesting 700 nodes.

cc: @piosz @fgrzadkowski @jsz

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1989)


fgrzadkowski:
I have two concerns. First of all I have a feeling that instead of having a plan for a proper fix we keep adding temporary hacky solutions. I'm worried that it will be more and more complicated to explain how CA works. Secondly I don't see how this PR would help. User will request 700 VMs anyway, just in 7 batches. What am I missing?

fgrzadkowski:
As discussed offline we should not introduce this workaround.

PR #1992, opened 2016-11-05
Cluster-autoscaler: delete empty nodes in bulk

mwielgus:
Regular nodes are deleted one by one. However, if a nodes are completely empty (only daemonsets or manifest-run pods) then they can easily be deleted in bulk. With this PR in scale down we will first try to delete empty nodes in bulk and if that fails, we will proceed with regular node deletion.

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1992)

### DIFF
--- a/cluster-autoscaler/scale_down.go
+++ b/cluster-autoscaler/scale_down.go
@@ -120,7 +120,8 @@ func ScaleDown(
 	predicateChecker *simulator.PredicateChecker,
 	oldHints map[string]string,
 	usageTracker *simulator.UsageTracker,
-	recorder kube_record.EventRecorder) (ScaleDownResult, error) {
+	recorder kube_record.EventRecorder,
+	maxEmptyBulkDelete int) (ScaleDownResult, error) {
 
 	now := time.Now()
 	candidates := make([]*kube_api.Node, 0)
@@ -163,6 +164,35 @@ func ScaleDown(
 		return ScaleDownNoUnneeded, nil
 	}
 
+	emptyNodes := simulator.FindEmptyNodesToRemove(candidates, pods)
+	if len(emptyNodes) > 0 {
+		limit := maxEmptyBulkDelete
+		if len(emptyNodes) < limit {
+			limit = len(emptyNodes)
+		}
+		emptyNodes = emptyNodes[:limit]
+		confirmation := make(chan error, len(emptyNodes))
+		for _, node := range emptyNodes {
+			glog.V(0).Infof("Scale-down: removing empty node %s", node.Name)
+			simulator.RemoveNodeFromTracker(usageTracker, node.Name, unneededNodes)
+			go func(nodeToDelete *kube_api.Node) {
+				confirmation <- deleteNodeFromCloudProvider(nodeToDelete, cloudProvider, recorder)
+			}(node)
+		}
+		var finalError error
+		for range emptyNodes {
+			if err := <-confirmation; err != nil {
+				glog.Errorf("Problem with empty node deletion: %v", err)
+				finalError = err
+			}
+		}
+		if finalError == nil {
+			return ScaleDownNodeDeleted, nil
+		} else {
+			return ScaleDownError, fmt.Errorf("failed to delete at least one empty node: %v", finalError)
+		}
+	}
+
 	// We look for only 1 node so new hints may be incomplete.
 	nodesToRemove, _, err := simulator.FindNodesToRemove(candidates, nodes, pods, client, predicateChecker, 1, false,
 		oldHints, usageTracker, time.Now())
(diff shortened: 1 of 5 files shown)


cluster-autoscaler/scale_down.go
```diff
@@ -163,6 +164,31 @@ func ScaleDown(
 		return ScaleDownNoUnneeded, nil
 	}
 
+	emptyNodes := filterEmptyNodes(simulator.FindEmptyNodesToRemove(candidates, pods), maxEmptyBulkDelete, cloudProvider)
+	if len(emptyNodes) > 0 {
+
```
fgrzadkowski: delete empty line

cluster-autoscaler/scale_down.go
```diff
@@ -163,6 +164,31 @@ func ScaleDown(
 		return ScaleDownNoUnneeded, nil
 	}
 
+	emptyNodes := filterEmptyNodes(simulator.FindEmptyNodesToRemove(candidates, pods), maxEmptyBulkDelete, cloudProvider)
```
fgrzadkowski: Can you please add a comment that will explain that in this section you try to delete empty nodes in bulks, and if this fails then you will try to delete 1 non-empty node? A comment in this next section would also help.

cluster-autoscaler/scale_down.go
```diff
@@ -163,6 +164,31 @@ func ScaleDown(
...
+			simulator.RemoveNodeFromTracker(usageTracker, node.Name, unneededNodes)
+			go func(nodeToDelete *kube_api.Node) {
+				confirmation <- deleteNodeFromCloudProvider(nodeToDelete, cloudProvider, recorder)
+			}(node)
+		}
+		var finalError error
+		for _ = range emptyNodes {
+			if err := <-confirmation; err != nil {
+				glog.Errorf("Problem with empty node deletion: %v", err)
+				finalError = err
```
fgrzadkowski: Should you combine somehow those errors?

cluster-autoscaler/scale_down.go
```diff
@@ -163,6 +164,31 @@ func ScaleDown(
...
+			}
+		}
+		if finalError == nil {
+			return ScaleDownNodeDeleted, nil
+		} else {
+			return ScaleDownError, fmt.Errorf("failed to delete at least one empty node: %v", finalError)
```
fgrzadkowski: This will only delete only for the last node that failed. Shouldn't we combine somehow the output?

cluster-autoscaler/scale_down.go
```diff
@@ -183,23 +209,64 @@ func ScaleDown(
...
+	err = deleteNodeFromCloudProvider(toRemove.Node, cloudProvider, recorder)
 	if err != nil {
 		return ScaleDownError, fmt.Errorf("Failed to delete %s: %v", toRemove.Node.Name, err)
 	}
+	return ScaleDownNodeDeleted, nil
+}
 
-	recorder.Eventf(toRemove.Node, kube_api.EventTypeNormal, "ScaleDown",
-		"node removed by cluster autoscaler")
+func filterEmptyNodes(nodes []*kube_api.Node, maxEmptyBulkDelete int, cloudProvider cloudprovider.CloudProvider) []*kube_api.Node {
```
fgrzadkowski: Arg name `maxEmptyBulkDelete` doesn't make much sense in this context. AFAIU it will basically limit how many empty nodes you will return. Please rename and add a comment explaining what this function does/returns.

fgrzadkowski: Only after reading the code I understood that this function assumes that `nodes` are already empty nodes and it will only choose those can be deleted (due to `minSize`). Comment is really needed.

mwielgus: Done.

cluster-autoscaler/scale_down.go
mwielgus: Done.

cluster-autoscaler/scale_down.go
mwielgus: The result is only important from the timing perspective. If we failed to delete some node we wait a bit more when trying to delete something more.

cluster-autoscaler/scale_down.go
mwielgus: Done.

cluster-autoscaler/scale_down.go
fgrzadkowski: (my first comment looks really hard to understand...)

Should we then just return error that one of them failed if we don't care about that message inside? I think it'd be cleaner.

fgrzadkowski:
LGTM. Please take a look at the last comment that's unresolved and add lgtm label when you consider it (either by changing error or by answering why not :) )

cluster-autoscaler/scale_down.go
mwielgus: Still don't understand. We return an error if at least one error occurred, but we wait for all statuses to arrive so we don't have unfinished jobs.

mwielgus:
Added comment. Applying LGTM.

PR #1999, opened 2016-11-07
Cluster-autoscaler: use binpacking estimation algorithm by default

mwielgus:
Should actually be better than basic for most of the users:
* If only 1 type of pod is pending CA is guaranteed to provide the best estimate.
* Only 1 iteration is needed to grow cluster to correct state.
* It should not be much worse than multi-phase iterative algorithm for mixed pending pods.

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/1999)


fgrzadkowski:
LGTM

PR #2008, opened 2016-11-08
mungegithub: Fixes #422

### LINKED ISSUE #422: Possible for submit queue to merge a sha other than what was tested
There's a race condition in the submit queue where the queue may merge a sha other than what was tested:
For example: kubernetes/kubernetes#19716 : this PR was 5/5 checks green, then needed rebase after the merge bot had called for a final test.  I rebased, repushed, and re-added the LGTM label, and the PR merged even though 1/5 checks was green (presumably CLA check) in github for that sha.

@eparis

apelisse:
At least attempts to fix one or two of the problems in #422.

Refresh the pull-request after we've done the tests to:
- Make sure it is still mergeable (doesn't have do-no-merge label or
still has lgtm label)
- The SHA hasn't changed, so that we don't commit something we haven't
tested

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/2008)

### DIFF
--- a/mungegithub/mungers/submit-queue.go
+++ b/mungegithub/mungers/submit-queue.go
@@ -798,6 +798,7 @@ const (
 	ghE2ERunning            = "Running github e2e tests a second time."
 	ghE2EFailed             = "Second github e2e run failed."
 	unmergeableMilestone    = "Milestone is for a future release and cannot be merged"
+	headCommitChanged       = "This PR has changed since we ran the tests"
 )
 
 func getEarliestApprovedTime(obj *github.MungeObject) *time.Time {
@@ -1157,6 +1158,12 @@ func (sq *SubmitQueue) doGithubE2EAndMerge(obj *github.MungeObject) bool {
 		return true
 	}
 
+	sha, _, ok := obj.GetHeadAndBase()
+	if !ok {
+		glog.Errorf("%d: Unable to get SHA", *obj.Issue.Number)
+		sq.SetMergeStatus(obj, unknown)
+		return true
+	}
 	if interruptedObj != nil {
 		if interruptedObj.hasSHAChanged() {
 			// This PR will have to be rested.
@@ -1165,8 +1172,33 @@ func (sq *SubmitQueue) doGithubE2EAndMerge(obj *github.MungeObject) bool {
 		}
 		glog.Infof("Skipping retest since head and base sha match previous attempt!")
 		atomic.AddInt32(&sq.retestsAvoided, 1)
-	} else if sq.retestPR(obj) {
+	} else {
+		if sq.retestPR(obj) {
+			return true
+		}
+
+		err := obj.Refresh()
+		if err != nil {
+			glog.Errorf("%d: unknown err: %v", *obj.Issue.Number, err)
+			sq.SetMergeStatus(obj, unknown)
+			return true
+		}
+	}
+
+	// We shouldn't merge if it's not valid anymore
+	if !sq.validForMerge(obj) {
+		glog.Errorf("%d: Not mergeable anymore. Do not merge.", *obj.Issue.Number)
 		return true
+	}
+
+	if newSha, _, ok := obj.GetHeadAndBase(); !ok {
+		glog.Errorf("%d: Unable to get SHA", *obj.Issue.Number)
+		sq.SetMergeStatus(obj, unknown)
+		return true
+	} else if newSha != sha {
+		glog.Errorf("%d: Changed while running the test. Do not merge.", *obj.Issue.Number)
+		sq.SetMergeStatus(obj, headCommitChanged)
+		return false
 	}
 
 	if !sq.e2eStable(true) {


mungegithub/mungers/submit-queue.go
```diff
@@ -1165,7 +1172,32 @@ func (sq *SubmitQueue) doGithubE2EAndMerge(obj *github.MungeObject) bool {
...
+		if err != nil {
+			glog.Errorf("%d: unknown err: %v", *obj.Issue.Number, err)
+			sq.SetMergeStatus(obj, unknown)
+			return true
+		}
+	}
+
+	// We shouldn't merge if it's not valid anymore
+	if !sq.validForMerge(obj) {
+		glog.Errorf("%d: Not mergeable anymore. Do not merge.", *obj.Issue.Number)
```
rmmh: we should probably do sq.setMergeStatus here with the "sha changed", and return false to do a retest.

mungegithub/mungers/submit-queue.go
```diff
@@ -1165,7 +1172,32 @@ func (sq *SubmitQueue) doGithubE2EAndMerge(obj *github.MungeObject) bool {
...
+	}
+
+	if newSha, _, ok := obj.GetHeadAndBase(); !ok {
+		glog.Errorf("%d: Unable to get SHA", *obj.Issue.Number)
+		sq.SetMergeStatus(obj, unknown)
+		return true
+	} else if newSha != sha {
+		glog.Errorf("%d: Changed while running the test. Do not merge.", *obj.Issue.Number)
+		sq.SetMergeStatus(obj, headCommitChanged)
 		return true
```
rmmh: this should return false to do a retest, maybe?

mungegithub/mungers/submit-queue.go
apelisse: `sq.validForMerge()` does it for us

mungegithub/mungers/submit-queue.go
apelisse: Good point

apelisse:
Updated @rmmh

rmmh:
/lgtm

Do the unit tests exercise this path? How disgusting would it be to test this functionality?

apelisse:
Yes, unit-tests are going there (I also run them successfully). I agree it'd be nice to be able to test that more thorougly.

apelisse:
Replaced tabs with space for alignment. Auto-lgtming ...

apelisse:
No, but I can add the label myself

mungegithub/mungers/submit-queue.go
eparis: Why is this a glog.Errorf() ?

mungegithub/mungers/submit-queue.go
```diff
@@ -1165,10 +1172,35 @@ func (sq *SubmitQueue) doGithubE2EAndMerge(obj *github.MungeObject) bool {
...
+	}
+
+	// We shouldn't merge if it's not valid anymore
+	if !sq.validForMerge(obj) {
+		glog.Errorf("%d: Not mergeable anymore. Do not merge.", *obj.Issue.Number)
 		return true
 	}
 
+	if newSha, _, ok := obj.GetHeadAndBase(); !ok {
+		glog.Errorf("%d: Unable to get SHA", *obj.Issue.Number)
```
eparis: Do we need to spam the logs? shouldn't we have already done that in github.go?

mungegithub/mungers/submit-queue.go
```diff
@@ -1165,10 +1172,35 @@ func (sq *SubmitQueue) doGithubE2EAndMerge(obj *github.MungeObject) bool {
...
+		sq.SetMergeStatus(obj, unknown)
+		return true
+	} else if newSha != sha {
+		glog.Errorf("%d: Changed while running the test. Do not merge.", *obj.Issue.Number)
```
eparis: Maybe this belongs in the logs, but its certainly not an Errorf...

mungegithub/mungers/submit-queue.go
```diff
@@ -1165,10 +1172,35 @@ func (sq *SubmitQueue) doGithubE2EAndMerge(obj *github.MungeObject) bool {
...
+		sq.SetMergeStatus(obj, headCommitChanged)
+		return false
+	}
+
 	if !sq.e2eStable(true) {
 		if sq.validForMerge(obj) {
```
eparis: does this still serve a purpose?

eparis (commented):
Anyone have time to clean this up?

eparis:
Also, this doesn't actually fix the problem...  It makes the race smaller, but it isn't fixed...

fixing it requires extending go-github. I actually have an open PR for that...  I'll try to get back to it next week.

I also didn't see an explanation why there was no unit test added for this case. It's certainly possible to do so...

apelisse:
I should have paid more attention to log verbosity. I will address that now.

Adding link to pull-request for the record: https://github.com/google/go-github/pull/456

OK So I will remove the part that checks that the sha hasn't changed, but we should probably keep the refresh and check for mergeability (we should still not merge if the lgtm is gone, or do-not-merge label has appeared).

mungegithub/mungers/submit-queue.go
apelisse: Removing this

mungegithub/mungers/submit-queue.go
apelisse: Removing this

mungegithub/mungers/submit-queue.go
apelisse: removed the condition

Issue #2010, opened 2016-11-09
[Ingress] 0.8.4 Release?


r2d4:
I understand ingress is being moved to kubernetes/ingress so if issues are being move there feel free to close this and I'll reopen there.  

Could we push a 0.8.4 release of the nginx-ingress controller?  The last one was released awhile ago

```
$ gcloud alpha container images list-tags gcr.io/google_containers/nginx-ingress-controller

820c338dc22e  0.8.3  2016-08-22T11:59:22
```

aledbf:
@r2d4 I'm sorry for the delay. As you realized Ingress is being moved to `kubernetes/ingress`.
The reason for the delay is the refactoring done to the controller (temporal place https://github.com/aledbf/ingress-controller)

whereisaaron:
Could the already merged changes here on `contrib` perhaps be pushed out as an 0.8.4.alpha1 or alternatively an alpha `kubernetes/ingress` release?

It seems like lots of important small improvements and fixes here over the last 6 months are neither being used nor tested while the refactoring project happens. If we had even an untested alpha or rc release from either `contrib` or `ingress` then we could contribute to testing and share some of that work.

aledbf:
> It seems like lots of important small improvements and fixes here over the last 6 months are neither being used nor tested while the refactoring project happens

That's not correct. Here's is where I put the temporal images when I close a PR
 https://quay.io/repository/aledbf/nginx-ingress-controller?tab=tags

0.55 contains the refactoring from the PR https://github.com/kubernetes/ingress/pull/95

whereisaaron:
> Here's is where I put the temporal images when I close a PR
https://quay.io/repository/aledbf/nginx-ingress-controller?tab=tags
0.55 contains the refactoring from the PR kubernetes/ingress#95

Thanks @aledbf that's great news they are builds we can test! And thanks for the 0.55 reference. Is there a way to map between those images and the PR that created them? Or the other direction, from the PR -> Travis CI -> image?

aledbf:
>  And thanks for the 0.55 reference. Is there a way to map between those images and the PR that created them? 

Not yet. I've been working in that too but still is a work in progress.
https://github.com/kubernetes/ingress/tree/master/hack/e2e-internal

r2d4:
Closing this, there are now pre-alpha releases in the kubernetes/ingress repo https://github.com/kubernetes/ingress/releases

PR #2015, opened 2016-11-10
Upgrade ingress controller image to 0.8.4

r2d4:
ref https://github.com/kubernetes/minikube/issues/611#issuecomment-259756556
ref https://github.com/kubernetes/contrib/issues/2010

cc @bprashanth

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/2015)

### DIFF
--- a/ingress/controllers/nginx/Makefile
+++ b/ingress/controllers/nginx/Makefile
@@ -1,7 +1,7 @@
 all: push
 
 # 0.0 shouldn't clobber any release builds
-TAG = 0.8.3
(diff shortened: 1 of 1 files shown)


r2d4:
Talked to @aledbf on slack - looks like ingress is being moved over to k8s/ingress right now and that makes a release difficult.

I guess we'll go ahead and either build it ourselves based on the k8s/contrib/ingress code for minikube or wait until a release of the newly refactored k8s/ingress code.

PR #2021, opened 2016-11-11
Add a github munger that knows how to map files (or dirs) to PRs

brendandburns:
@eparis @philips

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/2021)

### DIFF
--- a/mungegithub/github/github.go
+++ b/mungegithub/github/github.go
@@ -528,6 +528,15 @@ func (config *Config) GetObject(num int) (*MungeObject, error) {
 	return obj, nil
 }
 
+// GetFilesForPR gets the files associated with a particular PR
+func (config *Config) GetFilesForPR(num int) ([]*github.CommitFile, error) {
+	files, _, err := config.client.PullRequests.ListFiles(config.Org, config.Project, num, &github.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+	return files, nil
+}
+
 // NewIssue will file a new issue and return an object for it.
 // If "owner" is not empty, the issue will be assigned to "owner".
 func (config *Config) NewIssue(title, body string, labels []string, owner string) (*MungeObject, error) {
--- a/mungegithub/mungers/file-to-pr-map.go
+++ b/mungegithub/mungers/file-to-pr-map.go
@@ -1,0 +1,116 @@
+/*
+Copyright 2016 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package mungers
+
+import (
+	"encoding/json"
+	"net/http"
+	"sync"
+
+	"k8s.io/contrib/mungegithub/features"
+	"k8s.io/contrib/mungegithub/github"
+
+	"github.com/golang/glog"
+	"github.com/spf13/cobra"
+)
+
+var _ Munger = &FileToPRMap{}
+
+func init() {
+	RegisterMungerOrDie(&FileToPRMap{})
+}
+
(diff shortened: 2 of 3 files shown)


brendandburns:
I signed it!

mungegithub/github/github.go
```diff
@@ -528,6 +528,15 @@ func (config *Config) GetObject(num int) (*MungeObject, error) {
 	return obj, nil
 }
 
+// GetFilesForPR gets the files associated with a particular PR
+func (config *Config) GetFilesForPR(num int) ([]*github.CommitFile, error) {
+	files, _, err := config.client.PullRequests.ListFiles(config.Org, config.Project, num, &github.ListOptions{})
```
apelisse: Already exists: https://github.com/kubernetes/contrib/blob/master/mungegithub/github/github.go#L1284-L1318

mungegithub/mungers/file-to-pr-map.go
```diff
@@ -0,0 +1,137 @@
...
+
+// ServePRs serves the current PR list over HTTP
+func (f *FileToPRMap) ServePRs(res http.ResponseWriter, req *http.Request) {
+	prefix := req.URL.Query().Get("prefix")
+	f.lock.Lock()
+	defer f.lock.Unlock()
+	output := map[string][]int{}
+	for file, prNum := range f.fileToPR {
+		if strings.HasPrefix(file, prefix) {
+			output[file] = append(output[file], prNum...)
```
apelisse: Why do you need to append here? When would you already have something in `output[file]`?

mungegithub/mungers/file-to-pr-map.go
```diff
@@ -0,0 +1,137 @@
...
+	files, err := f.config.GetFilesForPR(*pr.Number)
+	if err != nil {
+		glog.Errorf("unexpected error getting files: %v", err)
+		return
+	}
+	f.updateFiles(files, *pr.Number)
+}
+
+func (f *FileToPRMap) updateFiles(files []*github_api.CommitFile, prNumber int) {
+	f.lock.Lock()
```
apelisse: Maybe you can clarify, but my understanding is that `lock` protects `fileToPR` only, not `nextFileToPR`.

mungegithub/mungers/file-to-pr-map.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+}
+
+// AddFlags implements the Munger interface
+func (f *FileToPRMap) AddFlags(cmd *cobra.Command, config *github.Config) {
+}
+
+// Name implements the Munger interface
+func (f *FileToPRMap) Name() string {
+	return "bulk-lgtm"
```
apelisse: Out of curiosity, does this PR belong to a bigger picture? Would you mind sharing what you plan is? :-)

I'm saying that because the name of the munger doesn't seem to match what it does

mungegithub/mungers/file-to-pr-map.go
```diff
@@ -0,0 +1,137 @@
...
+func (f *FileToPRMap) Munge(obj *github.MungeObject) {
+	if !obj.IsPR() {
+		return
+	}
+	pr, err := obj.GetPR()
+	if err != nil {
+		glog.Errorf("Unexpected error getting PR: %v", err)
+		return
+	}
+	glog.V(4).Infof("Found a PR: %#v", *pr)
```
apelisse: nit: I don't think this is super useful

mungegithub/mungers/file-to-pr-map.go
```diff
@@ -0,0 +1,137 @@
...
+var _ Munger = &FileToPRMap{}
+
+func init() {
+	RegisterMungerOrDie(&FileToPRMap{})
+}
+
+// FileToPRMap maintains a map from files (and directories) to a list of PRs
+type FileToPRMap struct {
+	config *github.Config
+	// protects the map
```
apelisse: See comment below, but this is what needs to be clarified: which map?

apelisse (changes requested):

eparis:
also, are we supposed to be taking these thing here, or over in test-infra now? I think test-infra. Who did that move? We should delete this code...

lavalamp:
Yes, it's moved, yes we should delete this code now. I think @apelisse did
the move.

foxish:
I had created https://github.com/kubernetes/contrib/pull/2335 to delete the mungegithub bot and add a redirect to its new home.

apelisse:
This has been merged in test-infra already. Closing, thanks

PR #2024, opened 2016-11-11
Ignore autogen files in sizer

thockin:
This is an action item from dev summit.  It builds but is not tested.  PTAL and help me test?

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/2024)


apelisse:
That LGTM.

Two things:
- Testing should probably be done by seperating into functions/class with smaller responsibility: `ignoreList = ParseIgnoreList(generatedFilesFile)` and later `ignoreList.IsIgnored(changedFile)`. I could be interested in doing that, but maybe not immediately.

- More interesting. I'm a little bit worried by the increasing number of "types". Could we have something like  this instead:
   - `path /exact/path/file` (doesn't end with `/`, assume full path, look for `pattern == file`)
   - `path /just/path/` (ends with `/`, assume path prefix, look for `file.HasPrefix(pattern)`)
   - `file name` (look for full name, can use `filepath.Match(pattern, filename)`)
   - `file nameprefix*` (look for partial name, can reuse `filepath.Match(pattern, filename)`)

thockin:
On review, the file itself should live in the repo.  https://github.com/kubernetes/kubernetes/pull/36854 moves the file there.  I will retool this to read that file.

thockin:
PTAL at this and https://github.com/kubernetes/kubernetes/pull/36854

thockin:
I just cargo culted that.  Can you handle the fix?

apelisse:
Sure

PR #2056, opened 2016-11-17
handle the case where release note not on same line as tag

### LINKED ISSUE #2054: Release Note Auto-Assign Doesn't Recognize If RN On Next Line
```
noteMatcher := regexp.MustCompile("Release note.*```(.+)```")
```

The wildcard matcher does not recognize newlines.  Need to set the (?s) flag.

grodrigues3:
The (?s) flag allows the wildcard operator to match newlines as well.

fixes #2054

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/2056)

### DIFF
--- a/mungegithub/mungers/release-note-label_test.go
+++ b/mungegithub/mungers/release-note-label_test.go
@@ -244,17 +244,22 @@ func TestGetReleaseNote(t *testing.T) {
 		expectedReleaseNoteVariable string
 	}{
 		{
-			body:                        "**Release note**: <other unimportant information> ```NONE```",
+			body:                        "**Release note**:  ```NONE```",
 			expectedReleaseNote:         "NONE",
 			expectedReleaseNoteVariable: releaseNoteNone,
 		},
 		{
-			body:                        "**Release note**: <other unimportant information> ```This is a description of my feature```",
+			body:                        "**Release note**:\n\n ```NONE```",
+			expectedReleaseNote:         "NONE",
+			expectedReleaseNoteVariable: releaseNoteNone,
+		},
+		{
+			body:                        "**Release note**:\n\n  ```This is a description of my feature```",
 			expectedReleaseNote:         "This is a description of my feature",
 			expectedReleaseNoteVariable: releaseNote,
 		},
 		{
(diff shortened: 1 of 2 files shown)


mungegithub/mungers/release-note-label_test.go
```diff
@@ -249,6 +249,11 @@ func TestGetReleaseNote(t *testing.T) {
 			expectedReleaseNoteVariable: releaseNoteNone,
 		},
 		{
+			body:                        "**Release note**:\n\n<other unimportant information> ```NONE```",
```
apelisse: isn't it weird that we skip "<other unimportant information>". What if it was important? (like being something completely different?). The .* becomes VERY lenient now that you have the `(?s)` modifier.

Should we use:
`Release Note:\s*BQ BQ BQ`.....

grodrigues3: also, check out the PR template https://github.com/kubernetes/kubernetes/edit/master/.github/PULL_REQUEST_TEMPLATE.md 

The goal of the regex is to match what's there explicitly.

grodrigues3: We ask specifically n the release-note template for the release note to be between three ticks.
E.g. ```Release Note Goes Here```

mungegithub/mungers/release-note-label_test.go
```diff
@@ -244,17 +244,22 @@ func TestGetReleaseNote(t *testing.T) {
...
+			expectedReleaseNoteVariable: releaseNoteNone,
+		},
+		{
+			body:                        "**Release note**:\n\n  ```This is a description of my feature```",
 			expectedReleaseNote:         "This is a description of my feature",
 			expectedReleaseNoteVariable: releaseNote,
 		},
 		{
-			body:                        "**Release note**: <other unimportant information> ```This is my feature. There is some action required for my feature.```",
+			body:                        "**Release note**: ```This is my feature. There is some action required for my feature.```",
```
apelisse: can we add a test that leads to no release note? One that shows that the regex is less lenient?

apelisse:
Looks good

PR #2060, opened 2016-11-18
SSL-enabled Nginx Ingress Controller integrated Hashicorp Vault 

devlinmr:
Based on the nginx-alpha controller, this adds functionality to configure SSL nginx servers for ingresses requiring it, pulling in certificates from Vault.

This is the controller demonstrated at KubeCon 2016 by Michael Ward (devoperandi.com) of Pearson.

See README.md.

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/2060)


bprashanth:
@devlinmr we're moving ingress out of contrib, can you integrate this either: 
1. as a backend to the freshly baked generic controller (https://github.com/kubernetes/ingress/pull/15)
2. as a cmdline option to the actual nginx ingress controller (https://github.com/kubernetes/ingress/tree/master/controllers/nginx)

?

bprashanth:
@aledbf

devlinmr:
Sure, it's an nginx-based controller that retrieves certs from Hashicorp Vault, so it's got a lot of Vault-specific functionality.

Which of 1 or 2 would you prefer with that in mind?

Issue #2026, opened 2016-11-12
[ansible] ansible fails with 'kube_binaries' is undefined when using packageManager on fedora 24


ObviusOwl:
Hi,
I'm on a fully updated fresh install of fedora 24. The configuration is almost default except enabled kube-ui and kube-dash.

Ansible fails with:

```
TASK [master : Set a list of kube master binaries to download] *****************
skipping: [kube-master.example.com] => {"changed": false, "skip_reason": "Conditional check failed", "skipped": true}

TASK [master : Extend the list of kube master binaries with kubelet] ***********
skipping: [kube-master.example.com] => {"changed": false, "skip_reason": "Conditional check failed", "skipped": true}

TASK [master : Download Kubernetes binaries] ***********************************
fatal: [kube-master.example.com]: FAILED! => {"failed": true, "msg": "'kube_binaries' is undefined"}
	to retry, use: --limit @/root/contrib/ansible/playbooks/deploy-cluster.retry
```

Deleting the content in `roles/master/tasks/download_bins.yml` helps.

The same applies for `roles/master/tasks/install_rpms.yml` where ansible complains about undefined `kube_rpms` (logs not included for brevity)

Deleting the content of the files still provides a working set-up, since the packages are in fact installed correctly and thus don't need external rpm's or github release.

Maybe this is a issue with ansible not ignoring correctly the tasks, unfortunately I'm completely new to ansible.

lvnilesh:
I see similar failure

    ./deploy-cluster.sh --tags=masters -vvv 

```
TASK [master : Download Kubernetes binaries] ***********************************
task path: /Users/cloudgenius/contrib/ansible/roles/master/tasks/download_bins.yml:17
fatal: [192.168.122.22]: FAILED! => {
    "failed": true, 
    "msg": "'kube_binaries' is undefined"
}
```

ingvagabund:
@ObviusOwl, @lvnilesh what is your version of ansible?

ingvagabund:
Most likely related to this: https://github.com/ansible/ansible/issues/14201

ObviusOwl:
My ansible version is 2.2.0.0, the issue mentioned seems to get fixed in 2.2.0.0-1

ingvagabund:
@ObviusOwl this needs to be fixed anyway based on the PR's comment.

I am not able to reproduce the scenario described in the PR with ansible-2.1.0.0-1.fc24.noarch. Testing with two files and localhost inventory:

```
---
[...]

PR #2065, opened 2016-11-19
ansible: Set empty array by default in kubernetes master role if tasks skipped

mkumatag:
Fix for https://github.com/kubernetes/contrib/issues/2026

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/2065)

### DIFF
--- a/ansible/roles/master/tasks/download_bins.yml
+++ b/ansible/roles/master/tasks/download_bins.yml
@@ -25,7 +25,7 @@
     https_proxy: "{{ https_proxy|default('') }}"
     no_proxy: "{{ no_proxy|default('') }}"
   with_items:
-    - "{{ kube_binaries }}"
+    - "{{ kube_binaries | default([]) }}"
 
 - name: Create symlinks
   file:
@@ -33,4 +33,4 @@
     dest: "{{ bin_dir }}/{{ item }}"
     state: link
   with_items:
-    - "{{ kube_binaries }}"
+    - "{{ kube_binaries | default([]) }}"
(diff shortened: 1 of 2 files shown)


ansible/roles/master/tasks/download_bins.yml
```diff
@@ -5,13 +5,9 @@
...
   file: path={{ kube_current_release_directory }} state=directory
 
-- name: Set a list of kube master binaries to download
-  set_fact:
-    kube_binaries: "{{ kube_master_binaries }}"
-
 - name: Extend the list of kube master binaries with kubelet
   set_fact:
-    kube_binaries: "{{ kube_master_binaries }} kubelet"
+    kube_master_binaries: "{{ kube_master_binaries + ['kubelet'] }}"
```
ingvagabund: This reason why this was not done in the introducing PR was the ``kube_master_binaries`` gets extended. And as each variable/fact has global scope, it could lead to unexpected behaviour if the ``kube_master_binaries`` gets used later after the ``download_bins``.

ingvagabund: I would like to keep the intermediate variable as the kubelet is not really a master binary. What about to use ``{{kube_binaries|default([])}}`` in the ``with_items``. Your solution is syntactically clean but semantically can lead to confusion.

mkumatag:
@ingvagabund pushed updated commit. ptal

ingvagabund:
@mkumatag LGTM, thanks.

PR #2067, opened 2016-11-21
Updated sed command to work on macOS

pabloguerrero:
A similar issue as fixed in kubernetes/kubernetes#20147

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/2067)


luxas:
From what I understand this will then break linux, right?
Can you come up with something that works on both, please?

for instance: http://stackoverflow.com/questions/5694228/sed-in-place-flag-that-works-both-on-mac-bsd-and-linux

pabloguerrero:
@luxas I've changed to [this](http://stackoverflow.com/a/36722778/746417) suggestion, but I am unable to test it in Linux. Would you be able to?

thockin:
code moved to ingress repo

PR #2071, opened 2016-11-22
mungegithub: Prometheus should not use gzipHandler

apelisse:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/2071)


apelisse:
It fails with content-length error ... (but it looks like it works from curl)

PR #2078, opened 2016-11-23
Implement batched merging for the submit queue.

rmmh:
A commandline option specifying a Prow data endpoint is used to collect
results for batch tests. If a batch has passed all necessary tests and
is applicable given the current state of the repo, it will merge all
included PRs at once, without individually testing them.

This batch goroutine runs in parallel with the normal goroutine that
processes PRs sequentially. A mutex guards the critical section for merges
to avoid conflicts.

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/2078)

### DIFF
--- a/mungegithub/github/github.go
+++ b/mungegithub/github/github.go
@@ -569,6 +569,28 @@ func (config *Config) NewIssue(title, body string, labels []string, owner string
 	return obj, nil
 }
 
+// GetBranchCommits gets recent commits for the given branch.
+func (config *Config) GetBranchCommits(branch string, limit int) ([]*github.RepositoryCommit, error) {
+	commits := []*github.RepositoryCommit{}
+	page := 0
+	for {
+		commitsPage, response, err := config.client.Repositories.ListCommits(
+			config.Org, config.Project,
+			&github.CommitsListOptions{ListOptions: github.ListOptions{PerPage: 100, Page: page}, SHA: branch})
+		config.analytics.ListCommits.Call(config, response)
+		if err != nil {
+			glog.Errorf("Error reading commits for branch %s: %v", branch, err)
+			return nil, err
+		}
+		commits = append(commits, commitsPage...)
+		if response.LastPage == 0 || response.LastPage <= page || len(commits) > limit {
+			break
+		}
+		page++
+	}
+	return commits, nil
+}
+
 // Branch returns the branch the PR is for. Return "" if this is not a PR or
 // it does not have the required information.
 func (obj *MungeObject) Branch() string {
--- a/mungegithub/mungers/submit-queue-batch.go
+++ b/mungegithub/mungers/submit-queue-batch.go
@@ -1,0 +1,306 @@
+/*
+Copyright 2016 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package mungers
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"strconv"
+	"strings"
+	"sync/atomic"
+	"time"
+
+	"github.com/golang/glog"
+	githubapi "github.com/google/go-github/github"
+	"k8s.io/contrib/mungegithub/github"
+)
+
+// xref k8s.io/test-infra/prow/cmd/deck/jobs.go
+type prowJob struct {
+	Type    string `json:"type"`
+	Repo    string `json:"repo"`
+	Refs    string `json:"refs"`
+	State   string `json:"state"`
+	Context string `json:"context"`
+}
+
+// getSuccessfulBatchJobs reads test results from Prow and returns
+// all batch jobs that succeeded for the current repo.
+func getSuccessfulBatchJobs(repo, url string) ([]prowJob, error) {
+	resp, err := http.Get(url)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+	body, err := ioutil.ReadAll(resp.Body)
+
+	allJobs := []prowJob{}
+	err = json.Unmarshal(body, &allJobs)
+	if err != nil {
+		return nil, err
+	}
+
+	jobs := []prowJob{}
+	for _, job := range allJobs {
+		if job.Repo == repo && job.Type == "batch" && job.State == "success" {
+			jobs = append(jobs, job)
+		}
+	}
+	return jobs, nil
+}
+
+type batchPull struct {
+	Number int
+	Sha    string
+}
+
+// Batch represents a specific merge state:
+// a base branch and SHA, and the SHAs of each PR merged into it.
+type Batch struct {
+	BaseName string
+	BaseSha  string
+	Pulls    []batchPull
+}
+
+func (b *Batch) String() string {
+	out := b.BaseName + ":" + b.BaseSha
+	for _, pull := range b.Pulls {
+		out += "," + strconv.Itoa(pull.Number) + ":" + pull.Sha
+	}
+	return out
+}
+
+// batchRefToBatch parses a string into a Batch.
+// The input is a comma-separated list of colon-separated ref/sha pairs,
+// like "master:abcdef0,123:f00d,456:f00f".
+func batchRefToBatch(batchRef string) (Batch, error) {
+	batch := Batch{}
+	for i, ref := range strings.Split(batchRef, ",") {
+		parts := strings.Split(ref, ":")
+		if len(parts) != 2 {
+			return Batch{}, errors.New("bad batchref: " + batchRef)
+		}
+		if i == 0 {
+			batch.BaseName = parts[0]
+			batch.BaseSha = parts[1]
+		} else {
+			num, err := strconv.ParseInt(parts[0], 10, 32)
+			if err != nil {
+				return Batch{}, fmt.Errorf("bad batchref: %s (%v)", batchRef, err)
+			}
+			batch.Pulls = append(batch.Pulls, batchPull{int(num), parts[1]})
+		}
+	}
+	return batch, nil
+}
+
+// getCompleteBatches returns a list of Batches that passed all
+// required tests.
+func (sq *SubmitQueue) getCompleteBatches(jobs []prowJob) []Batch {
+	// for each batch specifier, a set of successful contexts
+	batchContexts := make(map[string]map[string]interface{})
+	for _, job := range jobs {
+		if batchContexts[job.Refs] == nil {
+			batchContexts[job.Refs] = make(map[string]interface{})
+		}
+		batchContexts[job.Refs][job.Context] = nil
+	}
+	batches := []Batch{}
+	for batchRef, contexts := range batchContexts {
+		match := true
+		// Did this succeed in all the contexts we want?
+		for _, ctx := range sq.RequiredStatusContexts {
+			if _, ok := contexts[ctx]; !ok {
+				match = false
+			}
+		}
+		for _, ctx := range sq.RequiredRetestContexts {
+			if _, ok := contexts[ctx]; !ok {
+				match = false
+			}
+		}
+		if match {
+			batch, err := batchRefToBatch(batchRef)
+			if err != nil {
+				continue
+			}
+			batches = append(batches, batch)
+		}
+	}
+	return batches
+}
+
+// batchIntersectsQueue returns whether at least one PR in the batch is queued.
+func (sq *SubmitQueue) batchIntersectsQueue(batch Batch) bool {
+	sq.Lock()
+	defer sq.Unlock()
+	for _, pull := range batch.Pulls {
+		if _, ok := sq.githubE2EQueue[pull.Number]; ok {
+			return true
+		}
+	}
+	return false
+}
+
+// matchesCommit determines if the batch can be merged given some commits.
+// That is, does it contain exactly:
+// 1) the batch's BaseSha
+// 2) (optional) merge commits for PRs in the batch
+// 3) any merged PRs in the batch are sequential from the beginning
+// The return value is the number of PRs already merged, and any errors.
+func (b *Batch) matchesCommits(commits []*githubapi.RepositoryCommit) (int, error) {
+	if len(commits) == 0 {
+		return 0, errors.New("no commits")
+	}
+
+	shaToPR := make(map[string]int)
+
+	for _, pull := range b.Pulls {
+		shaToPR[pull.Sha] = pull.Number
+	}
+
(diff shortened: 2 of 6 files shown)


mungegithub/github/github.go
```diff
@@ -569,6 +569,28 @@ func (config *Config) NewIssue(title, body string, labels []string, owner string
...
+		commitsPage, response, err := config.client.Repositories.ListCommits(
+			config.Org, config.Project,
+			&github.CommitsListOptions{ListOptions: github.ListOptions{PerPage: 100, Page: page}, SHA: branch})
+		config.analytics.ListCommits.Call(config, response)
+		if err != nil {
+			glog.Errorf("Error reading commits for branch %s: %v", branch, err)
+			return nil, err
+		}
+		commits = append(commits, commitsPage...)
+		if response.LastPage == 0 || response.LastPage <= page || len(commits) > 200 {
```
spxtr: Why 200? Why not just stop at 100? Seems like you shouldn't even need a loop here.

mungegithub/mungers/submit-queue-batch.go
```diff
@@ -0,0 +1,282 @@
...
+			batch.Pulls = append(batch.Pulls, BatchPull{int(num), parts[1]})
+		}
+	}
+	return batch, nil
+}
+
+// getCompleteBatches returns a list of Batches that passed all
+// required tests.
+func (sq *SubmitQueue) getCompleteBatches(jobs []ProwJob) []Batch {
+	batchContexts := make(map[string]map[string]interface{})
```
spxtr: please add a comment saying what this is. it's hard to tell from the code

mungegithub/mungers/submit-queue-batch.go
```diff
@@ -0,0 +1,282 @@
...
+	}
+	return false
+}
+
+// matchesCommit determines if the batch can be merged given some commits.
+// That is, does it contain exactly:
+// 1) the batch's BaseSha
+// 2) (optional) merge commits for PRs in the batch
+// 3) any merged PRs in the batch are sequential from the beginning
+// The return value is -1 on error, else the number of PRs already merged.
```
spxtr: return `(int, error)`...

mungegithub/mungers/submit-queue-batch.go
```diff
@@ -0,0 +1,282 @@
...
+		}
+	}
+	return len(matchedPRs)
+}
+
+// batchIsApplicable returns whether a successful batch result can be used--
+// 1) some of the batch is still unmerged and in the queue.
+// 2) the recent commits are the batch head ref or merges of batch PRs.
+// 3) all unmerged PRs in the batch are still in the queue.
+// The return value is -1 on error, else the number of PRs already merged.
```
spxtr: `(int, error)`

mungegithub/mungers/submit-queue-batch.go
```diff
@@ -0,0 +1,282 @@
...
+}
+
+type Batch struct {
+	BaseName string
+	BaseSha  string
+	Pulls    []BatchPull
+}
+
+// batchRefToBatch parses a string into a Batch
+func batchRefToBatch(batchRef string) (Batch, error) {
```
apelisse: Could you explain the format?

mungegithub/mungers/submit-queue-batch.go
apelisse: Also, I'm pretty sure we already import k8s `sets.String`, so you could just use that.

mungegithub/mungers/submit-queue-batch.go
```diff
@@ -0,0 +1,282 @@
...
+func (sq *SubmitQueue) getCompleteBatches(jobs []ProwJob) []Batch {
+	batchContexts := make(map[string]map[string]interface{})
+	for _, job := range jobs {
+		if _, ok := batchContexts[job.Refs]; !ok {
+			batchContexts[job.Refs] = make(map[string]interface{})
+		}
+		batchContexts[job.Refs][job.Context] = nil
+	}
+	batches := []Batch{}
+	for batchRef, contexts := range batchContexts {
```
apelisse: In this loop:
- match is initialized to true
- it can go to false, but can't go back to true
- if match is false, it doesn't have any side effects

Could you:
- remove `match := true`
- replace `match = false` with `continue`
- remove `if match { }` ?

mungegithub/mungers/submit-queue-batch.go
```diff
@@ -0,0 +1,282 @@
...
+	repo := sq.githubConfig.Org + "/" + sq.githubConfig.Project
+	for range time.Tick(1 * time.Minute) {
+		jobs, err := getSuccessfulBatchJobs(repo, sq.BatchURL)
+		if err != nil {
+			glog.Errorf("Error reading batch jobs from Prow URL %v", sq.BatchURL)
+			continue
+		}
+		batches := sq.getCompleteBatches(jobs)
+		for _, batch := range batches {
+			if sq.batchIsApplicable(batch) < 0 {
```
apelisse: That condition is repeated in `sq.doBatchmerge`, should we have it once? where?

mungegithub/mungers/submit-queue-batch.go
```diff
@@ -0,0 +1,282 @@
...
+	ref := *commits[0].SHA
+	for {
+		if ref == batch.BaseSha {
+			break // found the base ref (condition #1)
+		}
+		commit, ok := dag[ref]
+		if !ok {
+			return -1 // ref not in partial list of commits
+		}
+		if len(commit.Parents) == 2 && commit.Message != nil && (*commit.Message)[:5] == "Merge" {
```
apelisse: `(*commit.Message)[:5] == "Merge"` :-/

Why is it necessary?

mungegithub/mungers/submit-queue-batch.go
apelisse: batchRef seems like to have an ugly format, but I can't even tell, could you describe it?

mungegithub/mungers/submit-queue-batch.go
apelisse: Also, `sets.String` is already imported in this project, you could probably use it

mungegithub/mungers/submit-queue-batch.go
```diff
@@ -0,0 +1,282 @@
...
+	batchContexts := make(map[string]map[string]interface{})
+	for _, job := range jobs {
+		if _, ok := batchContexts[job.Refs]; !ok {
+			batchContexts[job.Refs] = make(map[string]interface{})
+		}
+		batchContexts[job.Refs][job.Context] = nil
+	}
+	batches := []Batch{}
+	for batchRef, contexts := range batchContexts {
+		match := true
```
apelisse: Can you remove match and use continue?

mungegithub/github/github.go
apelisse: Why would we hard-code the value?

mungegithub/mungers/submit-queue-batch.go
apelisse: `(*commit.Message)[:5] == "Merge"` :-/

Do we really need that? Also, if we do, it should probably be `strings.HasPrefix(*commit.Message, "Merge")`.

mungegithub/mungers/submit-queue-batch.go
apelisse: You run this condition here, and directly in the following function. Could we do it once?

mungegithub/mungers/submit-queue-batch.go
```diff
@@ -0,0 +1,282 @@
...
+		}
+		if sha, _, ok := obj.GetHeadAndBase(); !ok {
+			glog.Errorf("error getting pr #%d sha", pull.Number, err)
+			return
+		} else if sha != pull.Sha {
+			glog.Errorf("error: batch PR #%d HEAD changed: %s instead of %s",
+				sha, pull.Sha)
+			return
+		}
+		if !sq.validForMergeExt(obj, false) {
```
apelisse: why don't we care about the CI status in this case? testing?

mungegithub/mungers/submit-queue-batch.go
```diff
@@ -0,0 +1,282 @@
...
+			return
+		}
+		prs = append(prs, obj)
```
apelisse: why do we need to create a temporary list and then merge?

mungegithub/mungers/submit-queue.go
```diff
@@ -1204,7 +1227,7 @@ func (sq *SubmitQueue) doGithubE2EAndMerge(obj *github.MungeObject) bool {
 
 	if obj.HasLabel(retestNotRequiredLabel) || obj.HasLabel(retestNotRequiredDocsOnlyLabel) {
 		atomic.AddInt32(&sq.instantMerges, 1)
-		sq.mergePullRequest(obj)
+		sq.mergePullRequest(obj, mergedSkippedRetest)
```
apelisse: should you do something with the newly returned err?

mungegithub/mungers/submit-queue.go
```diff
@@ -1259,7 +1285,7 @@ func (sq *SubmitQueue) doGithubE2EAndMerge(obj *github.MungeObject) bool {
 		return true
 	}
 
-	sq.mergePullRequest(obj)
+	sq.mergePullRequest(obj, merged)
```
apelisse: same as above

mungegithub/mungers/submit-queue-batch.go
spxtr: Although some people like `map[string]bool` since then the zero value is `false`.

mungegithub/mungers/submit-queue-batch.go
rmmh: No, because continue will only continue the inner for loop. I could use a goto here as well, but people have prejudices about that.

mungegithub/mungers/submit-queue.go
rmmh: eh, it was fine before when it was ignoring the error

all that will happen is that it fails the merge, removes it from the queue, and then re-adds it on the next loop

mungegithub/mungers/submit-queue.go
rmmh: ditto

spxtr:
The general idea looks excellent, but needs more comments and explanation.

mungegithub/mungers/submit-queue-batch.go
```diff
@@ -0,0 +1,303 @@
...
+		}
+		for _, ctx := range sq.RequiredRetestContexts {
+			if _, ok := contexts[ctx]; !ok {
+				match = false
+			}
+		}
+		if match {
+			batch, err := batchRefToBatch(batchRef)
+			if err != nil {
+				continue
```
spxtr: log the error?

mungegithub/mungers/submit-queue-batch.go
```diff
@@ -0,0 +1,303 @@
...
+
+// batchRefToBatch parses a string into a Batch.
+// The input is a comma-separated list of colon-separated ref/sha pairs,
+// like "master:abcdef0,123:f00d,456:f00f".
+func batchRefToBatch(batchRef string) (Batch, error) {
+	batch := Batch{}
+	for i, ref := range strings.Split(batchRef, ",") {
+		parts := strings.Split(ref, ":")
+		if len(parts) != 2 {
+			return Batch{}, errors.New("bad batchref")
```
spxtr: `fmt.Errorf("bad batchref: %s", batchRef)`, otherwise it's not very easy to debug.

mungegithub/mungers/submit-queue-batch.go
```diff
@@ -0,0 +1,303 @@
...
+		}
+		if i == 0 {
+			batch.BaseName = parts[0]
+			batch.BaseSha = parts[1]
+		} else {
+			num, err := strconv.ParseInt(parts[0], 10, 32)
+			if err != nil {
+				return Batch{}, err
```
spxtr: Possibly return a nicer error here.

mungegithub/mungers/submit-queue-batch.go
```diff
@@ -0,0 +1,303 @@
...
+	for {
+		if ref == batch.BaseSha {
+			break // found the base ref (condition #1)
+		}
+		commit, ok := dag[ref]
+		if !ok {
+			return 0, errors.New("ran out of commits (missing ref " + ref + ")")
+		}
+		if len(commit.Parents) == 2 && commit.Message != nil &&
+			strings.HasPrefix(*commit.Message, "Merge") {
```
spxtr: This is a little gross. Isn't it guaranteed to be a merge commit if it has multiple parents?

mungegithub/mungers/submit-queue-batch.go
```diff
@@ -0,0 +1,303 @@
...
+// The return value is the number of PRs already merged, and any errors.
+func (sq *SubmitQueue) batchIsApplicable(batch Batch) (int, error) {
+	// batch must intersect the queue
+	if !sq.batchIntersectsQueue(batch) {
+		return 0, errors.New("Batch has no PRs in Queue")
+	}
+	commits, err := sq.githubConfig.GetBranchCommits(batch.BaseName, 100)
+	if err != nil {
+		glog.Errorf("Error getting commits for batchIsApplicable: %v", err)
+		return 0, errors.New("Failed to get branch commits: " + err.Error())
```
spxtr: `fmt.Errorf("failed to get branch commits: %v", err)`

mungegithub/mungers/submit-queue-batch.go
```diff
@@ -0,0 +1,303 @@
...
+
+// batchIsApplicable returns whether a successful batch result can be used--
+// 1) some of the batch is still unmerged and in the queue.
+// 2) the recent commits are the batch head ref or merges of batch PRs.
+// 3) all unmerged PRs in the batch are still in the queue.
+// The return value is the number of PRs already merged, and any errors.
+func (sq *SubmitQueue) batchIsApplicable(batch Batch) (int, error) {
+	// batch must intersect the queue
+	if !sq.batchIntersectsQueue(batch) {
+		return 0, errors.New("Batch has no PRs in Queue")
```
spxtr: Errors usually start with lowercase.

mungegithub/mungers/submit-queue-batch_test.go
```diff
@@ -0,0 +1,166 @@
...
+		err      error
+	}{
+		{"m:a", Batch{"m", "a", nil}, nil},
+		{"m:a,1:b", Batch{"m", "a", []BatchPull{{1, "b"}}}, nil},
+		{"m:a,1:b,2:c", Batch{"m", "a", []BatchPull{{1, "b"}, {2, "c"}}}, nil},
+		{"asdf", Batch{}, errors.New("bad batchref")},
+		{"m:a,a:3", Batch{}, strconvErr},
+	} {
+		batch, err := batchRefToBatch(test.ref)
+		expectEqual(t, "error", err, test.err)
```
spxtr: It's bad to do this. If you want to test that the error is the proper type then define a new type for it and check that a type assertion works.

This test breaks if you change the text of the error which is annoying. Not a big deal, just annoying.

spxtr (approved):
Minor comments. Overall looks good.

mungegithub/mungers/submit-queue-batch_test.go
rmmh: I'll clean this up if we start distinguishing between errors in non-test code.

PR #2084, opened 2016-11-28
Rank PRs with milestones with due dates higher than all other PRs.

### LINKED ISSUE #2083: milestones should rank higher in the submit queue
For example if we have 2 PR(s) with ["P3","v1.5"] or ["P2","v1.6"]

Currently we are prioritizing the 1.6 above the 1.5... Can we please do the opposite?

rmmh:
Note that this does not affect milestones without due dates, (v1.6), so
normal priority ordering is maintained until we are in the release cycle.

Fixes kubernetes/contrib#2083.

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/2084)

### DIFF
--- a/mungegithub/mungers/submit-queue.go
+++ b/mungegithub/mungers/submit-queue.go
@@ -1070,21 +1070,21 @@ func (s queueSorter) Less(i, j int) bool {
 	a := s.queue[i]
 	b := s.queue[j]
 
-	aPrio := priority(a)
-	bPrio := priority(b)
-
-	if aPrio < bPrio {
-		return true
-	} else if aPrio > bPrio {
-		return false
-	}
-
 	aDue := a.ReleaseMilestoneDue()
 	bDue := b.ReleaseMilestoneDue()
 
 	if aDue.Before(bDue) {
 		return true
 	} else if aDue.After(bDue) {
+		return false
+	}
+
+	aPrio := priority(a)
+	bPrio := priority(b)
+
+	if aPrio < bPrio {
+		return true
+	} else if aPrio > bPrio {
 		return false
 	}
 
@@ -1445,19 +1445,20 @@ func (sq *SubmitQueue) servePriorityInfo(res http.ResponseWriter, req *http.Requ
 	res.WriteHeader(http.StatusOK)
 	res.Write([]byte(`The merge queue is sorted by the following. If there is a tie in any test the next test will be used. A P0 will always come before a P1, no matter how the other tests compare.
 <ol>
+  <li>Release milestone due date
+    <ul>
+      <li>Release milestones are of the form vX.Y where X and Y are integers</li>
+      <li>Milestones must have due dates.
+      <li>Other milestones are ignored.
+      <li>PR with no dated release milestone will be considered after any PR with a milestone</li>
+    </ul>
+  </li>
   <li>Priority
     <ul>
       <li>Determined by a label of the form 'priority/pX'
       <li>P0 -&gt; P1 -&gt; P2</li>
       <li>A PR with no priority label is considered equal to a P3</li>
       <li>A PR with the '` + retestNotRequiredLabel + `' or '` + retestNotRequiredDocsOnlyLabel + `' label will come first, before even P0</li>
-    </ul>
-  </li>
-  <li>Release milestone due date
-    <ul>
-      <li>Release milestones are of the form vX.Y where X and Y are integers</li>
-      <li>Other milestones are ignored.
-      <li>PR with no release milestone will be considered after any PR with a milestone</li>
     </ul>
   </li>
   <li>First time at which the LGTM label was applied.


luxas:
We could have decided to set v1.5 milestones on kubeadm PRs as well, and tried to get kubeadm into the "normal" releasing scheme like all the other binaries.

Instead, we chose to keep kubeadm out of that normal flow, and instead release the next stable kubeadm binary from master once the code freeze is over. So practically, the kubeadm PRs will belong to the v1.5, but will exist on master. But they haven't taken up SQ capacity before today.

And we'd like to get most kubeadm PRs merged quickly, so we can move it out from the main repo.

ref: https://github.com/kubernetes/kubernetes/issues/35314 https://github.com/kubernetes/kubeadm/issues/70 http://kubernetes.slackarchive.io/kubernetes-dev/

luxas:
@kubernetes/sig-cluster-lifecycle

bgrant0607:
cc @lavalamp @kubernetes/contributor-experience

bgrant0607:
cc @saad-ali

bgrant0607:
I agree with @eparis that for the most part people shouldn't be using priorities to game the queue. I think we need clear guidelines. IMO, prioritization should only be allowed in exceptional circumstances, such as for a fix for a broken non-queue-blocking test or for a PR that would otherwise likely need to be rebased.

Also note that I'm about to rename the priority labels. If we want to retain the queue-jumping behavior, we'll need to update the bot or create a new set of labels to use just for the queue.

deads2k:
> I agree with @eparis that for the most part people shouldn't be using priorities to game the queue. I think we need clear guidelines. IMO, prioritization should only be allowed in exceptional circumstances, such as for a fix for a broken non-queue-blocking test or for a PR that would otherwise likely need to be rebased.

Used judiciously, prioritization can allow PRs that are blocking additional work, particularly work across teams or areas to merge earlier and unblock them.  As a for instance, I have controller-manager pulls which block progress on authorizers from progressing and since it is cross people, being unable to work against master hurts.  It's more important to get those merged that my other fixes like a new CLI command or roundtripper.

Also, I think a "head of the queue" (which I thought was represented p0), is important for dealing with important flakes.

dims:
cc @saad-ali

saad-ali:
> I agree with @eparis that for the most part people shouldn't be using priorities to game the queue. I think we need clear guidelines. IMO, prioritization should only be allowed in exceptional circumstances, such as for a fix for a broken non-queue-blocking test or for a PR that would otherwise likely need to be rebased.

As much as I want this PR (it solves the immediate pain we're feeling in the submit queue), I agree that priority shouldn't be used to manipulate the queue.

> Also note that I'm about to rename the priority labels. If we want to retain the queue-jumping behavior, we'll need to update the bot or create a new set of labels to use just for the queue.

Ack. Since milestone labels will get priority, I'm ok with this.

eparis:
As we move away from priority numbers to named labels this was the last proposal I heard:

- priority/P0 -> priority/critical-urgent, kind/test-failure
- priority/P1 -> priority/important-soon, priority/important-longterm
- priority/P2 -> priority/backlog
- priority/P3 -> priority/awaiting-more-evidence

Which makes sense for labels, but not for PRs. So what makes sense for PRs to manage their place in the queue?

- `sumbit-queue/head` - Just goes to the top. Only used for major test breaks. Security fixes. This would be the big hammer.
- `submit-queue/multiple-rebases` - used on PRs after the 2nd rebase to get out of 'rebase hell'
- `submit-queue/blocks-multiple-others` - used on PRs if there are 2 other PRs that depend on this one
- `submit-queue/bug-fix` - I personally think bugs before features. Do we need a critical/urgent bug vs just bug?

What other valid reasons exist for a PR to need to 'jump' in the queue? I'd like to enumerate valid reasons for queue prioritization instead of a numeric system that is too ill defined to be used constently.

roberthbailey:
Are we planning to create multiple queues for each release branch (with different blocking tests)? What about different repos? We should apply labels, queue jumping logic, etc consistently across branches and repositories.

bgrant0607:
@eparis I like your proposed labels. I'd change `head` to `queue-fix` and add a `critical-fix`, which could include security fixes.

@roberthbailey Separate queues for each branch makes sense. I think there are already separate queues for separate repos.

eparis:
per repo already exits. Per branch would be a moderate amount of work.

bgrant0607:
cc @idvoretskyi @grodrigues3

bgrant0607:
I created queue/critical-fix, queue/multiple-rebases, queue/blocks-others, and queue/fix. We should update the bot to interpret those as P0-P3.

eparis:
@bgrant0607 I was seriously considering a label that meant 'fixes blocked e2e test' so I could be merged by the bot even though the tests were failed...

I'll try to get to these new names today/tomorrow, but I'm not sure.

bgrant0607:
I'd be ok with a test-fix label also

rmmh:
Priority and retest-not-required labels make batched merges less effective, and aren't very necessary with shallow queues.

eparis:
priorities are gone. retest-not-required I totally see the issue.

Is it a system design? How the tools interact?

rmmh:
Closing since the labels this would affect are now irrelevant.

PR #2107, opened 2016-12-01
Submit queue batch fixes

rmmh:
- Reveal BaseRef so batches can be filtered to just "master"
- Stop polling for status if the PR is closed underneath us.

### DIFF
--- a/mungegithub/github/github.go
+++ b/mungegithub/github/github.go
@@ -1228,6 +1228,12 @@ func (obj *MungeObject) doWaitStatus(pending bool, requiredContexts []string, c 
 			glog.V(4).Infof("PR# %d is pending, waiting for %f seconds", *obj.Issue.Number, sleepTime.Seconds())
 		}
 		time.Sleep(sleepTime)
+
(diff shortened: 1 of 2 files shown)


k8s-oncall:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/contrib/2107)

mungegithub/github/github.go
```diff
@@ -1228,6 +1228,12 @@ func (obj *MungeObject) doWaitStatus(pending bool, requiredContexts []string, c
 			glog.V(4).Infof("PR# %d is pending, waiting for %f seconds", *obj.Issue.Number, sleepTime.Seconds())
 		}
 		time.Sleep(sleepTime)
+
+		// If it has been closed, assume that we want to break from the poll loop early.
+		obj.Refresh()
```
rmmh: ~4 API calls per minute, which is pretty tiny when the SQ regularly uses 5 API calls per second.

spxtr: I'm not sure if that's a great argument, but okay.

PR #2113, opened 2016-12-02
Make stale-pending-ci not segfault when cleaning comments while disabled.

rmmh:


fejta:
Thanks Ryan!

eparis:
So disabling it broke it. I wonder how many others would have this problem? Maybe we should do something better around stale comment registration generically...

PR #2117, opened 2016-12-04
Cluster Autoscaler: Support other scale down poilcies

andrewsykim:
Seems like cluster autoscaler terminates the first valid candidate which works for most cases. There are situations where we might want to terminate the oldest nodes (like gradually upgrading cluster to new kubernetes version) or the least utilized node (to reduce rescheduling overhead). 

I wrote some preliminary code to get some initial feedback. After approval I will attempt to add some unit/integration/e2e tests for the different scaling policies (possibly in another PR).

### DIFF
--- a/cluster-autoscaler/scale_down.go
+++ b/cluster-autoscaler/scale_down.go
@@ -86,8 +86,7 @@ func FindUnneededNodes(nodes []*kube_api.Node,
 
 	// Phase2 - check which nodes can be probably removed using fast drain.
 	nodesToRemove, newHints, err := simulator.FindNodesToRemove(currentlyUnneededNodes, nodes, pods,
-		nil, predicateChecker,
-		len(currentlyUnneededNodes), true, oldHints, tracker, timestamp)
+		nil, predicateChecker, true, oldHints, tracker, timestamp)
 	if err != nil {
 		glog.Errorf("Error while simulating node drains: %v", err)
 		return map[string]time.Time{}, oldHints, map[string]float64{}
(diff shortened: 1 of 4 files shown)


cluster-autoscaler/scale_down.go
```diff
@@ -191,7 +190,7 @@ func ScaleDown(
 	}
 
 	// We look for only 1 node so new hints may be incomplete.
-	nodesToRemove, _, err := simulator.FindNodesToRemove(candidates, nodes, pods, client, predicateChecker, 1, false,
+	nodesToRemove, _, err := simulator.FindNodesToRemove(candidates, nodes, pods, client, predicateChecker, false,
```
mwielgus: This code (FindNodesToRemove with fastDrain=false) does quite expensive manual verification that rs/rc/jobs/daemonset exist. It should not go over all nodes. So I would rather sort the candidates according to some criteria. And then get 1 (or more) that best fit the strategy.

andrewsykim:
@mwielgus is this a desirable feature? If so I can up with a better implementation.

mwielgus:
@andrewsykim We can have some strategies there. Picking the least-utilized node for deletion sounds quite reasonable. Oldest node for upgrades - at least for me - not necessarily but you can try to convince me with a more detailed example. Plus I would add a random strategy.

andrewsykim:
I agree that the oldest node strategy won't be used often, but if the implementation is not too difficult I don't see why it would hurt to keep that option available. I'll use a similar implementation to what we used for the expander strategies.

mwielgus:
This needs to go to kubernetes/autoscaler if you want to get it in.

Issue #1921, opened 2016-10-24
Cluster-autoscaler: add a way to specify which node pool to expand


mwielgus:
Right now if there are two node pools that expanding which would make a pod schedulable a random one is chosen. This may result creating a 32core+ssd node for a single pod.

mumoshu:
Hi @mwielgus, thanks for developing clusterr-autoscaler!
I'm the primary maintainer of https://github.com/coreos/kube-aws and interested in using or even implementing this feature but how would you like to specify node pool?

Adding something like `--precedence ondemand>spot --precedence smaller>larger` to cluster autoscaler's command-line arg to set preference to ASGs with on-demand instances and/or smaller instances when there're multiple candidates would suffice?

mumoshu:
IMHO I believe this is related to #2066.
If you'd like to keep minimum/base-line capacity with on-demand instances while spreading some loads to cheaper but unreliable instances(from a spot fleet) for performance reason, you'd also like to specify which asg or spot fleet to expand.

thebigjc:
There is the idea of how well the pods that need to be scheduled fit. 

Real world problem I have:
- Some of my pods have moderate CPU / light RAM needs, so we have a node group of c3.2xlarge nodes. 
- Some pods have higher RAM needs, so we have a node group of r3.2xlarge nodes for those. 

Ideally, when needing to schedule the first group of nodes, the cluster-autoscaler would pick the c3.2xlarge nodes, and when needing to schedule the second, it would scale the second. 

An idea of 'fit' for the pods, or 'waste' on the node would satisfy this solution. In the first case, if you were to schedule the pods on the c3 nodes, you would have less wasted RAM than if the r3 were chosen.

Does this seem like a reasonable strategy to adopt?

mumoshu:
@thebigjc Sounds great!
Just my one cent but If there're 2 or more "fitted" node pools, selecting one randomly would be good.

thebigjc:
@mumoshu I agree if there is a tie we should still be random. I have a draft that I'll send as a PR today!

thebigjc:
Should we close this?

PR #2118, opened 2016-12-04
Provide multiple strategies for choosing which Node Group to increase

thebigjc:
As per #1921, the current cluster-autoscaler chooses a node group to increase at random among the groups that can fit at least one pod. This expands this to two additional strategies:

- Most Pods - choose the node group that can schedule the most pods, usually due to predicate rules
- Least Waste - choose the node group that has the least remaining CPU resources after a scale up, and if that is tied, the least remaining Memory

In the event of a tie, both of these strategies fall back to the current Random method.

This probably needs test coverage, to move to its own package, and there could be some weirdness in the algorithms as it needs manual testing.

Looking mostly to see if I'm on the right track and will proceed after.

### DIFF
--- a/cluster-autoscaler/expander/waste/waste.go
+++ b/cluster-autoscaler/expander/waste/waste.go
@@ -1,0 +1,96 @@
+/*
+Copyright 2016 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package waste
+
+import (
+	"github.com/golang/glog"
+	"k8s.io/contrib/cluster-autoscaler/expander"
+	"k8s.io/contrib/cluster-autoscaler/expander/random"
+	"k8s.io/kubernetes/pkg/api/resource"
+	apiv1 "k8s.io/kubernetes/pkg/api/v1"
+	"k8s.io/kubernetes/plugin/pkg/scheduler/schedulercache"
+)
+
+type leastwaste struct {
+	fallbackStrategy expander.Strategy
+}
+
+// NewStrategy returns a strategy that selects the best scale up option based on which node group returns the least waste
+func NewStrategy() expander.Strategy {
+	return &leastwaste{random.NewStrategy()}
+}
+
+// BestOption Finds the option that wastes the least fraction of CPU and Memory
+func (l *leastwaste) BestOption(expansionOptions []expander.Option, nodeInfo map[string]*schedulercache.NodeInfo) *expander.Option {
+	var leastWastedScore float64
+	var leastWastedOptions []expander.Option
+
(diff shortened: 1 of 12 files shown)


thebigjc:
Moved the strategies into their own package and added test coverage. Currently calling them 'expanders' but open to better names.

mwielgus:
Thank you for working on this. I guess we will have more than 2-3 estimators - yours are quite basic and eventually we will have to come up with something much more complex that incorporates  $$$ cost and balances the strategies. We need a way to keep everything in order. 

So, please:

* Define an interface for expander and a common way to build it (factory or whatever).
* Put each of the expanders into a separate directory, along with tests.
* Create a separate md page where the expanders are documented along with some guidance when to use which.

thebigjc:
@mwielgus generally sounds good. one question on 'Put each of the expanders into a separate directory, along with tests.' - are you thinking like the `cloudproviders` package? One main with the factory and interface, and one for each? Main reason I ask is they are currently quite simple, so it's a lot for very little code, but I'm happy to do it if it makes sense.

Agree this is a first step - solves a particular problem I have and lays the ground work for more in the future.

thebigjc:
@mwielgus changes are done except for the md file, in case you want to take a look here

thebigjc:
@mwielgus All done my changes. @andrewsykim and I are using this in our production cluster and it's working great so far!

cluster-autoscaler/expander/waste/waste.go
```diff
@@ -0,0 +1,99 @@
...
+type leastwaste struct {
+	next expander.Strategy
+}
+
+// NewStrategy returns a strategy that selects the best scale up option based on which node group returns the least waste
+func NewStrategy() expander.Strategy {
+	return &leastwaste{random.NewStrategy()}
+}
+
+// BestOption Finds the option that wastes the least amount of CPU, then the least amount of Memory, then random
```
mwielgus: Maybe we should make an average out of % of CPU and Mem wasted? WDYT?

mwielgus:
Mostly OK, one comment regarding least waste strategy. Rebase is needed.

thebigjc:
@mwielgus I like the idea - made the change, and rebased.

cluster-autoscaler/expander/waste/waste_test.go
```diff
@@ -0,0 +1,117 @@
...
+	nodeInfo := schedulercache.NewNodeInfo()
+	nodeInfo.SetNode(node)
+
+	return nodeInfo
+}
+
+func TestLeastWaste(t *testing.T) {
+	cpuPerPod := int64(500)
+	memoryPerPod := int64(1000 * 1024 * 1024)
+
```
mwielgus: Reduce the number of empty lines in this function.

cluster-autoscaler/expander/waste/waste.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/golang/glog"
+	"k8s.io/contrib/cluster-autoscaler/expander"
+	"k8s.io/contrib/cluster-autoscaler/expander/random"
+	"k8s.io/kubernetes/pkg/api/resource"
+	apiv1 "k8s.io/kubernetes/pkg/api/v1"
+	"k8s.io/kubernetes/plugin/pkg/scheduler/schedulercache"
+)
+
+type leastwaste struct {
+	next expander.Strategy
```
mwielgus: s/next/fallbackStrategy/

mwielgus:
Couple nits. Please fix, squash commits to just 1 and we are good to go.

thebigjc:
Good to go once Travis passes.

PR #2119, opened 2016-12-05
prometheus: update choices

brancz:
The previously described setup only does declarative blackbox
monitoring, which is a very small subset of what Prometheus is able to
do. Furthermore it is incorrect as Prometheus expects everything to be
whitebox monitored as `instance` labels are otherwise incorrectly set,
and therefore results in an ambiguous source of a metric.

Prometheus has a built-in Kubernetes Service Discovery mechanism which
can be used to discover targets to perform whitebox monitoring. However,
even though all targets can be discovered with that mechanism, it is not
a one-size-fits-all solution and thus still requires deep knowledge of
the way Prometheus is configured.

The Prometheus Operator manages Prometheus instances and integrates its
configuration natively into Kubernetes using third party resources.

The Prometheus Helm Chart provides another alternative.

@fabxc @matthiasr


piosz:
@fabxc could you please take a look?

fabxc:
LGTM. Short references to maintained documentation are more helpful than more elaborate descriptions that run out of date.

PR #2122, opened 2016-12-05
Relax OWNERSHIP rules if none found

eparis:
It was possible that a PR changing files whose only "leaf" owner was the
PR author would result in 1 OWNER. The PR author. That OWNER would be
eliminated. Thus the PR would get no owner.

The fix is to look for ALL owners of files if we can't find any.


foxish:
Adding @grodrigues3 who is more familiar with the recent changes in blunderbuss logic.

PR #2123, opened 2016-12-05
Specific ci failure status

eparis:


eparis:
What would have been better in different commits?

I broke CI, crud, let me fix that...

apelisse:
/lgtm

@eparis the independent changes

PR #2126, opened 2016-12-05
Delete cached github responses we know are bad

eparis:
Github seems to screw up their conditional responses when the last entry
is exactly on a page size boundary. Today, when we get the last page
which is exactly full we try the next page anyway even though github
says there isn't a next page.

This patch attempts to delete our cached entry when we find that case.


eparis:
@lavalamp you found this initially and wrote the first workaround. I can't prove that this helps without running it in production, because I can't seem to ever trigger it locally...

PR #2127, opened 2016-12-05
Fix accidental test limit merged when using make test

eparis:
I accidentally merged a patch which restricted which tests were run by
make test. Run them all.


eparis:
@apelisse

PR #2128, opened 2016-12-06
Keep looking up the tree if we found an empty set

eparis:
If we find an OWNERS file with only reviewers, but no assignee or
approvers we will insert an empty set into the assignees. If we find
that empty set later we should ignore it and look for the closest
assignable person.


apelisse:
/lgtm

Yes, I see now :-)

PR #2132, opened 2016-12-07
Adds --probe to export health checks and latency metrics for dns servers

bowei:
`--probe <label>,<server>,<dns name>,<interval>` will query the given
DNS server for `<dns name>` every `<interval>` seconds. The results will
be posted to a health check URL and DNS latencies and errors reported
via the /metrics URL.

Multple `--probe`s may be specified.

This replaces the functionality of the exec-healthz for the DNS pod.

### DIFF
--- a/dnsmasq-metrics/pkg/server/dnsprobe.go
+++ b/dnsmasq-metrics/pkg/server/dnsprobe.go
@@ -1,0 +1,199 @@
+/*
+Copyright 2016 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package server
+
+import (
+	"encoding/json"
+	"fmt"
+	"math/rand"
+	"net/http"
+	"sync"
+	"time"
+
+	"github.com/golang/glog"
+	"github.com/miekg/dns"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+// loopDelayer encapsulates the delay-loop timing logic. This
+// externalizes it for testing.
+type loopDelayer interface {
+	// Start the delay loop, may sleep.
+	Start(interval time.Duration)
+	// Sleep the required amount of time taking into account the
+	// `latency` of the loop body.
+	Sleep(latency time.Duration)
+}
+
+type defaultLoopDelayer struct {
+	interval time.Duration
+}
+
+func (d *defaultLoopDelayer) Start(interval time.Duration) {
+	d.interval = interval
+	// Stagger the start of the loop to avoid sending all probes at
+	// exactly the same time.
+	time.Sleep(time.Duration(rand.Int63n((int64)(d.interval))))
+}
+
+func (d *defaultLoopDelayer) Sleep(latency time.Duration) {
+	sleepInterval := d.interval - latency
+	if sleepInterval > 0 {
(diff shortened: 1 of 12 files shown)


bowei:
@MrHohn

dnsmasq-metrics/pkg/server/dnsprobe.go
```diff
@@ -0,0 +1,160 @@
...
+			w.Write(buf)
+		} else {
+			w.WriteHeader(500)
+			glog.Errorf("JSON Marshal error: %v", err)
+		}
+	} else {
+		response.IsOk = false
+		response.Err = p.lastError.Error()
+
+		w.Header().Add("Content-Type", "application/json")
```
MrHohn: subtle: may be move this out of if else block as we are calling it anyway?

dnsmasq-metrics/pkg/server/dnsprobe_test.go
```diff
@@ -0,0 +1,137 @@
...
+	addr, port := server.Init(t)
+	go server.Run(okResponseCallback)
+	<-server.StartChan
+
+	options := makeOptions("ok", addr, port)
+	probe := &dnsProbe{DNSProbeOption: options.Probes[0]}
+	probe.Start(options)
+
+	// This is ugly and flaky but we wait for some probes to go through by sleeping.
+	time.Sleep(250 * time.Millisecond)
```
MrHohn: Suggest to use a mock clock to prevent flakiness. As we did in https://github.com/kubernetes/contrib/blob/master/exec-healthz/cmd/exechealthz/exechealthz_test.go#L55-L84.

MrHohn (commented):
Mostly LG, suggest don't use a real clock for unit tests.

bowei:
- fixed racy unit test
- clean up of the http response code

dnsmasq-metrics/pkg/test/test.go
```diff
@@ -0,0 +1,83 @@
+/*
```
MrHohn: I found this file name is somehow confusing --- I thought there might be tests in it, but turned out it is like test utils...

dnsmasq-metrics/pkg/server/dnsprobe.go
```diff
@@ -0,0 +1,199 @@
...
+
+type defaultLoopDelayer struct {
+	interval time.Duration
+}
+
+func (d *defaultLoopDelayer) Start(interval time.Duration) {
+	d.interval = interval
+	// Stagger the start of the loop to avoid sending all probes at
+	// exactly the same time.
+	time.Sleep(time.Duration(rand.Int63n((int64)(d.interval))))
```
MrHohn: I like this. We should use similar mechanism for contrib/exec-healthz.

dnsmasq-metrics/pkg/server/dnsprobe_test.go
```diff
@@ -0,0 +1,140 @@
...
+		DNSProbeOption: options.Probes[0],
+		delayer:        delayer,
+	}
+
+	probe.Start(options)
+
+	// Wait for one loop to have been completed.
+	<-delayer.sleepDone
+
+	probe.lock.Lock()
```
MrHohn: It is a bit strange to use the probe.lock itself to pause the mock probing process explicitly. May be we should provide `mockLoopDelayer` the ability to exec one loop on our demand, but not just send out a notification when a loop is done?

MrHohn: This should be fine because channel in golang is buffered. Putting additional data into unconsumed channel will be blocked.

MrHohn:
LGTM

bowei:
@thockin

dnsmasq-metrics/pkg/server/dnsprobe_test.go
bowei: I think you mean that channel is *not* buffered by default.

MrHohn: Sorry, should be not buffered

thockin:
We should move DNS out to a new repo which includes this and kube-dns proper.

PR #2137, opened 2016-12-09
Cluster-Autoscaler: add node drain logic

mwielgus:
Currently a node is removed without any notice and pods  that live on the node killed without graceful termination. This PR adds drain logic to scale down.

### DIFF
--- a/cluster-autoscaler/cluster_autoscaler.go
+++ b/cluster-autoscaler/cluster_autoscaler.go
@@ -205,6 +205,12 @@ func run(_ <-chan struct{}) {
 					continue
 				}
 
+				// CA can die at any time. Removing taints that might have been left from the previous run.
+				if err := cleanToBeDeleted(nodes, kubeClient, recorder); err != nil {
+					glog.Warningf("Failed to clean ToBeDeleted information: %v", err)
+					continue
+				}
+
 				allUnschedulablePods, err := unschedulablePodLister.List()
 				if err != nil {
 					glog.Errorf("Failed to list unscheduled pods: %v", err)
--- a/cluster-autoscaler/scale_down.go
+++ b/cluster-autoscaler/scale_down.go
@@ -17,6 +17,7 @@ limitations under the License.
 package main
 
 import (
+	"encoding/json"
 	"fmt"
 	"reflect"
 	"strings"
@@ -24,6 +25,7 @@ import (
 
 	"k8s.io/contrib/cluster-autoscaler/cloudprovider"
 	"k8s.io/contrib/cluster-autoscaler/simulator"
+	"k8s.io/kubernetes/pkg/api/errors"
 	apiv1 "k8s.io/kubernetes/pkg/api/v1"
 	kube_client "k8s.io/kubernetes/pkg/client/clientset_generated/release_1_5"
 	kube_record "k8s.io/kubernetes/pkg/client/record"
@@ -44,6 +46,14 @@ const (
 	ScaleDownNoNodeDeleted ScaleDownResult = iota
 	// ScaleDownNodeDeleted - a node was deleted.
 	ScaleDownNodeDeleted ScaleDownResult = iota
+)
+
+const (
+	// ToBeDeletedTaint is a taint used to make the node unschedulable.
+	ToBeDeletedTaint = "ToBeDeletedByClusterAutoscaler"
+
+	// MaxGracefulTerminationTime is max gracefull termination time used by CA.
+	MaxGracefulTerminationTime = time.Minute
 )
 
 // FindUnneededNodes calculates which nodes are not needed, i.e. all pods can be scheduled somewhere else,
@@ -210,11 +220,14 @@ func ScaleDown(
 	glog.V(0).Infof("Scale-down: removing node %s, utilization: %v, pods to reschedule: ", toRemove.Node.Name, utilization,
 		strings.Join(podNames, ","))
 
+	// Nothing super-bad should happen if the node is removed from tracker prematurely.
 	simulator.RemoveNodeFromTracker(usageTracker, toRemove.Node.Name, unneededNodes)
-	err = deleteNodeFromCloudProvider(toRemove.Node, cloudProvider, recorder)
+
+	err = deleteNode(toRemove.Node, toRemove.PodsToReschedule, client, cloudProvider, recorder)
 	if err != nil {
 		return ScaleDownError, fmt.Errorf("Failed to delete %s: %v", toRemove.Node.Name, err)
 	}
+
 	return ScaleDownNodeDeleted, nil
 }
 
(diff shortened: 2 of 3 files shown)


fgrzadkowski:
Why do you re-implement drain instead of reusing existing mechanisms?

Phlamethrower:
FWIW I recently implemented my own take on this, although I haven't had much chance to test it on a live workload yet: https://github.com/Exient/contrib/commit/860f0ca0ad6f5d50e2e055dd2e28a501224eb1ba

As someone who's fairly new to Kubernetes it's encouraging to me that your approach is basically the same as mine (mark the node unschedulable and then terminate the pods), but I am a bit worried about what will happen if you have a replica set containing only a couple of pods and those pods are all located on the node that's being terminated. If the autoscaler terminates all the pods at once then there'll be a brief period where the replica set/service will have no pods available to respond to requests.

In my solution it's careful to only terminate one pod at a time from any given replica set, so as long as the desired replica count is >1 it shouldn't result in the replica dropping to zero active instances. In an ideal world we'd have a solution that would also work with replica sets which have a desired count of 1, but for us that wasn't a high priority (we're happy to run a minimum of two instances of any mission-critical pods)

mwielgus:
Couple reasons:
* The current drain logic sits inside kubectl and would require significant refactoring to get out of there. 
* Even after the refactoring the code is likely to be changed. This happened to the previous attempt to reuse drain node evaluation. It lasted a month and then someone changed the code in kubectl. CA code is in a different repo.
* We need to have an automatic rollback/retry logic should the drain failed in the middle.
* AFAIR there was a plan to move drain to the server side. So this code will eventually be removed as well.

mwielgus:
@Phlamethrower for your use case we have pod disruption budget. Eventually CA will respect PDB as well.

andrewsykim:
@mwielgus I noticed you mentioned fallbacks for failing to delete pods or nodes, have we considered what would happen if cluster-autoscaler pod dies in the middle of draining a node? It would have marked the node as unschedulable and when it starts up again it would not start any scaling operations because the expected and actual node counts will be different (since CA does not count unschedulable nodes).

andrewsykim:
Should we also consider removing the `skip-nodes-with-system-pods` flag entirely? Seems like an awful idea to set `--skip-nodes-with-system-pods=false` and have CA drain nodes before terminating. If it elects itself for termination it would delete itself after marking a node unschedulable

mwielgus:
@andrewsykim - for that reason we add an annotation to the node. On CA startup we will scan the nodes and set schedulable to true on nodes that have it set to false AND have the annotation. 

As of skipping nodes with system pods - for now we will keep it. System pods should eventually get pod disruption budget objects and then we will use pdb only.

As mentioned in the PR title this is WIP (work in progress). I started working on this and noticed that i should probably first update the kubernetes version and migrate client to clientset (what requires some grungy work). So I sent this for initial feedback.

andrewsykim:
Ahh, thanks for explaining, makes much more sense :) looking forward to seeing this merged!

davidopp:
> As of skipping nodes with system pods - for now we will keep it. System pods should eventually get pod disruption budget objects and then we will use pdb only.

Can you explain what you mean by this? You will set PodDisruptionBudget on system pods so they can never be evicted (thus their nodes can never be removed)? I'm not sure we want to allow this because then it would block node upgrade.

mwielgus:
@davidopp - Hmm I see your point but this is also true for user application. They can declare PDB of minAvailable=1 for replicaset of size 1 and effectively prevent all node upgrades. 
In the long run I would rather not have any custom logic in CA that somehow favors system pods. The situation that some components can evict system pods and some not (node upgrade vs ca) should be handled by (expanded) pdbs.

mwielgus:
Tests, cleanups and clientset added.

@fgrzadkowski  PTAL

cluster-autoscaler/scale_down.go
```diff
@@ -46,6 +47,13 @@ const (
 	ScaleDownNodeDeleted ScaleDownResult = iota
 )
 
+const (
+	// DrainTimeClusterAutoscaler is an annotation added to the nodes that CA marks as unschedulable.
+	DrainTimeClusterAutoscaler = "clusterautoscaler.beta.kubernetes.io/drain_time"
+	// MaxGracefulTerminationTime is max gracefull termination time used by CA.
+	MaxGracefulTerminationTime = time.Minute
```
mwielgus: I will do in the next PR. The next PR will be about reducing the number of arguments passed to scale up/down and putting all the config related params into a single struct.

cluster-autoscaler/cluster_autoscaler.go
```diff
@@ -205,6 +205,11 @@ func run(_ <-chan struct{}) {
 					continue
 				}
 
+				if err := cleanUnschedulable(nodes, kubeClient, recorder); err != nil {
```
piosz: Why do you need to clean unschedulable nodes here? My feeling is that you only mark them unschedulable right before removing them.

mwielgus: CA can die at any moment.

cluster-autoscaler/scale_down.go
```diff
@@ -260,6 +271,114 @@ func getEmptyNodes(candidates []*apiv1.Node, pods []*apiv1.Pod, maxEmptyBulkDele
...
+// Performs drain logic on the node. Marks the node as unschedulable and later removes all pods, giving
+// them up to MaxGracefulTerminationTime to finish.
+func drainNode(node *apiv1.Node, pods []*apiv1.Pod, client kube_client.Interface, recorder kube_record.EventRecorder) error {
+	if err := makeUnschedulable(node, client, recorder); err != nil {
+		return err
+	}
+
+	for _, pod := range pods {
+		recorder.Eventf(pod, apiv1.EventTypeNormal, "ScaleDown", "deleting pod for node scale down")
+		err := client.Core().Pods(pod.Namespace).Delete(pod.Name, &apiv1.DeleteOptions{})
```
mwielgus: What difference would it make? I mean, we kill the node after MaxGracefulTerminationTime anyway.

cluster-autoscaler/scale_down.go
piosz: SGTM

cluster-autoscaler/scale_down.go
piosz: Discussed offline

piosz:
As discussed offline please add a comment to migrate to taints to avoid potential race with other components using unschedulable bit.

cluster-autoscaler/cluster_autoscaler.go

mwielgus:
Switched to Taints. PTAL

cluster-autoscaler/scale_down.go
```diff
@@ -260,6 +273,153 @@ func getEmptyNodes(candidates []*apiv1.Node, pods []*apiv1.Pod, maxEmptyBulkDele
...
+
+	// Wait up to MaxGracefulTerminationTime.
+	for start := time.Now(); time.Now().Sub(start) < MaxGracefulTerminationTime; time.Sleep(5 * time.Second) {
+		allGone = true
+		for _, pod := range pods {
+			podreturned, err := client.Core().Pods(pod.Namespace).Get(pod.Name)
+			if err == nil {
+				glog.Errorf("Not deleted yet %v", podreturned)
+				allGone = false
+				continue
```
piosz: Shouldn't you break here?

piosz:
LGTM, just a nit

piosz (approved):

PR #2138, opened 2016-12-09
Make release-note-label work with newlines and ```release-note format.

rmmh:
There's an even split (109 "Release note**", 113 "```release-note") between the two styles of release note marking, so we might as well support both.

It didn't work with newlines at all before, so it was messing up in almost every case.


rmmh:
deployed

Issue #2145, opened 2016-12-13
Cluster AutoScaler - Azure support


uthark:
Hi all,

I'm working on azure support in cluster-autoscaler and main functionality is ready.

Code needs some polishing, though, I'm working on it.

Also, I'd like to get feedback for the code.

uthark:
I'm closing the issue because pull request was merged.

PR #2146, opened 2016-12-13
Support for Azure in cluster-autoscaler

uthark:
See issue #2145


djsly:
I looked at the code and I can't seem to see where the scaleset selection comes from?

Does this support multiple scaleSet for different kube-node groups ? e.g. 20xsmall_vm_nodes and 20xbig_vm_nodes

colemickens:
cc: @brendandburns

brendandburns:
@uthark Thanks for doing this!

I have a concern which is that this is based on the VMScaleSets API which isn't supported inside of Kubernetes proper (the VM and VMScaleSets APIs are different *sigh*)

I'm worried that this will confuse people and result in some broken clusters.

Would it be too much work to factor this so that it works _either_ with VM Scale Sets or the traditional VM API?

In the limit, we're going to move all of Kubernetes over to scale sets, but for now, given some of the limitations (e.g. you can't mount disks into scale sets) we can make the transition yet.

Let me know what you think, thanks again for the contribution!
--brendan

uthark:
@brendandburns Current implementation works with VM ScaleSet API only. If the node that is supposed to be shutdown is not a part of the scaleset — it will not be shutdown. Also, if there is a scale up operation — then node will be added in the scaleset which still has place to grow.

uthark:
@djsly Implementation is heavily based on existing implementations for GCE and AWS. User need to provide scaleset names to manage and corresponding min/max size in command-line options. The option name is `--nodes`, it can be provided several times, so several kube-node groups are supported (if we are speaking about the same). The mechanism to determine how many nodes should be removed or added is not modified, existing implementation is used.

uthark:
I'll rebase my commit and update pull request.

uthark:
I see that CI fails because of missing Azure dependencies. 

I tried to run `godep save ./...` but it fails with the following message:


```
godep save ./...                          
godep: Package (k8s.io/kubernetes/pkg/client/clientset_generated/release_1_5/fake) not found
```

Building with `godep go install` locally works correctly.

What am I missing?

uthark:
I performed a rebase, but I'm facing an issue that godep can't do a restore.
```
# cd /home/uthark/work/src/k8s.io/kubernetes; git checkout 9b7678f1086ce0f4980ecc78c7b6b20a97f8ee21
fatal: reference is not a tree: 9b7678f1086ce0f4980ecc78c7b6b20a97f8ee21
```
The full log is available here: https://gist.github.com/uthark/fa7ebc2c21bbeee8589e123756b2df8c

In `Godeps.json`
```
{
			"ImportPath": "k8s.io/kubernetes/cmd/kubeadm/app/apis/kubeadm",
			"Comment": "v1.6.0-alpha.0-1575-g9b7678f",
			"Rev": "9b7678f1086ce0f4980ecc78c7b6b20a97f8ee21"
},
```

I have full git clone of k8s in my work tree.

Am I missing something? Or is referenced revision is in unpublished branch?

Please help to resolve the issue.

mwielgus:
@uthark did you try to godep restore to a clean/new GOPATH?

mwielgus:
@uthark Something must have gone wrong with the last godep update - i will fix it.

uthark:
Ok, I'll do rebase on the master, and I see there are new methods in the NodeGroup (i.e. DecreaseTargetSize)

uthark:
Ok, I've fixed all go lint, go vet and boilerplates warnings. Travis is happy now. What to do with Github CI test? http://submit-queue.k8s.io/#/prs?prDisplay=2146&historyDisplay=2146 doesn't show any results. What are the next steps required from me?

mwielgus:
@brendandburns - Can you take another look?

mwielgus:
please squash the commits - there should be 1 or 2 commits (code + godeps).

uthark:
@mwielgus Rebased and squashed. Pull request updated, Travis is happy.

mwielgus:
LGTM, thank you for this contribution :)

PR #2164, opened 2016-12-16
Set the bazel build to be blocking.

spxtr:
It's been green, and we plan to remove unit tests from the test-go job since they're much more suited to the bazel job.


mikedanese:
LGTM

spxtr:
@rmmh who knows the magic incantations required to deploy the change.

PR #2165, opened 2016-12-16
Add more ingress tests to post merge blockers

bprashanth:
Presubmit is already running: https://github.com/kubernetes/kubernetes/pull/38141
Add the more extensive suite as a post merge blocker


bprashanth:
This has been stable (https://k8s-testgrid.appspot.com/google-gce#gci-gce-ingress), the intent is to run it like the slow suite (ignore flakes till they happen back to back, autofile issues, block merge on constant failure etc), except in its own project because it has specific requirements.

spxtr:
Seems fine to me.

/lgtm

PR #2169, opened 2016-12-16
Turn on branch protection in contrib

eparis:


foxish:
@grodrigues3, if I am an approver and I have put in `/lgtm`, would I need to do `/approve` separately?

grodrigues3:
yes, you would need to approve as well

PR #2173, opened 2016-12-17
handling the . case

grodrigues3:
The output of the call to `filepath.Dir()` for an empty string is `.` .  This results in an infinitely loop in the code.


mungegithub/features/repo-updates.go
```diff
@@ -265,7 +265,7 @@ func (o *RepoInfo) FindOwnersForPath(path string) string {
 
 	for {
 		// special case the root
-		if d == "" {
+		if d == "." || d == ""{
```
mtaufen: gofmt

mtaufen:
lgtm

PR #2182, opened 2016-12-20
Cluster-autoscaler: store scale down state in a struct.

mwielgus:


jszczepkowski:
LGTM

Consider adding unittests for TryToScaleDown method (in a subsequent PR).

PR #2184, opened 2016-12-20
Fix configMaps

eparis:
contrib configMap was out of date.
some configMaps didn't have protected-branches
update some help text


eparis:
@rmmh who did the approval-munger get turned on if the configmap wasn't updated?

eparis:
@foxish     preference for /approve and /lgtm to make sure the bot is running ok....

rmmh:
@grodrigues3 disabled the approval notifier because it was looping repeatedly, he should have more context.

eparis:
it's back off. and if we merge this it is off in git too!, so it won't come back.

@grodrigues3 this is the reason not to make changes to config without merging them to git...

eparis:
I'm an idiot. It was fixed int git, but not in production. So I put it back in git.  I'm an idiot.

foxish:
@eparis, Is this fix to the configmap already in the running version of the SQ?

eparis:
For contrib, yes.

PR #2185, opened 2016-12-20
Remove zoidbergwill from OWNERS

eparis:
@zoidbergwill is not in the org and thus the bot cannot assign them. So remove them from OWNERS


foxish:
@grodrigues3 This looks like another issue. The bot is suggesting the author as an approver.

eparis:
@emaildanwilson @zoidbergwill

zoidbergwill:
Thanks, I think?

PR #2187, opened 2016-12-20
Fix comment editing to use the right API.

rmmh:
It should be Issues.EditComment, not Repository.UpdateComment.

### DIFF
--- a/mungegithub/github/github.go
+++ b/mungegithub/github/github.go
@@ -232,7 +232,7 @@ type analytics struct {
 	ListReviewComments   analytic
 	CreateComment        analytic
 	DeleteComment        analytic
-	UpdateComment        analytic
+	EditComment          analytic
 	Merge                analytic
 	GetUser              analytic
 	SetMilestone         analytic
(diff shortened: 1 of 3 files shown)


rmmh:
Errors without this: 

    E1220 19:58:59.519577       1 github.go:1943] Error updating comment: PATCH https://api.github.com/repos/kubernetes/kubernetes/comments/268314989: 404 Not Found []

Woops!

eparis:
While you are in here....

When we glog.Infof() the `body` we need to truncate that a bunch.
Before we actually send it to github we need to truncate it to < 65536

We do both "correctly" in `WriteComment()`

eparis:
I also wonder if the truncating will mean that our comparator functions need some updating. But that's outside this particular area.

rmmh:
OK, I made it smarter about how it handles too-long comments.

We should probably also truncate the failure reason to a few KB-- nobody wants to read stacktrace novels in a comment.

mungegithub/github/github.go
```diff
@@ -1856,11 +1856,7 @@ func (obj *MungeObject) WriteComment(msg string) error {
 	config := obj.config
 	prNum := obj.Number()
 	config.analytics.CreateComment.Call(config, nil)
-	comment := msg
-	if len(comment) > 512 {
-		comment = comment[:512]
-	}
-	glog.Infof("Commenting in %d: %q", prNum, comment)
+	glog.Infof("Commenting in %d: %q", prNum, msg[:512])
```
eparis: Is this safe if len(msg) < 512?

eparis (changes requested):

mungegithub/github/github.go
rmmh: yes, it just becomes all of msg in that case

eparis: https://play.golang.org/p/Bl4MND4c4Q

rmmh: nope, I'm wrong :(

eparis (approved):

eparis:
This looks safe. I'm not convinced the sync-er code is 'correct' in the face of large comments, but this is definitely 'better'

eparis:
Who wrote these stupid bots? I said /lgtm AFTER the last code change (and before).  But it still removed my LGTM.

PR #2189, opened 2016-12-20
Fix docs and !GateApproval

eparis:
The main change is the "docs" to show that approval either is or isn't
required. These are displayed on the info tab of the bot.

While in here it became clear that we were hiding such simple logic
oddly and incorrectly. Even if `GateApproved` was false, if a PR had the
approved label and had been changed since we would still reject it.

This PR attempts to make the logic a bit more clear. Although we still
do not have any tests for the GateApproved=false case.

### DIFF
--- a/mungegithub/github/github.go
+++ b/mungegithub/github/github.go
@@ -945,6 +945,23 @@ func (obj *MungeObject) RemoveLabel(label string) error {
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


eparis:
hopefully goling will like me again.

grodrigues3:
@spxtr can I add comments after lgtm as above?

PR #2202, opened 2016-12-22
Use default etcd port (2379) in systemd service

GheRivero:
etcd changed its default port to 2379. Update apiserver configuration
for systemd to add a etcd node listening in that port, and not removing
the previous one for a smooth transition.


GheRivero:
I signed it!

PR #2204, opened 2016-12-22
Detailed NGINX-Ingress-Controller status page config

ysolt:
Add detailed example for NGINX-Ingress-Controller status page configuration


bprashanth:
Thanks for the pr, can you please move it to https://github.com/kubernetes/ingress/tree/master/examples (in examples/status-page/nginx/, we would presumably add a example/status-page/{aws,gce..etc} as well, though I'm not sure it's worth it since it's pretty obvious how to visit the status page on a cloudprovider)

ysolt:
@bprashanth alright! Already opened a new PR, please find it here https://github.com/kubernetes/ingress/pull/84

PR #2208, opened 2016-12-22
symbol errors

tim-zju:
fix symbol errors: change ‘ to ' 
Signed-off-by: tim-zju <21651152@zju.edu.cn>


foxish:
@tim-zju This changes files in vendor. I don't think we want that. 
Can you explain why you think this change is needed?

tim-zju:
do you mean that i don't need to change files in vendor folders? How about other files?

eparis:
Changes to files under `vendor` or `Godeps/` directories are prohibited. Those files are actually copies from other upstream projects. Any fixes like these would need to go directly to those upstream projects.

Changes like this to files outside of `vendor` and `Godeps/` are very welcome! If you can revert any changes in those two directories it would be greatly appreciated. Thank you!

foxish:
Closing. Please reopen if you're updating the PR. @tim-zju

PR #2211, opened 2016-12-22
Fixed issue with "become" behavior in checking the local addons direc…

jberkus:
…tory, in cases where the user sets ansible_become in global_vars.


jberkus:
Sorry, thought I was creating this against my fork.  This isn't tested yet, I'll reopen it once it is.

jberkus:
OK, testing complete now.

This fix resolves an issue where, if the user sets ansible_become in global_vars, the current Ansible role will try to sudo on localhost before checking for the addons directory, causing the check to fail.  After this change, the directory check will continue to work even if the user sets ansible_become globally.

dustymabe:
tested and works for me

PR #2212, opened 2016-12-22
Correctly forward X-Forwarded-Port

alexw23:
It appears this package is assuming the port running on the Ingress Controller as the port.

In the current situation, if my LB sends `X-Forwarded-Port: 443` and the ingress controller is running on port 80. `X-Forwarded-Port: 80` will end up in my app.

Now it's correctly routing.


alexw23:
I signed it!

aledbf:
@alexw23 please check the new version https://github.com/kubernetes/ingress
We are moving the code to that repository.

PR #2213, opened 2016-12-22
Added default_server option for the default server

alexw23:
If you're using wildcard host names the default backend will no longer be available due to the way nginx has predence on wildcards. By providing `default_server` it indicates a catch-all so regardless of wildcards if there is no direct match they will end up in default server block.


aledbf:
@alexw23 please check the new version https://github.com/kubernetes/ingress
We are moving the code to that repository.

PR #2224, opened 2016-12-28
Release version 0.6

aledbf:
Check https://quay.io/repository/aledbf/ubuntu-slim?tab=tags


ixdy:
/lgtm

though I don't know if that's good enough anymore.

grodrigues3:
@ixdy It should be good enough.  There's seems to be some problem with the approvers-handlers detecting its old comments, so turning it off for now.

ixdy:
Pushed `gcr.io/google_containers/ubuntu-slim:0.6`.

aledbf:
@ixdy thanks!

ixdy:
though it occurs to me that you might not need `RUN sed -i 's/bash/sh/g' /etc/passwd` anymore.

aledbf:
@ixdy you are right. PR #2267 fixes that

PR #2225, opened 2016-12-28
[nginx-slim]: Release v0.12

aledbf:


aledbf:
@lavalamp please assign this to @bprashanth

bprashanth:
Any reason to not move this to the ingress repo?

aledbf:
@bprashanth release history? I can try to do the same procedure when I migrated the ingress directory.

aledbf:
@bprashanth I think kubernetes/ingress is not the right place for this. Right now exists a directory only for images here. If we move only this image or with ubuntu-slim then the other images in this repo will depend on a different repo

ixdy:
`gcr.io/google_containers/nginx-slim:0.12` has been pushed.

aledbf:
@ixdy thanks!

PR #2227, opened 2016-12-28
Removing dnsmasq-metrics; it now lives in kubernetes/dns

bowei:
@thockin @MrHohn


thockin:
One less directory in contrib!

thockin:
Two, in fact!

PR #2242, opened 2017-01-03
Updates the TempFile Location from NGINX Ingress Controller

rikatz:
This commit updates the TempFile Location from SSL Certificates. This is necessary, as the os.Rename directive doesn't work on different mount points.

As this program is being used also on bare metal machines (and not only containers), the temp file should be created on the same directory of the final SSL certificates.

### DIFF
--- a/ingress/controllers/nginx/nginx/ssl.go
+++ b/ingress/controllers/nginx/nginx/ssl.go
@@ -36,7 +36,7 @@ func (nginx *Manager) AddOrUpdateCertAndKey(name string, cert string, key string
 	temporaryPemFileName := fmt.Sprintf("%v.pem", name)
 	pemFileName := fmt.Sprintf("%v/%v.pem", config.SSLDirectory, name)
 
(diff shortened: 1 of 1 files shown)


aledbf:
@rikatz the development of ingress is now located in [kubernetes/ingress](https://github.com/kubernetes/ingress/blob/master/core/pkg/net/ssl/ssl.go)
Do you mind opening the PR in that repo and adding a note about the rename restriction?

rikatz:
@aledbf Sure I can open there.

I've noticed that there is no pointing from k8s.io/ingress to the repo github.com/kubernetes/ingress, will do some symlinks here, but would be nice if someone could make this link!

I'll close this PR and work on that repo! Thanks!

aledbf:
> I've noticed that there is no pointing from k8s.io/ingress to the repo github.com/kubernetes/ingress, 

that's because most of the repos in the kubernetes organization follow the [development workflow](https://github.com/kubernetes/community/blob/master/contributors/devel/development.md#workflow)

PR #2243, opened 2017-01-03
ansible: restore selinux to enforcing

jasonbrooks:
Rather than setting selinux to permissive on all nodes, leave selinux enforcing and specify the unconfined spc_t container type for the kube-addons containers. For more on spc_t, see: http://danwalsh.livejournal.com/74754.html. See also https://github.com/kubernetes/kubernetes/pull/37327.

I tested this with centos atomic host.


ingvagabund:
@redhat-k8s-bot test please

jasonbrooks:
It looks like this failed for fedora on an error syncing the updates-testing repo on one of the nodes -- I don't think this is related to the change. Also, is updates-testing what we want for fedora?

PR #2245, opened 2017-01-03
SQ: Add kops-aws back to blocking

zmerlynn:
Since I attempted this last, we have..
- logging on Up failures as well
- An AWS resource janitor to make sure we don't explode completely

This is a follow-on to kubernetes/test-infra#1471, which reenables reporting of the PR builder. ~~I'm waiting for the latest failure to finish clearing before merging that one, and I'll wait for the PR builder to go (more) green before merging this. (The `kops-aws-updown` job has cleared already, the `kops-aws` job should be green in the next half hour.)~~

I'd like to get this into blocking today, since basically after every weekend I end up fighting the next breakage in `kops-aws`.

Along the way: Reformat `required-retest-contexts` as well (separate commit).


zmerlynn:
cc @justinsb @chrislovecnm @kris-nova

PR #2252, opened 2017-01-04
Switch to new batch URL in submit queue.

spxtr:


rmmh:
This isn't equivalent. Previously, it would link to just the currently running batch. This shows all batch results.

rmmh:
If you're going to make this change, you should remove the newly dead code that figured out what to link to.

rmmh:
And probably put the link in "CURRENTLY RUNNING BATCH" instead of the specific PRs-- those PRs could then become Github links.

spxtr:
Removed the dead code. I'm not sure that it's right to link anywhere else.

PR #2226, opened 2016-12-28
cluster-autoscaler: Dynamic Reconfiguration via ConfigMaps

mumoshu:
This is the first iteration to address #2181.
Please read the discussion in the issue to see more context.

@andrewsykim @Raffo @mwielgus Would you mind reviewing this? :bow:

TODOs:

- [x] Add unit tests

Non-TODOs:

- Automatic creation of configmaps filled with defaults, like cluster-proportional-autoscaler does with `--configmap` with `--default-params`. IMHO doing so is a completely orthogonal issue therefore won't be addressed in this PR.

---

This PR adds a new optional flag named `configmap` to specify the name of a configmap containing node group specs.

The configmap is polled every `scan-interval` seconds to reconfigure cluster-autoscaler dynamically at runtime.

Example usage:

```
./cluster-autoscaler --v=4 --cloud-provider=aws --skip-nodes-with-local-storage=false --logtostderr --leader-elect=false --configmap=cluster-autoscaler --logtostderr
```

The configmap would look like:

```yaml
kind: ConfigMap
apiVersion: v1
metadata:
  name: cluster-autoscaler
  namespace: kube-system
data:
  settings: |-
    {
      "nodeGroups": [
        {
          "minSize": 1,
          "maxSize": 2,
          "name": "kubeawstest-nodepool1-AutoScaleWorker-1VWD4GAVG35L5"
        }
      ]
    }
 ```

PR #2255, opened 2017-01-05
Cluster-autoscaler: add stop channel to listers

mwielgus:
So that the whole CA loop can be easily stopped and restarted with new settings. This will be handy for ConfigMap PR: https://github.com/kubernetes/contrib/pull/2226


jszczepkowski:
LGTM

PR #2229, opened 2016-12-29
Cluster-autoscaler: cluster state registry proposal

mwielgus:
The proposal that addresses #2228.

cc: @fgrzadkowski @jszczepkowski @piosz @mumoshu @andrewsykim @kubernetes/sig-autoscaling-misc

Issue #2228, opened 2016-12-29
Cluster-autoscaler: handle unready nodes


mwielgus:
### Introduction

Currently ClusterAutoscaler stops working when the number of nodes observed on the cloud provider side doesn’t match to the number of ready nodes on the Kubernetes side. This behavior was introduced in the first version of CA in order to prevent CA from breaking more the already broken cluster. However the client reported feedback indicated that this behavior is oftentimes suboptimal and leads to confusion. This document sketches how the problem of unready nodes will be solved in the next release of K8S.

### Use cases

The number of ready nodes can be different than on the mig side in the following situations:

* [UC1] A new node is being added to the cluster. The node group has been increased but the node has not been created/started/registered in K8S yet. On GCP this usually takes couple minutes. 
Indicating factors:
 -- There was a scale up in the last couple minutes.
 -- The number of missing node is at most the size of executed scale-up.
Suggested action: Continue operations, however include all yet-to-arrive nodes in all scale-up considerations.

* [UC2] A new node is being added to the cluster. The node has registered on the cluster but has not yet switched its state to ready. This should be fixed in couple seconds. Indicating factors:
 -- The unready node is new. CreateTime in the last couple minutes.
Suggested action: Continue operations, however include all yet-to-arrive nodes in all scale-up considerations.

* [UC3] A new node was added to the cluster but failed to start within the reasonable time. There is little chance that it will start anytime soon. Indicating factors:
 -- Node is unready
 -- CreateTime == unready NodeCondition.LastTransitionTime
Suggested action: Continue operations, however do not expand this node pool. The probable scenario is that the node will be picked by scale down soon.

* [UC4] A new node is being added to the cluster. However the cloud provider cannot provision the node within the reasonable time due to either no quota or technical problems. Indicating factors:
 -- The target number of nodes on the cloud provider side is greater than the number of nodes in K8S for the prolonged time (more than couple minutes) and the difference doesn’t change.
Suggested action: Reduce the target size of the problematic node group to the current size. 

* [UC5] A node is in an unready state for quite a while (+20min) and the total number of unready/not-present nodes is low (less than XX%). It could either not switched from unready to ready on node registration or something crashed on the node and could not be recovered. Indicating factors:
-- Node condition is unready and last transition time is >= 20 min.
-- The number of TOTAL nodes in K8S is equal to the target number of nodes on the cloud provider side. 
Suggested action: Include the node in scale down, although with greater (configurable) unneeded time.

[...]

PR #2260, opened 2017-01-09
Cluster-autoscaler: skip unready node groups in scale up

mwielgus:
Ref: #2228 #2229

cc: @jszczepkowski @fgrzadkowski @piosz


jszczepkowski:
LGTM

PR #2270, opened 2017-01-10
[mungegithub] Edit the approval notification

grodrigues3:
- rather than create a new approval notification and potentially spam
  PRs, use edit api functionality to edit original comment

### DIFF
--- a/mungegithub/mungers/approval-handler.go
+++ b/mungegithub/mungers/approval-handler.go
@@ -128,8 +128,10 @@ func (h *ApprovalHandler) updateNotification(obj *github.MungeObject, ownersMap 
 	notifications := c.FilterComments(comments, notificationMatcher)
 	latestNotification := notifications.GetLast()
 	if latestNotification == nil {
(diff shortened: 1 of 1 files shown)


grodrigues3:
@rmmh when you edit a comment on github, it doesn't send a new email right?

eparis:
That is correct. No new comment.

Which might ruin the whole point, since you were using that comment stream to tell people they needed to approve.

Now we're back to the original problem with approvals. A good way to see who needs/can approve a given a PR (we'll at least have a subset) and a good way to know what PRs I need to approve.

I think edit + some non-existant dashboard is the best solution, but I might have just not thought of the best...

grodrigues3:
The comment will be updated and there was never the intention to have a stream of comments.  I think one clear comment indicating which files are approved and which need approval (and by whom).  This PR achieves that.  What will the dashboard give us that this comment won't?

rmmh:
/lgtm

we can integrate approvals into the gubernator dashboard once it's stable

PR #2277, opened 2017-01-11
make docker configuration  more robust

xialonglee:
this commit makes sure the line `. /run/flannel/subnet.env` was written before `DOCKER_OPTS...` in the docker configuration file

**why we need this**
1. People could install k8s without install docker by adding flag `--skip-tags=docker`
So here comes the potential problem, once the docker was installed before and configured, the prefix string `DOCKER_OPTS=` may already exsit , the string ` /run/flannel/subnet.env` may not exsit.
in this situation,  ` /run/flannel/subnet.env`  will written in the end of file after `DOCKER_OPTS=` so that the network configuration not work.

2. this commit will first remove strings that satisfied the regex pattern and then written the line we want and make lines sequence is right.


xialonglee:
cc @ingvagabund

ingvagabund:
Hey @xialonglee 

it looks fine. Just wondering how to test this in CI. I don't have automatic test for Ubuntu. Aware of any tips how to test Ubuntu's side of the ansible?

xialonglee:
@ingvagabund 

i tested it manually using my local ubuntu 14.04 enviroment with command 
`./deploy-cluster.sh --skip-tags=docker,addons -c paramiko` .

 `-c paramiko` was added to solve a problem may happened in Ubuntu 14.04, see isues:  https://github.com/ansible/ansible/issues/14020 
https://github.com/ansible/ansible/issues/13278 
Ubuntu which main version under 15 uses `Upstart` and greater or equal to 15 uses `systemd` as init manager.
This pr only work under Ubuntu which use `Upstart` and only tested it in ubuntu 14.04 env is enough, because this task was originally added for Ubuntu 14.04 support ,see the pr which generated this file #588   
i think you could make a virtual box of Ubuntu 14.04 to test it. i don't know how to make a automatic test for Ubuntu.

ingvagabund:
What version of ansible did you use to deploy the cluster?

xialonglee:
@ingvagabund 
It's ansible 2.2.0.0.

xialonglee:
ping

ingvagabund (approved):

PR #2279, opened 2017-01-11
Redis cache

dhilipkumars:
Hi

Initial PR for redis-cache (no persistence) with automatic master-slave promotion.  

If you think this looks okay, i will add the source-code and docker files for other utilities.  

for this petset to work, i have changed the peer-finder little bit, i think it will be better if we factor-out peer-finder as a package instead of a utility.  I could do this as a part of separate PR.

Thanks in Advance,
Dhilip
@erictune @bprashanth


dhilipkumars:
I signed it!

dhilipkumars:
Closing in favor of https://github.com/kubernetes/contrib/pull/2301, an attempt to squash a couple of commits resulted in an ugly PR.

PR #2286, opened 2017-01-11
Enabled approvers and turned of cache for contrib

grodrigues3:
Looks like the http cache was responsible for the duplicate comments on contrib as ListComments() was returning an empty set of Comments on Issues where the bot had commented before.  This is not a permanent/good solution as we will hit the github rate limit on kubernetes/kubernetes but for the purposes of testing on contrib, it may be okay.

*Verified that the bot is no longer double commenting on PRs

### DIFF
--- a/mungegithub/submit-queue/deployment/contrib/configmap.yaml
+++ b/mungegithub/submit-queue/deployment/contrib/configmap.yaml
@@ -4,10 +4,10 @@ metadata:
   name: contrib-sq-config
 data:
   # basic config options.
-  config.http-cache-dir: /cache/httpcache
+  config.http-cache-dir:
   config.organization: kubernetes
   config.project: contrib
-  config.pr-mungers: blunderbuss,lgtm-after-commit,submit-queue,needs-rebase
+  config.pr-mungers: blunderbuss,lgtm-after-commit,submit-queue,needs-rebase,approval-handler
   config.state: open
   config.token-file: /etc/secret-volume/token
   gitrepo.repo-dir: /gitrepos
@@ -50,4 +50,4 @@ data:
   alias.alias-file: "/gitrepos/contrib/OWNERS_ALIASES"
 
   # Temporary gate approval process
-  approval-activated: "false"
+  approval-activated: "true"


grodrigues3:
Looking at http://contrib.submit-queue.k8s.io/#/info
 
The last count I saw for uncached calls to github's api was 676 and since the bot runs every 10 minutes, that will result in (60/10)*676 = 4056 calls per hour.  Not great, but under the limit

foxish:
If rhe contrib submit queue alone makes that many Api calls, that would be undesirable. The 5k per hour limit applies across all the submit queue.

foxish:
@grodrigues3 and I discussed and we'll turn the cache back on if we start seeing the limit being hit. Monitoring API token usage at velodrome.kubernetes.io/dashboard/db/monitoring

@lavalamp

eparis:
I haven't looked at this yet, but turning off the cache is almost certainly wrong. We should be working with github support and getting them to fix their api...

eparis:
So I removed LGTM. Can you show me the reason we are blaming the cache and the response we got from github?

lavalamp:
That was very sloppy of me, I didn't notice the cache being turned off. I agree w/ @eparis, we shouldn't turn that off.

foxish:
I think it was meant to be experimental while we figure out the cache issue; and I think the currently running contrib submit queue has its cache turned off. http://velodrome.kubernetes.io/dashboard/db/monitoring shows we didn't run close to the limits.
@grodrigues3 can confirm.

grodrigues3:
correct, it was intended to be experimental.

grodrigues3:
@eparis @lavalamp 
So I need some debugging help regarding the double postings.  This log is from a run with the cache disabled and approvers running (as merged in master)

For #2289 the approvers-handler commented twice, but it seems that it only did so after the PR was merged.  Given the approvers code that handles [notifications](https://github.com/kubernetes/contrib/blob/master/mungegithub/mungers/approval-handler.go#L130), it should only ever add a new comment when it does not find an older comment (i.e `latestNotification = nil`). 

As implemented, if the PR is modified after approval or if someone approves, approvers only edits the existing comment (`EditComment`) and it should add information why it's editing the comment (either "Latest approve was after last time notified" or "PR Modified After Last Notification").    Neither of those happen in the following log 

http://pastebin.com/W3wL9ydS

So I'm hypothesizing that ListComments is returning an empty list for some reason (seemingly after the PR is modified, since it's not double commenting at any other time).  I've seen it happen once before, but I can't reproduce regularly.  I'd really appreciate any debugging help, suggestions.

grodrigues3:
On another note, is anyone else in favor of creating a dummy repo for bot testing purposes?  Since I can't always reproduce the bots behavior with local runs, it'd be great if I could deploy it against a repo for experiments like these without disabling, harming the SQ.

grodrigues3:
Code I added to approval-handler when running against my repo https://github.com/grodrigues3/contrib/pull/3 

`glog.Infof("Found %v comments for issue %d, %d of which were written by the bot", len(comments), *obj.Issue.Number, len(notifications))
`

Output
`

> I0112 12:35:40.298398  108848 approval-handler.go:131] Found 0 comments for issue 3, 0 of which were written by the bot


`

eparis:
So that proves it's not a cache issue?  Are we using the right comment lister? I know there is a PR comments list and an issue comments list. I think both are exposed in github.go...

grodrigues3:
I'm using 
https://github.com/kubernetes/contrib/blob/master/mungegithub/github/github.go#L1860

grodrigues3:
Closing because investigation showed the http cache wasn't the problem

PR #2290, opened 2017-01-12
Cluster-autoscaler: return node names in format of gce://proj/zone/name from GceProvider

mwielgus:
cc: @jszczepkowski  @fgrzadkowski


jszczepkowski:
LGTM

PR #2291, opened 2017-01-12
Cluster-autoscaler: unregistered nodes in cluster state registry

mwielgus:
Ref: #2228 #2229

### DIFF
--- a/cluster-autoscaler/clusterstate/clusterstate.go
+++ b/cluster-autoscaler/clusterstate/clusterstate.go
@@ -24,7 +24,9 @@ import (
 
 	"k8s.io/contrib/cluster-autoscaler/cloudprovider"
 	"k8s.io/contrib/cluster-autoscaler/utils/deletetaint"
+
 	apiv1 "k8s.io/kubernetes/pkg/api/v1"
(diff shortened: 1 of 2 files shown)


cluster-autoscaler/clusterstate/clusterstate.go
```diff
@@ -401,3 +449,34 @@ func (csr *ClusterStateRegistry) GetUpcomingNodes() map[string]int {
 	}
 	return result
 }
+
+// Calculates which of the existing cloud provider nodes are not registered in Kuberenetes.
+func getNotRegisteredNodes(allNodes []*apiv1.Node, cloudProvider cloudprovider.CloudProvider, time time.Time) ([]UnregisteredNode, error) {
+	registered := make(map[string]struct{})
```
jszczepkowski: Please, use sets.String from pkg/util/sets

jszczepkowski:
Just one comment, LGTM after applying it.

cluster-autoscaler/clusterstate/clusterstate.go
mwielgus: Done.

PR #2292, opened 2017-01-12
Cluster-autoscaler: remove nodes that have been unregistered for a long time

mwielgus:
Ref: #2228

### DIFF
--- a/cluster-autoscaler/cluster_autoscaler.go
+++ b/cluster-autoscaler/cluster_autoscaler.go
@@ -95,14 +95,15 @@ var (
 		"Node utilization level, defined as sum of requested resources divided by capacity, below which a node can be considered for scale down")
 	scaleDownTrialInterval = flag.Duration("scale-down-trial-interval", 1*time.Minute,
 		"How often scale down possiblity is check")
-	scanInterval               = flag.Duration("scan-interval", 10*time.Second, "How often cluster is reevaluated for scale up or down")
-	maxNodesTotal              = flag.Int("max-nodes-total", 0, "Maximum number of nodes in all node groups. Cluster autoscaler will not grow the cluster beyond this number.")
-	cloudProviderFlag          = flag.String("cloud-provider", "gce", "Cloud provider type. Allowed values: gce, aws")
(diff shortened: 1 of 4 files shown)


cluster-autoscaler/cluster_autoscaler.go
```diff
@@ -273,6 +275,24 @@ func run(_ <-chan struct{}) {
...
 				}
 
+				// Check if there are any nodes that failed to register in kuberentes
+				// master.
+				unregisteredNodes := autoscalingContext.ClusterStateRegistry.GetUnregisteredNodes()
+				if len(unregisteredNodes) > 0 {
+					glog.V(1).Infof("%d unregistered nodes present", len(unregisteredNodes))
+					removedAny, err := removeOldUnregisteredNodes(unregisteredNodes, &autoscalingContext, time.Now())
+					// There was a problem with removing unregistered nodes. Retry in the next loop.
+					if err != nil {
```
jszczepkowski: You are not handling correctly the case when some nodes were removed, but error was observed afterwards:  "Some unregistered nodes were removed" should be printed before printing error.

cluster-autoscaler/utils_test.go
```diff
@@ -51,3 +55,46 @@ func TestFilterOutSchedulable(t *testing.T) {
...
+	context := &AutoscalingContext{
+		CloudProvider:               provider,
+		ClusterStateRegistry:        clusterState,
+		UnregisteredNodeRemovalTime: 45 * time.Minute,
+	}
+	unregisteredNodes := clusterState.GetUnregisteredNodes()
+	assert.Equal(t, 1, len(unregisteredNodes))
+
+	// Nothing should be removed. The unregistered node is not old enough.
+	removed, err := removeOldUnregisteredNodes(unregisteredNodes, context, time.Now().Add(-50*time.Minute))
```
jszczepkowski: Nit: do not call Now() multiple times. Call it once and remember the value.

jszczepkowski:
Just two minor comment, after applying them: LGTM

cluster-autoscaler/cluster_autoscaler.go
mwielgus: Do you mean the error itself message. Yeah - we can have 2 different here.

PR #2296, opened 2017-01-13
Don't cache results across function calls for ListComments

grodrigues3:
So good news, I found the problem with ListComments and it wasn't with the httpcache, with github's api, or the approval handler.

Turns out we had implemented an incorrect cache in the ListComments wrapper, and it was returning the result from call to the function with different options.

More Info:
1. The first time `ListComments` gets called in the approvers implementation (via `getCommentsAfterLastModified`) it has the since parameter set in the `IssueListCommentsOptions` 
1. The second time we call` ListComments()` with no parameters expecting ALL Comments on the PR but instead we get the results from the first call

### DIFF
--- a/mungegithub/github/github.go
+++ b/mungegithub/github/github.go
@@ -1857,7 +1857,7 @@ func (obj *MungeObject) ListReviewComments() ([]*github.PullRequestComment, bool
 type WithListOpt func(*github.IssueListCommentsOptions) *github.IssueListCommentsOptions
 
 // ListComments returns all comments for the issue/PR in question
-func (obj *MungeObject) ListComments(withListOpts ...WithListOpt) ([]*github.IssueComment, bool) {
+func (obj *MungeObject) ListComments() ([]*github.IssueComment, bool) {
 	config := obj.config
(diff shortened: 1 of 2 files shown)


rmmh:
This is an important optimization for a lot of other mungers, I would add an explicit comment clearing call instead.

rmmh:
Maybe make it only store or use obj.comments if `len(withListOptions)==0` ?

grodrigues3:
interesting should we try and require that `obj.comments` if there are no listoptions

rmmh:
No, because it's valid to call this without a previous cached call.

eparis:
If we have obj.comments set, and we have listoptions, should we do that matching ourselves? or make a github api call?

grodrigues3:
Alright, then I'm going to set `obj.ListComments` to always return a list of all the comments on an issue and we can do the sorting and filtering separately on our end.  Does that sound reasonable?  That way `obj.comments` always has all of the comments associated with an issue or PR

eparis:
/approve
just in case

PR #2299, opened 2017-01-13
Cluster-autoscaler: clear to be deleted taints only at the startup

mwielgus:
We want to clean the taint only at the startup. Currently it will remove the taint also from nodes that have just been deleted but not marked as unready by node controller. 


jszczepkowski:
LGTM

PR #2300, opened 2017-01-13
Cluster-autoscaler: decrease target size function in cloud provider interface

mwielgus:
Sometimes the cloud provider has troubles with providing the requested amount of nodes either due to quota or to (temporary) technical limitations. In such a case CA should be able to change the previously made request without killing any of the existing nodes. 

@andrewsykim - will this PR work on AWS?

cc: @jszczepkowski @andrewsykim @fgrzadkowski

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
+++ b/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
@@ -142,6 +142,30 @@ func (asg *Asg) IncreaseSize(delta int) error {
 	return asg.awsManager.SetAsgSize(asg, size+int64(delta))
 }
 
+// DecreaseTargetSize decreases the target size of the node group. This function
+// doesn't permit to delete any existing node and can be used only to reduce the
+// request for new nodes that have not been yet fulfilled. Delta should be negative.
+// It is assumed that cloud provider will not delete the existing nodes if the size
+// when there is an option to just decrease the target.
+func (asg *Asg) DecreaseTargetSize(delta int) error {
+	if delta >= 0 {
+		return fmt.Errorf("size decrease size must be negative")
+	}
+	size, err := asg.awsManager.GetAsgSize(asg)
+	if err != nil {
+		return err
+	}
+	nodes, err := asg.awsManager.GetAsgNodes(asg)
+	if err != nil {
+		return err
+	}
+	if int(size)+delta < len(nodes) {
+		return fmt.Errorf("attempt to delete existing nodes targetSize:%d delta:%d existingNodes: %d",
+			size, delta, len(nodes))
+	}
+	return asg.awsManager.SetAsgSize(asg, size+int64(delta))
+}
+
 // Belongs returns true if the given node belongs to the NodeGroup.
 func (asg *Asg) Belongs(node *apiv1.Node) (bool, error) {
 	ref, err := AwsRefFromProviderId(node.Spec.ProviderID)
--- a/cluster-autoscaler/cloudprovider/gce/gce_cloud_provider.go
+++ b/cluster-autoscaler/cloudprovider/gce/gce_cloud_provider.go
@@ -146,6 +146,28 @@ func (mig *Mig) IncreaseSize(delta int) error {
 	return mig.gceManager.SetMigSize(mig, size+int64(delta))
 }
 
+// DecreaseTargetSize decreases the target size of the node group. This function
+// doesn't permit to delete any existing node and can be used only to reduce the
+// request for new nodes that have not been yet fulfilled. Delta should be negative.
+func (mig *Mig) DecreaseTargetSize(delta int) error {
+	if delta >= 0 {
+		return fmt.Errorf("size decrease must be netative")
+	}
+	size, err := mig.gceManager.GetMigSize(mig)
+	if err != nil {
+		return err
+	}
+	nodes, err := mig.gceManager.GetMigNodes(mig)
+	if err != nil {
+		return err
+	}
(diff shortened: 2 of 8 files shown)


cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
```diff
@@ -142,6 +142,30 @@ func (asg *Asg) IncreaseSize(delta int) error {
...
+	size, err := asg.awsManager.GetAsgSize(asg)
+	if err != nil {
+		return err
+	}
+	nodes, err := asg.awsManager.GetAsgNodes(asg)
+	if err != nil {
+		return err
+	}
+	if int(size)+delta < len(nodes) {
+		return fmt.Errorf("attempt to delete existing nodes targetSize:%d delta:%d existintNodes: %d",
```
andrewsykim: `s/existintNodes/existingNodes`

cluster-autoscaler/cloudprovider/gce/gce_cloud_provider.go
```diff
@@ -146,6 +146,28 @@ func (mig *Mig) IncreaseSize(delta int) error {
 	return mig.gceManager.SetMigSize(mig, size+int64(delta))
 }
 
+// DecreaseTargetSize decreases the target size of the node group. This function
+// doesn't permit to delete any existing node and can be used only to reduce the
+// request for new nodes that have not been yet fulfilled. Delta should be negative.
+func (mig *Mig) DecreaseTargetSize(delta int) error {
+	if delta <= 0 {
```
andrewsykim: should be `delta >= 0 `?

andrewsykim (commented):
@mwielgus I don't think this approach would work on AWS as well as it would on GCE. Though from my experience its rare to get rate limited for ec2 instances on AWS unless you're on spot instances. 

If you increase the ASG size from 5 -> 10 and end up with 8 nodes after the timeout and then set the desired size back to 8 there could be a chance the autoscaling group may attempt to delete nodes from the existing 8 nodes. There are a lot of edge cases to consider. I'll get into contact with some AWS support people to confirm.

cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
mwielgus: Done.

cluster-autoscaler/cloudprovider/gce/gce_cloud_provider.go
mwielgus: Yes, obviously. Thanks for catching this.

mwielgus:
@andrewsykim - I made some experiments/checks on GCE and it seems to delete "uncreated" instances first. I will double check with people who actually wrote this code but it seems to be a reasonable strategy to recover from quota overrun. How otherwise it should be done?

andrewsykim:
@mwielgus I discussed this with AWS support. Decreasing the ASG size would work for the most part if there are no instances launching. I'm still a bit hesitate because its not entirely clear when a instance is "launching" in the AWS world (when the request was received vs when the VM is actually launching). There's a race condition to consider, it seems possible that you could decrease the ASG size as soon as the instances you waited for launched, in which case its very likely that you would terminate an existing node (with no drain). However, this race condition seems possible in GCE too right? There are a few alternatives I can think of at the moment but it would involve a fair bit of refactoring. Given that its unlikely, it might be worth it to go forward with the current implementation and reconsider those alternatives once we understand the likelyhood of those race conditions.

mwielgus:
Yes, I'm aware of the potential race condition here. To mitigate the risk we will only do it if the difference between the number of started nodes and the target was constant for a prolonged time (like 15 min). 

What alternatives do you have on your mind?

mwielgus:
On GCE you can find all instances managed by a node-group and delete the uncreated one 

$ gcloud compute instance-groups managed list-instances gke-cluster-1-default-pool-96a2740c-grp 

NAME                                      ZONE            STATUS   ACTION    LAST_ERROR
gke-cluster-1-default-pool-96a2740c-13xf  europe-west1-c           CREATING  Error QUOTA_EXCEEDED: Instance 'gke-cluster-1-default-pool-96a2740c-13xf' creation failed: Quota 'IN_USE_ADDRESSES' exceeded.  Limit: 23.0

$ gcloud compute instance-groups managed delete-instances gke-cluster-1-default-pool-96a2740c-grp --instances=gke-cluster-1-default-pool-96a2740c-48g8

Is there such a way in AWS?

andrewsykim:
@mwielgus not that I'm aware of. The alternative I had in mind was enabling [instance protection](http://docs.aws.amazon.com/autoscaling/latest/userguide/as-instance-termination.html#instance-protection). Only instances that are `InService` can qualify for instance protection and we can enable/disable instance protection for individual nodes. If we enable instance protection for any node visible to CA, it can't be terminated by decreasing the desired size unless we explicity do the [TerminateInstance](https://github.com/kubernetes/contrib/blob/master/cluster-autoscaler/cloudprovider/aws/aws_manager.go#L155) operation.

mwielgus:
@andrewsykim What about turning on instance protection automatically on all registered/ready nodes?

mwielgus:
@andrewsykim Well that what we want there - to delete the nodes about which we are sure that we want to delete them. Do you think that such a side effect (automatic protection on all registered/ready) nodes would be acceptable for the AWS users?

andrewsykim:
At first thought, if its well documented and there's logging for it on the CA, then I don't see it being a problem. But let me think about it and ask others for their opinions. :)

jszczepkowski:
LGTM

PR #2301, opened 2017-01-13
redis cache Statefulset

dhilipkumars:
Hi

Initial PR for redis-cache statefulset

1) Automatic slave-promotion in case of master failure.
2) Around 3 times faster writes than the current [petset version](https://github.com/kubernetes/contrib/tree/master/pets/redis)
3) Extremely light weight, installs in around 30 - 45 secs
4) over all image size is less than 30MB.
5) Is a statefulset with anti affinity and PodDisruption Budget

#### Quick Benchmark comparing with and without persistence. 
Redis Petset with persistance enabled
```
$ k exec rediscli -- redis-benchmark -q -h rd-0.redis.default.svc.cluster.local -p 6379 -t set,get -n 100000 -d 100 -r 1000000
SET: 22026.43 requests per second
GET: 76161.46 requests per second
```
Redis Statefulset configured without persistance
```
$ k exec rediscli -- redis-benchmark -q -h cache-0.cache.default.svc.cluster.local -p 6379 -t set,get -n 100000 -d 100 -r 1000000
 SET: 60060.06 requests per second
 GET: 89285.71 requests per second
```

Thanks in Advance,
Dhilip

@erictune @kow3ns

PS: Was originall raised as https://github.com/kubernetes/contrib/pull/2279


kow3ns:
First, one major concern. You have code here called Redis Sentinel Micro, **but this code is not Redis source code and it is not Redis Sentinel**. I think this should be renamed because it misleading.  Also, do you have permission from the Redis authors to use this name? If not, that is probably a larger issue.

Second, you've implemented your own leader election software for Redis replicas. I think that this is a contribution that is worthy of its own repository, and not something that should be hosted in the K8s contrib repository. Also, I have concerns about using a non-standard leader elector for Redis in general. I understand that your objective is to provide the smallest image possible, but replacing the standard, well known, and well tested Sentinel implementation with your own implementation is an optimization that most other users would not want. Would it be possible to default to the standard Sentinel implementation for this contribution?

 

---

Reviewed 5 of 10 files at r1, 1 of 1 files at r2.
Review status: 6 of 10 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.kubernetes.io:443/reviews/kubernetes/contrib/2301#-:-KbBNoA5hOFwxetG22AA:bsbinyu)*

dhilipkumars:
@kow3ns  Thanks for looking at the PR,  

First concern:- I will get back to you on that.  (I've no permission to use their names, i will try to contact them if they do not agree, we can change the name. I did not know if that would become a problem)

Second concern:- It was originally hosted as a [separate ](https://github.com/dhilipkumars/redis-sentinel-micro) project, but it was exclusively  written only for Kubernetes and petset usecase in mind,  The original redis-sentinel does more than just slave promotion, it performs fault tolerance, maintians configuration.  For each redis cluster you need to form a quorum of redis-sentinel to make it work effectively.   In addition to that redis recommends not using redis-sentinels for named redis instances when used without persistence.   From our analysis we found that sentinel i used best when redis is provisioned without any scheduler like Mesos or Kubernetes.  

When redis is provisioned via K8s, the only thing we want from the sentinel would be master-slave promotion.  This implementation is a stripped-down version of sentinel which does master-slave promotion.  It is idempotent such that if you run it against a well configured redis-master slave setup, it does nothing. 

Agreed, its a new software and needs feedback from people to make it better, could we mark this Statefulset alpha ? 

This utility simplifies configuring redis master-slave clusters via Statefulset to a great extent.

kow3ns:
If you're referring to the issue https://github.com/antirez/redis/issues/2087 wrt to Sentinel configurations with no persistence enabled, then yes, Sentinel has a gap there, and I've seen no motion on it in 2 years or more, but is it your intention to ** only** support Redis deployed as a cache? As replication with AOF/RDB persistence and Sentinel based cluster management is the configuration that is supported "out of the box", shouldn't we support this configuration first? 

What I mean is that if users install a helm chart with Redis + Sentinel they would expect, as you point out above, a persistent replicated solution managed by 3 or 5 sentinels. They would not expect a cache deployed that provides no durability guarantees.

Would it be possible to either provide a template that allows for both the Redis supported sentinel based configuration and for the custom cache configuration, or to move this to a redis-cache directory, to allow for others to contribute a redis StatefulSet that more closely conforms to the cluster configuration supported by Redis?

With respect to hosting the Sentinel + Micro, have you considered making this a contribution to the Redis organization https://github.com/redis? If they accept it, it would fall under their license, and you'd be able to keep the name as is.

---

Review status: 6 of 10 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.kubernetes.io:443/reviews/kubernetes/contrib/2301#-:-KbG5aLHdOmyiMX02fFA:bihc0yb)*

dhilipkumars:
@kow3ns  Thanks again for responding.

I was thinking on the same lines, could we make a directory hierarchy like below? 
```
/redis
   |---redis_as_cache    [this one, in the readme we mention as Alpha because of this new slave-promotor utility]
   |---redis_as_db       [the one that is available in petsets directory, i could move it from petset to here if you want as a part of seperate PR]
   |---redis_cluster     [in the future , a statefulset for redis 3.0 cluster/shards setup]
```
I'm more concerned about this behaviour of redis as explained in this [documentation](https://redis.io/topics/replication#safety-of-replication-when-master-has-persistence-turned-off),

This means in kubernetes statefulset you can never configure redis without persistence (even using redis-sentinels).   An in-memory master if crashed will be automatically restarted by K8s without any data and all slaves will attempt to replicate from it and the whole master-slave cluster becomes empty.  

This utility allows us to configure in-memory master-slave in an automatically restarted environment like K8s. 

If not used in K8s environment, then this utility has very limited value in comparison to actual redis-sentinel.  Moreover if we manage to start a quorum of redis-sentinels `(atleast 3)` for each master-slave setup in k8s then among the 4 main functions of sentinel (listed below) two are done out-of-the-box by K8s for us.

*  **Monitoring**. `(Provided by K8s)` Sentinel constantly checks if your master and slave instances are working as expected. 
* **Notification**. `(Provided by K8s)` Sentinel can notify the system administrator, another computer programs, via an API, that something is wrong with one of the monitored Redis instances. 
* **Automatic failover**. If a master is not working as expected, Sentinel can start a failover process where a slave is promoted to master, the other additional slaves are reconfigured to use the new master, and the applications using the Redis server informed about the new address to use when connecting.
* **Configuration provider**. Sentinel acts as a source of authority for clients service discovery: clients connect to Sentinels in order to ask for the address of the current Redis master responsible for a given service. If a failover occurs, Sentinels will report the new address.

PS: I'm planning to write to the redis team to get their opinion with regards to the name.  Or could we change the name to slave-promoter or something?

Thanks in Advance,
Dhilip

dhilipkumars:
@kow3ns the redis author does not have any problem with the name 'redis-sentinel-micro' if you agree with the directory structure, i would append the PR (and rebase it.)

dhilipkumars:
@kow3ns the utility is now changed to redis-sentinel-micro sentinel-k8s.

dhilipkumars:
@janetkuo @bprashanth @foxish @enisoc Hi it appears that Kenneth is busy, it would be great if any of the other owners provide any feedback on this PR please?

erictune:
I prefer that we don't add more examples to "statefulsets".  I'd like new
config bundles go into kubernetes/charts rather than kubernetes/contrib, if
at all possible.  That will keep examples more consistent, tested,
disccoverable and supported.

dhilipkumars:
@erictune thanks Eric, that makes a lot of sense will they accept along with the utility programs like redis-sentinel-k8s?

erictune:
No.  Usually the source to binaries is kept in the maintainers own repo.

Here is an example: https://github.com/jtblin/kube2iam
Makefile for that source:
https://github.com/jtblin/kube2iam/blob/master/Makefile
Running "make release" pushes a new image to Docker Hub (jtblin/kube2iam)
The config for running the image on Kubernetes lives in the Charts repo:
  https://github.com/kubernetes/charts/tree/master/stable/kube2iamThat file
The Chart contains parameterized Kubernetes Config:

https://github.com/kubernetes/charts/blob/master/stable/kube2iam/templates/daemonset.yaml#L23
So, the image is a reference to a variable (so that users can build their
own image FROM yours if they want to).
The default value of that variable is set here:

https://github.com/kubernetes/charts/blob/master/stable/kube2iam/values.yaml#L12
Which points back to Docker image name jtblin/kube2iam


By analogy:
Your code can stay in: https://github.com/dhilipkumars/redis-sentinel-micro
You push new images to Docker Hub.
Your chart can live in https://github.com/kubernetes/charts/tree/master
(make a new redis directory).
Each time you change your code, you release a new image.  Then you update
your chart to use the new image.

dhilipkumars:
Closing this PR in favor is redis-cache merged as a chart.

PR #2303, opened 2017-01-13
Shouldn't modify approvers comment

grodrigues3:
Definitely want to modify the last notification from the bot and not the last time a person wrote "/approve"


apelisse:
/lgtm

That's more than needed :-)

PR #2310, opened 2017-01-16
Cluster-autoscaler: decrease node group target size if no new nodes arrive in reasonable time

mwielgus:
cc: @jszczepkowski @andrewsykim @fgrzadkowski

### DIFF
--- a/cluster-autoscaler/utils.go
+++ b/cluster-autoscaler/utils.go
@@ -244,3 +244,31 @@ func removeOldUnregisteredNodes(unregisteredNodes []clusterstate.UnregisteredNod
 	}
 	return removedAny, nil
 }
+
+// Sets the target size of node groups to the current number of nodes in them
+// if the difference was constant for a prolonged time. Returns true if managed
+// to fix something.
+func fixNodeGroupSize(contetxt *AutoscalingContext, currentTime time.Time) (bool, error) {
+	fixed := false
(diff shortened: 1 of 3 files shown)


cluster-autoscaler/utils.go
```diff
@@ -244,3 +244,31 @@ func removeOldUnregisteredNodes(unregisteredNodes []clusterstate.UnregisteredNod
...
+		if incorrectSize.FirstObserved.Add(contetxt.UnregisteredNodeRemovalTime).Before(currentTime) {
+
+			delta := incorrectSize.CurrentSize - incorrectSize.ExpectedSize
+			if delta < 0 {
+				glog.V(0).Infof("Decreasing size of %s, expected=%d current=%d delta=%d", nodeGroup.Id(),
+					incorrectSize.ExpectedSize,
+					incorrectSize.CurrentSize,
+					delta)
+				if err := nodeGroup.DecreaseTargetSize(delta); err != nil {
+					return fixed, fmt.Errorf("Failed to decrease %s: %v", nodeGroup.Id(), err)
```
jszczepkowski: If there is a one node group w/too large delata, the hole autoscaler algorithm for all groups will be stopped. Is it the intended behavior?

jszczepkowski: Discussed offline

jszczepkowski:
LGTM

PR #2311, opened 2017-01-17
fix misspell initalize in predicates.go

fate-grand-order:


mwielgus:
The misspell was fixed in a different PR. Diff is zero.

PR #2320, opened 2017-01-17
turn on approvers in contrib

grodrigues3:


apelisse:
/lgtm

I'm happy with this

apelisse:
I like the irony of this. You need to approve, so that we can turn on
approval ;-)

PR #2327, opened 2017-01-19
improve the logging and error messages in un/assign handler

grodrigues3:
I noticed a few log messages that were wrong.  

Not sure if this is the best way to handle the error from `AssignPR` from the githubapi.

### DIFF
--- a/mungegithub/mungers/assign-unassign-handler.go
+++ b/mungegithub/mungers/assign-unassign-handler.go
@@ -24,6 +24,7 @@ import (
 	"k8s.io/contrib/mungegithub/github"
 	"k8s.io/kubernetes/pkg/util/sets"
 
+	"github.com/golang/glog"
 	githubapi "github.com/google/go-github/github"
 	"github.com/spf13/cobra"
 	c "k8s.io/contrib/mungegithub/mungers/matchers/comment"
@@ -87,9 +88,18 @@ func (h AssignUnassignHandler) Munge(obj *github.MungeObject) {
 
 	toAssign, toUnassign := h.getAssigneesAndUnassignees(obj, comments, fileList, potentialOwners)
 	for _, username := range toAssign.List() {
-		obj.AssignPR(username)
+		err := obj.AssignPR(username)
+		if err != nil {
+			glog.Errorf("Could not assign %v to PR %v because %v", username, obj.Issue.Number, err)
+		}
 	}
-	obj.UnassignPR(toUnassign.List()...)
+	for _, username := range toUnassign.List() {
+		err := obj.UnassignPR(username)
+		if err != nil {
+			glog.Errorf("Could not unassign %v from PR %v because %v", username, obj.Issue.Number, err)
+		}
+	}
+
 }
 
 // getAssigneesAndUnassignees checks to see when someone comments "/assign" or "/unassign"
--- a/mungegithub/github/github.go
+++ b/mungegithub/github/github.go
@@ -1563,7 +1563,7 @@ func (obj *MungeObject) UnassignPR(assignees ...string) error {
 	config := obj.config
 	prNum := *obj.Issue.Number
 	config.analytics.UnassignPR.Call(config, nil)
-	glog.Infof("Unassigning %v from PR# %d  to %v", assignees, prNum)
+	glog.Infof("Unassigning %v from PR #%d", assignees, prNum)
 	if config.DryRun {
 		return nil
 	}


mungegithub/mungers/assign-unassign-handler.go
```diff
@@ -84,10 +85,14 @@ func (h AssignUnassignHandler) Munge(obj *github.MungeObject) {
 
 	//get ALL (not just leaf) the people that could potentially own the file based on the blunderbuss.go implementation
 	potentialOwners, _ := getPotentialOwners(*obj.Issue.User.Login, h.features, fileList, false)
+	glog.Infof("potential owners %+v", potentialOwners)
```
apelisse: That's good for debug, but it's too verbose for normal run

mungegithub/mungers/assign-unassign-handler.go
```diff
@@ -84,10 +85,14 @@ func (h AssignUnassignHandler) Munge(obj *github.MungeObject) {
...
 
 	toAssign, toUnassign := h.getAssigneesAndUnassignees(obj, comments, fileList, potentialOwners)
 	for _, username := range toAssign.List() {
-		obj.AssignPR(username)
+		err := obj.AssignPR(username)
+		if err != nil {
+			glog.Infof("Could not assign %v to PR %v because %v", username, obj.Issue.Number, err)
```
apelisse: That seems to be an error, maybe it should be `glog.Error`

apelisse (changes requested):

mungegithub/mungers/assign-unassign-handler.go
```diff
@@ -87,9 +88,18 @@ func (h AssignUnassignHandler) Munge(obj *github.MungeObject) {
...
+			glog.Errorf("Could not assign %v to PR %v because %v", username, obj.Issue.Number, err)
+		}
+	}
+	for _, username := range toUnassign.List() {
+		err := obj.UnassignPR(username)
+		if err != nil {
+			glog.Errorf("Could not unassign %v from PR %v because %v", username, obj.Issue.Number, err)
+		}
 	}
-	obj.UnassignPR(toUnassign.List()...)
```
apelisse: Why did you remove that?

grodrigues3: see line 97.  Instead of removing them all at once, it's done a person by person basis to 1) be consistent with the assign call and 2) so that the error is specific to a single person that wrote `/unassign` rather than for a group

apelisse: I'd much rather the consistency be fixed the other way around: have `AssignPR` take a list than `Unassign` take individual. I'm not too worried about the error not reporting exactly what individual couldn't be assigned (if it's not already reported by github)

grodrigues3: As I was re implementing it as you described, I realize another downside.  If assigns and unassigns are aggregated, then if someone not on the org/project comments `/assign` or `/unassign` and github's api returns an error, they effectively prevent everyone else from being assigned/unassigned.

mungegithub/mungers/assign-unassign-handler.go
```diff
@@ -87,9 +88,18 @@ func (h AssignUnassignHandler) Munge(obj *github.MungeObject) {
 
 	toAssign, toUnassign := h.getAssigneesAndUnassignees(obj, comments, fileList, potentialOwners)
 	for _, username := range toAssign.List() {
-		obj.AssignPR(username)
+		err := obj.AssignPR(username)
```
eparis: Typically we do the glog.Errorf() in the github.go function and be silent here...

mungegithub/mungers/assign-unassign-handler.go
```diff
@@ -87,9 +88,18 @@ func (h AssignUnassignHandler) Munge(obj *github.MungeObject) {
...
+		if err != nil {
+			glog.Errorf("Could not assign %v to PR %v because %v", username, obj.Issue.Number, err)
+		}
+	}
+	for _, username := range toUnassign.List() {
+		err := obj.UnassignPR(username)
+		if err != nil {
+			glog.Errorf("Could not unassign %v from PR %v because %v", username, obj.Issue.Number, err)
```
eparis: same comment here about where to do the Errorf

grodrigues3:
Closing this PR because changing the behavior to do the assigns all at once has unintended consequences (it's possible to prevent anyone form being assigned with one non project member writing `/assign` )

PR #2332, opened 2017-01-20
[mungegithub] AssignPR has same interface as Unassign

grodrigues3:
Make AssignPR a variadic function like UnassignPR


grodrigues3:
Tested it quickly on PR in my repo.  Seems to work

https://github.com/grodrigues3/contrib/pull/4

eparis:
```
--- FAIL: TestAssignFixes (0.00s)
	assign-fixes_test.go:83: Patching the incorrect Assignee  instead of dev45
```

grodrigues3:
Closing this PR because I don't think we ever will want to assign multiple people in the same function call because if one of the desired assignees cannot be assigned (not in the org/project), then the entire call fails.  It makes sense to do this on a person by person basis.

PR #2334, opened 2017-01-20
allow maintainers to add approved label explicitly

grodrigues3:
At least temporarily, we should allow people with label-adding privileges to add the approved label.

The process is designed so that only when all files are approved the approved label can be added to the bot but with this "fix", human's will be able to manually apply the approved label.

### DIFF
--- a/mungegithub/mungers/approval-handler.go
+++ b/mungegithub/mungers/approval-handler.go
@@ -30,6 +30,7 @@ import (
 	"k8s.io/contrib/mungegithub/features"
 	"k8s.io/contrib/mungegithub/github"
 	c "k8s.io/contrib/mungegithub/mungers/matchers/comment"
+	"k8s.io/contrib/mungegithub/mungers/matchers/event"
 	"k8s.io/kubernetes/pkg/util/sets"
 )
 
@@ -106,7 +107,8 @@ func (h *ApprovalHandler) Munge(obj *github.MungeObject) {
 
 	for _, approverSet := range ownersMap {
 		if approverSet.Len() == 0 {
-			if obj.HasLabel(approvedLabel) {
+			// if a human explicitly added the label, allow it through
+			if obj.HasLabel(approvedLabel) && !humanAddedApproved(obj) {
 				obj.RemoveLabel(approvedLabel)
 			}
 			return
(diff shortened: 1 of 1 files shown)


mungegithub/mungers/approval-handler.go
```diff
@@ -118,6 +120,16 @@ func (h *ApprovalHandler) Munge(obj *github.MungeObject) {
...
 }
 
+func humanAddedApproved(obj *github.MungeObject) bool {
+	events, ok := obj.GetEvents()
+	if !ok {
+		return false
+	}
+	approveAddedMatcher := event.And([]event.Matcher{event.AddLabel{}, event.LabelName(approvedLabel)})
+	labelEvents := event.FilterEvents(events, approveAddedMatcher)
+	return *labelEvents.GetLast().Actor.Login != botName
```
foxish: do we need some null checks here?
Can labelEvents be nil, or Login be nil?

apelisse: You're dereferencing a pointer, are you 100% sure it's never nil?

eparis:
Why aren't we forcing you to get a higher level person to approve? Why do we need this?

foxish:
@eparis It seemed like a good in-between place to go to. The goal of the PR is to let maintainers apply the "approved" label via the github interface (and not just via the bot command). I think this is the expectation.
This was echoed by davidopp. In addition, if we look at how /lgtm is being used, I think a lot of people still apply the label from the GitHub UI rather than use the bot command.

apelisse:
This code doesn't address this specific use-case, but it might be useful to be able to remove the label manually ...

fejta:
I would expect an owner to be able to both /lgtm and /approve, and in fact I would expect this to be the standard case: the review goes to the owner, and the owner gives an /approve which also gives an /lgtm

grodrigues3:
This is a provision for the transition phase, while people get used to `/approve`.   The current implementation requires all files to be approved and manually applying the approved label without all files being approved will result in the bot removing the label (without this PR).  This will allow people to explicitly add it

foxish:
@eparis Does the change sound reasonable?

eparis:
my 'issue' is that both the `LGTM` label and the `/lgtm` command are done by one person for the whole PR. So a person being able to set that label makes total sense.

But approve is different. `/approve` is for a subset of a PR. Whereas `approve` label is for the entire PR. So a single person being able to set the label is not a similar workflow.

I understand the transition is going to be painful, and this would make it less painful. I worry it will mean that we have a much harder time getting people to use approvers, since I'm betting many people will just set the label, quite possibly without knowing why that was a bad thing.

I'm fine with this merging, but I think we'll need to track how often it gets used and by whom. We'll need to make a concerted education campaign for any user that sets the label. Then after some period of time we might need to turn it back off....

foxish:
> We'll need to make a concerted education campaign for any user that sets the label. Then after some period of time we might need to turn it back off.

Ack. That sounds good.

PR #2336, opened 2017-01-22
fix misspell "contains" in nginx/main.go

fate-grand-order:


bprashanth:
Thanks for the pr, we're migrating this code to ingress/ where I think this typo is fixed (https://github.com/kubernetes/ingress/search?q=containes&type=Code&utf8=%E2%9C%93), if not, please reopen against that repo. Sorry for the confusion.

eparis:
@bprashanth you probably should delete the code in here, to keep people from finding it....

bprashanth:
Yeah Im replacing all readmes with links to the main repo, so places we've posted them on don't completely break. I plan to delete the code when we do a release on that repo.

PR #2342, opened 2017-01-25
rename skydns to kubedns

xialonglee:
`skydns` was no longer used, rename to `kubedns`.


xialonglee:
cc @eparis

ansible/roles/kubernetes-addons/templates/dns/kube-svc.yaml.j2
```diff
@@ -1,4 +1,4 @@
-# This file should be kept in sync with cluster/gce/coreos/kube-manifests/addons/dns/skydns-svc.yaml
+# This file should be kept in sync with cluster/gce/coreos/kube-manifests/addons/dns/kubedns-svc.yaml
```
eparis: This file was renamed to `kube-svc.yaml.j2`  shouldn't it be `kubeDNS-svc.yaml.j2` ?

xialonglee: see this https://github.com/kubernetes/kubernetes/tree/master/cluster/addons/dns , the name of the yaml source files are with prefix `kubedns`.

xialonglee:
sorry, i was a little thoughtless of this pr

see https://github.com/kubernetes/contrib/blob/master/ansible/roles/kubernetes/defaults/main.yml

it use version 1.4.5 of kubernetes by default, so for compatibility reasons,  these files should keep sync with kubernetes branch 1.4 and they still named with `skydns` prefix.
see https://github.com/kubernetes/kubernetes/tree/release-1.4/cluster/addons/dns

These files name changed with `kubedns` prefix were happened in branch master.
and kubedns-rc.yaml should named kubedns-controller.yaml (my mistake).
see https://github.com/kubernetes/kubernetes/tree/master/cluster/addons/dns

though these files name will change sooner or later, if you think that it is not appropriate to change these files name so early, feel free to close this pr.

but it should rename skydns to kubedns in README.md
@ingvagabund @eparis

ansible/roles/kubernetes-addons/templates/dns/kube-svc.yaml.j2
eparis: The problem is not inside the file. The problem is the filename itself. Do an `ls` in the `roles/kubernetes-addons/templates/dns/` directory to see the problem. You renamed `skydns-svc.yaml.j2` to `kube-svc.yaml.j2`. I believe we both agree it should have been called `kubedns-svc.yaml.j2`

xialonglee:
@eparis 
oh... so sorry for misunderstood your mean and i didn't notice that...
also rename `kubedns-rc.yaml.j2` to `kubedns-controller.yaml.j2` for consistency with kubernetes master branch.
i fixed it now and tested locally, seems it's ok now.

PR #2343, opened 2017-01-25
support setup etcd cluster with diff interfaces

xialonglee:
Use fact ·ansible_default_ipv4.address· can setup etcd cluster on different network interfaces

for example
setup etcd cluster with 2 nodes A (default_ipv4.interface is `eth0`), B (default_ipv4.interface is `virbr0`, it's a virtual machine)
use fact `['ansible_' + etcd_interface]` will failed, variable `etcd_interface` is related to the host that is executing, and node A does not have `virbr0` and B does not have `eth0` respectively.
use fact `ansible_default_ipv4` can solve this.

\======================

The method solve problem above was outdated.
Adapt @rutsky 's suggestion, simply using `if` statement to solve the problem.

### DIFF
--- a/ansible/roles/etcd/defaults/main.yaml
+++ b/ansible/roles/etcd/defaults/main.yaml
@@ -29,12 +29,14 @@ etcd_initial_cluster_state: new
 etcd_initial_cluster_token: etcd-k8-cluster
 
 # Interface on which etcd listens.
-# Useful on systems when default interface is not connected to other machines,
+# It is useful when you want define common network interface for etcd clusters and the interface is not connected to other machines (i.e. use same network interface for setup a etcd clusters),
 # for example as in Vagrant+VirtualBox configuration.
 # Note that this variable can't be set in per-host manner with current implementation.
-etcd_interface: "{{ ansible_default_ipv4.interface }}"
+# If not specified, fact "ansible_default_ipv4.address" will be applied by every host itself.
+etcd_interface: ""
 
-etcd_machine_address: "{{ hostvars[inventory_hostname]['ansible_' + etcd_interface].ipv4.address }}"
+etcd_machine_address: "{% if etcd_interface != '' %}{{ hostvars[inventory_hostname]['ansible_' + etcd_interface].ipv4.address }}
+{%- else %}{{ hostvars[inventory_hostname].ansible_default_ipv4.address }}{% endif %}"
 etcd_initial_advertise_peer_urls: "{{ etcd_peer_url_scheme }}://{{ etcd_machine_address }}:{{ etcd_peer_port }}"
 etcd_listen_peer_urls: "{{ etcd_peer_url_scheme }}://0.0.0.0:{{ etcd_peer_port }}"
 etcd_advertise_client_urls: "{{ etcd_url_scheme }}://{{ etcd_machine_address }}:{{ etcd_client_port }}"
--- a/ansible/roles/etcd/templates/etcd.conf.j2
+++ b/ansible/roles/etcd/templates/etcd.conf.j2
@@ -1,6 +1,12 @@
 {% macro initial_cluster() -%}
 {% for host in groups[etcd_peers_group] -%}
-  {{ hostvars[host]['ansible_hostname'] }}={{ etcd_peer_url_scheme }}://{{ hostvars[host]['ansible_' + etcd_interface].ipv4.address }}:{{ etcd_peer_port }}
+  {{ hostvars[host]['ansible_hostname'] }}={{ etcd_peer_url_scheme }}://
+  {%- if etcd_interface != "" -%}
+    {{ hostvars[host]['ansible_' + etcd_interface].ipv4.address }}
+  {%- else -%}
+    {{ hostvars[host].ansible_default_ipv4.address }}
+  {%- endif -%}
+:{{ etcd_peer_port }}
   {%- if not loop.last -%},{%- endif -%}
 {%- endfor -%}
 {% endmacro -%}


ingvagabund:
@redhat-k8s-bot test please

eparis:
@redhat-k8s-bot test please

eparis:
See https://github.com/kubernetes/contrib/pull/724 for where this came from. @rutsky added this. What you did should, I believe, be a no-op, as etcd_interface should be set to the default ipv4 interface per host by default...

ingvagabund:
> Internal error

because setuptools-34.0.2.zip installed via pip is most likely broken

ingvagabund:
https://github.com/pypa/setuptools/issues/937

rutsky:
This PR will break Vagrant+VirtualBox configuration and according to [this comment](https://github.com/kubernetes/contrib/blob/master/ansible/roles/etcd/defaults/main.yaml#L34):

```
# Interface on which etcd listens.
# Useful on systems when default interface is not connected to other machines,
# for example as in Vagrant+VirtualBox configuration.
# Note that this variable can't be set in per-host manner with current implementation.
etcd_interface: "{{ ansible_default_ipv4.interface }}"
```

it indeed doesn't allow to specify different interfaces on different hosts.

@xialonglee please suggest solution that doesn't break Vagrant configuration.

xialonglee:
@rutsky 
i see your point... so variable `etcd_interface` is used for Vagrant configuration.( i usaully setup a k8s clusters directly on physical machines so didn't notice that... maybe i should try it)

maybe register a fact that assigned `etcd_interface` value could solve it and without any side effect. 
What's your opinion? 
if that would work i will update this pr after few days.( because i am in vacation now and there is no environment to test)

rutsky:
@xialonglee I don't see easy way to support setting per-host custom public etcd interface, and, which is more important, I don't see need for this currently.

I suggest to cover two use cases: custom single etcd interface on all hosts (i.e. Vagrant case), and use of `default_ipv4.interface` (i.e. "common" case for typical physical/vm setups).
To do so I suggest to make `etcd_interface` not defined by default and in all dependent places (like `etcd_machine_address: "{{ hostvars[inventory_hostname]['ansible_' + etcd_interface].ipv4.address }}"`) check is `etcd_interface` defined --- if it is, use it, if it is not, use `default_ipv4.interface`.

xialonglee:
@rutsky 
That's good and much better then my suggestion. it's the simplest way to solve the problem in setting etcd cluster with different network interface name.
Thanks for your suggestion, i will update my code.

rutsky:
@xialonglee `etcd_initial_advertise_peer_urls` uses `etcd_interface` too, you haven't covered this.

xialonglee:
@rutsky 
My local test shows that there's no problem with `etcd_initial_advertise_peer_urls`.

I see the `etcd_advertise_client_urls` was assigned by one value of 
"{{ etcd_url_scheme }}://**{{ etcd_machine_address }}**:{{ etcd_client_port }}".

Here `etcd_machine_address` was assigned by `{{ hostvars[inventory_hostname]['ansible_' + etcd_interface].ipv4.address }}`.

There is only one value of `etcd_advertise_client_urls`, not related to other machines, in non-Vagrant case, it equals to `default_ipv4.interface`(default `etcd_interface` is `default_ipv4.interface`) of the host that is executing.

So i think that is no need to covered.

rutsky:
@xialonglee isn't in your case `etcd_machine_address` will have invalid IP address: IP of `eth0` instead of IP `virbr0`?

xialonglee:
@rutsky 
No, it is not. It works correctly.
Actually, i got two machines node A and node B to be etcd clusters.
Node A is a physical machine and Node B is a virtual machine running on Node A.
Node A network interfaces is:
<pre>virbr0
docker0 
flannel.1 
.... </pre>

Node B network interfaces is:
<pre>eth0
docker0 
flannel.1 
lo
.... </pre>

type
`# ansible all -m setup`
I see `ansible_default_ipv4` is `virbr0` for node A and `eth0` for node B.
Since `etcd_machine_address` using `etcd_interface` which default is `default_ipv4.interface` , it is not problem in my case.

rutsky:
@xialonglee I reviewed  once again all `etcd_interface` implementation and highly suggest do more serious refactoring then you provide currently.
With your change logic of interface selection starts to be twisted and error-prone --- `etcd_interface` doesn't mean now "public interface to use for etcd", but "try to use `etcd_interface` interface, and if it is not available try default interface".

Lets stick with my [initial suggestion](https://github.com/kubernetes/contrib/pull/2343#issuecomment-276036118):

1. Make `etcd_interface` empty by default (or unset):

```
# defaults/main.yml
etcd_interface: ""
```

2. In all places that depends on `etcd_interface` check is `etcd_interface` is set, and use either `etcd_interface` or default interface:

```
# defaults/main.yml
etcd_machine_address: "{{ hostvars[inventory_hostname]['ansible_' + etcd_interface].ipv4.address if etcd_interface else ansible_default_ipv4.address }}"
```

```
# etcd/templates/etcd.conf.j2
  {{ hostvars[host]['ansible_hostname'] }}={{ etcd_peer_url_scheme }}://{{ hostvars[host]['ansible_' + etcd_interface].ipv4.address if etcd_interface else hostvars[host].ansible_default_ipv4.address }}:{{ etcd_peer_port }} 
```

rutsky (changes requested):
Please update your patch according to my comment.

xialonglee:
@rutsky 
You are right, i think i've got your point.
I've updated the codes as your suggestion and i also added some comments in addition, PTAL.

ansible/roles/etcd/defaults/main.yaml
```diff
@@ -29,12 +29,14 @@ etcd_initial_cluster_state: new
 etcd_initial_cluster_token: etcd-k8-cluster
 
 # Interface on which etcd listens.
-# Useful on systems when default interface is not connected to other machines,
+# It is useful when you want define common network interface for etcd clusters and the interface is not connected to other machines(i.e. use same network interface for setup a etcd clusters),
```
rutsky: Please add space between "machines" and opening bracket.

rutsky (changes requested):
Except minor missed space this LGTM

rutsky:
@xialonglee please rebase.

xialonglee:
@rutsky 
I've updated the codes. Thanks!

rutsky (approved):
LGTM.
Thanks @xialonglee !

rutsky:
@xialonglee still you need to rebase you PR.

ingvagabund:
@redhat-k8s-bot test please

ingvagabund:
CentOS repo mirrors related issue, not PR related

ingvagabund:
Hm, for some reason the Ci for fedora does not work. Anyway, the RHEL is fine.
/lgtm

PR #2344, opened 2017-01-25
Add cmdline flag --watch-all-namespaces to kube-keepalived-vip

ghost:
This is a workaround for clientcmd not allowing --namespace= to override the cluster context namespace (e.g. POD_NAMESPACE) to listen for services on all namespaces.


isavin:
I signed it!

xeor:
Will this fix the problem that I'm not able to use multiple namespaces in the vip-config? Or might that be another problem?

isavin:
@xeor Yes, with these change you will be able to add in vip-config services from all namespaces. Otherwise you can only expose services from one namespace.

xeor:
Awesome.. any ETA when this pr will get merged? The docs are a little confusing on this problem. Since the ns is a part of the config, I would think that it supported multiple namespaces out of the box..

Glad to see this pr taking care of this!

isavin:
Not sure if/when it will merged. Needs to be reviewed first. /cc @bprashanth @mikedanese @bowei

bowei (approved):

xeor:
What is the process for when the `gcr.io/google_containers/kube-keepalived-vip` image gets rebuilt based on this fix and a new version is out?

PR #2345, opened 2017-01-25
Set verbosity levels for addon-resizer

### LINKED ISSUE #2337: [addon-resizer] Set logging levels to reduce amount of logs produced
Currently, nanny produces a lot of meaningful messages by default, like

```go
log.Infof("The number of nodes is %d", num)
```

or

```go
log.Infof("Resources are within the expected limits.")
```

IMO, logging levels should be set up according to [the developer's guide](https://github.com/kubernetes/community/blob/master/contributors/devel/logging.md) and such messages shouldn't be visible by default.

CC @piosz

crassirostris:
Fix https://github.com/kubernetes/contrib/issues/2337

Set verbosity levels according to the [developer's logging guide](https://github.com/kubernetes/community/blob/master/contributors/devel/logging.md)

CC @Q-Lee @piosz

### DIFF
--- a/addon-resizer/nanny/nanny_lib.go
+++ b/addon-resizer/nanny/nanny_lib.go
@@ -87,7 +87,7 @@ func PollAPIServer(k8s KubernetesClient, est ResourceEstimator, contName string,
 			log.Error(err)
 			continue
 		}
-		log.Infof("The number of nodes is %d", num)
+		log.V(4).Infof("The number of nodes is %d", num)
 
 		// Query the apiserver for this pod's information.
 		resources, err := k8s.ContainerResources()
@@ -95,18 +95,17 @@ func PollAPIServer(k8s KubernetesClient, est ResourceEstimator, contName string,
 			log.Errorf("Error while querying apiserver for resources: %v", err)
 			continue
 		}
-		log.Infof("The container resources are %+v", *resources)
 
 		// Get the expected resource limits.
 		expResources := est.scaleWithNodes(num)
-		log.Infof("The expected resources are %+v", *expResources)
 
 		// If there's a difference, go ahead and set the new values.
 		if !shouldOverwriteResources(int64(threshold), resources.Limits, resources.Requests, expResources.Limits, expResources.Requests) {
-			log.Infof("Resources are within the expected limits.")
+			log.V(4).Infof("Resources are within the expected limits. Actual: %+v Expected: %+v", *resources, *expResources)
 			continue
 		}
-		log.Infof("Resources are not within the expected limits: updating the deployment.")
+
+		log.Infof("Resources are not within the expected limits, updating the deployment. Actual: %+v Expected: %+v", *resources, *expResources)
 		if err := k8s.UpdateDeployment(expResources); err != nil {
 			log.Error(err)
 			continue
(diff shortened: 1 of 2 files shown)


addon-resizer/nanny/nanny_lib.go
```diff
@@ -87,26 +87,25 @@ func PollAPIServer(k8s KubernetesClient, est ResourceEstimator, contName string,
...
 
 		// If there's a difference, go ahead and set the new values.
 		if !shouldOverwriteResources(int64(threshold), resources.Limits, resources.Requests, expResources.Limits, expResources.Requests) {
-			log.Infof("Resources are within the expected limits.")
+			log.V(4).Infof("Resources are within the expected limits. Actual: %+v Expected: %+v", *resources, *expResources)
 			continue
 		}
-		log.Infof("Resources are not within the expected limits: updating the deployment.")
+
+		log.V(0).Infof("Resources are not within the expected limits, updating the deployment. Actual: %+v Expected: %+v", *resources, *expResources)
```
piosz: Just Infof, remove V(0).

addon-resizer/nanny/nanny_lib.go
```diff
@@ -87,26 +87,25 @@ func PollAPIServer(k8s KubernetesClient, est ResourceEstimator, contName string,
...
-		log.Infof("The container resources are %+v", *resources)
 
 		// Get the expected resource limits.
 		expResources := est.scaleWithNodes(num)
-		log.Infof("The expected resources are %+v", *expResources)
 
 		// If there's a difference, go ahead and set the new values.
 		if !shouldOverwriteResources(int64(threshold), resources.Limits, resources.Requests, expResources.Limits, expResources.Requests) {
-			log.Infof("Resources are within the expected limits.")
+			log.V(4).Infof("Resources are within the expected limits. Actual: %+v Expected: %+v", *resources, *expResources)
```
piosz: Optional: As for now I can see resources in the following format
```
cpu:{i:{value:88 scale:-3} d:{Dec:<nil>} s:88m Format:DecimalSI}
```
How about making it more readable for a human reader?

piosz (changes requested):

addon-resizer/nanny/nanny_lib.go
Crassirostris: This a good question, whether to use V(0) explicitly or not. I looked up and it seems like just infof is a much more popular approach.

ACK, that should also be mentioned in guide

addon-resizer/nanny/nanny_lib.go
Crassirostris: I tried to touch as little code as possible. I think I will spend much more time making it better-looking

addon-resizer/nanny/main/pod_nanny.go
```diff
@@ -53,7 +53,7 @@ var (
 
 func main() {
 	// First log our starting config, and then set up.
-	log.Infof("Invoked by %v", os.Args)
+	log.V(1).Infof("Invoked by %v", os.Args)
```
Crassirostris: This contradicts logging guide, where configuration details should be under V(1).

However, I tend to agree that O(1) logs should be on level 0. ACK

Crassirostris:
@piosz PTAL

addon-resizer/nanny/nanny_lib.go
piosz: Sure. Created #2351

piosz (approved):

piosz:
cc @Q-Lee

piosz:
Also please bump the version in Kubernetes

Crassirostris:
@k8s-bot test this

PR #2348, opened 2017-01-26
Add Least-Cost Node Group Expander

Capitrium:
#2118 added support for node group expansion strategies, allowing the cluster autoscaler to determine the best node group to expand instead of just selecting a node group at random. This adds a new expansion strategy that selects the node group with the lowest price per node as the group to be expanded, falling back to random selection if the price per node is equal across all configured node groups.

The main purpose of this PR is to improve support for AWS spot instance nodes; as GCE does not provide a bidding . Consider the following snippet from a cluster-autoscaler deployment manifest:
```
        - ./cluster-autoscaler
        - --v=4
        - --cloud-provider=aws
        - --expander=least-cost
        - --nodes=1:10:spot-node-1
        - --nodes=1:10:spot-node-2
        - --nodes=1:10:spot-node-3
        - --nodes=1:10:spot-node-4
        - --nodes=1:10:on-demand-node:0.10
```
Each of the spot worker groups spans a single Availability Zone on AWS. As instances are added to the spot worker groups, the current market price for spot instances in a given AZ is likely to change. Using the least-cost expander allows the cluster autoscaler to take the market price of spot instances into account when scaling, causing it to select the node group with the cheapest spot price at the time the scale-up was requested.

Note the additional on-demand node group; since there is no simple API for getting the current on-demand price of an AWS instance type, the node cost is manually specified as part of the node group. This way, if the spot price in all AZs is higher than the max bid price configured for the spot node groups, the cluster autoscaler can fall back to the on-demand node group.

Since GCE doesn't use a variable pricing model for their discounted Preemptible VMs, configuring the cluster autoscaler to add support for them should be trivial:
```
        - ./cluster-autoscaler
        - --v=4
        - --cloud-provider=gce
        - --expander=least-cost
        - --nodes=1:10:preemptible-node-1:0.20
        - --nodes=1:10:preemptible-node-2:0.20
        - --nodes=1:10:preemptible-node-3:0.20
        - --nodes=1:10:preemptible-node-4:0.20
        - --nodes=1:10:regular-node:1.00
```

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
+++ b/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
@@ -139,12 +139,6 @@ func (asg *Asg) NodeCost() (float64, error) {
 		glog.V(4).Infof("Returning user-defined cost on NodeGroup %s: %+v", asg.Id(), asg.nodeCost)
 		return *asg.nodeCost, nil
 	}
-
-	if asg.launchConfig == nil {
-		if err := asg.awsManager.regenerateCache(); err != nil {
-			return 0, fmt.Errorf("Error while looking for LaunchConfiguration of ASG %s, error: %v", asg.Name, err)
-		}
-	}
 	cost, err := asg.awsManager.GetAsgSpotInstanceCost(asg)
 	if err != nil {
 		return 0, fmt.Errorf("Error calculating nodeCost: %v", err)
--- a/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
+++ b/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
@@ -22,6 +22,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/golang/glog"
 	"k8s.io/contrib/cluster-autoscaler/cloudprovider"
 	apiv1 "k8s.io/kubernetes/pkg/api/v1"
 )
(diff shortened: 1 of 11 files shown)


mwielgus:
Thank you for this contribution. Having the cost included is definitely a big step forward. 

However I would try to polish the cost definition a bit. Is it cost per hour? Per minute? How do we handle situation when there is a flat fee of X for Y minutes and then the cost is Z pre minute or per hour? Obviously there is no reason to immediately delete an instances for which we have to pay for the next hour anyway and whose capacity may be needed soon.
And with these spot instances there might be situation where the current cost is prohibitive and having the pods pending for a while wouldn't be, in the end, that bad.

So there is a lot of "numbers" to be passed and configured. And for these numbers I would try to come up with some more elaborate config flag/file with (json/csv/whatever) than nodes so that we don't have to change the config format totally once we get couple more cloud providers or we try to include the cost in some other calculations. And I would try to store all of these information in a separate structure to not over-pollute nodegroup.

Capitrium:
Yeah, that all makes sense; I'll work on those changes as soon as I can. I'm mainly an AWS guy (if you hadn't already guessed) so I definitely made a few assumptions around the cost definition and how long people are willing to have `Pending` pods, etc. The current implementation was definitely designed with our own cluster's fairly specific requirements in mind.

mwielgus:
Closing in favor of https://github.com/kubernetes/autoscaler/issues/82

PR #2364, opened 2017-02-02
parser now parses deletion and volume events

dashpole:
Add deletion events and volume mount/unmount events to the node perf dash
cc @Random-Liu


dashpole:
cc @dchen1107

dashpole:
this is still useful.
/remove-lifecycle stale

PR #2366, opened 2017-02-02
Base kubelet-to-gcm on scratch, set user to nobody, and bump tag to 1.2.2

ixdy:
This also fixes changes the prefix back to the correct default.

I haven't pushed 1.2.2 yet, but I don't expect any major changes, besides using an updated base image and being built with go1.7.5.

### DIFF
--- a/kubelet-to-gcm/Dockerfile
+++ b/kubelet-to-gcm/Dockerfile
@@ -12,15 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM ubuntu:14.04
-MAINTAINER Quintin Lee "qlee@google.com"
+FROM scratch
 
-# Disable prompts from apt.
-ENV DEBIAN_FRONTEND noninteractive
+COPY build/monitor /
 
-# # Install the Fluentd agent that knows how to send logs to Google Cloud Logging.
-RUN apt-get -q update && \
-    apt-get install -y curl && \
-    apt-get clean
-
-ADD build/monitor monitor
+# nobody:nobody
+USER 65534:65534
+ENTRYPOINT ["/monitor"]
--- a/kubelet-to-gcm/Makefile
+++ b/kubelet-to-gcm/Makefile
@@ -24,9 +24,8 @@
 
 OUT_DIR = build
 PACKAGE = k8s.io/contrib/kubelet-to-gcm
-#PREFIX = gcr.io/google_containers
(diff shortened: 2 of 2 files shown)


ixdy:
ping @Q-Lee

ixdy:
Actually, I'm confused - why is this image based on ubuntu? Why does it install curl? Is there any reason this can't just be from `scratch`?

ixdy:
Changed the base image to scratch. I still haven't pushed this anywhere yet. Not sure how to test this.
cc @timstclair

kubelet-to-gcm/Dockerfile
```diff
@@ -12,15 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM ubuntu:14.04
+FROM scratch
 MAINTAINER Quintin Lee "qlee@google.com"
```
timstclair: Mind removing this? We're removing `MAINTAINER`s from Dockerfiles, since they're deprecated and tend to get out of date.

kubelet-to-gcm/Dockerfile
```diff
@@ -12,15 +12,7 @@
...
 
-# Disable prompts from apt.
-ENV DEBIAN_FRONTEND noninteractive
-
-# # Install the Fluentd agent that knows how to send logs to Google Cloud Logging.
-RUN apt-get -q update && \
-    apt-get install -y curl && \
-    apt-get clean
-
 ADD build/monitor monitor
```
timstclair: nit: prefer `COPY` over `ADD` (https://www.ctl.io/developers/blog/post/dockerfile-add-vs-copy/)

timstclair: It looks like this image should work fine running as `nobody`. Mind adding a `USER` directive? See https://github.com/kubernetes/heapster/pull/1514/commits/c0f5ab124765c09c0d7a679e843d8ffd1aaca016 for an example.

timstclair (commented):
Thanks Jeff.

ixdy:
Done. Also added an `entrypoint`.

timstclair:
Thanks, LGTM (please squash).

ixdy:
squashed.

ixdy:
I guess this PR needs approval too

ixdy:
Thanks. `gcr.io/google_containers/kubelet-to-gcm:1.2.2` pushed.

timstclair:
Thanks. Will you follow up with a change to use the updated image?

PR #2370, opened 2017-02-06
Return correct ProviderID format from GetAsgNodes

philk:
This matches the way that the [gce does it][1] and fixes issues with the
[cluster state][2].

[1]: https://github.com/kubernetes/contrib/blob/master/cluster-autoscaler/cloudprovider/gce/gce_manager.go#L247
[2]: https://github.com/kubernetes/contrib/issues/2228#issuecomment-274896211

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/aws_manager.go
+++ b/cluster-autoscaler/cloudprovider/aws/aws_manager.go
@@ -224,7 +224,8 @@ func (m *AwsManager) GetAsgNodes(asg *Asg) ([]string, error) {
 		return []string{}, err
 	}
 	for _, instance := range group.Instances {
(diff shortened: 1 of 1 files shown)


philk:
I signed it!

philk:
@mwielgus friendly ping

mwielgus:
cc @andrewsykim

andrewsykim (approved):
LGTM based on https://github.com/kubernetes/contrib/blob/master/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go#L95. I think the instance representations should encapsulated into a struct instead of a string like "aws:///zone/instance-id" but out of the scope of this PR.

PR #2372, opened 2017-02-07
Hack for debugging FROM scratch containers

timstclair:
Hacky solution to https://github.com/kubernetes/kubernetes/issues/10834 and https://github.com/kubernetes/kubernetes/issues/27140 to unblock migration of system pods to `scratch` base images (https://github.com/kubernetes/kubernetes/issues/40248). This is not meant to be a long term solution, but rather a temporary workaround until we have native Kubernetes support.

The script works by bringing up a busybox pod on the same node as the debug target, mounting the node's root fs, and calling docker directly to copy busybox into the target container. Once the "install" is complete, the target can be debugged through a standard `kubectl exec`.

/cc @verb @thockin

### DIFF
--- a/scratch-debugger/debug.sh
+++ b/scratch-debugger/debug.sh
@@ -1,0 +1,196 @@
+#!/bin/bash
+
+# Copyright 2017 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o nounset
+set -o errexit
+set -o pipefail
+
+# Customizable parameters
+TMP_SUBDIR="${TMP_SUBDIR:-debug-tools}"
+CONTEXT="${KUBECONTEXT:-}"
+ARCH="${ARCH:-amd64}"
+
+# Parse arguments & flags
+while [[ $# -gt 0 ]]; do
+  case $1 in
+    -n|--namespace)
+      NAMESPACE=$2
+      shift
+      ;;
+    -c|--container)
+      CONTAINER_NAME=$2
+      shift
+      ;;
+    *)
+      NAME=$1
+      ;;
+  esac
+  shift
+done
+
+if [[ -z ${NAME:-} ]]; then
+  echo >&2 "USAGE: $0 POD_NAME [-n POD_NAMESPACE] [-c CONTAINER_NAME]"
+  exit 1
+fi
+
+# Internal variables
+KUBECTL="kubectl"
+[ -z "${CONTEXT}" ] || KUBECTL="${KUBECTL} --context=${CONTEXT}"
+[ -z "${NAMESPACE:-}" ] || KUBECTL="${KUBECTL} --namespace=${NAMESPACE}"
+
+NAMESPACE=${NAMESPACE:-default}
+
+TARGET_CNAME=${CONTAINER_NAME:-}
+if [ -z "${CONTAINER_NAME:-}" ]; then
+  CONTAINER_NAME=$(${KUBECTL} get pod ${NAME} -o jsonpath='{.spec.containers[0].name}')
+fi
+
+INSTALL_DIR="/tmp/${TMP_SUBDIR}"
+CONTAINER_ID=$(${KUBECTL} get pod ${NAME} -o jsonpath="{.status.containerStatuses[?(@.name==\"${CONTAINER_NAME}\")].containerID}")
+RUNTIME=${CONTAINER_ID%://*}
+CONTAINER_ID=${CONTAINER_ID#*://}
+NODE=$(${KUBECTL} get pod ${NAME} -o jsonpath='{.spec.nodeName}')
+
+if [[ ${RUNTIME} != docker ]]; then
+  echo >&2 "Error: $0 only works with a docker runtime. Found: ${CONTAINER_ID}"
+  exit 1
+fi
+
+# Construct the command to debug the target container.
+DEBUG_CMD="${KUBECTL} exec -i -t ${NAME}"
+[ -z "${TARGET_CNAME}" ] || DEBUG_CMD="${DEBUG_CMD} -c ${TARGET_CNAME}"
+DEBUG_CMD="${DEBUG_CMD} -- ${INSTALL_DIR}/sh -c 'PATH=\${PATH}:${INSTALL_DIR} sh'"
+
+if ${KUBECTL} exec ${NAME} -c ${CONTAINER_NAME} -- ${INSTALL_DIR}/echo &>/dev/null; then
+  echo "Debug tools already installed. Dumping you into the pod container now."
+  eval "${DEBUG_CMD}"
+  exit 0
+fi
+
+cat <<EOF
+Debug Target Container:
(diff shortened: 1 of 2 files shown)


scratch-debugger/debug.sh
```diff
@@ -0,0 +1,120 @@
...
+[ -z "$TARGET_CNAME" ] || DEBUG_CMD="$DEBUG_CMD -c $TARGET_CNAME"
+DEBUG_CMD="$DEBUG_CMD -- ${INSTALL_DIR}/sh -c 'PATH=\$PATH:${INSTALL_DIR} sh'"
+
+if $KUBECTL exec ${NAME} -c $CONTAINER_NAME -- ${INSTALL_DIR}/echo &>/dev/null; then
+  echo "Debug tools already installed. Dumping you into the pod container now."
+  eval "$DEBUG_CMD"
+  exit 0
+fi
+
+echo "Installing busybox to $INSTALL_DIR ..."
```
bprashanth: I'm guessing the busybox binary is statically compiled? in which case don't you want to be more explicity about which version of glibc/uclibc/musl it contains?

scratch-debugger/debugger.yaml
```diff
@@ -0,0 +1,31 @@
...
+apiVersion: v1
+kind: Pod
+metadata:
+  name:        ARG_DEBUGGER
+  namespace:   ARG_NAMESPACE
+spec:
+  nodeName:    ARG_NODENAME
+  containers:
+    - name:    debugger
+      image:   busybox
```
bprashanth: don't you want a version ?

bprashanth: And ARCH?

bprashanth:
I think we've decomissioned contrib entirely, so maybe we should find a better home? 
Re-assigning to the 2 people you atted

bprashanth:
Oh, can't assign to @verb, Lee Verberne PTAL when you have a minute

scratch-debugger/debugger.yaml
timstclair: Added ARCH. For these purposes, I'd rather use `latest` than pin to a specific version. I can add a parameter for it if you think it's necessary...

timstclair:
> I think we've decomissioned contrib entirely, so maybe we should find a better home?

Last I heard contrib was still the right place for debug / development scripts that are never intended to be supported or productionized. Skimming through the other repo's, I didn't see an obvious alternative. Maybe @thockin knows...

scratch-debugger/debug.sh
timstclair: Yes, it is statically compiled with uclibc. For the purposes here (1-off debugging), I don't think version should matter. Note that most of our _production_ images use busybox without version pinning...

luxas:
cc myself

verb:
This is a clever solution to a difficult problem, LGTM

scratch-debugger/debug.sh
```diff
@@ -0,0 +1,143 @@
...
+    ;;
+  *)
+    echo "Unsupported architecture: $ARCH"
+    exit 1
+esac
+
+# If the debugger is not already running...
+if ! ${KUBECTL} get pod ${DEBUGGER_NAME} &>/dev/null; then
+  # Start the debug pod
+  cat ${BASH_SOURCE[0]%/*}/debugger.yaml | \
```
verb: Suggestion: it might be nice to inline debugger.yaml so the script is stand-alone. I suspect this might see use outside of Kubernetes developers.

timstclair: Good point, done. This actually cleaned up the parameters too, so I moved the install script execution into the container command.

scratch-debugger/debug.sh
```diff
@@ -0,0 +1,179 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o nounset
+set -o errexit
+set -o pipefail
+
+if [[ $# == 0 ]]; then
+  echo >&2 "USAGE: $0 POD_NAME [POD_NAMESPACE CONTAINER_NAME]"
```
thockin: would be nice if namespace was `-n NAMESPACE` to be consistent with kubectl..

thockin: and -c CONTAINER

scratch-debugger/debug.sh
```diff
@@ -0,0 +1,179 @@
...
+  exit 1
+fi
+
+DOCKERCMD="/mnt/rootfs/usr/bin/docker -H unix:///run/docker.sock"
+
+${KUBECTL} create -f - <<EOF
+apiVersion: v1
+kind: Pod
+metadata:
+  name:        ${DEBUGGER_NAME}
```
thockin: why not use `generateName` to ensure no conflict?

scratch-debugger/debug.sh
```diff
@@ -0,0 +1,179 @@
...
+  containers:
+    - name:    debugger
+      image:   ${IMAGE}
+      securityContext:
+        privileged: true
+      command:
+        - sh
+        - -c
+        - "set -x; mkdir -p ${INSTALL_DIR} && \
+           ${DOCKERCMD} cp /tmp ${CONTAINER_ID}:/ && \
```
thockin: I don't grok this line?

scratch-debugger/debug.sh
```diff
@@ -0,0 +1,179 @@
...
+ARCH="${ARCH:-amd64}"
+
+# Name & namespace of target pod
+NAME=$1
+NAMESPACE=${2:-}
+CONTAINER_NAME=${3:-}
+
+# Internal variables
+KUBECTL="kubectl"
+[ -z "$CONTEXT" ] || KUBECTL="$KUBECTL --context=$CONTEXT"
```
thockin: inconsistent use of ${FOO} vs $FOO.  I hate "${FOO}" but it is what most shell in kube uses.

thockin (commented):
This is cute.  I'm not sure where to stick it.  We really don't want new stuff in contrib, but this certainly doesn't warrant a repo.  I guess I'm OK.

This does open an interesting design avenue for the "real" solution - this feels a bit like 'toolbox' on CoreOS and COS.  `kubectl toolbox POD -c CONTAINER --toolbox=busybox` --into=/tmp/toolbox` ?

timstclair:
@kubernetes/sig-contributor-experience-pr-reviews - Do we have a place for tools aimed at Kubernetes developers? I don't intend for this to be a tool for users (though there's nothing stopping people from using it, we want to build a more native experience version for them). Would it make sense to start a 'devel' repo, or rebrand contrib as that place? I bet we all have some scripts that could be useful to other Kubernetes devs.

ixdy:
It's not really a fit for `kubernetes/test-infra`, since that's more focused on things like CI.
It might fit in `kubernetes/repo-infra`, though that repo is pretty young, so I'm not sure what its intended use is.
I'd suggest `kubernetes/community` (e.g. in the `contributors/devel` subdir), though that repo seems to be more about docs than scripts or other code.

scratch-debugger/debug.sh
timstclair: Ugh, parsing flags in bash.... done.

scratch-debugger/debug.sh
timstclair: Done.

scratch-debugger/debug.sh
timstclair: Done.

scratch-debugger/debug.sh
timstclair: Commented each line. This specific one is funny, since there is no mkdir in the scratch container to create the target directory structure.

timstclair (commented):
Thanks Tim.

thockin:
LGTM, needs rebase

timstclair:
Thanks. Squashed & rebased. Reapplying LGTM.

PR #2378, opened 2017-02-11
Fix wrong task tags name

xialonglee:
The later tags name overrides the former, fixed this.


xialonglee:
cc @ingvagabund , i touch your code.

ingvagabund:
@xialonglee thanks :)

PR #2384, opened 2017-02-14
[keepalived-vip] Support for configurable virtual router id

cdemers:
Hi all, I submit this small PR.

It adds support for configurable keepalived _virtual router id_.   This functionality is necessary in order to be able to run multiple keepalived "clusters" in the same subnet without having them stepping on each other's toes (having one node from one group confusing the other group into thinking it has a master).

The configuration of the _virtual router id_ is done using the `--router-id` command line flag. I considered using the project's _configmap_ but the structure of the existing _configmap_ is not easily expandable, so I had to use a flag (I would have added support for environment variables, but it didn't seems to be in line with the way this application is thought out, so I refrained).

The flag can be specified in a k8s object with the `args` key, for example:
`- args: ["--services-configmap=admin/vip-configmap", "--router-id=45"]`

If not specified, it will default to router id _50_ as it was before.

That's pretty much all there is to it, thanks in advance for your time!


cdemers:
I signed it!

cdemers:
hah, used the wrong email (cncf-cla) for this commit... not sure there is a reasonable way to fix that without creating a new PR.

I would be very grateful if you might give me your opinion as to wether you might find this PR valuable, if so I'll submit a new PR with a cncf-cla valid email.

Thanks again!

cdemers:
I refactored a little, adding a clearer flag description, sanity check on the VRID and renaming variables.

cdemers:
Hi @mikedanese, any chance you might have a look at this? I'm not sure I follow the right process, I'm pretty new at this.
Thanks!

bowei (commented):
/lgtm

bowei (approved):

bowei:
LGTM

bowei:
I think you need to rebase and squash your commits: http://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git

cdemers:
Thanks!! I will rebase and squash, I'll also update the documentation.

cdemers:
My apologies, it looks like I broke the link to the "from" branch. Can't re-open it now, I submitted a new PR (#2508).

PR #2385, opened 2017-02-15
fix misspell "environment" in helpers.go

fate-grand-order:


Random-Liu:
@fate-grand-order Thanks for the fix!
/lgtm
/approve

PR #2386, opened 2017-02-15
Rebase 404-server on scratch & don't run as root

timstclair:
As part of the effort behind kubernetes/kubernetes#40248, I'm cutting out unnecessary dependencies from system containers by rebasing them on scratch (for containers without any external dependencies). In addition, this does not require any elevated privileges, so run as the nobody user instead of root.

For debugging scratch containers, see kubernetes/contrib#2372

/cc @ixdy

### DIFF
--- a/404-server/Makefile
+++ b/404-server/Makefile
@@ -20,40 +20,21 @@
 
 all: push
 
-TAG=1.2
+TAG=1.3
 PREFIX?=gcr.io/google_containers/defaultbackend
 ARCH?=amd64
-GOLANG_VERSION=1.6
+GOLANG_VERSION=1.7
 TEMP_DIR:=$(shell mktemp -d)
 
-# Set default base image dynamically for each arch
-ifeq ($(ARCH),amd64)
-	BASEIMAGE?=busybox
-endif
-ifeq ($(ARCH),arm)
-	BASEIMAGE?=armel/busybox
-endif
-ifeq ($(ARCH),arm64)
-	BASEIMAGE?=aarch64/busybox
-endif
-ifeq ($(ARCH),ppc64le)
-	BASEIMAGE?=ppc64le/busybox
-endif
-
 server: server.go
-	CGO_ENABLED=0 GOOS=linux GOARCH=$(ARCH) GOARM=6 go build -a -installsuffix cgo -ldflags '-w' -o server ./server.go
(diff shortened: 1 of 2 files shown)


404-server/Makefile
```diff
@@ -20,33 +20,19 @@
...
-ifeq ($(ARCH),ppc64le)
-	BASEIMAGE?=ppc64le/busybox
-endif
-
 server: server.go
 	CGO_ENABLED=0 GOOS=linux GOARCH=$(ARCH) GOARM=6 go build -a -installsuffix cgo -ldflags '-w' -o server ./server.go
 
 container:
 	# Copy the whole directory to a temporary dir and set the base image
 	cp -r ./* $(TEMP_DIR)
```
thockin: since we're not pre-processing Dockerfile why not simplify this to `docker build .`

thockin:
LGTM, but one comment.  Not sure if we care about the cross-platform for this, but can always bring it back.

404-server/Makefile
luxas: +1

timstclair: Done.

timstclair (commented):
Also updated to go1.7, and strip the debug symbols (`-s` linker flag).

404-server/Makefile
```diff
@@ -20,40 +20,20 @@
...
 container:
-	# Copy the whole directory to a temporary dir and set the base image
-	cp -r ./* $(TEMP_DIR)
-	cd $(TEMP_DIR) && sed -i "s|BASEIMAGE|$(BASEIMAGE)|g" Dockerfile
-
 	# Compile the binary inside a container for reliable builds
-	docker run --rm -it -v $(TEMP_DIR):/build golang:$(GOLANG_VERSION) /bin/bash -c "make -C /build server ARCH=$(ARCH)"
-
-	docker build --pull -t $(PREFIX)-$(ARCH):$(TAG) $(TEMP_DIR)
+	docker run --rm -it -v $(PWD):/build golang:$(GOLANG_VERSION) /bin/bash -c "make -C /build server ARCH=$(ARCH)"
```
ixdy: maybe `docker pull` the golang image first? otherwise you may get stuck on `1.7.1` or whatever.

(alternately, use `1.7.5` instead of `1.7` for the tag.)

timstclair: Done.

timstclair:
Rebased & squashed.

timstclair:
Pushed images.

PR #2390, opened 2017-02-16
Change base image for rescheduler to busybox

ixdy:
Tested in my own cluster against the `[k8s.io] Rescheduler [Serial] should ensure that critical pod is scheduled in case there is no resources available` test and it passed.

Note that because we write to `/var/log/rescheduler.log` on the master via stdout/stderr redirect, we need busybox and we need to run as root.

cc @timstclair


ixdy:
ping @piosz

piosz:
LGTM

PR #2391, opened 2017-02-17
Add Docker upstart dependency on Flannel for Ubuntu 14.04

unicell:
Also adding respawn delay to avoid Docker respawning too fast

[1] http://upstart.ubuntu.com/cookbook/#delay-respawn-of-a-job

### DIFF
--- a/ansible/roles/flannel/tasks/ubuntu.yml
+++ b/ansible/roles/flannel/tasks/ubuntu.yml
@@ -13,3 +13,16 @@
   when: ansible_distribution_major_version|int < 15
   notify:
     - restart flannel
+
+- name: Ubuntu | Configure Docker / Flannel upstart dependency
+  lineinfile:
+       dest: "/etc/init/docker.conf"
+       state: "{{ item.state }}"
+       line: "{{ item.line }}"
+       regexp: "{{ item.regexp }}"
+  with_items:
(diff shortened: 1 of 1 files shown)


ingvagabund:
What happens in you run the flannel role without docker?

unicell:
@ingvagabund The whole `roles/flannel/tasks/ubuntu.yml` file is supposed to update Docker so that it can work together with Flannel, for Ubuntu distro. My change is to add service dependency to ensure Docker start after Flannel. That's in accordance to what this playbook supposed to handle.

Also, in `roles/flannel/tasks/main.yml` it will also try bounce docker once flanneld started. So to your question, I don't think current flannel playbook allows you to run without Docker.

ingvagabund:
> Also, in `roles/flannel/tasks/main.yml` it will also try bounce docker once flanneld started. So to your question, I don't think current flannel playbook allows you to run without Docker.

That's unfortunate. The docker should be de-coupled from the flannel role. Think you can do that? E.g. introduce a mechanism to detect if the docker is installed and iff notify the handler and run your new task? Introducing `docker_on_system` or similar variable should be sufficient for the moment (set to `False` be default. Then updating [1] and setting the variable to `True`. Unless you know a way how to detect the docker universally. I would not want to run `rpm -q docker` on Fedora distros and `apt-get ...` on other. Less if-else structures the better.

[1] https://github.com/kubernetes/contrib/blob/master/ansible/playbooks/deploy-cluster.yml#L26

PR #2393, opened 2017-02-17
Remove old logging info and replace with a link to the docs

crassirostris:
Container images and the general approach are obsolete and duplicating the documentation.


Crassirostris:
@mikedanese Could you please approve?

PR #2395, opened 2017-02-20
generate a list of years from 2014 to the current year

ingvagabund:
Not the best solution but at least we don't have to extend the list each year.


eparis:
can you take this change to https://github.com/kubernetes/repo-infra/  please?

PR #2396, opened 2017-02-20
Cluster-autoscaler: api type object for status reporting

mwielgus:
As the PR with api (https://github.com/kubernetes/kubernetes/pull/40513) rose some controversy I would like to add it temporarily add it to CA codebase so that CA can use the struct to build the status information. If the PR is approved the api struct will simply be replaced with the one from K8S codebase. If not - a http endpoint using these structs will be exposed in CA.


jszczepkowski:
LGTM

PR #2398, opened 2017-02-21
Cluster-autoscaler: GetState() - health condition

mwielgus:
This PR is the 1st out of 3 providing ClusterAutoscalerStatus object populated with the current CA state. 


jszczepkowski:
Just one comment: please add unittests for unhealthy status. After it: LGTM

jszczepkowski:
It seems that there is already test for unhealthy.

LGTM

PR #2404, opened 2017-02-22
Cluster-autoscaler: skip pods that are being deleted in node drain

mwielgus:
Pods that are still running but are being deleted should not be deleted/moved in CA. 
This also fixes CA against the change in K8S after which the node controller stopped hard-deleting pods (it just sets deletionTimestamp) from unready nodes. With this PR CA will see nodes with all pods under deletion as empty.

cc: @jszczepkowski @gmarek @maciekpytel


gmarek:
LGTM

PR #2407, opened 2017-02-23
Add launch ansible-playbook example in Vagrant README

harobed:
Add launch ansible-playbook example in Vagrant README.

### DIFF
--- a/ansible/vagrant/README.md
+++ b/ansible/vagrant/README.md
@@ -199,6 +199,19 @@ kube-master | SUCCESS => {
 ...
 ```
 
+Example to launch one playbooks:
+
+```
+$ ansible-playbook ../playbooks/deploy-cluster.yml
+...
+PLAY RECAP *********************************************************************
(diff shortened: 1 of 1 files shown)


harobed:
CLA signed.

ansible/vagrant/README.md
```diff
@@ -199,6 +199,19 @@ kube-master | SUCCESS => {
 ...
 ```
 
+Example to launch one playbooks:
```
ingvagabund: Why do we need it? The `deploy-cluster.yml` playbook is run when ``vagrant up`` is executed.

harobed: If user want launch some playbooks after vagrant provisioning.

ingvagabund: Can you mentioned that information with the example? Plus, I don't think it will work unless you specify vagrant's inventory file. Which is at ``.vagrant/provisioners/ansible/inventory/vagrant_ansible_inventory``.

PR #2410, opened 2017-02-23
Add generic fluentd image for Stackdriver Logging

crassirostris:
This PR creates a fluentd image which can be configured in any way to export logs to Stackdriver Logging.

CC @igorpeshansky @qingling128

### DIFF
--- a/kube-fluentd/README.md
+++ b/kube-fluentd/README.md
@@ -1,0 +1,16 @@
+# Overview
+
+This directory contains docker images with fluentd, which can be later configured to collect
+log from any source.
+
+# Usage
+
+Each image is built with its own set of plugins which you can later use in configuration. Set of
+plugin is enumerated in Gemfile. You can find details about fluentd configuration on the
+[official site](http://docs.fluentd.org/articles/config-file).
+
+In order to configure fluentd image, you should mount directory with `.conf` files to
+`/etc/fluent/config` or add files to this directory by building a new image on top. All `.conf`
+files will be included to the final fluentd configuration.
+
+Command line arguments to the fluentd can be passed via environment variable `FLUENTD_ARGS`.
--- a/kube-fluentd/README.md
+++ b/kube-fluentd/README.md
@@ -1,16 +1,19 @@
 # Overview
 
-This directory contains docker images with fluentd, which can be later configured to collect
-log from any source.
+This directory contains docker images with fluentd, which can be later
+configured to collect and ingest logs.
 
 # Usage
 
-Each image is built with its own set of plugins which you can later use in configuration. Set of
-plugin is enumerated in Gemfile. You can find details about fluentd configuration on the
+Each image is built with its own set of plugins which you can later use
+in the configuration. The set of plugin is enumerated in a Gemfile in the
+image's directory. You can find details about fluentd configuration on the
 [official site](http://docs.fluentd.org/articles/config-file).
 
-In order to configure fluentd image, you should mount directory with `.conf` files to
-`/etc/fluent/config` or add files to this directory by building a new image on top. All `.conf`
-files will be included to the final fluentd configuration.
+In order to configure fluentd image, you should mount directory with `.conf`
+files to `/etc/fluent/config.d` or add files to that directory by building
+a new image on top. All `.conf` files in the `/etc/fluent/config.d` directory
+will be included to the final fluentd configuration.
 
-Command line arguments to the fluentd can be passed via environment variable `FLUENTD_ARGS`.
+Command line arguments to the fluentd executable are passed
+via environment variable `FLUENTD_ARGS`.
--- a/kube-fluentd/gcp/Dockerfile
+++ b/kube-fluentd/gcp/Dockerfile
@@ -1,0 +1,130 @@
+# Copyright 2017 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This Dockerfile will build an image that is configured
+# to use Fluentd to collect all Docker container log files
+# and then cause them to be ingested using the Google Cloud
+# Logging API. This configuration assumes that the host performning
+# the collection is a VM that has been created with a logging.write
+# scope and that the Logging API has been enabled for the project
+# in the Google Developer Console.
+
+FROM gcr.io/google_containers/ubuntu-slim:0.7
+
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
+        bzip2 \
+        ca-certificates \
+        libffi-dev \
+        libgdbm3 \
+        libssl-dev \
+        libyaml-dev \
+        procps \
+        zlib1g-dev \
+        libjemalloc1 \
+    && rm -rf /var/lib/apt/lists/*
+
+# skip installing gem documentation
+RUN mkdir -p /usr/local/etc \
+    && { \
+        echo 'install: --no-document'; \
+        echo 'update: --no-document'; \
+    } >> /usr/local/etc/gemrc
+
+ENV RUBY_MAJOR 2.3
+ENV RUBY_VERSION 2.3.3
+ENV RUBY_DOWNLOAD_SHA256 1a4fa8c2885734ba37b97ffdb4a19b8fba0e8982606db02d936e65bac07419dc
+ENV RUBYGEMS_VERSION 2.6.10
+
+ENV BUNDLER_VERSION 1.14.4
+
+# install things globally, for great justice
+# and don't create ".bundle" in all our apps
+ENV GEM_HOME /usr/local/bundle
+ENV BUNDLE_PATH="$GEM_HOME" \
+    BUNDLE_BIN="$GEM_HOME/bin" \
+    BUNDLE_SILENCE_ROOT_WARNING=1 \
+    BUNDLE_APP_CONFIG="$GEM_HOME"
+ENV PATH $BUNDLE_BIN:$PATH
+RUN mkdir -p "$GEM_HOME" "$BUNDLE_BIN" \
+    && chmod 777 "$GEM_HOME" "$BUNDLE_BIN"
+
+# Gemfile with dependencies
+COPY Gemfile Gemfile
+
+# some of ruby's build scripts are written in ruby
+#   we purge system ruby later to make sure our final image uses what we just built
+RUN set -ex \
+    \
+    && buildDeps=' \
+        autoconf \
+        bison \
+        gcc \
+        libbz2-dev \
+        libgdbm-dev \
+        libglib2.0-dev \
+        libncurses-dev \
+        libreadline-dev \
+        libxml2-dev \
+        libxslt-dev \
+        make \
+        ruby \
+        wget \
+        xz-utils \
+    ' \
+    && apt-get update \
+    && apt-get install -y --no-install-recommends $buildDeps \
+    && rm -rf /var/lib/apt/lists/* \
+    \
+    && wget -O ruby.tar.xz "https://cache.ruby-lang.org/pub/ruby/${RUBY_MAJOR%-rc}/ruby-$RUBY_VERSION.tar.xz" \
+    && echo "$RUBY_DOWNLOAD_SHA256 *ruby.tar.xz" | sha256sum -c - \
+    \
+    && mkdir -p /usr/src/ruby \
+    && tar -xJf ruby.tar.xz -C /usr/src/ruby --strip-components=1 \
+    && rm ruby.tar.xz \
+    \
+    && cd /usr/src/ruby \
+    \
+# hack in "ENABLE_PATH_CHECK" disabling to suppress:
+#   warning: Insecure world writable dir
+    && { \
+        echo '#define ENABLE_PATH_CHECK 0'; \
+        echo; \
+        cat file.c; \
+    } > file.c.new \
+    && mv file.c.new file.c \
+    \
+    && autoconf \
+    && ./configure --disable-install-doc --enable-shared \
+    && make -j"$(nproc)" \
+    && make install \
+    \
+    && cd / \
+    && rm -r /usr/src/ruby \
+    \
+    && gem update --system "$RUBYGEMS_VERSION" \
+    \
+    && gem install bundler --version "$BUNDLER_VERSION" \
+    \
+    && bundle install \
+    \
+    && apt-get purge -y --auto-remove $buildDeps
+
+# copy the base Fluentd configuration
+COPY fluent.conf /etc/fluent/fluent.conf
+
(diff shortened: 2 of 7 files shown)


kube-fluentd/README.md
```diff
@@ -0,0 +1,16 @@
+# Overview
+
+This directory contains docker images with fluentd, which can be later configured to collect
+log from any source.
```
igorpeshansky: "logs from any source"

kube-fluentd/README.md
```diff
@@ -0,0 +1,16 @@
+# Overview
+
+This directory contains docker images with fluentd, which can be later configured to collect
```
igorpeshansky: [Optional] "fluentd", or "Google Cloud Logging agent"?

kube-fluentd/README.md
```diff
@@ -0,0 +1,16 @@
...
+log from any source.
+
+# Usage
+
+Each image is built with its own set of plugins which you can later use in configuration. Set of
+plugin is enumerated in Gemfile. You can find details about fluentd configuration on the
```
igorpeshansky: "The set of plugins is enumerated"

kube-fluentd/README.md
```diff
@@ -0,0 +1,16 @@
...
+[official site](http://docs.fluentd.org/articles/config-file).
+
+In order to configure fluentd image, you should mount directory with `.conf` files to
+`/etc/fluent/config` or add files to this directory by building a new image on top. All `.conf`
+files will be included to the final fluentd configuration.
```
igorpeshansky: All files *in that directory* will be included...

kube-fluentd/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+Command line arguments to the fluentd can be passed via environment variable `FLUENTD_ARGS`.
```
igorpeshansky: Either "to the fluentd executable" or just "to fluentd".

kube-fluentd/gcp/Dockerfile
```diff
@@ -0,0 +1,130 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This Dockerfile will build an image that is configured
+# to use Fluentd to collect all Docker container log files
+# and then cause them to be ingested using the Google Cloud
+# Logging API. This configuration assumes that the host performning
+# the collection is a VM that has been created with a logging.write
```
igorpeshansky: The logging agent also allows using private key credentials (https://cloud.google.com/logging/docs/agent/authorization). Is that supported by this image? If yes, let's document this. If not, how can we make it supported?

Crassirostris:
@bprashanth @eparis Could you please take a look and approve it?

kube-fluentd/gcp/Dockerfile
```diff
@@ -0,0 +1,130 @@
...
+    BUNDLE_APP_CONFIG="$GEM_HOME"
+ENV PATH $BUNDLE_BIN:$PATH
+RUN mkdir -p "$GEM_HOME" "$BUNDLE_BIN" \
+    && chmod 777 "$GEM_HOME" "$BUNDLE_BIN"
+
+# Gemfile with dependencies
+COPY Gemfile Gemfile
+
+# some of ruby's build scripts are written in ruby
+#   we purge system ruby later to make sure our final image uses what we just built
```
igorpeshansky: Is this extra indentation intentional?

kube-fluentd/gcp/Dockerfile
```diff
@@ -0,0 +1,130 @@
...
+    \
+    && gem update --system "$RUBYGEMS_VERSION" \
+    \
+    && gem install bundler --version "$BUNDLER_VERSION" \
+    \
+    && bundle install \
+    \
+    && apt-get purge -y --auto-remove $buildDeps
+
+# copy the base Fluentd configuration
```
igorpeshansky: Why do you capitalize "fluentd"?

kube-fluentd/gcp/Gemfile
```diff
@@ -0,0 +1,7 @@
+source "https://rubygems.org"
+
+gem 'fluentd', '~>0.12.31'
```
igorpeshansky: `fluent-plugin-google-cloud` depends on fluentd version 0.14. Are you intentionally downgrading it? Won't installing the other gem pull in both?

kube-fluentd/gcp/fluent.conf
```diff
@@ -0,0 +1,8 @@
+# This is the root config file, which only includes components of the actual configuration
+
+# Do not collect fluentd's own logs to avoid infinite loops.
+<match fluent.**>
+  type null
+</match>
+
+@include /etc/fluent/config/*.conf
```
igorpeshansky: You should at least configure the google-cloud output plugin here, else users will have to configure it themselves whenever they mount the whole config directory.

kube-fluentd/gcp/run.sh
```diff
@@ -0,0 +1,31 @@
...
+
+# If you attach host lib directory, this will copy the systemd lib
+# This may be necessary in order for journald to work
+if [ ! -z "`ls /host/lib/libsystemd* 2>/dev/null`" ]
+then
+  rm /lib/x86_64-linux-gnu/libsystemd*
+  cp -a /host/lib/libsystemd* /lib/x86_64-linux-gnu/
+fi
+
+# Jemalloc is a better malloc implementation
```
igorpeshansky: I think it's more that fluentd requires jemalloc to avoid memory fragmentation...

kube-fluentd/gcp/run.sh
```diff
@@ -0,0 +1,31 @@
...
+LD_PRELOAD=/usr/lib/x86_64-linux-gnu/libjemalloc.so.1
+
+fluentd $@
```
igorpeshansky: Might be best to actually specify the main config file path, rather than rely on the compiled-in default...

kube-fluentd/README.md
```diff
@@ -0,0 +1,16 @@
...
+log from any source.
+
+# Usage
+
+Each image is built with its own set of plugins which you can later use in configuration. Set of
+plugin is enumerated in Gemfile. You can find details about fluentd configuration on the
+[official site](http://docs.fluentd.org/articles/config-file).
+
+In order to configure fluentd image, you should mount directory with `.conf` files to
+`/etc/fluent/config` or add files to this directory by building a new image on top. All `.conf`
```
igorpeshansky: This gem comes with no default configuration. If users mount the config files over `/etc/fluentd/config`, they'll overwrite/hide whatever config you have there. It might be best to split configurations into the main config (in `/etc/fluent/fluent.conf` or whatever name you pick for your main config file), vendor config (e.g., in `/etc/fluentd/config`), and user config (e.g., in `/etc/fluentd/config/extra`).

igorpeshansky: Traditionally, directories with extra configuration files have the `.d` suffix, so how about `/etc/fluentd/config.d`?

igorpeshansky (changes requested):

kube-fluentd/gcp/Gemfile
Crassirostris: Are you sure? Judging from the https://rubygems.org/gems/fluent-plugin-google-cloud/versions/0.5.6, minor fluentd version is not fixed

That's true for all gems, so I'm installing 0.12 explicitly, because 0.14 doesn't work yet (I think I mentioned it to you, the problem with `failed to write data into buffer by buffer overflow`)

kube-fluentd/gcp/fluent.conf
Crassirostris: I intentionally left it this way, because for now we want to able to tweak these parameters, including in tests, as they influence the stability of fluentd container

This part will be removed from the configuration with switching to the Stackdriver container image.

kube-fluentd/gcp/run.sh
Crassirostris: The problem with that, as with `--no-supervisor` is that someone might want to override it, I think it's better to be flexible here. Also, this should not be the problem, because the dependencies are fixed, implicit conventions should not change across patch releases

kube-fluentd/README.md
Crassirostris: I plan to move fluentd image for ES here, so fluentd

Crassirostris:
@igorpeshansky Thanks a lot for your comments, I will address them shortly!

kube-fluentd/gcp/run.sh
Crassirostris: ACK

kube-fluentd/gcp/Dockerfile
Crassirostris: ACK

kube-fluentd/gcp/Dockerfile
Crassirostris: Originally it is, but I will make it conform to the other pieces of comments we have

ACK

kube-fluentd/README.md
Crassirostris: ACK

kube-fluentd/README.md
Crassirostris: ACK

kube-fluentd/README.md
Crassirostris: > Traditionally, directories with extra configuration files have the .d suffix, so how about /etc/fluentd/config.d?

ACK

> If users mount the config files over `/etc/fluentd/config`, they'll overwrite/hide whatever config you have there. It might be best to split configurations into the main config (in `/etc/fluent/fluent.conf` or whatever name you pick for your main config file), vendor config (e.g., in `/etc/fluentd/config`), and user config (e.g., in `/etc/fluentd/config/extra`).

As I mentioned somewhere below, I wanted to be able to control the output plugin for now, so it doesn't make sense to separate those two.

What you describe is actually achievable by building a new image on top of one of these, to infer some additional structure.

Crassirostris: Nested includes should work, right?

kube-fluentd/README.md
Crassirostris: ACK

kube-fluentd/README.md
Crassirostris: ACK

Crassirostris:
> The logging agent also allows using private key credentials (https://cloud.google.com/logging/docs/agent/authorization). Is that supported by this image? If yes, let's document this. If not, how can we make it supported?

Won't it work just by mounting the credentials to the image?

kube-fluentd/gcp/Dockerfile
```diff
@@ -0,0 +1,126 @@
...
+ENV BUNDLE_PATH="$GEM_HOME" \
+    BUNDLE_BIN="$GEM_HOME/bin" \
+    BUNDLE_SILENCE_ROOT_WARNING=1 \
+    BUNDLE_APP_CONFIG="$GEM_HOME"
+ENV PATH $BUNDLE_BIN:$PATH
+RUN mkdir -p "$GEM_HOME" "$BUNDLE_BIN" \
+    && chmod 777 "$GEM_HOME" "$BUNDLE_BIN"
+
+# Gemfile with dependencies
+COPY Gemfile Gemfile
```
dhrupadb: @Crassirostris Probably a silly question but could you explain where this Gemfile is coming from? It wasn't obvious by reading the Dockerfile.

Crassirostris: It's copied from the local folder on build, you can see its contents in the diff

kube-fluentd/gcp/Gemfile
dhrupadb: Ah okay it's from this repo itself. Understood thanks!

kube-fluentd/gcp/Dockerfile
```diff
@@ -0,0 +1,126 @@
...
+        libxml2-dev \
+        libxslt-dev \
+        make \
+        ruby \
+        wget \
+        xz-utils \
+    ' \
+    && apt-get update \
+    && apt-get install -y --no-install-recommends $buildDeps \
+    && rm -rf /var/lib/apt/lists/* \
```
Crassirostris: It has to be done every time for the sake of image size. Each RUN is a new layer for union filesystem. If we leave these files for the next layer to delete, resulting image will still contain them in the first layer.

kube-fluentd/gcp/Dockerfile
```diff
@@ -0,0 +1,126 @@
...
+    && cd /usr/src/ruby \
+    \
+# Hack in "ENABLE_PATH_CHECK" disabling to suppress:
+# warning: Insecure world writable dir
+    && { \
+        echo '#define ENABLE_PATH_CHECK 0'; \
+        echo; \
+        cat file.c; \
+    } > file.c.new \
+    && mv file.c.new file.c \
```
dhrupadb: This may be cleaner perhaps? 

`&&  sed -i '1s/^/\#define ENABLE_PATH_CHECK 0\n/' file.c  \`

kube-fluentd/gcp/Dockerfile
```diff
@@ -0,0 +1,126 @@
...
+    && rm -r /usr/src/ruby \
+    \
+    && gem update --system "$RUBYGEMS_VERSION" \
+    \
+    && gem install bundler --version "$BUNDLER_VERSION" \
+    \
+    && bundle install \
+    \
+    && apt-get purge -y --auto-remove $buildDeps
+
```
dhrupadb: Would it be easier to abstract away the process of installing Ruby into a separate script and having the Dockerfile invoke that instead?

kube-fluentd/gcp/Dockerfile
Crassirostris: This is copied from https://github.com/docker-library/ruby/blob/17f666371fa5c88c48a38bc53c611f26c7455778/2.3/slim/Dockerfile I would prefer to keep it close to the original.

However, I will add a comment, pointing to this Dockerfile

kube-fluentd/gcp/Dockerfile
Crassirostris: That's possible, but that would split the code into two files, which is not great. I think it's better to have everything about build in one file

Just in case, `run.sh` is a different story, it's not invoked on build

kube-fluentd/gcp/Dockerfile
dhrupadb: Sounds good. Thanks for explaining all of this!

kube-fluentd/gcp/Dockerfile
dhrupadb: 👍

kube-fluentd/gcp/Dockerfile
dhrupadb: 👍

Crassirostris:
Closing, reasoning is given in https://github.com/kubernetes/kubernetes/pull/41998

kube-fluentd/gcp/Dockerfile
timstclair: I think when the install commands get complex enough it is much more readable to break out a separate script, even at the cost of the extra file. Especially since the 3 run commands in this Dockerfile will result in 3 separate image layers.

piosz:
Reopened due to https://github.com/kubernetes/kubernetes/pull/41998#issuecomment-282126884

piosz:
Please continue working on this.

Crassirostris:
Closing in favor of https://github.com/kubernetes/contrib/pull/2426

PR #2411, opened 2017-02-23
Cluster-autoscaler: expand half-deleted pod skip logic in drain

mwielgus:
We should not immediately discard deleting nodes from drain to considering a node as empty immediately after the pod deletion was started.

cc: @jszczepkowski @fgrzadkowski @maciekpytel


jszczepkowski:
Please add unittest, then LGTM

mwielgus:
Test added.

PR #2413, opened 2017-02-23
Cluster-autoscaler: implemented LogCollector

MaciekPytel:


cluster-autoscaler/clusterstate/api/utils.go
```diff
@@ -90,3 +99,65 @@ func (status ClusterAutoscalerStatus) GetReadableString() string {
 	}
 	return buffer.String()
 }
+
+// LogItem is a single entry in log managed by LogCollector.
+type LogItem struct {
+	// Log is the logged message body.
+	Log string
+	// Timestamp when the Log was created.
+	Timestamp time.Time
```
mwielgus: Add log level (Debug, Info, Warning, Error)

cluster-autoscaler/clusterstate/api/utils.go
```diff
@@ -90,3 +99,65 @@ func (status ClusterAutoscalerStatus) GetReadableString() string {
 	}
 	return buffer.String()
 }
+
+// LogItem is a single entry in log managed by LogCollector.
+type LogItem struct {
```
mwielgus: Move to types.

cluster-autoscaler/clusterstate/api/utils.go
```diff
@@ -90,3 +99,65 @@ func (status ClusterAutoscalerStatus) GetReadableString() string {
...
+	// Log is the logged message body.
+	Log string
+	// Timestamp when the Log was created.
+	Timestamp time.Time
+}
+
+// LogCollector keeps recent log events. It is automatically truncated on each access based on predefined set of conditions.
+type LogCollector struct {
```
mwielgus: This is not a part of any api. Lets move the whole file to ./clusterstate/utils/

cluster-autoscaler/clusterstate/api/utils.go
```diff
@@ -90,3 +99,65 @@ func (status ClusterAutoscalerStatus) GetReadableString() string {
...
+	sync.Mutex
+	// MaxItems is the maximum size of the log. Adding further messages will cause the oldest ones to be dropped.
+	MaxItems int
```
mwielgus: no need for public.

cluster-autoscaler/clusterstate/api/utils.go
```diff
@@ -90,3 +99,65 @@ func (status ClusterAutoscalerStatus) GetReadableString() string {
...
+	// ItemLifetime is the maximum time an item can be stored in log.
+	ItemLifetime time.Duration
```
mwielgus: same here and below.

cluster-autoscaler/clusterstate/api/utils.go
MaciekPytel: In case of MaxItems, and ItemLifetime - should I add a setter for them then?

cluster-autoscaler/clusterstate/api/utils.go
```diff
@@ -90,3 +99,65 @@ func (status ClusterAutoscalerStatus) GetReadableString() string {
...
+}
+
+// To be executed under a lock.
+func (lc *LogCollector) compact(now time.Time) {
+	firstIndex := 0
+	if len(lc.Store) > lc.MaxItems {
+		firstIndex = len(lc.Store) - lc.MaxItems
+	}
+	for ; firstIndex < len(lc.Store); firstIndex++ {
+		lifetimeEnd := lc.Store[firstIndex].Timestamp.Add(lc.ItemLifetime)
```
mwielgus: what about:
threshold := now.Add(-lc.ItemLifetime)
...
if lc.Store[firstIndex].After(threshold) { break }

In the current form it is a bit tricky to understand.

cluster-autoscaler/clusterstate/api/utils.go
```diff
@@ -90,3 +99,65 @@ func (status ClusterAutoscalerStatus) GetReadableString() string {
...
+			break
+		}
+	}
+	if firstIndex > 0 {
+		lc.Store = append(lc.Store[:0], lc.Store[firstIndex:]...)
+	}
+}
+
+// Log logs a single provided message in LogCollector.
+func (lc *LogCollector) Log(msg string, timestamp time.Time) {
```
mwielgus: Log level.

cluster-autoscaler/clusterstate/api/utils.go
```diff
@@ -90,3 +99,65 @@ func (status ClusterAutoscalerStatus) GetReadableString() string {
...
+	lc.Lock()
+	defer lc.Unlock()
+	lc.Store = append(lc.Store, LogItem{Log: msg, Timestamp: timestamp})
+	lc.compact(timestamp)
```
mwielgus: A wrong timestamp from the user can break the whole thing. We should not pass it.

cluster-autoscaler/clusterstate/api/utils.go
```diff
@@ -90,3 +99,65 @@ func (status ClusterAutoscalerStatus) GetReadableString() string {
...
+		firstIndex = len(lc.Store) - lc.MaxItems
+	}
+	for ; firstIndex < len(lc.Store); firstIndex++ {
+		lifetimeEnd := lc.Store[firstIndex].Timestamp.Add(lc.ItemLifetime)
+		if now.Before(lifetimeEnd) {
+			break
+		}
+	}
+	if firstIndex > 0 {
+		lc.Store = append(lc.Store[:0], lc.Store[firstIndex:]...)
```
mwielgus: Just allocate a new array and copy data. With this it is a bit hard to tell what exactly is going on here. Clarity is better than minor performance gain.

cluster-autoscaler/clusterstate/api/utils.go
mwielgus: You can have an internal version of Log that takes msg, level and timestamp for testing purposes.

cluster-autoscaler/clusterstate/api/utils.go
MaciekPytel: I took that from go wiki: https://github.com/golang/go/wiki/SliceTricks :)

cluster-autoscaler/clusterstate/utils/types.go
```diff
@@ -0,0 +1,60 @@
...
+	DefaultLogCollectorMaxItems = 50
+	// DefaultLogCollectorItemLifetime is the default time after which LogItem will be removed from LogCollector.
+	DefaultLogCollectorItemLifetime = 15 * time.Minute
+)
+
+type LogLevel int
+
+const (
+	// Debug log level.
+	Debug = iota
```
mwielgus: Only first const should have iota. 
We try not to have iotas in apis - one extra type and whole sequence is broken.
Lets have strings here - it will be easier to write the print function and make the consts explicitly typed.

cluster-autoscaler/clusterstate/utils/types.go
```diff
@@ -0,0 +1,60 @@
+/*
```
mwielgus: I was rather thinking about putting log to api. But we don't have to - it can stay here but please combine logging.go and types.go - there is no reason to keep them separate now.

MaciekPytel: I am confused now. I understood from your earlier comment that I should move it here?

> This is not a part of any api. Lets move the whole file to ./clusterstate/utils/

cluster-autoscaler/clusterstate/utils/types.go
MaciekPytel: good points, will do

cluster-autoscaler/clusterstate/utils/types.go
mwielgus: I meant:

* LogItem types to ./clusterstate/api. It may eventually go to the end user some day via an api.
* LogCollector code ./clusterstate/utils (or ./utils). It will never get to the user.

Maybe I wasn't clear enough. Sorry for confusion.

cluster-autoscaler/clusterstate/utils/logging.go
```diff
@@ -0,0 +1,107 @@
...
+	"sync"
+	"time"
+)
+
+// LogLevel describes severity of log entry.
+type LogLevel string
+
+const (
+	// Debug log level.
+	Debug = "DEBUG"
```
mwielgus: LogLevel Debug = "DEBUG"

PR #2416, opened 2017-02-23
Fix nil pointer exception when nodegroupfornode is nil

andrewstuart:
This fix should prevent a segmentation fault when a NodeGroup comes back as nil from the cloudprovider. I've not been able to track down the root cause, but I figured this might be useful for others.

In case the following is relevant or helpful:

K8s version: 1.5.2
Cluster tool: kops 1.5.1
Cloud: AWS
cluster-autoscaler version: master


MaciekPytel:
This has been fixed in https://github.com/kubernetes/contrib/pull/2471, closing the old PR.

PR #2420, opened 2017-02-25
[Rescheduler] Added Prometheus metrics

piosz:
ref https://github.com/kubernetes/kubernetes/issues/40736

cc @fabxc @brancz


rescheduler/rescheduler.go
```diff
@@ -112,6 +123,11 @@ func main() {
 				if len(criticalPods) > 0 {
 					for _, pod := range criticalPods {
 						glog.Infof("Critical pod %s is unschedulable. Trying to find a spot for it.", podId(pod))
+						k8sApp := "unknown"
+						if l, found := pod.ObjectMeta.Labels["k8s-app"]; found {
```
brancz: Is this an actual formalized convention? If so can you point it out? I've seen some cluster bootstrapping tools that don't do this according to this convention, and it would be nice to be able to point at something for reference.

brancz:
lgtm

rescheduler/rescheduler.go
piosz: I'm not sure whether it's formalized but `k8s-app` is applied to every critical addon. I understand that it's not ideal but I'd like to have some metadata about problems with different addons and haven't found anything better (pod name is changed when a pod is restarted).

piosz:
@eparis can you please approve?

rescheduler/rescheduler.go
brancz: Oh don't get me wrong, I'm all for doing this, I've been trying to push for this convention in [some](https://github.com/kubernetes/kops/issues/1226) bootstrapping tools. What I'm saying is, that it would be neat if we can formalize this to get consistency between clusters even if created by different tools. But that should rather be tracked somewhere else, I was just wondering if something like that already exists.

piosz: I totally agree that it makes sense. You can try to open an issue in the main repo and cc Cluster Lifecycle team and Brian Grant. But let's start with something. In the worsts case it will be just "unknown".

brancz: Yep sgtm :+1:

rutsky: @brancz have you opened issue to introduce standard/best practices for labelling k8s-apps?

rutsky: Ok, I see it: https://github.com/kubernetes/kubernetes/issues/42887

PR #2424, opened 2017-02-26
Add ability to set multiple services on the same ip

gkopylov:
Now you can use single ip for multiple services.


keepalived-vip/controller.go
```diff
@@ -170,44 +171,55 @@ func (ipvsc *ipvsControllerController) getServices(cfgMap *api.ConfigMap) []vip
...
-		if err != nil {
-			glog.Warningf("error getting service %v: %v", nsSvc, err)
-			continue
-		}
+	for externalIP, LnsSvcLvs := range cfgMap.Data {
+		usedPorts := map[int]string{}
+		for _, nsSvcLvs := range strings.Split(LnsSvcLvs, "\n") {
+			ns, svc, lvsm, err := parseNsSvcLVS(nsSvcLvs)
+			if err != nil {
+				glog.Warningf("%v", err)
```
bowei: `glog.Errorf("Error parsing namespace/service entry (%q): %v", nsSvcLvs, err)`

keepalived-vip/controller.go
```diff
@@ -170,44 +171,55 @@ func (ipvsc *ipvsControllerController) getServices(cfgMap *api.ConfigMap) []vip
...
+			}
 
-		if !svcExists {
-			glog.Warningf("service %v not found", nsSvc)
-			continue
-		}
+			nsSvc := fmt.Sprintf("%v/%v", ns, svc)
+			svcObj, svcExists, err := ipvsc.svcLister.Indexer.GetByKey(nsSvc)
+			if err != nil {
+				glog.Warningf("error getting service %v: %v", nsSvc, err)
```
bowei: These should really be errors, not warnings...

keepalived-vip/controller.go
```diff
@@ -170,44 +171,55 @@ func (ipvsc *ipvsControllerController) getServices(cfgMap *api.ConfigMap) []vip
...
+				continue
+			}
 
-		s := svcObj.(*api.Service)
-		for _, servicePort := range s.Spec.Ports {
-			ep := ipvsc.getEndpoints(s, &servicePort)
-			if len(ep) == 0 {
-				glog.Warningf("no endpoints found for service %v, port %+v", s.Name, servicePort)
+			if !svcExists {
+				glog.Warningf("service %v not found", nsSvc)
```
bowei: same here

keepalived-vip/controller.go
```diff
@@ -170,44 +171,55 @@ func (ipvsc *ipvsControllerController) getServices(cfgMap *api.ConfigMap) []vip
...
+			s := svcObj.(*api.Service)
+			for _, servicePort := range s.Spec.Ports {
+				ep := ipvsc.getEndpoints(s, &servicePort)
+				if len(ep) == 0 {
+					glog.Warningf("no endpoints found for service %v, port %+v", s.Name, servicePort)
+					continue
+				}
+
+				if usedPort, ok := usedPorts[int(servicePort.Port)]; ok {
+					glog.Warningf("port %v of the service %v has already been allocated by the service %v",
```
bowei: error

bowei (changes requested):
Can you write a unit test for this logic?

Take a look at `contrib/cluster-autoscaler/core/autoscaler_test.go`, it uses the `fake` interfaces from client-go to mock the cluster interaction.

PR #2427, opened 2017-02-27
Change ansible files to respect etcd_client_port variable

bryanbecker:
The current implementation is inconsistent, with some roles using `etcd_client_port`, and some roles hard-coding "2379" as the port.

This patch fixes those situations.  I also added `etcd_client_port` into `group_vars` so the user only needs to set it once (instead of setting it per role)


bryanbecker:
CLA signed

ingvagabund:
@redhat-k8s-bot test please

ingvagabund:
Cool, thanks for the patch :)
/lgtm
/approve

PR #2431, opened 2017-02-28
Ca configmap cleanup

MaciekPytel:


cluster-autoscaler/run.sh
```diff
@@ -0,0 +1,30 @@
...
+# cluster-autoscaler binary.
+
+if [ -z "$LOG_OUTPUT" ]; then
+  LOG_OUTPUT="/var/log/cluster_autoscaler.log"
+fi
+
+./cluster-autoscaler $@ 1>>$LOG_OUTPUT 2>&1 &
+pid="$!"
+trap "kill -15 $pid" 15
+while kill -0 $pid > /dev/null 2>&1; do
```
mwielgus: Add a comment about -0.

mwielgus: And why do we have a loop here - it is not obvious.

MaciekPytel: done

PR #2436, opened 2017-03-02
[Ansible]fix etcd start twice

xialonglee:
There are two register variables use the same name `etcd_started`, the first one was defined in `etcd-start.yml` and the second one was in `etcd2-start.yml`.
The later var will override the previous var even though the task was skipped, see 

http://docs.ansible.com/ansible/playbooks_variables.html#registered-variables

> Note
If a task fails or is skipped, the variable still is registered with a failure or skipped status, the only way to avoid registering a variable is using tags.

In not-coreos case, this will cause etcd restart immediately during its first start process. 
It is not harmless, it will occationally cause etcd unhealthy and could not start normally again , see https://github.com/kubernetes-incubator/kargo/issues/342

### DIFF
--- a/ansible/roles/etcd/tasks/main.yml
+++ b/ansible/roles/etcd/tasks/main.yml
@@ -15,5 +15,5 @@
 
 - name: Restart etcd
   include: restart.yml
-  when: etcd_modified is not defined or (etcd_modified is defined and etcd_started.changed == false)
+  when: etcd_modified is not defined or etcd_started.changed == false
   tags: restart
--- a/ansible/roles/etcd/tasks/start.yml
+++ b/ansible/roles/etcd/tasks/start.yml
@@ -1,6 +1,2 @@
 ---
-- include: etcd-start.yml
-  when: not is_coreos
-
-- include: etcd2-start.yml
-  when: is_coreos
+- include: "{% if not is_coreos %}etcd-start{% else %}etcd2-start{% endif %}.yml"


xialonglee:
cc @ingvagabund

ansible/roles/etcd/tasks/main.yml
```diff
@@ -15,5 +15,5 @@
 
 - name: Restart etcd
   include: restart.yml
-  when: etcd_modified is not defined or (etcd_modified is defined and etcd_started.changed == false)
+  when: etcd_modified is not defined or (etcd_modified is defined and etcd_started.changed == false and etcd_started2.changed == false)
```
galexrt: Removing the redundant `(etcd_modified is defined and` in this PR would be good, so it's done.

galexrt (changes requested):
:+1: except the small "change" comment I added.

xialonglee:
@galexrt fixed fixed : )

galexrt (approved):

ingvagabund:
I would rather use dynamic includes and get rid of the etcd[0-9]_modifed style of variables.

xialonglee:
fixed it, looks better : ) @ingvagabund

ansible/roles/etcd/tasks/start.yml
```diff
@@ -1,6 +1,2 @@
 ---
-- include: etcd-start.yml
-  when: not is_coreos
-
-- include: etcd2-start.yml
-  when: is_coreos
+- include: "{% if not is_coreos %}etcd-start{% else %}etcd2-start{% endif %}.yml"
```
ingvagabund: Given all the roles need some additional refactoring, this is ok.

ingvagabund:
@redhat-k8s-bot test please

ingvagabund (approved):

ingvagabund:
@redhat-k8s-bot test please

ansible/roles/etcd/tasks/start.yml
xialonglee: You mean refactoring all roles that include tasks like this to dynamic include? or just in 'etcd' role?

ingvagabund: Yeah, all roles. The dynamic include is a recent feature of the ansible. So is `block`. I see a lot of places where the roles can be improved.

xialonglee: I think it should take a little time to do this and testing ...
Should i just refactoring these include statement of the 'etcd' role this pr?

ingvagabund: It is ok. Let's keep the PR as it is.

ingvagabund:
@redhat-k8s-bot test please

ansible/roles/etcd/tasks/start.yml
xialonglee: okay : )

PR #2442, opened 2017-03-03
[Ansible] Use slurp and copy instead of synchronize

galexrt:
Ansible synchronize module is a huge trouble for users with a ssh gateway or just extremely strict sshd config(s), not allowing direct server to server, and only playbook-runner to server connections.
Only the know_tokens file is copied, but for **other** masters more isn't needed.

/cc @xialonglee Do you think this is "okay" to change to only copy the `know_tokens.csv`?

### DIFF
--- a/ansible/roles/kubernetes/tasks/gen_tokens.yml
+++ b/ansible/roles/kubernetes/tasks/gen_tokens.yml
@@ -4,16 +4,19 @@
     src=kube-gen-token.sh
     dest={{ kube_script_dir }}
     mode=u+x
+  run_once: true
 
 - name: Generate tokens for master components
   command: "{{ kube_script_dir }}/kube-gen-token.sh {{ item[0] }}-{{ item[1] }}"
   environment:
     TOKEN_DIR: "{{ kube_token_dir }}"
+  delegate_to: "{{ groups['masters'][0] }}"
+  run_once: true
   with_nested:
     - [ "system:controller_manager", "system:scheduler", "system:kubectl" ]
     - "{{ groups['masters'] }}"
-  register: gentoken
-  changed_when: "'Added' in gentoken.stdout"
+  register: gentoken_master
+  changed_when: "'Added' in gentoken_master.stdout"
   notify:
     - restart daemons
 
@@ -21,10 +24,12 @@
   command: "{{ kube_script_dir }}/kube-gen-token.sh {{ item[0] }}-{{ item[1] }}"
   environment:
     TOKEN_DIR: "{{ kube_token_dir }}"
+  delegate_to: "{{ groups['masters'][0] }}"
+  run_once: true
   with_nested:
     - [ 'system:kubelet', 'system:proxy' ]
     - "{{ groups['nodes'] }}"
-  register: gentoken
-  changed_when: "'Added' in gentoken.stdout"
+  register: gentoken_node
+  changed_when: "'Added' in gentoken_node.stdout"
   notify:
     - restart daemons
--- a/ansible/roles/kubernetes/tasks/secrets.yml
+++ b/ansible/roles/kubernetes/tasks/secrets.yml
@@ -81,15 +81,41 @@
 
 - include: gen_tokens.yml
   when: inventory_hostname == groups['masters'][0]
+  tags:
+    - secrets
+    - configure
 
-- name: Copy tokens to other masters
-  synchronize:
-    src: /etc/kubernetes/tokens/
-    dest: /etc/kubernetes/tokens
+- name: Slurp known_tokens.csv from first master
+  slurp:
+    src: "{{ kube_token_dir }}/known_tokens.csv"
+  register: known_tokens
   delegate_to: "{{ groups['masters'][0] }}"
-  when: inventory_hostname in groups['masters'] and inventory_hostname != groups['masters'][0] 
+  run_once: true
+  tags:
+    - secrets
+    - configure
+
+- name: Copy known_tokens.csv to other masters
+  copy:
+    content: "{{ known_tokens.content|b64decode }}"
+    dest: "{{ kube_token_dir }}/known_tokens.csv"
+  when: inventory_hostname in groups['masters'] and inventory_hostname != groups['masters'][0]
   notify:
     - restart daemons
   tags:
-   - secrets
-   - configure
+    - secrets
+    - configure
+
+- name: Copy token files to other masters
+  copy:
+    content: "{{ item.stdout }}"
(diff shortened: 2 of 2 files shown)


xialonglee:
yeah... i found the problem too and i going to make a PR, you are ahead of me XD
My method is use `local_action` module to sync all the token files to the host which the ansible scripts are running on, and then copy files to other masters from localhost.

Because my modified ansible scripts can also be able to add a k8s node into a existing kubernetes cluster, only copy `known_tokens.csv` may required keeping the master[0] host not changed when adding a new node next time. (I am not pretty sure, maybe it doesn't matter).

Anyway, keep all masters got the same token files ( include these files, e.g. system:scheduler-{{ inventory_name }}.token ) makes you don't consider the sequence of masters and if i want add a node which was deleted from a existing k8s cluster again, it wouldn't create a new token for the node.

I preferred keep all masters got all tokens files from master[0].

galexrt:
@xialonglee The problem is to get a list of all tokens created, as the `kubernetes-addons` role creates tokens too.. after the `master` role has already run.

galexrt:
@xialonglee These changes should do the trick better.
How about now? All tokens except the kubernetes-addons generated are synced.
(With synchronize the kubernetes-addons tokens would also **not** be synced)

ansible/roles/kubernetes/tasks/secrets.yml
```diff
@@ -80,16 +80,39 @@
     - restart daemons
 
 - include: gen_tokens.yml
-  when: inventory_hostname == groups['masters'][0]
```
xialonglee: Why you delete this condition and add the same condition to every task in the `gen_tokens.yml` ?  
I can not see any difference of what you have changed...

xialonglee: i get it, you consider the case that master[0] is the cluster node too.

galexrt: There is/was also a problem that when using "when" for the include that the other masters only got the `gen_tokens` var with "empty" (skipped) values.

ansible/roles/kubernetes/tasks/gen_tokens.yml
```diff
@@ -4,27 +4,35 @@
...
+  when: inventory_hostname == groups['masters'][0]
+  run_once: true
 
 - name: Generate tokens for master components
   command: "{{ kube_script_dir }}/kube-gen-token.sh {{ item[0] }}-{{ item[1] }}"
   environment:
     TOKEN_DIR: "{{ kube_token_dir }}"
+  delegate_to: "{{ groups['masters'][0] }}"
+  run_once: true
+  when: inventory_hostname == groups['masters'][0]
```
xialonglee: Did other masters except master[0] will get an empty ( skipped value ) register variable of  `gentoken_master` and `gentoken_node`?

ansible/roles/kubernetes/tasks/secrets.yml
```diff
@@ -80,16 +80,39 @@
...
-    src: /etc/kubernetes/tokens/
-    dest: /etc/kubernetes/tokens
+- name: Slurp known_tokens.csv from first master
+  slurp:
+    src: "{{ kube_token_dir }}/known_tokens.csv"
+  register: known_tokens
   delegate_to: "{{ groups['masters'][0] }}"
-  when: inventory_hostname in groups['masters'] and inventory_hostname != groups['masters'][0] 
+  run_once: true
+  when: inventory_hostname in groups['masters'] and inventory_hostname == groups['masters'][0]
```
galexrt: The slurp needs to be done on the first master.

ansible/roles/kubernetes/tasks/gen_tokens.yml
galexrt: I'm testing this right now.

ansible/roles/kubernetes/tasks/secrets.yml
xialonglee: You have use `delegate_to` already.
And i guess file `known_tokens.csv` in other masters didn't get the right content.

ansible/roles/kubernetes/tasks/gen_tokens.yml
xialonglee: I think use condition `inventory_hostname in groups['masters']` will work.

galexrt:
@xialonglee I implemented your feedback.

ansible/roles/kubernetes/tasks/secrets.yml
```diff
@@ -81,15 +81,41 @@
 
 - include: gen_tokens.yml
   when: inventory_hostname == groups['masters'][0]
```
xialonglee: Should it be `inventory_hostname == groups['masters']` ?

galexrt: `gen_tokens.yml` should only run on the first master and not any other master.

xialonglee: As you mention previously, does these register variables `gentoken_master` and `gentoken_node` wouldn't got skipped values ?

galexrt: It looked like I got skipped values. But I just ran the playbook and it worked.

xialonglee: Actually, these token files of `system:scheduler-{{ inventory_name }}.token` do nothing. If you check the script of `kube-gen-token.sh`, you will find that these files only use to check weather a token was already created. 
So even with skipped values, it works well. I just think taht it looks not so "perfect", because these files are different from master[0] to others ...

xialonglee (approved):

ansible/roles/kubernetes/tasks/secrets.yml
galexrt: @xialonglee The `know_tokens.csv` file has to same on all masters and the `*.token` files are copied too, so when changing the "first" master it won't regenerate the tokens and make the running cluster unusable because of the new tokens for the components.

galexrt:
@ingvagabund Could you please test this?

ingvagabund:
@redhat-k8s-bot test please

ingvagabund:
@redhat-k8s-bot test please

galexrt:
Rebased on latest changes.

ingvagabund:
@galexrt if there is no conflict, there is no need to rebase even if the github says it needs rebase :)

ingvagabund:
@redhat-k8s-bot test please

galexrt:
@ingvagabund The rebase-habit got the better of me ;)

PR #2445, opened 2017-03-03
Cluster-autoscaler: precheck that the api server link is ok

mwielgus:
The logs from leader election are super vague. An explicit check is needed to let the user know that the connection could not be established.

cc: @jszczepkowski @MaciekPytel


jszczepkowski:
LGTM

PR #2446, opened 2017-03-03
Cluster-Autoscaler: fix ignoring node groups config

MaciekPytel:
This bug makes CA pretty much useless.

@mwielgus please review


mwielgus:
/lgtm
/approve

The bug got in with #2226. In future we should be explicitly requesting proof of manual tests (like log dumps) on bigger changes.

PR #2449, opened 2017-03-04
Pass Namespace information about services to template expansion, such…

kskalski:
… that loadbalancing can be scoped / customized using namespace of a service.


kskalski:
I signed it!

kskalski:
Again, should be signed now

kskalski:
ping

PR #2451, opened 2017-03-06
Cluster-autoscaler: include PodDisruptionBudget in drain - part 1/2

mwielgus:
In part 1 or 2 we skip nodes that have a pod with 0 poddisruptionallowed. Part 2/2 will delete pods using evict.

cc: @jszczepkowski @MaciekPytel @davidopp @fgrzadkowski

### DIFF
--- a/cluster-autoscaler/core/static_autoscaler.go
+++ b/cluster-autoscaler/core/static_autoscaler.go
@@ -82,6 +82,7 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) {
 	allNodeLister := a.AllNodeLister()
 	unschedulablePodLister := a.UnschedulablePodLister()
 	scheduledPodLister := a.ScheduledPodLister()
+	pdbLister := a.PodDisruptionBudgetLister()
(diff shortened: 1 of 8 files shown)


cluster-autoscaler/core/static_autoscaler.go
```diff
@@ -215,6 +216,12 @@ func (a *StaticAutoscaler) RunOnce(currentTime time.Time) {
 	if a.ScaleDownEnabled {
 		unneededStart := time.Now()
 
+		pdbs, err := pdbLister.List()
+		if err != nil {
+			glog.Errorf("Failed to list pod disruption budgets: %v", err)
+			return
```
jszczepkowski: Add event for this?

jszczepkowski:
Just one comment, then LGTM

cluster-autoscaler/core/static_autoscaler.go
mwielgus: Next PR - we should have events around other listeners as well.

PR #2455, opened 2017-03-08
Cluster-Autoscaler: Update timestamps in status configmap

MaciekPytel:
Update LastProbeTime and LastTransitionTime fields in ClusterStateRegistry (previously they weren't used and always showed as epoch in status). Update scale down part of status whenever list of unneeded nodes in CA changes.


mwielgus:
Add the PR description. Otherwise seems OK.

mwielgus:
LGTM

PR #2457, opened 2017-03-09
fluent-plugin-systemd to 0.0.8 & cleanup libc6-dev

timstclair:
Diff: https://github.com/timstclair/kubernetes/commit/e0b766b63b4ebafb3df8acd2cbc4d36abfe81b1e

Changes:

1. The fluent-plugin-systemd regression was fixed in 0.0.8 (https://github.com/reevoo/fluent-plugin-systemd/issues/32)
2. `libc-dev` is an alias for `libc6-dev`, which works on install but not purge for some reason, so a couple build dependencies were being left behind. Changing to `libc6-dev` solves this.
3. All other gem & package changes are just updates since the last image was built.

/cc @piosz

### DIFF
--- a/fluentd/fluentd-gcp-image/Dockerfile
+++ b/fluentd/fluentd-gcp-image/Dockerfile
@@ -28,7 +28,7 @@ COPY Gemfile /Gemfile
 # 2. Install fluentd via ruby.
 # 3. Remove build dependencies.
 # 4. Cleanup leftover caches & files.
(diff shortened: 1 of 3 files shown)


crassirostris:
@timstclair Thanks a lot!

timstclair:
Pushed gcr.io/google-containers/fluentd-gcp:2.0.2

piosz:
cc @igorpeshansky

PR #2458, opened 2017-03-09
Fixup ca-cert, private key needs to be kept

### LINKED ISSUE #651: Accessing Kubernetes UI via browser, Getting Unauthorized
[Stackoverflow Issue](http://stackoverflow.com/questions/36270602/how-to-access-kubernetes-ui-via-browser)

### LINKED ISSUE #2394: Cannot have a valid client certificate
As in #651 I will explain my problem here with more details.

Installed on CentOS 7, kubernetes is now working. I want to access UI from my computer and give access to other peaple.

I tried this:

```bash
# on my computer
openssl genrsa -out client.key 1024
openssl req -key client.key -new -out client.csr

# Passing my key and csr to kubernetes server, I now
# create a pem file
openssl x509 -req -in client.csr \
    -CA /etc/kubernetes/certs/server.crt \
    -CAkey /etc/kubernetes/certs/server.key \
    -CAcreateserial \
    -out client.pem \
    -days 3650
```

It's OK, I now get back my client.pem file and create a pkcs12 file on my computer:

```bash
openssl pkcs12 -export -out client.pfx -in client.pem -inkey client.key 
```

I finally import that pfx file into Chrome, the given certificate is prompted while I'm connecting to my UI. But I always get "Unauthorized" response.

I tried different options, when I generating pkcs12 file I tried to add "-certfile"  with ca.crt, server.crt, and so on...

Is anyone here can tell me if there is a solution, or if I do something wrong.

metal3d:
To be able to create user certificates, we need to have ca.key on one server (masters).

Fixes #2394
Fixes #651

### DIFF
--- a/ansible/inventory/group_vars/all.yml
+++ b/ansible/inventory/group_vars/all.yml
@@ -107,6 +107,14 @@ dns_setup: true
 # How many replicas in the Replication Controller
 dns_replicas: 1
 
+# Certificate authority private key should not be kept on server
+# but you probably want to keep it to generate user certificates. Set
+# that value to "true" to keep ca.key file on {{ kube_cert_dir}}.
+# It's recommanded to remove the private key from the server. So if you set
+# kube_cert_keep_ca to true, please copy the ca.key file somewhere that
+# is secured, and remove it from server.
+kube_cert_keep_ca: false
+
 # There are other variable in roles/kubernetes/defaults/main.yml but changing
 # them comes with a much higher risk to your cluster. So proceed over there
 # with caution.
--- a/ansible/roles/kubernetes/files/make-ca-cert.sh
+++ b/ansible/roles/kubernetes/files/make-ca-cert.sh
@@ -27,6 +27,7 @@ set -o pipefail
 # MASTERS - DNS name for the masters
 # DNS_DOMAIN - which will be passed to minions in --cluster-domain
 # SERVICE_CLUSTER_IP_RANGE - where all service IPs are allocated
+# KUBE_CERT_KEEP_CA - to keep ca.key file or not, "true"|"false"
 
 # Also the following will be respected
 # CERT_DIR - where to place the finished certs
@@ -38,6 +39,7 @@ service_range="${SERVICE_CLUSTER_IP_RANGE:="10.0.0.0/16"}"
 dns_domain="${DNS_DOMAIN:="cluster.local"}"
 cert_dir="${CERT_DIR:-"/srv/kubernetes"}"
 cert_group="${CERT_GROUP:="kube-cert"}"
+keep_ca_priv_key="${KUBE_CERT_KEEP_CA:="false"}"
 
 # The following certificate pairs are created:
 #
@@ -142,6 +144,9 @@ cp -p pki/issued/kubelet.crt "${cert_dir}/kubelet.crt"
 cp -p pki/private/kubelet.key "${cert_dir}/kubelet.key"
 
 CERTS=("ca.crt" "server.key" "server.crt" "kubelet.key" "kubelet.crt" "kubecfg.key" "kubecfg.crt")
+if [[ "${keep_ca_priv_key}" == "true" ]]; then
+    CERTS+=("ca.key")
+fi
 for cert in "${CERTS[@]}"; do
   chgrp "${cert_group}" "${cert_dir}/${cert}"
   chmod 660 "${cert_dir}/${cert}"
--- a/ansible/roles/kubernetes/tasks/gen_certs.yml
+++ b/ansible/roles/kubernetes/tasks/gen_certs.yml
@@ -38,6 +38,7 @@
     HTTPS_PROXY: "{{ https_proxy|default('') }}"
     CLUSTER_HOSTNAME: "{{ master_cluster_hostname|default('') }}"
     CLUSTER_PUBLIC_HOSTNAME: "{{ master_cluster_public_hostname|default('') }}"
+    KUBE_CERT_KEEP_CA: "{{ kube_cert_keep_ca | default('false') | lower }}"
 
 - name: Verify certificate permissions
   file:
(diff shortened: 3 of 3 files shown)


metal3d:
@eparis it seems I need you to approve this PR, or maybe I didn't undestand the bot message.

eparis:
I think we've historically not retained the .key for security reasons. I understand why one might want to keep it somewhere to do things by hand later...

ingvagabund:
@metal3d can you make it configurable and disabled by default?

metal3d:
The main problem is that we cannot authenticate to api, or dashboard, without to have that CA.key to provide user certs. See #2394 where we've discussed that problem.

At this time, we need to change manually the make-ca-cert script to keep ca.key, so we need to change the repo source (and that's not that easy to find where the ca.key were removed). If we leave that contrib/ansible in that state, all users will complain to not be able to authenticate or to create user certs.

Maybe we can:

- document this
- let ca.key on master and write an alert message after installtion
- add a configutation variable in group_vars to activate ca.key
- add a configuration variable to write ca.key on localhost...

I can modify my PR as needed.

metal3d:
I just seen @ingvagabund answer :) I can make an option, right !

I do it quickly

ansible/inventory/group_vars/all.yml
```diff
@@ -107,6 +107,15 @@ dns_setup: true
 # How many replicas in the Replication Controller
 dns_replicas: 1
 
+# Certificate authority private key should not be kept on server
+# but you probably want to keep it to generate user certificates. Set
+# that value to "true" to keep ca.key file on {{ kube_cert_dir}}.
+# It's recommanded to remove the private key from the server. So if you set
+# ca_keep_private_key to true, please copy the ca.key file somewhere that
+# is secured, and remove it from server.
+ca_keep_private_key: false
```
galexrt: I think the name should be more like `kube_cert_keep_ca`.

ansible/roles/kubernetes/tasks/gen_certs.yml
```diff
@@ -51,3 +51,17 @@
...
+
+- name: Remove CA private key if ca_keep_private_key is false
+  file:
+    path={{ kube_cert_dir}}/ca.key
+    state=absent
+  when: not ca_keep_private_key
+
+- name: Check CA private key permissions if ca_keep_private_key is True
+  file:
+    path={{ kube_cert_dir }}/ca.key
```
galexrt: You should use `:` instead of `=`.
That is currently not 100% enforced in all of the files here in contrib/ansible but from my point of view should be as there should be only **one** style for args.

ansible/roles/kubernetes/files/make-ca-cert.sh
```diff
@@ -134,6 +134,7 @@ fi
 mkdir -p "$cert_dir"
 
 cp -p pki/ca.crt "${cert_dir}/ca.crt"
+cp -p pki/private/ca.key "${cert_dir}/ca.key"
```
galexrt: Why not add a condition here and use an environment variable for this in the Ansible task?

galexrt (changes requested):
I have added some comments for changes I think should be added.

metal3d:
That's à fois Idea. I just didn't want to break something. I will make the
changes (comments and script env)

Le jeu. 9 mars 2017 17:39, Alexander Trost <notifications@github.com> a
écrit :

ansible/roles/kubernetes/tasks/gen_certs.yml
metal3d: I just followed same syntax as above. I agree (I prefer `:` instead of `=`)

ansible/inventory/group_vars/all.yml
metal3d: no problem, I'll do it

metal3d:
I just fixed following your comments. I tried both installation and this is ok here.

I precise that @galexrt comment asking to use a parameter fixed a problem I didn't seen before. Without that, I kept etcd ca.key file. Now, because I only set argument to the kube cert generation, the ca.key file for etcd isn't kept.

I've also seen that kubelet.cert and key hasn't the "kube" user (root instead) - I will make another PR as soon as this one is acceptable.

ansible/roles/kubernetes/files/make-ca-cert.sh
```diff
@@ -146,3 +148,9 @@ for cert in "${CERTS[@]}"; do
   chgrp "${cert_group}" "${cert_dir}/${cert}"
   chmod 660 "${cert_dir}/${cert}"
 done
+
+if [[ "${keep_ca_priv_key}" == "true" ]]; then
```
galexrt: To make it look/"feel" cleaner:
```
CERTS=("ca.crt" "server.key" "server.crt" "kubelet.key" "kubelet.crt" "kubecfg.key" "kubecfg.crt")
if [[ "${keep_ca_priv_key}" == "true" ]]; then
    CERTS+=("ca.key")
fi
for [...]
```
(Remove the below added conditional stuff)

galexrt (changes requested):
This one more change and it looks good to me.
One small thing, I think it would be nice if you would squash your commits into one using rebase.

ansible/roles/kubernetes/files/make-ca-cert.sh
metal3d: wow nice idea ! I do it in a couple of minutes. Thanks !

galexrt:
@metal3d About the etcd ca key not being preserved. Is it really required to keep the ca key? Do you want to create new client certs for etcd? As the one that gets created is enough for authenticating any client.

metal3d:
@galexrt no, I meant that my first PR kept ca.key for etcd and that was not wanted. Using a command argument to make-ca-cert.sh script fixed the problem because now, the ca.key file is generated only when the argument is TRUE, and that only the case when we create certificates for kubernetes.

galexrt (approved):

ingvagabund:
@redhat-k8s-bot test please

ingvagabund:
@redhat-k8s-bot test please

ingvagabund:
@redhat-k8s-bot test please

ingvagabund:
@redhat-k8s-bot test please

metal3d:
I'm glade to help. Thanks for acceptance :)

ingvagabund:
Thanks for improving the ansible :)

PR #2464, opened 2017-03-10
Cluster-Autoscaler: evict pods instead of deleting them

MaciekPytel:
This should make CA respect PodDisruptionBudget.

### DIFF
--- a/cluster-autoscaler/core/scale_down.go
+++ b/cluster-autoscaler/core/scale_down.go
@@ -58,6 +58,10 @@ const (
 	MaxKubernetesEmptyNodeDeletionTime = 3 * time.Minute
 	// MaxCloudProviderNodeDeletionTime is the maximum time needed by cloud provider to delete a node.
 	MaxCloudProviderNodeDeletionTime = 5 * time.Minute
+	// MaxPodEvictionTime is the maximum time CA tries to evict a pod before giving up.
+	MaxPodEvictionTime = 2 * time.Minute
+	// EvictionRetryTime is the time after CA retries failed pod eviction.
+	EvictionRetryTime = 10 * time.Second
 )
 
 // ScaleDown is responsible for maintaining the state needed to perform unneded node removals.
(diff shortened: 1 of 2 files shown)


cluster-autoscaler/core/scale_down.go
```diff
@@ -342,35 +346,85 @@ func getEmptyNodes(candidates []*apiv1.Node, pods []*apiv1.Pod, maxEmptyBulkDele
...
 	return deleteNodeFromCloudProvider(node, context.CloudProvider, context.Recorder, context.ClusterStateRegistry)
 }
 
+func evictPod(podToEvict *apiv1.Pod, client kube_client.Interface, recorder kube_record.EventRecorder,
+	maxGratefulTerminationSec int, retryUntil time.Time, waitBetweenRetries time.Duration) error {
+	recorder.Eventf(podToEvict, apiv1.EventTypeNormal, "ScaleDown", "deleting pod for node scale down")
+	maxGraceful64 := int64(maxGratefulTerminationSec)
+	var lastError error
+	first := true
+	for first || time.Now().Before(retryUntil) {
```
mwielgus: you can combine, first, check and sleep in this single line.

cluster-autoscaler/core/scale_down.go
```diff
@@ -342,35 +346,85 @@ func getEmptyNodes(candidates []*apiv1.Node, pods []*apiv1.Pod, maxEmptyBulkDele
...
-		})
+		go func(podToEvict *apiv1.Pod) {
+			confirmations <- evictPod(podToEvict, client, recorder, maxGratefulTerminationSec, retryUntil, waitBetweenRetries)
+		}(pod)
+	}
+
+	evictionErrs := make([]error, 0)
+
+	for range pods {
+		err := <-confirmations
```
mwielgus: In theory this can hung indefinitely. You try to retry until maxPodEvictionTime) but it may happen that the send takes forever. 

select {
    case <- time.After(retryUntil.Sub(time.Now) + 5 * time.sec): 
       glog.Errorf("WARNING - SOMETHING STUCKED  UP - should have stopped all pod evictions long time ago) 
       evictionErrors = append(evictionErrors, something stucked up)
   case err := <- contirmations
     ....
}

cluster-autoscaler/core/scale_down.go
```diff
@@ -387,13 +441,11 @@ func drainNode(node *apiv1.Node, pods []*apiv1.Pod, client kube_client.Interface
 		}
 		if allGone {
 			glog.V(1).Infof("All pods removed from %s", node.Name)
-			break
+			drainSuccessfull = true
```
mwielgus: do we need to set this?

MaciekPytel: This is used in defered closure - I prefer to set it here once if we succeeded, instead of caring about cleanup in every place where we may fail.

I can add a comment here if you think it will be more readable that way.

mwielgus: yes please add a comment.

cluster-autoscaler/core/scale_down.go
MaciekPytel: done

cluster-autoscaler/core/scale_down.go
MaciekPytel: done

cluster-autoscaler/core/scale_down.go
MaciekPytel: added comment

davidopp:
ref/ kubernetes/kubernetes#25321

PR #2490, opened 2017-03-23
Return error if kube client can't be initialized

redbaron:


piosz:
@redbaron thanks for the fix

piosz:
please fix
```
# k8s.io/contrib/rescheduler
./rescheduler.go:205: not enough arguments to return
godep: go exit status 2
```

PR #2496, opened 2017-03-24
Cluster-autoscaler: add information about which version is supported in which k8s

mwielgus:
cc: @andrewsykim @MaciekPytel @fgrzadkowski


jszczepkowski:
LGTM

PR #2508, opened 2017-03-29
Configurable keepalived virtual router id

cdemers:
Hi all, this PR is an up to date version of this previous PR  #2384, it's the same but squashed and merged with the latest master revision.

It adds support for configurable keepalived virtual router id. This functionality is necessary in order to be able to run multiple keepalived "clusters" in the same subnet without having them stepping on each other's toes (having one node from one group confusing the other group into thinking it has a master).

The configuration of the virtual router id is done using the --vrid command line flag. I considered using the project's configmap but the structure of the existing configmap is not easily expandable, so I had to use a flag (I would have added support for environment variables, but it didn't seems to be in line with the way this application is thought out, so I refrained).

The flag can be specified in a k8s object with the args key, for example:
- args: ["--services-configmap=admin/vip-configmap", "--vrid=45"]

If not specified, it will default to router id 50 as it was before.

That's pretty much all there is to it, thanks in advance for your time!

### DIFF
--- a/hack/verify-flags/known-flags.txt
+++ b/hack/verify-flags/known-flags.txt
@@ -162,6 +162,7 @@ use-reviewers
 use-unicast
 user-whitelist
 verify-unschedulable-pods
+vrid
 vrrp-password
 watch-namespace
 weak-stable-jobs
--- a/keepalived-vip/controller.go
+++ b/keepalived-vip/controller.go
@@ -284,7 +284,7 @@ func (ipvsc *ipvsControllerController) Stop() error {
 }
 
 // newIPVSController creates a new controller from the given config.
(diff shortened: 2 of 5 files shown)


cdemers:
@bowei My apologies for closing the previous PR 2384.

cdemers:
Hi Eric (@eparis), I'm not sure I follow this process properly, but would it be that we still need your approval as an approver of the hack/OWNERS file? Sorry for my beginners questions.

eparis:
The comment from the bot seems reasonably clear that you are waiting on @bprashanth for /approve. Yes I am capable of (and do) approving the hack change I do not know the the keepalive stuff at all and am not qualified to approve that portion. Since I have no way to approve just one part I will not approve this BZ and we will have to wait on @bprashanth, like the bot suggested.

cdemers:
Thanks Eric! :)

Issue #2408, opened 2017-02-23
Why keep kube-ui in Ansible roles if it is deprecated?


harobed:
Why keep kube-ui in [Ansible roles](https://github.com/kubernetes/contrib/blob/master/ansible/roles/kubernetes-addons/tasks/kube-ui.yml) if it is [deprecated](https://github.com/kubernetes/kube-ui) ?

If it is deprecated I suggest to delete it.

ingvagabund:
If you create a PR, I will review it.

PR #2510, opened 2017-03-30
[Ansible] removed kubeui

alxndr13:
remove kubeui as proposed in #2408


alxndr13:
@ingvagabund as proposed in #2408

PR #2514, opened 2017-03-31
[Ansible] Add missing default var for gen_certs keep ca var

galexrt:
This adds the missing default var to the when condition for `kube_cert_keep_ca` var.


galexrt:
@eparis and @ingvagabund can we get this merged? This isn't a major change that breaks anything.

PR #2532, opened 2017-04-06
Vertical Pod Autoscaler : updater first version

GFilipek:
Updater component for Vertical Pod Autoscaler described in  https://github.com/kubernetes/community/pull/338

Runs in loop. On one iteration performs:
 - Fetching Vertical Pod Autoscaler configuration 
 - Fetching Pods information and resource allocation recommendations for pods. 
 - Calculating if pod update is required and how many replicas can be evicted
 - Evicting pods if recommended resources significantly vary from actual resources allocation.

Mocked parts - to be implemented:
 - Recommendation API for fetching data from Vertical Pod Autoscaler Recommender.
 - Vertical Pod Autoscaler lister for fetching Vertical Pod Autoscaler config.

cc: @mwielgus @kgrygiel @KarolKraskiewicz


bgrant0607:
@GFilipek 

Please don't add something this big to contrib. See #762.

We should create a repo in kubernetes-incubator instead.

The process is described here:

https://github.com/kubernetes/community/blob/master/incubator.md

PR #2539, opened 2017-04-10
Cluster-Autoscaler FAQ entry describing events

MaciekPytel:
While writing the FAQ I noticed one event is named inconsistently with the rest, so I updated the name. It's a very minor change so I included it in this PR, but I can move it to a separate one if you think it makes more sense @mwielgus


MaciekPytel:
@mwielgus this includes https://github.com/kubernetes/contrib/pull/2537 because I really don't want to merge those two later on. So it should be reviewed after #2537.

PR #2441, opened 2017-03-03
[Ansible] Added *_options var to apiserver, controller-manager, scheduler, kubelet (and flannel)

galexrt:
Changes:
* Added *_options var to apiserver, controller-manager, scheduler, kubelet and proxy
* Renamed the flannel_opts to be like the others, flannel_options
* Added the defaults for *_options to the defaults file and not the template

PR #2546, opened 2017-04-13
[ansible]fix var name

xialonglee:
Before #2441 , it called `apiserver_extra_args`, so after refactoring, it should called `kube_apiserver_additional_options` for consistency.

Using `kube_apiserver_options` will override the same named var defined in `roles/master/defaults/main.yml` .
cc @galexrt , I touch your PR.


galexrt (approved):
LGTM

PR #2548, opened 2017-04-13
fix openssl vulnerability

naarani:
the image is based on previous versions, I don't think this should be done in the future or it will add too many layers and all the packages should be updated manually to prevent bugs and vulnerability


thelinuxfoundation:
bump for CLA check

Issue #2181, opened 2016-12-20
cluster-autoscaler: Dynamic reconfiguration via ConfigMap


mumoshu:
> **Is this a request for help?** (If yes, you should use our troubleshooting guide and community support channels, see http://kubernetes.io/docs/troubleshooting/.):

No.

> **What keywords did you search in Kubernetes issues before filing this one?** (If you have found any duplicates, you should instead reply there.):

I've searched for:

* cluster-autoscaler
* dynamic reconfiguration
* configmap
* config map
* cluster autoscaler

and their combinations but found nothing.

---

> **Is this a BUG REPORT or FEATURE REQUEST?** (choose one):

FEATURE REQUEST

> If this is a FEATURE REQUEST, please:
>  - Describe *in detail* the feature/behavior/change you'd like to see.

As a maintainer of [kube-aws](https://github.com/coreos/kube-aws), I'm planning to add out-of-box support for/integration with cluster-autoscaler into [kube-aws](https://github.com/coreos/kube-aws) and got to want a feature to reconfigure cluster-autoscaler without deleting it and then creating it with new `--nodes` parameters, hence "dynamic reconfiguration".

cc @andrewsykim @mwielgus 

## My use-case

My use-case is to reconfigure cluster-autoscaler automatically when a kube-aws user added/removed node pool(s) associated to a k8s cluster.

More complete context can be seen in [an corresponding issue](https://github.com/coreos/kube-aws/issues/148) of kube-aws.

### What could be done WITHOUT the requested feature

Currently, an user has to delete and then create cluster-autoscaler completely out of band from a k8s cluster created by deployment tools like kops and kube-aws, to reconfigure cluster-autoscaler to:

* stop auto-scaling on already removed node pool(s)
* start auto-scaling on newly created node pool(s)

With cluster-autoscaler as of today, we can achieve dynamic reconfiguration via writing e.g.:

* a shell script to `inotify-wait` on a configmap volume,
  * read the latest configuration from it,
  * and then updating cluster-autoscaler by restarting it while updating `--nodes` parameters,
     * via `kubectl` or inside the container running cluster-autoscaler.

Though it is technically possible, doing so in that way seems to involve more moving parts than necessary:
* a correct init process(a shell script as pid=0 may result in zombie processes)
* process supervisor(to keep cluster-autoscaler up and running)

and makes testing harder.

### What could be done WITH the requested feature

[...]

PR #2549, opened 2017-04-14
cluster-autoscaler: Re: AWS Autoscaler autodiscover ASG names and sizes

mumoshu:
@mwielgus @Raffo @andrewsykim @osxi Would you mind reviewing this?

---

This is an alternative implementation of #1982

Notable differences from the original PR are:

* A new flag named `--node-group-auto-discovery` is introduced for opting in to enable the auto-discovery feature.
  * For example, specifying `--cloud-provider aws --node-group-auto-discovery asg:tag=k8s.io/cluster-autoscaler/enabled` instructs CA to auto-discover ASGs tagged with `k8s.io/cluster-autoscaler/enabled` to be used as target node groups
* The new code path introduced by this PR is executed only when `node-group-auto-discovery` is specified. There is relatively less chance to break existing features by this change

Other notes:

* ~~I thought it might be a good idea to implement this feature on top of the dynamic reconfiguration #2181 initially~~
  * ~~However it turned out it is easier and less verbose to just implement an alternative aws cloud provider like what I've done for this PR~~
* We rely mainly on the `DescribeTags` API rather than `DescribeAutoScalingGroups` so that AWS can filter out unnecessary ASGs which doesn't belong to the k8s cluster, for us.
  * If we relied on `DescribeAutoScalingGroups` here, as it doesn't support `Filter`ing, we'd need to iterate over ALL the ASGs available in an AWS account, which isn't desirable due to unnecessary excessive API calls and network usages

Possible future improvements before recommending this to everyone:

* Cache the result of an auto-discovery for a configurable period, so that we won't invoke DescribeTags and DescribeAutoScalingGroup APIs too many times

### DIFF
--- a/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
+++ b/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
@@ -21,20 +21,74 @@ import (
 	"regexp"
 	"strings"
 
+	"errors"
 	"k8s.io/contrib/cluster-autoscaler/cloudprovider"
 	"k8s.io/contrib/cluster-autoscaler/config/dynamic"
 	apiv1 "k8s.io/kubernetes/pkg/api/v1"
 )
 
-// AwsCloudProvider implements CloudProvider interface.
-type AwsCloudProvider struct {
+// awsCloudProvider implements CloudProvider interface.
+type awsCloudProvider struct {
 	awsManager *AwsManager
 	asgs       []*Asg
 }
 
+// autoDiscoveringProvider implements CloudProvider interface.
+type autoDiscoveringProvider struct {
+	*awsCloudProvider
+}
+
 // BuildAwsCloudProvider builds CloudProvider implementation for AWS.
-func BuildAwsCloudProvider(awsManager *AwsManager, specs []string) (*AwsCloudProvider, error) {
-	aws := &AwsCloudProvider{
+func BuildAwsCloudProvider(awsManager *AwsManager, discoveryOpts cloudprovider.NodeGroupDiscoveryOptions) (cloudprovider.CloudProvider, error) {
+	if len(discoveryOpts.NodeGroupSpecs) > 0 {
+		return buildStaticallyDiscoveringProvider(awsManager, discoveryOpts.NodeGroupSpecs)
+	}
+	if discoveryOpts.NodeGroupAutoDiscoverySpec != "" {
+		return buildAutoDiscoveringProvider(awsManager, discoveryOpts.NodeGroupAutoDiscoverySpec)
+	}
+	return nil, errors.New("Failed to build an aws cloud provider: Either node group specs or node group auto discovery spec must be specified")
+}
+
+func buildAutoDiscoveringProvider(awsManager *AwsManager, spec string) (*autoDiscoveringProvider, error) {
+	tokens := strings.Split(spec, ":")
+	if len(tokens) != 2 {
+		return nil, fmt.Errorf("Invalid node group auto discovery spec specified: %s", spec)
+	}
+	discoverer := tokens[0]
+	if discoverer != "asg" {
+		return nil, fmt.Errorf("Unsupported discoverer specified: %s", discoverer)
+	}
+	param := tokens[1]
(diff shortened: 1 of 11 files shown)


cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
```diff
@@ -21,20 +21,74 @@ import (
...
-func BuildAwsCloudProvider(awsManager *AwsManager, specs []string) (*AwsCloudProvider, error) {
-	aws := &AwsCloudProvider{
+func BuildAwsCloudProvider(awsManager *AwsManager, discoveryOpts cloudprovider.NodeGroupDiscoveryOptions) (cloudprovider.CloudProvider, error) {
+	if len(discoveryOpts.NodeGroupSpecs) > 0 {
+		return buildStaticallyDiscoveringProvider(awsManager, discoveryOpts.NodeGroupSpecs)
+	}
+	if discoveryOpts.NodeGroupAutoDiscoverySpec != "" {
+		return buildAutoDiscoveringProvider(awsManager, discoveryOpts.NodeGroupAutoDiscoverySpec)
+	}
+	return nil, errors.New("failed to build an aws cloud provider: either --nodes or --node-group-auto-discovery must be specified")
```
Raffo: I would not put the flags here as this part of code is not really aware of how the flags are defined.

cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
```diff
@@ -21,20 +21,74 @@ import (
...
+}
+
+func buildAutoDiscoveringProvider(awsManager *AwsManager, spec string) (*autoDiscoveringProvider, error) {
+	tokens := strings.Split(spec, ":")
+	if len(tokens) != 2 {
+		return nil, fmt.Errorf("invalid value for node-group-auto-discovery flag: %s", spec)
+	}
+	discoverer := tokens[0]
+	if discoverer != "asg" {
+		return nil, fmt.Errorf("unsupported discoverer specified via --node-group-auto-discovery flag: %s", discoverer)
```
Raffo: Same as above

cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
```diff
@@ -21,20 +21,74 @@ import (
...
+	}
+	hint := tokens[1]
```
Raffo: "hint" doesn't sound like a good name for this variable.

Raffo (commented):
Hey, thanks for jumping on this, as you probably imagined, I haven't found the time to completely refactor my original PR. 
I left you a couple of minor comments, but this looks very similar to what we discussed before, so looks mostly okay for me from a quick scan. 
A couple of other comments:

Shouldn't you also add a test for making sure that the polling functionality works?

cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
mumoshu: Thanks! Fixed the message not to include the flag name.

cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
mumoshu: Agreed. Fixed the message not to include the flag name.

cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go
mumoshu: Certainly. After rethinking a bit, I changed it to "param"/"parameter" which seemed nicer.

mumoshu:
@Raffo Thanks for reviewing 👍 
Excuse me for mentioning you with suddenly submitting this PR.
I just meant to help moving things forward by writing a working example of what I've suggested myself - I'm glad if this helped.

> Shouldn't you also add a test for making sure that the polling functionality works?

Definitely! I've just added `polling_autoscaler_test.go` accordingly.

mumoshu:
Closing this in favor of https://github.com/kubernetes/autoscaler/pull/11

PR #2550, opened 2017-04-14
Cluster-autoscaler: add default credentials to faq.

mwielgus:
cc: @MaciekPytel @fgrzadkowski


cluster-autoscaler/FAQ.md
```diff
@@ -291,6 +291,10 @@ the scale-up-triggering pods are removed.
     go run hack/e2e.go -- -v --test --test_args="--ginkgo.focus=\[Feature:ClusterSizeAutoscaling"
     ```
     It will take >1 hour to run the full suite. You may want to redirect output to file, as there will be plenty of it.
+    Test runner may be missing default credentials. In such a case invoke:
```
MaciekPytel: Those 2 separate notes render as a single paragraph on github. I think this can be fixed by adding an empty line in between.

Also do we assume people only run CA e2e on GCE? This note is pretty specific, maybe we should start it with "If you're running tests on GCE the test runner may be missing..."

mwielgus: Done.

PR #2557, opened 2017-04-19
fix typos (re-applying pull/2362)

dluc:


fgrzadkowski:
Cluster autoscaler has been moved to https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler. Please send this PR against that repository.

@mwielgus when we do plan to delete code in contrib?

Issue #2533, opened 2017-04-06
Create a separate repo for cluster autoscaler


fgrzadkowski:
We need a separate repo to a) conform to new guidelines of deprecating contrib and more importantly to b) properly do release management (branches, tags, release notes etc)

@MaciekPytel

mwielgus:
Done with #2558.

PR #2558, opened 2017-04-19
Cluster-Autoscaler: remove all code - the project has moved to https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler

mwielgus:
Ref: #2533

cc: @MaciekPytel @piosz @jszczepkowski @fgrzadkowski @andrewsykim @mumoshu @Raffo


MaciekPytel:
@mwielgus maybe it would be also worth adding a "redirect" FAQ.md? I believe I've seen people send direct links to faq already.

mwielgus:
@MaciekPytel  good point - fixed.

fgrzadkowski:
Is anyone using cluster autoscaler? For example have you updated rescheduler?

mwielgus:
Rescheduler has CA in vendor - it will have to be updated on the earliest convenience. 
The whole contrib compiled so I guess noone used CA directly.

fgrzadkowski:
Great. Thanks for checking.

--
Filip

PR #2575, opened 2017-05-02
Fix reading metrics with empty whitelisted-metrics flag

### LINKED ISSUE #2574: [prometheus-to-sd] Service does not send any metrics without whitelisted-metrics
In description to this param:
`Comma-separated list of whitelisted metrics. If empty all metrics will be exported.` 
But it does not send any metrics because of this line:
https://github.com/kubernetes/contrib/blob/master/prometheus-to-sd/main.go#L82
the result array will always have at least one element.

gaplyk:
this pr fixes issue when you do launch
```
./monitor
```
without param --whitelisted-metrics

Expected behavior: monitor service will send all metrics.
Actual: Error message - No metrics to send to Stackdriver

### DIFF
--- a/prometheus-to-sd/main.go
+++ b/prometheus-to-sd/main.go
@@ -71,6 +71,11 @@ func main() {
 	}
 	glog.V(4).Infof("Successfully created Stackdriver client")
 
+	var whitelistedList []string
+	if *whitelisted != "" {
+		whitelistedList = strings.Split(*whitelisted, ",")
+	}
+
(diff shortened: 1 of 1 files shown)


piosz:
@gaplyk thanks for the fix!

piosz:
Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.kubernetes.io:443/reviews/kubernetes/contrib/2575)*

PR #2576, opened 2017-05-02
Metadata proxy

Q-Lee:
Adding a container to proxy the gce metadata service.

### DIFF
--- a/metadata-proxy/Dockerfile
+++ b/metadata-proxy/Dockerfile
@@ -1,4 +1,4 @@
-# Copyright 2016 The Kubernetes Authors All rights reserved.
+# Copyright 2017 The Kubernetes Authors.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -12,16 +12,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-all: push
+FROM gcr.io/google-containers/debian-base-amd64:0.1
+LABEL maintainer "qlee@google.com"
 
-TAG = 0.2
-PREFIX = gcr.io/google_containers/redis-install
+ARG DEBIAN_FRONTEND=noninteractive
+RUN apt-get update && apt-get install -y \
+  apt-utils && clean-install iptables nginx
 
-container:
-	docker build --pull -t $(PREFIX):$(TAG) .
+# Give nginx its config file.
+COPY nginx.conf /etc/nginx/nginx.conf
+# Place our wrapper script into the image.
+COPY start-proxy.sh /
 
-push: container
-	gcloud docker -- push $(PREFIX):$(TAG)
-
-clean:
-	docker rmi $(PREFIX):$(TAG)
+ENTRYPOINT ["/start-proxy.sh"]
--- a/metadata-proxy/Makefile
+++ b/metadata-proxy/Makefile
@@ -1,4 +1,4 @@
-# Copyright 2016 The Kubernetes Authors All rights reserved.
+# Copyright 2017 The Kubernetes Authors.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -12,16 +12,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-all: push
+.PHONY:	build push
 
-TAG = 0.2
-PREFIX = gcr.io/google_containers/redis-install
+# TAG is the version to build and push to.
+PREFIX = gcr.io/google_containers
+TAG = 0.1
 
-container:
-	docker build --pull -t $(PREFIX):$(TAG) .
+build:
+	# We explicitly add "--pull" flag to always fetch the latest version
+	# of the base image. This is necessary to avoid using cached local
+	# versions of image e.g. when updating insecure base images.
+	docker build --pull -t ${PREFIX}/metadata-proxy:$(TAG) .
 
-push: container
-	gcloud docker -- push $(PREFIX):$(TAG)
-
-clean:
-	docker rmi $(PREFIX):$(TAG)
+push: build
+	gcloud docker -- push ${PREFIX}/metadata-proxy:$(TAG)
--- a/metadata-proxy/nginx.conf
+++ b/metadata-proxy/nginx.conf
@@ -1,0 +1,63 @@
+# Copyright 2017 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+user www-data;
+worker_processes 4;
+pid /run/nginx.pid;
+
+events {
+  worker_connections 20;
+}
+
+http {
+  server {
+    listen 127.0.0.1:988;
+
+    # By default, return 403. This protects us from new API versions.
+    location / {
+        return 403;
+    }
+
+    # Allow for REST discovery.
+    location = / {
+        proxy_pass http://169.254.169.254;
+    }
+    location = /computeMetadata/ {
+        proxy_pass http://169.254.169.254;
+    }
+
+    # By default, allow the v0.1, v1beta1, and v1 APIs.
+    location /0.1/ {
+        proxy_pass http://169.254.169.254;
+    }
+    location /computeMetadata/v1beta1/ {
+        proxy_pass http://169.254.169.254;
+    }
+    location /computeMetadata/v1/ {
+        proxy_pass http://169.254.169.254;
+    }
+
+    # Return a 403 for the kube-env attribute in all allowed API versions.
+    location /0.1/meta-data/attributes/kube-env {
+        return 403;
+    }
+    location /computeMetadata/v1beta1/instance/attributes/kube-env {
+        return 403;
+    }
+    location /computeMetadata/v1/instance/attributes/kube-env {
+        return 403;
+    }
+
+  }
+}
--- a/metadata-proxy/start-proxy.sh
+++ b/metadata-proxy/start-proxy.sh
@@ -1,0 +1,31 @@
+#!/bin/dash
+
+# Copyright 2017 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+_term() {
+  iptables -D -t filter -I KUBE-METADATA-SERVER -j ACCEPT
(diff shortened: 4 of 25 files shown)


metadata-proxy/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM gcr.io/google_containers/nginx
```
timstclair: Please don't use this image. I can't tell what it's sources are, and it hasn't been updated in 2 years. This should use either gcr.io/google-containers/debian-base-amd64 or gcr.io/google-containers/debian-iptables

metadata-proxy/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+MAINTAINER Quintin Lee <qlee@google.com>
```
timstclair: nit: maintainer tags are deprecated

metadata-proxy/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+ARG DEBIAN_FRONTEND=noninteractive
+RUN apt-get update && apt-get install -y \
+  iptables \
+  sudo
```
timstclair: I think it's weird to run sudo in a container... it should probably just run as root. If you're trying to avoid running nginx as root, I would just switch to a different user for that command.

metadata-proxy/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+# Give nginx its config file.
+COPY nginx.conf /etc/nginx/nginx.conf
+# Place our wrapper script into the image.
+COPY start-proxy.sh /
+RUN chmod +x /start-proxy.sh
```
timstclair: nit: I think `COPY` preserves permissions, so you could just add this permission to the file checked into git and skip this step.

metadata-proxy/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+CMD ["/start-proxy.sh"]
```
timstclair: nit: prefer `ENTRYPOINT`

metadata-proxy/Makefile
```diff
@@ -0,0 +1,15 @@
+.PHONY:	build push
```
timstclair: I think you might need license headers on each file.

metadata-proxy/Makefile
```diff
@@ -0,0 +1,15 @@
...
+
+# TAG is the version to build and push to.
+#PREFIX = gcr.io/google_containers
```
timstclair: delete this

metadata-proxy/Makefile
```diff
@@ -0,0 +1,15 @@
...
+PREFIX = gcr.io/google-containers
+TAG = 1.0
```
timstclair: nit: do you want to start with a pre-release version? 0.1?

metadata-proxy/nginx.conf
```diff
@@ -0,0 +1,47 @@
...
+http {
+  server {
+    listen 127.0.0.1:2020;
+
+    # By default, proxy everything.
+    location / {
+        proxy_pass http://169.254.169.254;
+    }
+
+    # Return 403 for anything too sensitive for pods. Would a regex be better?
```
timstclair: Can we use a whitelist instead of a blacklist approach? My concern is that, for example, when new API versions (or new APIs) come out, suddenly every node has access.

metadata-proxy/start-proxy.sh
```diff
@@ -0,0 +1,20 @@
...
+/usr/bin/sudo iptables -t nat -I PREROUTING -p tcp -d 169.254.169.254 --dport 80 -j DNAT --to-destination 127.0.0.1:2020
+/usr/bin/sudo iptables -t filter -I KUBE-METADATA-SERVER -j ACCEPT
+
+# Clean up the iptables rule if we're exiting gracefully.
+trap _term SIGTERM
+
+# Run nginx in the foreground.
+nginx -g 'daemon off;'
+child=$!
+wait $child
```
timstclair: I don't think you need this with the `daemon off` setting?

metadata-proxy/nginx.conf
```diff
@@ -0,0 +1,47 @@
...
+worker_processes 4;
+pid /run/nginx.pid;
+
+events {
+  worker_connections 20;
+}
+
+http {
+  server {
+    listen 127.0.0.1:2020;
```
timstclair: how did you pick :2020?

timstclair (changes requested):
Can you add the daemonset manifest too? Even if it's not the final version, I think including the manifest provides a good example of how some of the options work (e.g. I assume this needs host networking)

metadata-proxy/Dockerfile
Q-Lee: Ok, I changed it to be a label.

Q-Lee:
@timstclair - the daemonset will live in the main repo as an addon. I haven't sent it out yet, because I don't want to add anything to "google-containers" until it's been reviewed and checked in.

metadata-proxy/Dockerfile
Q-Lee: Ok, I'm following fluentd's lead and using "gcr.io/google-containers/debian-base-amd64:0.1".

metadata-proxy/Makefile
Q-Lee: Added.

metadata-proxy/Makefile
Q-Lee: Done.

metadata-proxy/nginx.conf
Q-Lee: Nothing was using it, and dynamic allocation would be considerably more difficult. Do you have another proposal?

timstclair: No, just curious. Can you use variables in the nginx configuration? It's not too bad since this is optional, but what if a customer has something running on that port? I would be nice to provide a way to override it.

metadata-proxy/Dockerfile
Q-Lee: Ok, I removed sudo. The nginx master process starts as root, and then forks www-data worker processes.

metadata-proxy/Dockerfile
Q-Lee: Ok

metadata-proxy/Dockerfile
Q-Lee: Ok

metadata-proxy/nginx.conf
Q-Lee: I don't think nginx configs support variables. In fact, I think their developers staunchly oppose that feature. IIRC, they recommend sed.

I could move it to a privileged port, and then only privileged pods w/host networking could conflict, instead of just host networking. I'll also drop this to v0.1, in case we decide the value should be configurable.

metadata-proxy/nginx.conf
Q-Lee: That's fair. I've switched to a whitelist. It makes a few things (e.g., REST discovery) weird, so I've aimed for clarity over brevity.

BTW, this config is overwritten by a configmap in the daemonset. I'm included this for the same reason as the "ENTRYPOINT": sane defaults. This allows users with sufficient permissions to overwrite the behavior.

metadata-proxy/start-proxy.sh
Q-Lee: I don't.

metadata-proxy/Makefile
Q-Lee: Ok

Q-Lee:
@timstclair - does this look better?

Take a look at the configmap and daemonset here:
https://github.com/Q-Lee/kubernetes/tree/mds/cluster/addons/metadata-proxy

metadata-proxy/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM gcr.io/google-containers/debian-base-amd64:0.1
+LABEL maintainer "qlee@google.com"
+
+ARG DEBIAN_FRONTEND=noninteractive
+RUN apt-get update && apt-get install -y \
```
timstclair: Use `clean-install iptables nginx`

metadata-proxy/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+  apt-utils && apt-get install -y \
+  iptables \
+  nginx \
+  sudo
```
timstclair: I thought you were removing sudo?

metadata-proxy/nginx.conf
```diff
@@ -0,0 +1,63 @@
...
+worker_processes 4;
+pid /run/nginx.pid;
+
+events {
+  worker_connections 20;
+}
+
+http {
+  server {
+    listen 127.0.0.1:988;
```
timstclair: So I'll let you decide if it's worth it in this PR or not, but this is what you could do here:

1. Add variables for `PORT` and `METADATA_IP` (e.g. `__PORT__` and `__METADATA_IP__`)
2. Update the start-proxy.sh script to replace those variables in the config with the value from an environment variable, using sed, e.g. `sed -i s/__PORT__/${PORT:-988}/g /etc/nginx/nginx.conf`
3. Populate the env vars from the pod spec

metadata-proxy/start-proxy.sh
```diff
@@ -0,0 +1,32 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+_term() {
+  iptables -D -t filter -I KUBE-METADATA-SERVER -j ACCEPT
+  iptables -D -t nat -I PREROUTING -p tcp -d 169.254.169.254 --dport 80 -j DNAT --to-destination 127.0.0.1:988
+  kill -TERM "$child" 2>/dev/null
```
timstclair: delete this?

metadata-proxy/Dockerfile
Q-Lee: I sure did.

metadata-proxy/Dockerfile
Q-Lee: What's "clean-install"?

timstclair: A script that encapsulates `apt-get update && apt-get install` and also does cleanup: https://github.com/kubernetes/kubernetes/blob/master/build/debian-base/clean-install

metadata-proxy/nginx.conf
Q-Lee: Even if make the port configurable, the addon is still a static file on the master. I think that choosing an unassigned privileged port does more to prevent conflicts than hiding an easy-to-configure setting behind a hard-to-impossible-to-configure implementation.

We can change this if it becomes a problem, but I don't see it being useful until we decide on, implement, and deploy a serious overhaul to the addon manager.

metadata-proxy/start-proxy.sh
Q-Lee: Whoops.

metadata-proxy/Dockerfile
Q-Lee: The "apt-utils" package must be installed first, and "apt-get install" doesn't guarantee the order of installation.

I imagine the next two commands (clean and rm -rf) are simply to shrink the image size?

Q-Lee: Interesting.

no cleanup:        fb786fa5fc50        112 MB
"clean-install":        730191ef3d6f        61.94 MB
"apt-get clean -y":        7f7e7eac1fb7        112 MB
gcr.io/google-containers/debian-base-amd64:        61a750f1056b        40.64 MB

The "rm -rf" seems to be removing about 50MB of junk. I'm convinced.

timstclair:
lgtm. Please squash & rebase.

jbeda:
This image should probably be named `gce-metadata-proxy` as it is GCE specific.  Might be a good idea to fix that up before the close of 1.7.

PR #2583, opened 2017-05-06
[Ansible] remove conflicted configure task

xizhibei:
I found the task **Get Systemd config files from Kubernetes repository** is conflicted with following tasks, and it override the correct config, causing the cluster deploy failed.

- https://github.com/kubernetes/contrib/blob/master/ansible/roles/kubernetes/tasks/configure.yml
- https://github.com/kubernetes/contrib/blob/master/ansible/roles/node/tasks/kubelet-configure.yml
- https://github.com/kubernetes/contrib/blob/master/ansible/roles/node/tasks/proxy-configure.yml

### DIFF
--- a/ansible/roles/node/tasks/retrieve-systemd-files.yml
+++ b/ansible/roles/node/tasks/retrieve-systemd-files.yml
@@ -15,23 +15,6 @@
     https_proxy: "{{ https_proxy|default('') }}"
     no_proxy: "{{ no_proxy|default('') }}"
 
-- name: Get Systemd config files from Kubernetes repository
-  get_url:
-    url=https://raw.githubusercontent.com/kubernetes/contrib/master/init/systemd/environ/{{ item }}
-    dest={{ kube_config_dir }}/{{ item }}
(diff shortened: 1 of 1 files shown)


xizhibei:
@k8s-ci-robot CLA signed

ingvagabund:
The task is run conditionally, see [1]. What are your installation requirements? Are you setting `kube_source_type` to `github-release` or running on Ubuntu 16?

[1] https://github.com/kubernetes/contrib/blob/master/ansible/roles/node/tasks/install.yml#L9

xizhibei:
Yes, I set `kube_source_type` to `github-release`, and I install k8s on the 1 master and 3 nodes cluster. 

The machines is CentOS 7.2 with kernel `3.10.0-327.22.2.el7.x86_64`

I think the config files under https://raw.githubusercontent.com/kubernetes/contrib/master/init/systemd/environ/ , which contains `127.0.0.1`, is not suitable for multi machines cluster situation.

Or, maybe there should be some conditions ?

PR #2586, opened 2017-05-08
Enabled rescheduler help include listen address.

### LINKED ISSUE #2585: The latest rescheduler do not include listen address flag
The test steps are as following:

1) Check out `contrib` code
2) go to `contrib` and `make build`
3) build succeed, but the help info do not include flag `listen-address`.

The source code is here https://github.com/kubernetes/contrib/blob/master/rescheduler/rescheduler.go#L83-L84

My test
```
root@k8s001:~/go/src/k8s.io/contrib/rescheduler# ./rescheduler  --help
Usage of rescheduler: rescheduler --running-in-cluster=true:
      --housekeeping-interval duration   How often rescheduler takes actions. (default 10s)
      --initial-delay duration           How long should rescheduler wait after start to make sure
		 all critical addons had a chance to start. (default 2m0s)
      --kube-api-content-type string     Content type of requests sent to apiserver. (default "application/vnd.kubernetes.protobuf")
      --pod-scheduled-timeout duration   How long should rescheduler wait for critical pod to be scheduled
		 after evicting pods to make a spot for it. (default 10m0s)
      --running-in-cluster               Optional, if this controller is running in a kubernetes cluster, use the
		 pod secrets for creating a Kubernetes client. (default true)
      --system-namespace string          Namespace to watch for critical addons. (default "kube-system")
```

/cc @piosz  can you help?

gyliu513:
Fixed #2585

### DIFF
--- a/rescheduler/rescheduler.go
+++ b/rescheduler/rescheduler.go
@@ -80,7 +80,7 @@ var (
 		`How long should rescheduler wait for critical pod to be scheduled
 		 after evicting pods to make a spot for it.`)
 
-	listenAddress = flag.String("listen-address", "localhost:9235",
+	listenAddress = flags.String("listen-address", "localhost:9235",
 		`Address to listen on for serving prometheus metrics`)
 )
 


piosz:
@gyliu513 thanks for the fix!

PR #2592, opened 2017-05-10
Fix perfdash after moving performance data to separate files

gmarek:
FYI @Random-Liu @dchen1107 - I want to wait for main repo changes to go in first, so that I can actually test this one. But this in theory should be enough.


gmarek:
It's done and working. @wojtek-t PTAL

PR #2601, opened 2017-05-15
[prom-to-sd] Added support for histogram

piosz:


prometheus-to-sd/translator/translator.go
```diff
@@ -123,11 +125,69 @@ func translateOne(config *config.GceConfig, component string, name string, mType
...
+}
+
+func getHistogramValue(h *dto.Histogram) *v3.Distribution {
+	count := int64(h.GetSampleCount())
+	mean := float64(0)
+	dev := float64(0)
+	buckets := []float64{}
+	values := []int64{}
+
+	if count > 0 {
```
gmarek: Wouldn't it be simpler to do fast return in case when `count == 0`?

prometheus-to-sd/translator/translator.go
```diff
@@ -123,11 +125,69 @@ func translateOne(config *config.GceConfig, component string, name string, mType
...
+		point.Value.Int64Value = &val
+		point.ForceSendFields = append(point.ForceSendFields, "Int64Value")
+	}
+}
+
+func getHistogramValue(h *dto.Histogram) *v3.Distribution {
+	count := int64(h.GetSampleCount())
+	mean := float64(0)
+	dev := float64(0)
+	buckets := []float64{}
```
gmarek: this is very misleading name - please rename to 'bounds' or 'bucketBorders' or something more explicit.

gmarek: Discussed f2f, it's fine as it is.

piosz: done

prometheus-to-sd/translator/translator.go
piosz: We need to export bucket bounds anyway

PR #2606, opened 2017-05-20
Use correct setting name for purgeInterval

bjoernhaeuser:
Correct setting name for "purgeInterval". Otherwise the config will not be picked up and zookeeper is running without purging, eventually causing out of disk.


bjoernhaeuser:
contrib seems pretty dead :)

PR #2610, opened 2017-05-24
[event-exporter] Add event exporter image sources

crassirostris:
Implement tool for exporting events from Kubernetes to Stackdriver Logging


crassirostris:
@wojtek-t Could you please take a look at the watching part? It's in the `watchers` subfolder

fluentd/event-exporter/watchers/consumer.go
```diff
@@ -0,0 +1,24 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package watchers
+
+// WatchConsumer interface provides a way to act upon signals from watcher.
+type WatchConsumer interface {
```
wojtek-t: Instead of introducing this interface, please reuse "ResourceEventHandler" interface from from client-go/tools/cache.

fluentd/event-exporter/watchers/events/consumer_test.go
```diff
@@ -0,0 +1,185 @@
...
+package events
+
+import (
+	"testing"
+
+	api_v1 "k8s.io/client-go/pkg/api/v1"
+)
+
+type fakeEventConsumer struct {
+	EventWatchConsumer
```
wojtek-t: Please remove - I don't think you need it.

fluentd/event-exporter/watchers/events/watcher.go
```diff
@@ -0,0 +1,69 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/kubernetes"
+	api_v1 "k8s.io/client-go/pkg/api/v1"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/contrib/fluentd/event-exporter/watchers"
+)
+
+const (
+	eventStorageTTL = 2 * time.Hour
```
wojtek-t: Why 2 hours?
I'm not saying it's wrong - just add a comment why.

fluentd/event-exporter/watchers/storage.go
```diff
@@ -0,0 +1,123 @@
...
+
+func (s *watcherStore) Delete(obj interface{}) error {
+	if err := s.cacheStorage.Delete(obj); err != nil {
+		return err
+	}
+	s.consumer.Delete(obj)
+	return nil
+}
+
+func (s *watcherStore) List() []interface{} {
```
wojtek-t: Since you did:
```
type watcherStore struct {
	cache.Store
        ....
```

I don't think you need to redefine any of those methods - they should already be there...

fluentd/event-exporter/watchers/watcher.go
```diff
@@ -0,0 +1,56 @@
...
+type WatcherConfig struct {
+	ListerWatcher cache.ListerWatcher
+	ExpectedType  interface{}
+	StoreConfig   *WatcherStoreConfig
+	ResyncPeriod  time.Duration
+}
+
+// Watcher is an interface of the generic proactive API watcher.
+type Watcher interface {
+	Start()
```
wojtek-t: In general, I would like to have an option also to stop it.
Ideally, I would like to get rid of "Run()" method of Reflector and leave only "RunUntil(stopCh)" - that's a better patter, even if you initially don't need it.

fluentd/event-exporter/watchers/storage.go
```diff
@@ -0,0 +1,123 @@
...
+	return nil
+}
+
+func (s *watcherStore) Update(obj interface{}) error {
+	oldObj, ok, err := s.cacheStorage.Get(obj)
+	if err != nil {
+		return err
+	}
+	if !ok {
+		oldObj = nil
```
wojtek-t: This should hopefully be like that anyway...

crassirostris: I was under impression, that the pattern implies that we shouldn't assume anything about the return value if the call wasn't successful (ok is false or err is not nil)

Is that not true?

wojtek-t: I guess it's safer to leave, yes. But I think all implementations actually return nil in case of false.

But sure - you can leave what you have - it's obviously correct.

crassirostris:
@wojtek-t Thanks a lot! PTAL

fluentd/event-exporter/utils/stop_helpers.go
```diff
@@ -0,0 +1,48 @@
...
+
+// StoppableFunc is a type of function that blocks on run, but can be
+// stopped via stop channel passed as an argument
+type StoppableFunc func(<-chan struct{})
+
+// RunConcurrentlyUntil runs several stoppable functions in parallel and blocks
+// until all of them are executed.
+func RunConcurrentlyUntil(stopCh <-chan struct{}, funcs ...StoppableFunc) {
+	var wg sync.WaitGroup
+	stopChs := make([]chan struct{}, len(funcs))
```
wojtek-t: Let's not create multiple stop channels. Instead, what you should do is:
```
stopCh := make(chan struct{})
for i, f := range funcs {
   wg.Add(1)
   go func() {
    defer wg.Done()       <- this is to make this run also if e.g. f panics.
    f(stopCh)
   }
}

close(stopCh)
wg.Wait()
}
```

So close(stopCh) does the whole trick.

Another option is to not have this function at all and simply pass the same stopCh to all functions.

fluentd/event-exporter/watchers/events/watcher.go
```diff
@@ -0,0 +1,74 @@
...
+	Handler        EventHandler
+}
+
+// NewEventWatcher create a new watcher that only watches the events resource.
+func NewEventWatcher(client kubernetes.Interface, config *EventWatcherConfig) watchers.Watcher {
+	return watchers.NewWatcher(&watchers.WatcherConfig{
+		ListerWatcher: &cache.ListWatch{
+			ListFunc: func(options meta_v1.ListOptions) (runtime.Object, error) {
+				list, err := client.CoreV1().Events(meta_v1.NamespaceAll).List(options)
+				list.Items = config.FilterListFunc(list.Items)
```
wojtek-t: I started thinking about this and now I'm not sure if it would work as you expect it to work.
Basically, assume that you are starting your component, and a bunch of events already exist.
But it seems that all filter functions you already implemented simply drop all the content. So this will return empty result.

This means, that every modification of one of those events, even though this is modification, it will be delivered to consumers as "OnAdd" event (since the previous version of object is computed based on contents of the store).

Maybe for your component it's not a big deal, but it definitely deserves a comment.

fluentd/event-exporter/utils/stop_helpers.go
crassirostris: Ack, done

> Another option is to not have this function at all and simply pass the same stopCh to all functions.

Each `Run` call is blocking, so there always should be something to block on both of them, and I'd prefer not to inline this to make the code in the high-level class clearer

fluentd/event-exporter/watchers/events/watcher.go
crassirostris: Ack

crassirostris: I've realized that it's unnecessary to drop those events, b/c reflector uses Replace on the list, which in turn doesn't call Add

wojtek-t: Actually, this deserves a comment, because it's not necessary obvious.

crassirostris: OK, agree, thanks for pointing that out!

crassirostris: @wojtek-t Added a comment above & in the sink interface

crassirostris:
@wojtek-t Extremely appreciate your comments, it's super useful! PTAL

fluentd/event-exporter/sinks/stackdriver/sink.go
```diff
@@ -0,0 +1,163 @@
...
+				break
+			}
+		}
+	}()
+}
+
+func (s *sdSink) flushBuffer() {
+	entries := s.currentBuffer
+	s.currentBuffer = nil
+	s.concurrencyChannel <- struct{}{}
```
piosz: Why do you need it? IIUC you want to send only one request at the same time. Please add a comment about it.

fluentd/event-exporter/sinks/stackdriver/sink_config.go
```diff
@@ -0,0 +1,81 @@
...
+	"strings"
+	"time"
+
+	"cloud.google.com/go/compute/metadata"
+	"github.com/golang/glog"
+	sd "google.golang.org/api/logging/v2"
+)
+
+const (
+	defaultFlushDelay     = 5 * time.Second
```
piosz: Can you briefly explain in the comments why you have such values of those consts? I don't mean they are wrong.

fluentd/event-exporter/sinks/stackdriver/sink_config.go
```diff
@@ -0,0 +1,81 @@
...
+	FlushDelay     time.Duration
+	MaxBufferSize  int
+	MaxConcurrency int
+	LogName        string
+	Resource       *sd.MonitoredResource
+}
+
+func newGceSdSinkConfig() (*sdSinkConfig, error) {
+	if !metadata.OnGCE() {
+		return nil, errors.New("not running on GCE, which is not yet supported for Stackdriver sink")
```
piosz: remove yet

fluentd/event-exporter/sinks/stackdriver/writer.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+func (w sdWriterImpl) Write(entries []*sd.LogEntry, logName string, resource *sd.MonitoredResource) int {
+	req := &sd.WriteLogEntriesRequest{
+		Entries:  entries,
+		LogName:  logName,
+		Resource: resource,
+	}
+
+	for {
```
piosz: ... and will actually break the whole flow.

piosz (changes requested):
I looked into Stackdriver part (without tests). I assume @wojtek-t will review the communication with Kubernetes part.

BTW very nice concurrency model:)

fluentd/event-exporter/sinks/stackdriver/writer.go
crassirostris: It will block the whole pipeline until Stackdriver is available. If there's something wrong with entries, SD will return 400 and that'll end the loop

> ... and will actually break the whole flow.

I didn't understand that, sorry, could you please elaborate?

fluentd/event-exporter/sinks/stackdriver/log_entry_factory.go
```diff
@@ -0,0 +1,103 @@
...
+
+func (f *sdLogEntryFactory) FromEvent(event *api_v1.Event) *sd.LogEntry {
+	payload, err := f.serializeEvent(event)
+	if err != nil {
+		glog.Warningf("Failed to encode event %v: %v", event, err)
+	}
+
+	return &sd.LogEntry{
+		JsonPayload: payload,
+		Severity:    "INFO",
```
crassirostris: Ack

fluentd/event-exporter/sinks/interface.go
```diff
@@ -0,0 +1,44 @@
...
+	// sink factory to this map.
+	KnownSinks = make(map[string]SinkFactory)
+)
+
+// Sink interface represents a generic sink that is responsible for handling
+// actions upon the event objects and filter the initial events list.
+type Sink interface {
+	events.EventWatchConsumer
+
+	FilterList([]api_v1.Event) []api_v1.Event
```
crassirostris: To send a log entry about some potentially lost events iff there are some events present in Stackdriver

fluentd/event-exporter/sinks/stackdriver/sink.go
crassirostris: No, that's not true, we want to be able to keep up with the stream of events and one thread may not be enough

crassirostris: Sorry, didn't understand, comment about what?

fluentd/event-exporter/sinks/stackdriver/sink_config.go
crassirostris: 5 sec - b/c fluentd has 5 sec flush by default
1k entries - kinda takes the whole 100Mb/s bandwidth for 5 sec
10 concurrency - if request is on average 10MB, process will consume on average 100MB

As you can see, no particular reason, they just seem kinda reasonable

crassirostris: I'm not sure these "reasons" should be mentioned

fluentd/event-exporter/sinks/stackdriver/sink_config.go
crassirostris: Ack

fluentd/event-exporter/OWNERS
```diff
@@ -0,0 +1,4 @@
+approvers:
+- crassirostris
```
crassirostris: Ack

fluentd/event-exporter/README.md
```diff
@@ -0,0 +1,66 @@
+# Event Exporter
+
+This tool is used to export Kubernetes events. It effectively runs a watch on
+the apiserver, detecting as granular as possible all changes to the event
+objects. Event exporter currently exports only to Stackdriver.
```
crassirostris: Ack

fluentd/event-exporter/sinks/stackdriver/log_entry_factory.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+}
+
+func newEncoder() runtime.Encoder {
+	jsonSerializer := json.NewSerializer(json.DefaultMetaFactory, scheme.Scheme, scheme.Scheme, false)
+	directCodecFactory := serializer.DirectCodecFactory{CodecFactory: scheme.Codecs}
+	return directCodecFactory.EncoderForVersion(jsonSerializer, api_v1.SchemeGroupVersion)
+}
+
+func (f *sdLogEntryFactory) serializeEvent(event *api_v1.Event) ([]byte, error) {
```
crassirostris: With field blacklist? I'm sure we don't

fluentd/event-exporter/sinks/stackdriver/sink.go
```diff
@@ -0,0 +1,163 @@
...
+	s.logEntryChannel <- logEntry
+}
+
+func (s *sdSink) Update(oldEvent *api_v1.Event, newEvent *api_v1.Event) {
+	if newEvent.Count != oldEvent.Count+1 {
+		// Sink doesn't send a LogEntry to Stackdriver, b/c event compression might
+		// indicate that part of the watch history was lost, which may result in
+		// multiple events being compressed. This may create an unecessary
+		// flood in Stackdriver.
+		glog.V(2).Infof("Event count has increased by something but one.\n"+
```
crassirostris: Ack

fluentd/event-exporter/README.md
```diff
@@ -0,0 +1,66 @@
...
+    Reflector resync period (default 1m0s)
+-sink-opts string
+    Parameters for configuring sink
+-sink-type string
+    Name of the sink used for events ingestion (default "stackdriver")
+```
+
+## Deploy
+
+Example deployment:
```
crassirostris: How about I'll add a single kubectl command to run this deployment? WDYT?

fluentd/event-exporter/main.go
```diff
@@ -0,0 +1,107 @@
...
+
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/contrib/fluentd/event-exporter/sinks"
+	"k8s.io/contrib/fluentd/event-exporter/sinks/stackdriver"
+)
+
+var (
+	resyncPeriod = flag.Duration("resync-period", 1*time.Minute, "Reflector resync period")
+	sinkType     = flag.String("sink-type", stackdriver.SdSinkName, "Name of the sink "+
```
crassirostris: Mb we'll reuse it somewhere else, e.g. by sending events to pub/sub

fluentd/event-exporter/sinks/interface.go
crassirostris: Actually, I've just realized that it makes sense to send a message even if the list is empty, e.g. if all events were gc'd while event exporter was down

fluentd/event-exporter/README.md
crassirostris: Decided to add yamls with the RBAC deployment, since it seems to be more needed

fluentd/event-exporter/OWNERS
crassirostris: Done

fluentd/event-exporter/watchers/events/handler.go
```diff
@@ -0,0 +1,69 @@
...
+func (c *eventHandlerWrapper) OnUpdate(oldObj interface{}, newObj interface{}) {
+	oldEvent, oldOk := c.convert(oldObj)
+	newEvent, newOk := c.convert(newObj)
+	if oldOk && newOk {
+		c.handler.OnUpdate(oldEvent, newEvent)
+	}
+}
+
+func (c *eventHandlerWrapper) OnDelete(obj interface{}) {
+	if event, ok := c.convert(obj); ok {
```
wojtek-t: Actually, one thing that I missed previous time is that for delete, it can be something different tha *v1.Event.
You should have something like this:
https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/replication/replication_controller.go#L348

crassirostris: Wow, thanks, that's not obvious

fluentd/event-exporter/sinks/stackdriver/sink.go
piosz: just add comment explaining this

crassirostris: Added a comment near the field definition

fluentd/event-exporter/sinks/stackdriver/writer.go
piosz: As discussed offline let's have consistency with fluentd here, but I think this is not ideal approach to wait forever.

fluentd/event-exporter/watchers/events/handler.go
crassirostris: Done, PTAL

fluentd/event-exporter/sinks/stackdriver/writer.go
crassirostris: Added a comment about it

fluentd/event-exporter/main.go
piosz: The we will introduce them. It totally doesn't make sense to maintain code "just in case".

fluentd/event-exporter/README.md
piosz: cool

fluentd/event-exporter/sinks/stackdriver/log_entry_factory.go
piosz: Oh, I see

fluentd/event-exporter/sinks/stackdriver/sink_config.go
piosz: Totally reasonable.

fluentd/event-exporter/example/event-exporter-service-account.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: event-exporter-sa
```
piosz: new line

piosz (commented):
Please remove sinkFactory and then I'll LGTM.

crassirostris:
@piosz PTAL

piosz:
LGTM, please squash commits a bit, to have 2-3 of them.

mwielgus:
approve for known flags.

crassirostris:
Applying approve from Marcin's comment

piosz (approved):

piosz:
Thanks @crassirostris for the PR!

PR #2612, opened 2017-05-26
Send timeserieses to stackdriver in batches.

loburm:
Stackdriver has a limit of 200 timeserieses per request. To be able to
send more we need to send them in batches.

### DIFF
--- a/prometheus-to-sd/translator/stackdriver.go
+++ b/prometheus-to-sd/translator/stackdriver.go
@@ -18,11 +18,16 @@ package translator
 
 import (
 	"fmt"
+	"sync"
 
 	"github.com/golang/glog"
 	v3 "google.golang.org/api/monitoring/v3"
 
 	"k8s.io/contrib/prometheus-to-sd/config"
+)
+
+const (
+	maxTimeseriesesPerRequest = 200
 )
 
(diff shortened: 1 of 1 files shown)


prometheus-to-sd/translator/stackdriver.go
```diff
@@ -18,27 +18,49 @@ package translator
...
-	req := &v3.CreateTimeSeriesRequest{TimeSeries: ts}
 	proj := fmt.Sprintf("projects/%s", config.Project)
 
-	_, err := service.Projects.TimeSeries.Create(proj, req).Do()
-	if err != nil {
-		glog.Errorf("Error while sending request to Stackdriver %v", err)
-	} else {
-		glog.V(4).Infof("Successfully sent %v timeserieses to Stackdriver", len(req.TimeSeries))
+	var wg sync.WaitGroup
+	wg.Add(numberOfBatches(len(ts)))
```
piosz: You don't need this function: you can just have `wg.Add(1)` right before `go func(...)` below. It'd be less error prone.

loburm: Thanks, I have updated the code.

PR #2616, opened 2017-05-30
Fix flaky TestTranslatePrometheusToStackdriver.

mborsz:
It assumes particular order of result, which is not guaranteed by
TestTranslatePrometheusToStackdriver as it uses maps internally.

Tested:
```
cd prometheus-to-sd; i=0; while godep go test ./...; do echo $i; i=$((i+1)); done
```
Without fix: It was failing after 3-9 attempts
Now: Not able to reproduce failure (>100 attempts succeeded)


piosz:
/lgtm
/approve
Nice catch! Thanks!

PR #2620, opened 2017-05-31
Update fluentd-gcp gems

crassirostris:
To address problem from https://github.com/kubernetes/kubernetes/pull/45892

/cc @piosz @igorpeshansky


crassirostris:
@igorpeshansky Could you please approve the new set of gems?

fluentd/fluentd-gcp-image/Gemfile
```diff
@@ -1,9 +1,9 @@
 source 'https://rubygems.org'
 
-gem 'fluentd', '~>0.12.32'
-gem 'fluent-plugin-record-reformer', '~>0.8.3'
+gem 'fluentd', '~>0.14.17'
```
igorpeshansky: We distribute 0.14.14, but I think that's just because that was the latest available... Just an FYI.

crassirostris: OK, thanks! I will keep this in mind

PR #2622, opened 2017-05-31
Remove leading and trailing whitespaces from cluster name.

### LINKED ISSUE #2619: [prometheus-to-sd] Cluster name field has a newline character
Here be Trim: https://github.com/kubernetes/contrib/blob/master/prometheus-to-sd/config/gce_config.go#L50

loburm:
It may happen that GCE is going to return cluster-name metadata with
trailing or leading whitespaces, to avoid errors remove them from the
final configuration.

fix #2619

### DIFF
--- a/prometheus-to-sd/config/gce_config.go
+++ b/prometheus-to-sd/config/gce_config.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 
 	gce "cloud.google.com/go/compute/metadata"
+	"strings"
 )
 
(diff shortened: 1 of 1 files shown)


piosz:
```
# k8s.io/contrib/prometheus-to-sd/config
config/gce_config.go:23: imported and not used: "k8s.io/contrib/prometheus-to-sd/vendor/github.com/golang/glog"
```
please fix

piosz:
ping

loburm:
Removed import.

PR #2623, opened 2017-05-31
Rewrite addon-resizer nanny

x13n:
There will be a single estimator class, providing two ranges: acceptable
and recommended range. As long as current pod requirements and limits fall into
acceptable range, nothing happens. Once either limits or requirements
fall out of acceptable range, they are both updated to lower (when
upscaling) or higher (when downscaling) end of recommended range. This
approach prevents flapping, which took place in previous implementation,
when cluster size oscillated around certain values.

More details: https://docs.google.com/a/google.com/document/d/1T0A7GHwNU_w6gpq_eCN166aRth6usbRvxgJJqSZeESU/edit?usp=sharing (shared with kubernetes-sig-instrumentation and kubernetes-sig-autoscaling)


x13n:
(Ping robot for CLA refresh.)

mwielgus:
Please migrate this PR to kubernetes/autoscaler

x13n:
Done: https://github.com/kubernetes/autoscaler/pull/128, closing this.

PR #2625, opened 2017-06-01
Fix fluentd-gcp image

crassirostris:
To keep fluentd 0.12


crassirostris:
@igorpeshansky Could you please take a look?

igorpeshansky:
@crassirostris Looks fine, but I'm curious as to what didn't work with fluentd 0.14... Are you running into https://github.com/fluent/fluentd/issues/1555, or are there other issues?

crassirostris:
@igorpeshansky `@type` and a change in one of the filter plugins broke our config and this patch is required for a cherry-pick into an old release

igorpeshansky:
Ah, makes sense, thanks. LGTM, then. :shipit:

crassirostris:
Thanks!

crassirostris:
@igorpeshansky BTW, version with 0.14 has a larger memory footprint. Where 0.12 works without any issues, 0.14 gets OOM-killed

PR #2628, opened 2017-06-09
Refactor node-perf-dash to read perf and time series data from json files

yguo0905:
Fixes the node-dash-perf issue in https://github.com/kubernetes/kubernetes/issues/44003, and refactors the code.

- Rename `data.go` to `types.go` and move the global variables from `types.go` to `node-perf-dash.go` so that `types.go` only contains types.
- Move the parsing logic from `downloader.go` to `parser.go`.
- `parser.go` now reads the perf and time series data from json files, generated in  https://github.com/kubernetes/kubernetes/pull/47260.

/assign @Random-Liu


node-perf-dash/kubelet-parser.go
```diff
@@ -188,7 +194,7 @@ func GrabTracingKubelet(d Downloader, job string, buildNumber int, nodeName stri
 			"test": sortedTestTime[currentTestIndex].TestName,
 			"node": nodeName,
 		},
-		Version: tracingVersion,
+		Version: perftype.CurrentKubeletPerfMetricsVersion,
```
Random-Liu: I don't understand, why do we add version on the node-perf-dash side?

The version is used to identify whether the version of the original data, it doesn't seem to be the original purpose of the versioning stuff.

yguo0905: This is because node-perf-dash itself also generates perf data logs by parsing the content of the `kubelet.log`.

E.g.,
```
[Result:TimeSeries]{"labels":{"node":"n1-standard-1-coreos-alpha-1122-0-0-v20160727-e4490a91","test":"resource_35"},"version":"v1","op_series":{"container_start_pleg":[1496646375134107000, ...
...
[Finish:TimeSeries]
```

yguo0905:
I've added `supportedMetricVersion` and set it to `v2`. The metrics of unsupported versions will be ignored. PTAL.

node-perf-dash/kubelet-parser.go
Random-Liu: Maybe we could leave this empty now, and we don't need to expose `CurrentKubeletPerfMetricsVersion`, either.

Random-Liu:
> This branch is out-of-date with the base branch

And seems the PR need rebase.

node-perf-dash/kubelet-parser.go
yguo0905: Yeah, I forgot to remove this.

yguo0905:
I will wait for https://github.com/kubernetes/kubernetes/pull/47260 to be merged and test this PR again with the new test runs.

yguo0905:
Rebased, tested and bumped the image version. Will need to release it.

Random-Liu:
LGTM besides the poll interval. May want to squash the first 2 commits.

Random-Liu:
LGTM

PR #2629, opened 2017-06-10
Make echoheaders nginx container escape &<> to prevent XSS.

rmmh:
Tested locally by doing `make container; docker run --rm -p 8080:8080 gcr.io/google_containers/echoserver:1.4` and then `curl 'localhost:8080/xss-<&>'`.

Intentionally not bumping the image tag because that would require multiple patches against old branches to pick up the newer version, and this change is relatively small and safe!

### DIFF
--- a/ingress/echoheaders-redirect/nginx.conf
+++ b/ingress/echoheaders-redirect/nginx.conf
@@ -24,13 +24,21 @@ http {
         location / {
             lua_need_request_body on;
             content_by_lua_block {
+                -- replace HTML special characters to prevent XSS.
+                function sanitize(s)
+                    local s, _ = string.gsub(s or "", "[&<>]",
+                        function(m) return string.format("&#x%x;", string.byte(m, 1)) end)
+                    return s
+                end
+
                 ngx.say("CLIENT VALUES:")
                 ngx.say("client_address=", ngx.var.remote_addr)
-                ngx.say("command=", ngx.req.get_method())
-                ngx.say("real path=", ngx.var.request_uri)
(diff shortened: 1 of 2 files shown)


bprashanth:
Please add `X-XSS-Protection: 1; mode=block` to force the browser to render placeholders instead of sanitizing content. If that's not enough, please check if there's a standard markup sanitization library you can use instead of doing it yourself, to offload the burden of proof. 

Since this is just a toy container, I'm fine with those going in as follow ups whenever you have time. LGTM.

bprashanth:
@aledbf might also have some sage suggestions for the follow up

rmmh:
Those headers won't help, this is specifically a problem with an overly-sensitive XSS scanner for sniffing in old IE, which doesn't support CSP. The page is served as `text/plain`, so any modern browser won't be vulnerable to XSS. And XSS against a random IP isn't useful anyways, so...

@bprashanth can you do /approve too?

rmmh:
Replaced with escape function from a Lua templating engine. It was simpler to inline than to figure out the packaging system for nginx.

bprashanth:
Ah, no i mean apply the headers for the case they would help (i.e typical xss). 
/approve

aledbf:
@rmmh @bprashanth we moved the image to https://github.com/kubernetes/ingress/tree/master/images/echoheaders

aledbf:
this is the new output https://github.com/kubernetes/ingress/pull/785

aledbf:
@rmmh can you open the PR in the new repository?

rmmh:
Ported to kubernetes/ingress#865

PR #2633, opened 2017-06-13
Bump images that use debian-base image

ixdy:
I recently updated the `debian-base-*` image off upstream with fixes for a number of CVEs.
The downstream images now need to be updated, which I've done in this PR, bumping the patch version for each.

I haven't yet pushed any of these images.
After doing so, I'll need to follow up with additional changes in the manifests.

x-ref https://github.com/kubernetes/kubernetes/issues/47386
cc @Q-Lee @crassirostris


timstclair:
Thanks Jeff!

/lgtm

ixdy:
These have all been pushed to gcr.io/google-containers.

PR #2641, opened 2017-06-22
[event-exporter] Fix event exporter spam

crassirostris:
Watch framework periodically resets the connections (~one every 15 minutes) which causes a lot of messages about lost events in Stackdriver. Though this change is raising the chances that some events will be lost without a note, the probability of this event is extremely low (it requires to lose whole apiserver event history, e.g. in case of a really long network outage). Note, that in case container is restarted, the message will still be generated.

/cc @piosz


crassirostris:
@piosz Thanks!

PR #2642, opened 2017-06-27
Enabling Multi-arch support in Ansible scripts

Pensu:
This PR enables multi-arch support for ansible scripts when github-release option is chosen. 

 - By default for x86_64 machines the packages have amd64 as arch, so we need to set ansible_architecture as amd64 when it's x86_64. 

 - Flannel new version has a different directory structure compared to older version, so had to change the version too with flannel.

### DIFF
--- a/ansible/roles/common/defaults/main.yaml
+++ b/ansible/roles/common/defaults/main.yaml
@@ -13,3 +13,7 @@ is_coreos: false
 has_rpm: false
 has_firewalld: false
 has_iptables: false
+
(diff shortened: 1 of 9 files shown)


Pensu:
I have signed the CLA, not sure why is it saying as CLA required?

Pensu:
@ingvagabund @eparis Hi, can you please review this?

Pensu:
@piosz Can you please have a look?

Issue #2643, opened 2017-06-27
pets/peer-finder could have a better default domain, messaging


venezia:
Right now, peer-finder will presume `cluster.local` as the domain name of the kubernetes cluster unless manually overridden.  This, in conjunction with some popular helm charts (including mongodb-replicaset) can lead to an odd circumstance where peer-finder will seemingly hang forever without any message to the user if the domain name is different.

Given that peer-finder is relying heavily on DNS, it would seem reasonable to reference `/etc/resolv.conf` to better guess the domain name if not supplied.

Furthermore, instead of just looping without any message to the user while the presumed host entry is not in the service's peer list, it would be nice to notify the user what peers are there and what the presumed host entry is.

Peer-entry is currently used in an init container in [kubernetes' chart repo](https://github.com/kubernetes/charts/tree/master/stable/mongodb-replicaset) - and init containers are a bit of a pain to debug - so the more information provided to the user the better.

PR #2644, opened 2017-06-27
Improve domain default, add messaging when in loop to help anyone debugging

venezia:
This is a proposed patch to peer-finder per issue #2643

### DIFF
--- a/pets/peer-finder/peer-finder.go
+++ b/pets/peer-finder/peer-finder.go
@@ -20,15 +20,16 @@ package main
 import (
 	"flag"
 	"fmt"
+	"io/ioutil"
 	"log"
 	"net"
 	"os"
 	"os/exec"
+	"regexp"
 	"sort"
 	"strings"
 	"time"
 
-	"io/ioutil"
 	"k8s.io/apimachinery/pkg/util/sets"
 )
 
@@ -41,7 +42,7 @@ var (
 	onStart   = flag.String("on-start", "", "Script to run on start, must accept a new line separated list of peers via stdin.")
 	svc       = flag.String("service", "", "Governing service responsible for the DNS records of the domain this pod is in.")
 	namespace = flag.String("ns", "", "The namespace this pod is running in. If unspecified, the POD_NAMESPACE env var is used.")
-	domain    = flag.String("domain", "", "The Cluster Domain which is used by the Cluster.")
+	domain    = flag.String("domain", "", "The Cluster Domain which is used by the Cluster, if not set tries to determine it from /etc/resolv.conf file.")
 )
 
 func lookup(svcName string) (sets.String, error) {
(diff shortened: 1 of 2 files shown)


pets/peer-finder/peer-finder.go
```diff
@@ -74,17 +75,42 @@ func main() {
...
+
+	// If domain is not provided, try to get it from resolv.conf
+	if (*domain == "") || (ns == "") {
+		resolvConfBytes, err := ioutil.ReadFile("/etc/resolv.conf")
+		resolvConf := string(resolvConfBytes)
+		if err != nil {
+			log.Fatal("Unable to read /etc/resolv.conf")
+		}
+		if ns != "" {
+			tempResolv := strings.Split(resolvConf, ("search " + ns + "."))
```
foxish: This logic is quite brittle. The resolv.conf may not contain the cluster domain, depending on the way the cluster was setup. I'm in favor of making logging more explicit.

pets/peer-finder/peer-finder.go
```diff
@@ -40,7 +41,7 @@ var (
 	onStart   = flag.String("on-start", "", "Script to run on start, must accept a new line separated list of peers via stdin.")
 	svc       = flag.String("service", "", "Governing service responsible for the DNS records of the domain this pod is in.")
 	namespace = flag.String("ns", "", "The namespace this pod is running in. If unspecified, the POD_NAMESPACE env var is used.")
-	domain    = flag.String("domain", "cluster.local", "The Cluster Domain which is used by the Cluster.")
+	domain    = flag.String("domain", "", "The Cluster Domain which is used by the Cluster.")
```
foxish: Can we keep the default as-is and just improve logging and add documentation to help debugging? Although the user must know of the domain option's existence, I think `cluster.local` works for a majority of clusters out there.

foxish (changes requested):

pets/peer-finder/peer-finder.go
venezia: I think we can come up with a better answer than `cluster.local` (as you see in the lower part of the patch) - the more that we make it easy on the user the better.  I suspect many developers wanting to release their application are not super familiar with their cluster domain setting, they just want their app running as quickly as possible, regardless of what setting some other person set when creating the cluster

pets/peer-finder/peer-finder.go
venezia: If the `resolv.conf` does not contain the cluster domain, I believe the dns e2e conformance test will fail - so I believe we can expect this behavior.  From a user perspective, there is a presumption in kubernetes that you can skip the FQDN and use the shorter domain name, and many scripts rely on that.  That only works if there is a search line.

I could be wrong but have you ever seen resolv.conf _not_ have the pod's domain in it?

https://github.com/kubernetes/kubernetes/blob/master/test/e2e/dns.go shows that a conformance test searches for `kubernetes`, `kubernetes.default`, `kubernetes.default.svc`, and then finally the FQDN.

foxish: > I think we can support the current options and have it deduce the "default" domain name if none of these variables are supplied

This makes sense to me. Based on that conformance test, it seems safe enough to try and deduce the domain from the `resolv.conf.` I'll let @kubernetes/sig-network-misc take a quick look just in case.

OTOH, I think we can refactor this bit of the code to rely less on string parsing and perhaps use regular expressions instead.

venezia: I'll give it a shot.

MrHohn: I think "deducing cluster domain from resolv.conf when domain not set" makes sense --- as I don't have better ideas. But also be aware of that resolv.conf will not have cluster domain set up if `dnsPolicy` is set to `Default` ([example](https://github.com/kubernetes/kubernetes/blob/v1.6.6/cluster/addons/dns/kubedns-controller.yaml.base#L171)).

venezia: > But also be aware of that resolv.conf will not have cluster domain set up if dnsPolicy is set to Default

I think peer-finder doesn't work right currently if this is true - because the programs that are using peer-finder (like mongodb-replicaset) will try to use cluster dns to have each peer talk to eachother.

Probably should add something that says "This program relies on `dnsPolicy` set to `ClusterFirst` - would that work?

MrHohn: > Probably should add something that says "This program relies on dnsPolicy set to ClusterFirst - would that work?

Yeah we should definitely clarify this in README. I think peer-finder works right currently in most of the cases --- the default value for dnsPolicy is `ClusterFirst` instead of `Default` :)

venezia:
Updated with regex and improved the readme

pets/peer-finder/peer-finder.go
joejulian: I rather disagree with regex parsing. A much simpler bufio scanner loop around a split loop would be much easier to read and is considered idiomatic go.

venezia:
pinging @foxish for /lgtm, @bprashanth for /approve

joejulian:
@spiffxp Can you help get some attention here?

pwittrock:
@foxish does this look good?

foxish:
Sorry about the delay here. This looks fine, although some testing would make me more confident.
@venezia have you tested this locally for your use-case?

/lgtm

foxish:
Owners files are outdated. Sending update in https://github.com/kubernetes/contrib/pull/2681

foxish (approved):

foxish:
Discussed offline with @spiffxp. 
Tests coming in a future PR.

PR #2645, opened 2017-06-28
Fix cni-plugins/to_docker/bin/c2d: return code should be an int

### LINKED ISSUE #2646: Return code in cni-plugins/to_docker should be an int
Currently a string is being returned instead of an int, in the error case.

MikeSpreitzer:
It was a string.

Resolves #2646


MikeSpreitzer:
Help, has something fallen between the cracks here?  On the Submit Queue line above I see a complaint that continuous-integration/travis-ci/pr is not green --- but it _is_ green!

PR #2652, opened 2017-07-05
Update labels of metric descriptors if they change

loburm:
This PR includes few minor changes that fixes issues in the prometheus-to-sd component:

1. Component is not trying to update metric descriptor if this metric was not whitelisted.
2. Updates metric descriptor if new label was added.
3. Doesn't send broken metrics to the stackdriver (for example when name is longer than 100 characters or number of labels bigger than 10)
4. If sending of batch of timeserieses to the Stackdriver fails, correct number of sent timeserieses is logged.

### DIFF
--- a/prometheus-to-sd/config/source_config.go
+++ b/prometheus-to-sd/config/source_config.go
@@ -22,6 +22,7 @@ import (
 	"net"
 	"strconv"
 	"strings"
+	"github.com/golang/glog"
 )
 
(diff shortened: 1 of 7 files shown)


loburm:
Closing this PR, because component is moved to different repo. Corresponding PR in new repo:
https://github.com/GoogleCloudPlatform/k8s-stackdriver/pull/5

PR #2662, opened 2017-07-10
use publicly available event-exporter image

gytisgreitai:
Change url in examples from private `google.com/vmik-k8s-testing-0` to public `gcr.io/google-containers`

### DIFF
--- a/fluentd/event-exporter/README.md
+++ b/fluentd/event-exporter/README.md
@@ -67,7 +67,7 @@ spec:
     spec:
       containers:
       - name: event-exporter
-        image: gcr.io/google.com/vmik-k8s-testing-0/event-exporter:v0.1.0
+        image: gcr.io/google-containers/event-exporter:v0.1.4
         command:
         - '/event-exporter'
 ```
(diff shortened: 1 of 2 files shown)


crassirostris:
Whoops. Thanks!

/lgtm
/approve no-issue

gytisgreitai:
@crassirostris since you're here, could you explain where under SD those logs should appear?

crassirostris:
@gytisgreitai Hm, I thought I documented it. Under GKE Cluster. I'll add this to the readme, thanks!

gytisgreitai:
Ok, this seems to be working. The one thing that confused me is the buffer size is quite large if you are testing this on a small cluster. I had to do `-sink-opts="-max-buffer-size=5"` to make this more feasable for testing

crassirostris:
@gytisgreitai What was problematic? 5 seconds of flush delay or memory usage?

gytisgreitai:
ah ok, you are flushing after 5 seconds or 100 events by default whichever comes first? 
Initially I thought that you were flushing only after 100 events, so I decreased that to max-buffer-size=5 and started seeing events, but probably this was just some timing/lag issue. 

Sorry for the confusion. All working fine, thanks!

crassirostris:
Thank you! :)

piosz:
/lgtm
Thanks!

PR #2664, opened 2017-07-10
Implement ListFilesInBuild

dashpole:
In order to use the node-perf-dash on local runs, it needs to be able to find local files.   A previous refactor left the ListFilesInBuild function with a TODO to implement it.
This PR implements the ListFilesInBuild function, which should allow using the node-perf-dash with local runs again.

I was not sure if this would follow symbolic links or not, so I added a max depth of 10 to avoid an infinite loop.

/assign @yguo0905

### DIFF
--- a/node-perf-dash/downloader.go
+++ b/node-perf-dash/downloader.go
@@ -20,10 +20,12 @@ import (
 	"bufio"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"log"
 	"os"
 	"path"
 	"strconv"
+	"strings"
 
 	"k8s.io/contrib/test-utils/utils"
 )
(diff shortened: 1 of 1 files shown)


node-perf-dash/downloader.go
```diff
@@ -69,10 +71,43 @@ func (d *LocalDownloader) GetLastestBuildNumber(job string) (int, error) {
...
+	filesInBuild := []string{}
+	for _, file := range filesInDir {
+		if strings.HasPrefix(file, prefix) {
+			filesInBuild = append(filesInBuild, file)
+		}
+	}
+	return filesInBuild, nil
+}
+
+func filesInLocalDir(baseDir, extraDirs string, currentDepth, maxDepth int) ([]string, error) {
```
yguo0905: The `ListFilesInBuild` function is called with `prefix` set to `artifacts/performance-*`
(https://github.com/kubernetes/contrib/blob/master/node-perf-dash/parser.go#L91-L92), so we only need to search the files under `$MY_TEST_RESULT_PATH/$BUILD_NUMBER/artifacts`, rather than recursively?

dashpole: I could do that instead.  I didn't realize that that was the only prefix we searched for.  It would be simpler, but the path would be hard-coded.

yguo0905: I think the `prefix` can be anything like `my/relative/path/my-prefix`. In `ListFilesInBuild`, we could just search the files with `my-prefix` under `$MY_TEST_RESULT_PATH/$BUILD_NUMBER/my/relative/path/`?

dashpole: done.

yguo0905:
Thanks @dashpole for implementing this function!

PR #2670, opened 2017-07-11
Changes: Updating BASEIMAGE for s390x from s390x/busybox to s390x/alp…

guirish:
I have create this PR to resolve below listed issue:
https://github.com/kubernetes/contrib/issues/2661

**Description**: Updated Exechealthz s390x BASEIMAGE to **s390x/alpine** which has **nobody** user in **nobody** group. **s390x/busybox** image has **nobody** user in **nogroup** group.

### DIFF
--- a/exec-healthz/Makefile
+++ b/exec-healthz/Makefile
@@ -46,7 +46,7 @@ ifeq ($(ARCH),ppc64le)
     BASEIMAGE?=ppc64le/busybox
 endif
 ifeq ($(ARCH),s390x)
-    BASEIMAGE?=s390x/busybox
+    BASEIMAGE?=s390x/alpine
 endif
 
 IMAGE := $(REGISTRY)/$(BIN)-$(ARCH)


luxas:
@MrHohn @bowei  Actually I think `nobody:nobody` in the yaml/dockerfile specs is a bug. I think it should be `nobody:nogroup`, isn't that pretty much a standard?

bowei:
I seem to recall that there were some base images that required nobody:nobody, but am not sure which image or whether that is actually true.

MrHohn:
Humm, seems like we are using nobody:nobody in for the latest kube-dns s390x images ([rule.mk#L30-L51](https://github.com/kubernetes/dns/blob/master/rules.mk#L30-L51)), would that be broken as well?

MrHohn:
From https://github.com/kubernetes/dns/issues/61#issuecomment-288391559 looks like this specific issue only happens on exechealthz.
cc @gajju26

gajju26:
@MrHohn
I think for dns we did not face this issue because in the Dockerfile they are not explicitly switching to user "nobody" as they are doing in exechealthz

MrHohn:
@gajju26 Thanks for the reply. Though from https://github.com/kubernetes/dns/blob/master/Dockerfile.sidecar#L21 we could see at least `sidecar` image is switching to `nobody:ARG_NOBODY`, for s390x that corresponding to `nobody:nobody`.

MrHohn:
Agreed that we shouldn't hardcode `nobody:nobody` in Dockfile. Sent https://github.com/kubernetes/contrib/pull/2675 to change the user:group configurations instead of switching base image.

guirish:
@MrHohn referring to the link which you had mentioned earlier https://github.com/kubernetes/dns/blob/master/rules.mk#L30-L51
 for kube-dns too we will have to either update all the BASEIMAGE's to respective alpine images for all ARCH or set
**NOBODY?=nogroup** if we continuing with respective busybox images for PowerPC and s390x ARCH.

I had encountered the nobody user specific issue when I had executed e2e-test of kube-dns which  spawns sidecar container. 

CC: @gajju26 @bowei @luxas

guirish:
@MrHohn as my suggested changes has gone in a new PR https://github.com/kubernetes/contrib/pull/2675 marking this PR as closed. Kindly let us know what the plan for suggested Kube-dns changes in rules.mk files.

Thanks.
cc: @gajju26

MrHohn:
@guirish Thanks for the reminder, I will send PR to fix the kube-dns images.

PR #2675, opened 2017-07-12
exec-healthz: Use alpine as base image on all platforms

### LINKED ISSUE #2661: exechealth-s390x pod fails to start on kubernetes local cluster
I built **exechealthz-s390x:9098dc6-dirty** image locally from master, later I used this image for testing kube-dns and kube-dnsmasq add-on features with Kubernetes local cluster setup. I observed that exechealth pod was failing to start. On debugging the exechealth docker image for s390x, I found that it is using **s390x/busybox** image as its BASEIMAGE in which nobody user is associated with **nogroup** group instead of **nobody** group as expected in latest docker file for exechealth. So I modified the BASEIMAGE for s390x to **s390x/alpine** and rebuilt exechealth. On using the latest built image of exechealth everything is working as expected.

I will be raising a PR to consider the BASEIMAGE change for s390x soon. So after my PR is it possible for you'll to build exechealth-s390x docker image and push it to google repository? So that we will have a working exechealth docker image for s390x.

MrHohn:
Fixes #2661. See discussions on #2670.

<s>Some others user:group configurations are imported from kubernetes/dns#31.</s> (This PR does not change user:group anymore.)

/assign @bowei @luxas 
@guirish Could you please build/test s390x image with this? I don't have the correct environment. Thanks :)

### DIFF
--- a/exec-healthz/Makefile
+++ b/exec-healthz/Makefile
@@ -37,16 +37,16 @@ ifeq ($(ARCH),amd64)
     BASEIMAGE?=alpine
 endif
 ifeq ($(ARCH),arm)
-    BASEIMAGE?=armel/busybox
+    BASEIMAGE?=arm32v6/alpine
 endif
 ifeq ($(ARCH),arm64)
-    BASEIMAGE?=aarch64/busybox
+    BASEIMAGE?=arm64v8/alpine
 endif
 ifeq ($(ARCH),ppc64le)
-    BASEIMAGE?=ppc64le/busybox
+    BASEIMAGE?=ppc64le/alpine
 endif
 ifeq ($(ARCH),s390x)
-    BASEIMAGE?=s390x/busybox
+    BASEIMAGE?=s390x/alpine
 endif
 
 IMAGE := $(REGISTRY)/$(BIN)-$(ARCH)


luxas (changes requested):
This seems veeery wacky tbh.

Isn't there _anything_ (like a uid/gid) that works across every distro?

In that case, we should set this setting in the k8s yaml spec IMO, not rely on implicit defaults from the Dockerfile

MrHohn:
@luxas Thanks for the comment. Unfortunately, doesn't seem like there is a consistent uid/gid we could use across every distro: https://en.wikipedia.org/wiki/User_identifier#Special_values, at least `nobody` has had a few different IDs.

Also just learned a bit how we set user in k8s yaml spec, found https://kubernetes.io/docs/tasks/configure-pod-container/security-context/. Don't think it is obvious enough that every user will follow :(

luxas:
> Also just learned a bit how we set user in k8s yaml spec, found https://kubernetes.io/docs/tasks/configure-pod-container/security-context/. Don't think it is obvious enough that every user will follow :(

If they don't, they don't.

The best way to solve this is to go all-in on either alpine or busybox I think and set this value in the k8s spec to make it visible for the admin.

I've never liked that we treat different platforms differently without a real reason...

@MrHohn What do you think about that?

bowei:
Is that possible given all of the platforms?

luxas:
alpine exists for all arches, yes

MrHohn:
> The best way to solve this is to go all-in on either alpine or busybox I think and set this value in the k8s spec to make it visible for the admin.

Make base images consistent sounds good to me. Though it still makes sense to configure a non-root `user:group` in Dockerfile by default,  in that way user will be better protected even if they don't set it in k8s spec.

guirish:
@MrHohn the build is successful for ARCH=s390x. As it is the same change which was suggested in PR #2670 for s390x, I did not encounter any issues while testing it on s390x. I will close PR #2670 once you merge these changes.

luxas:
This change LGTM

/cc @ixdy @timstclair looks good to you?

tallclair:
Nice. We should make this change to other images that use a mix of alpine & busybox.

bowei (approved):
lgtm

luxas (commented):
The new patch set LGTM

guirish:
@bowei @luxas @MrHohn For building exechealthz for s390x ARCH, I had to add correct 'BUILD_IMAGE' in make command for build to pass. 
`make ARCH=s390x container VERSION=1.X BUILD_IMAGE=s390x/golang:1.7-alpine3.6`

Can we update the Makefile to use BUILD_IMAGE specific to the architecture (similar to BASEIMAGE)?

MrHohn:
@guirish Did it fail during the build or the built image is broken? I'm not seeing error on my end:
```

.../contrib/exec-healthz$ ARCH=s390x make container                                                                                                                            
building: bin/s390x/exechealthz
Sending build context to Docker daemon 22.89 MB
Step 1 : FROM s390x/alpine
latest: Pulling from s390x/alpine

d45fd9d3c4f1: Pull complete 
Digest: sha256:3297a9b30b666eb3e2a926fbbe49dbdbc60a457deb1487a0df7838a9beb02916
Status: Downloaded newer image for s390x/alpine:latest
 ---> 63a4f4f24775
Step 2 : MAINTAINER Prashanth B <beeps@google.com>
 ---> Running in 2e202c14b409
 ---> bcde52cf8617
Removing intermediate container 2e202c14b409
Step 3 : ADD bin/s390x/exechealthz /exechealthz
 ---> 21577bbbe99f
Removing intermediate container 925c33b85fab
Step 4 : USER nobody:nobody
 ---> Running in 46dfb6d0d358
 ---> 6754df8bdf9f
Removing intermediate container 46dfb6d0d358
Step 5 : ENTRYPOINT /exechealthz
 ---> Running in 5f09e17600cb
 ---> 5fa7374c14b4
Removing intermediate container 5f09e17600cb
Successfully built 5fa7374c14b4
container: gcr.io/google_containers/exechealthz-s390x:b0d0b90e
```

luxas:
@MrHohn He's on s390x himself I think and you're cross-compiling, that's the difference.

MrHohn:
@guirish We may not want to bind BUILD_IMAGE with ARCH, as we usually cross compile images for all arch on the same machine. Another thought is to add a variable say BUILD_ARCH, but that sounds not much different from explicitly specifying the BUILD_IMAGE.

guirish:
@MrHohn Thanks for the updates. Could you kindly push a new docker images of exechealth for s390x on google repository, so that we can have a working copy of exechealthz for our reference.

MrHohn:
@guirish Sure, checked a bit and found we never released s390x image for exec-healthz. So I pushed one with the newest tag we have on the other platforms (`gcr.io/google_containers/exechealthz-s390x:1.2`).

PR #2694, opened 2017-07-26
[Ansible] Docker insecure registry task conditional error and formatting

### LINKED ISSUE #2693: [Ansible] Docker insecure registry task conditional error and formatting
The ansible task for "docker : Add any insecure registrys to docker config" tries to compare a list to an int.

Error:
```
fatal: [kube-master-1]: FAILED! => {"failed": true, "msg": "The conditional check 'insecure_registrys is defined and insecure_registrys > 0' failed. The error was: Unexpected templating type error occurred on ({% if insecure_registrys is defined and insecure_registrys > 0 %} True {% else %} False {% endif %}): unorderable types: list() > int()\n\nThe error appears to have been in '/home/themistymay/dev/datamachines/kube-contrib/ansible/roles/docker/tasks/main.yml': line 71, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Add any insecure registrys to docker config\n  ^ here\n"}
```

It also has a poorly formatted `lineinfile` that does not conform to similar `lineinfile` for this same task set.

themistymay:
Fixes #2693

### DIFF
--- a/ansible/roles/docker/tasks/main.yml
+++ b/ansible/roles/docker/tasks/main.yml
@@ -70,7 +70,7 @@
 
 - name: Add any insecure registrys to docker config
   lineinfile: dest={{ docker_config_dir }}/docker regexp=^INSECURE_REGISTRY= line=INSECURE_REGISTRY="{% for reg in insecure_registrys %}--insecure-registry={{ reg }} {% endfor %}"
-  when: insecure_registrys is defined and insecure_registrys > 0
+  when: insecure_registrys is defined and insecure_registrys|length > 0
   notify:
     - restart docker
   tags: configure
--- a/ansible/roles/docker/tasks/main.yml
+++ b/ansible/roles/docker/tasks/main.yml
@@ -69,7 +69,7 @@
   tags: configure
 
 - name: Add any insecure registrys to docker config
-  lineinfile: dest={{ docker_config_dir }}/docker regexp=^INSECURE_REGISTRY= line=INSECURE_REGISTRY="{% for reg in insecure_registrys %}--insecure-registry={{ reg }} {% endfor %}"
+  lineinfile: dest={{ docker_config_dir }}/docker regexp="^INSECURE_REGISTRY=" line="INSECURE_REGISTRY={% for reg in insecure_registrys %}--insecure-registry={{ reg }} {% endfor %}"
   when: insecure_registrys is defined and insecure_registrys|length > 0
   notify:
     - restart docker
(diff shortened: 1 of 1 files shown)


ansible/roles/docker/tasks/main.yml
```diff
@@ -69,8 +69,8 @@
   tags: configure
 
 - name: Add any insecure registrys to docker config
-  lineinfile: dest={{ docker_config_dir }}/docker regexp=^INSECURE_REGISTRY= line=INSECURE_REGISTRY="{% for reg in insecure_registrys %}--insecure-registry={{ reg }} {% endfor %}"
-  when: insecure_registrys is defined and insecure_registrys > 0
+  lineinfile: dest={{ docker_config_dir }}/docker regexp="^INSECURE_REGISTRY=" line="INSECURE_REGISTRY={% for reg in insecure_registrys %}--insecure-registry={{ reg }} {% endfor %}"
```
ingvagabund: Can you reformat it to:
```
lineinfile:
  dest: "{{ docker_config_dir }}/docker"
  regexp: "^INSECURE_REGISTRY="
  line: "INSECURE_REGISTRY={% for reg in insecure_registrys %}--insecure-registry={{ reg }} {% endfor %}"
```
? This way, it is more clear what is what.

ansible/roles/docker/tasks/main.yml
```diff
@@ -69,8 +69,8 @@
...
+  when: insecure_registrys is defined and insecure_registrys|length > 0
```
ingvagabund:
```yaml
when:
- insecure_registrys is defined
- insecure_registrys|length > 0
```
please.

ansible/roles/docker/tasks/main.yml
themistymay: I agree this would be more clear. This does not conform to the other `lineinfile`s in the `main.yml`.

I tend to write my ansible as you have.

I can do this for all of them or just this one. If figured such a change would be a separate issue and pull request (happy to open and fix).

themistymay: @ingvagabund what do you think about comments above?

ingvagabund: I prefer transparency over file scoped coding style. If you don't mind, can you alter all other lineinfile occurrences and squash the commits? I will merge then. Thanks.

themistymay:
Due to the default configuration of **insecure_registrys** and the ansible nuance here: https://github.com/ansible/ansible/issues/13877, Line 95 has to have a default value that will enable ansible 2.3 to work. https://github.com/kubernetes/contrib/pull/2694/commits/5d3c13fd0e249d6468fab0da3a2911af900810ef#diff-8a4c758b73a215ce668fd13badf3fb17R95

themistymay:
Pinging @ingvagabund for a review when you get a chance

themistymay:
waiting on review.

Issue #1709, opened 2016-09-07
[service-loadbalancer] Routed services


Philmod:
Hey!

Why does the service-loadbalancer automatically open routes for heapster, kube-dns:53, kubernetes-dashboard, kubernetes:443, and one for my redis service I created?
![openroutes](https://dl.dropboxusercontent.com/u/45971143/service-loadbalancer.png)

I clearly don't want to open these. Is there a way to avoid it and specify the services I want to route through service-loadbalancer?

Thanks,
Philmod

aledbf:
> Is there a way to avoid it and specify the services I want to route through service-loadbalancer?

No. The existing implementation exports all the services. 
You can use `--namespace` to limit services in one namespace

Philmod:
Ok, thanks.

PR #2700, opened 2017-08-02
[service-loadbalancer] Update README on namespaces

gsaslis:
I got really confused by the README stating that the service loadbalancer only watches the `default` namespace by default. My experience was that I was getting errors about services from other namespaces not playing nicely with haproxy. 

#1709 helped me confirm that, by default, all namespaces are watched, not just `default`, so I thought I'd submit a quick PR to fix this in the docs and hopefully save others some head-bashing... : )


gsaslis:
signed cla

PR #2701, opened 2017-08-03
Update examples and manifests for 0.10

claytono:


claytononeill:
I don't believe the CI failures are related to my change from reading the logs.   Not sure what the best way to proceed with that is.

claytononeill:
CI failure is unrelated and fixed by #2707

PR #2702, opened 2017-08-03
Rename OWNERS assignees: to approvers:

spiffxp:
They are effectively the same, assignees is deprecated

ref: https://github.com/kubernetes/test-infra/issues/3851

I broke this into two commits around `vendor/` because I'm not sure it's kosher to be editing vendored OWNERS directly. ref: https://github.com/kubernetes/test-infra/issues/3694

### DIFF
--- a/images/OWNERS
+++ b/images/OWNERS
@@ -1,2 +1,2 @@
-assignees:
+approvers:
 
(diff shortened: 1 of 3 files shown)


spiffxp:
I've removed the `vendor/` commit, we'll opt to have mungegithub ignore vendor OWNERS later.  In the meantime they'll be effectively ignored once `assignees` is no longer parsed.

PR #2705, opened 2017-08-04
keepalived-vip: fix deprecated NodeLegacyHostIP

claytono:
Replace NodeLegacyHostIP with NodeInternalIP.  This is required for
keepalived-vip to work on bare metal, since it has no cloud provider to set
ExternalIP on each node resource.

### DIFF
--- a/keepalived-vip/utils.go
+++ b/keepalived-vip/utils.go
@@ -100,7 +100,7 @@ func getPodDetails(kubeClient *unversioned.Client) (*podInfo, error) {
 			}
 		}
 
-		if externalIP == "" && address.Type == api.NodeLegacyHostIP {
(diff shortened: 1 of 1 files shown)


claytononeill:
CI failure is unrelated and fixed by #2707

barclaac:
This fix worked perfectly against my k8s 1.7 cluster.

claytononeill:
@barclaac I'd recommend using https://github.com/aledbf/kube-keepalived-vip instead of the version in contrib.  There is no one in the OWNER file for this project that actually does reviews and @aledbf is the original author and is maintaining his version.

bowei:
@aledbf If you want to be part of the owners file for keepalived-vip, feel free to send a PR

PR #2707, opened 2017-08-05
Fix service-loadbalancer test failures

claytono:
The previous test case of http://www.k8s.io is now failing becuase it
redirects to https://www.k8s.io, but the certificate being served for
that site isn't valid for that hostname anymore.  This changes it to use
the main kubenetes page which fixes the test failures.


aledbf:
@claytononeill apologies for the delay

PR #2708, opened 2017-08-06
Fallback to node internal IP if external IP is not set

munnerz:
Ref. this comment: https://github.com/kubernetes/kubernetes/issues/42125#issuecomment-320137328

Right now kube-keepalived does not work if using a cloud provider as the external IP field cannot be set.

I'm using my own cloud provider (https://github.com/munnerz/keepalived-cloud-provider) that automatically configures kube-keepalived-vip for service `LoadBalancer` support. Without this change, it will not work with any version of Kubernetes that doesn't use the old `NodeLegacyHostIP`.

### DIFF
--- a/keepalived-vip/utils.go
+++ b/keepalived-vip/utils.go
@@ -100,7 +100,7 @@ func getPodDetails(kubeClient *unversioned.Client) (*podInfo, error) {
 			}
 		}
 
-		if externalIP == "" && address.Type == api.NodeLegacyHostIP {
+		if externalIP == "" && (address.Type == api.NodeLegacyHostIP || address.Type == api.NodeInternalIP) {
 			externalIP = address.Address
 		}
 	}


munnerz:
Built image of this PR: `quay.io/munnerz/kube-keepalived-vip:0.10`

Tests appear to be failing due to `service-loadbalancer` tests failing. Not sure how best to resolve other than a rebase once it's fixed in master :)

claytononeill:
@munnerz I think we both fixed this in slightly different ways.  I have PR #2705 which just removes the NodeLegacyHostIP comparison, since that is now deprecated and removed.  I've also got a PR up to fix the CI failures at #2707

munnerz:
Ah sorry - I didn't see your fix! Glad somebody else is onto it...

I'm not really sure which approach is best - I just didn't want to change existing behaviour, as some users on older clusters may still depend on this field?

Either way, thanks and happy to close this as a dupe!

claytononeill:
Fair point about older clusters.  I'm ok either way.

cmluciano:
@munnerz Are you still working on this patch?

munnerz:
@cmluciano 

Since https://github.com/kubernetes/contrib/pull/2705 has been merged I think we can close this instead.

I'm assuming we don't want to support older clusters that set NodeLegacyHostIP any more. If we do, I'm happy to rebase this PR.

cmluciano:
Sure thing, feel free to close

PR #2709, opened 2017-08-07
Add GoDoc Widget

asifdxtreme:


asifdxtreme:
@foxish @eparis can you please help me to re-trigger the CI, I believe this CI failure is not because of any code change, there might be an issue with Travis env

PR #2712, opened 2017-08-08
Use pod graceful termination time in rescheduler

jsravn:
So rescheduler can be non-disruptive to existing pods.

It currently uses a hardcoded 10s timeout. In our use case, we want to rely on the value specified in each pod so we don't terminate pods abruptly.

This change adds a max grace period argument that will be set on delete if the pod's grace period is greater. It's set to a default of 10s to preserve the previous behaviour, although in my opinion we should default to the pod's value.

### DIFF
--- a/hack/verify-flags/known-flags.txt
+++ b/hack/verify-flags/known-flags.txt
@@ -53,6 +53,7 @@ gcm-endpoint
 gcs-bucket
 gcs-logs-dir
 generated-files-config
+grace-period
 health-check-path
 healthz-port
 history-url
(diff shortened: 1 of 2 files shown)


jsravn:
@piosz ping - can you have a look?

piosz:
@mwielgus could you please approve flag?

mwielgus:
/approve 
for flag

sebbonnet:
@mwielgus @piosz can this be fix be released please? 
The latest release `v0.3.2` was done before the PR was merged, so does not include the fix.

PR #2714, opened 2017-08-10
Migrate _workspace to vendor

mkumatag:
This PR will address following two things:

1. Bump go version to `1.8`
2. Migrate _workspace to vendor


lavalamp:
FYI I consider contrib closed for new development; I will approve PRs adding other people into individual subdirectories, though, as long as they agree to eventually move the contents elsewhere.

PR #2719, opened 2017-08-14
Remove 404 server

jonpulsifer:
According to https://github.com/kubernetes/contrib/issues/762 we should be trying to purge this repository of code that has a better home. I can think of none other to absorb the 404-server than https://github.com/kubernetes/ingress/

I adjusted `404-server/README.md` in the same fashion as those in contrib/ingress/controllers/ with a link to the new repository.

xref https://github.com/kubernetes/ingress/issues/1072


luxas:
LGTM
@nicksardo @aledbf can you sign off please?

Issue #2724, opened 2017-08-17
replace rescheduler default lister address to 127.0.0.1


hchenxa:
currently, the rescheduler default --listen-address was localhost:9235, but this depend on the dns solve result and will have some issue if dns server have localhost record. 

so suggest to replace the localhost with 127.0.0.1.

PR #2725, opened 2017-08-17
replace localhost to 127.0.0.1 as dns resolve localhost may have issue

hchenxa:
related to issue https://github.com/kubernetes/contrib/issues/2724

in some cases, resolve the localhost may have some problems like below:
root@hchenk8s5:/var/log# host localhost
localhost.eng.platformlab.ibm.com has address 9.111.159.90
root@hchenk8s5:/var/log# ping -c 1 localhost
PING localhost.localdomain (127.0.0.1) 56(84) bytes of data.

so we may need to use 127.0.0.1:9235 as default listen address when rescheduler started.

### DIFF
--- a/rescheduler/rescheduler.go
+++ b/rescheduler/rescheduler.go
@@ -81,7 +81,7 @@ var (
 		`How long should rescheduler wait for critical pod to be scheduled
 		 after evicting pods to make a spot for it.`)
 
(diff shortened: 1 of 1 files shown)


piosz:
/lgtm
/approve
Thanks for catching and fixing it.

PR #2608, opened 2017-05-23
keepalived: Update keepalived to 1.3.5 and use client-go

aledbf:
replaces #2041

PR #2733, opened 2017-08-24
Rebase of #2608 with master

dseeley:
A rebase of the #2608 PR to master, as well as restoring a couple of bits (DR, vrid in template) that were removed, and upgrading keepalived to v1.3.6. This PR requested in the thread: https://github.com/kubernetes/contrib/pull/2608#issuecomment-324109028


jonaz:
Would be really nice to get this one merged as we need updated version of keepalived.

PR #2738, opened 2017-08-31
Add a container & job spec to check metadata concealment correctness

ikehz:
This checks for some of the fixes made in https://github.com/kubernetes/kubernetes/pull/51302.  Ref #8867.

I'm unsure of where this should live, but this seems like as good a place as any?  @spxtr We discussed, and I'd rather just check it in here now.

### DIFF
--- a/metadata-proxy/test/Dockerfile
+++ b/metadata-proxy/test/Dockerfile
@@ -1,0 +1,19 @@
+# Copyright 2016 The Kubernetes Authors. All rights reserved
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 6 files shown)


ihmccreery:
Ping @lavalamp @eparis.

ihmccreery:
Ref #8867.

lavalamp:
I don't know what this is, and AFAIK contrib is not accepting new code. I would consider approving a change adding someone to the OWNERs file of this metadata-proxy thing if you can find someone who can convince me they will feed and water this thing and find it a new home.

ihmccreery:
@lavalamp Done, see #2751.  Please assign #2750 to me.  I'll make sure this gets moved out of kubernetes/contrib, but in the meantime, can I check in these tests?

ihmccreery:
@lavalamp with #2751 merged, are you comfortable lgtm-ing this?  What else can I do to get this merged?

ihmccreery:
Ping @lavalamp

PR #2739, opened 2017-08-31
Fix nginx.conf to block recursive calls and identity endpoints

ikehz:
These changes reflect https://github.com/kubernetes/kubernetes/pull/51302.

We should make these changes here and get rid of the configmap in https://github.com/kubernetes/kubernetes/pull/51302, because configmaps have no good update strategy.  Specifically, if the master version changes, the configmap will change, but pods making up the daemonset won't get restarted (unless the version of the container changed as well), so they won't update their config (the configmap volume may change, but nginx won't reload it).

cc @ahmetb @destijl @mikedanese @Q-Lee.

### DIFF
--- a/metadata-proxy/Dockerfile
+++ b/metadata-proxy/Dockerfile
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 FROM gcr.io/google-containers/debian-base-amd64:0.1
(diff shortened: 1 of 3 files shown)


ihmccreery:
Xref https://github.com/kubernetes/kubernetes/pull/51730

PR #2742, opened 2017-08-31
Don't have a default nginx.conf in the metadata proxy

ikehz:
This avoid confusion as to what config the pod is using.  See https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/metadata-proxy/gce/metadata-proxy-configmap.yaml


ihmccreery:
Ref #8867.

ihmccreery:
Ping @foxish @eparis

ihmccreery:
Ping @foxish @eparis

ihmccreery:
Ping @foxish @eparis

PR #2743, opened 2017-09-01
It looks like *x86_64 images are not longer available

yarilc:
This is to work-around an issue I have notificed today. It looks like the x86_64 are all failing the pull while amd64 are working.


ingvagabund:
@yarilc can you check your CLA? Without that I can not merge the PR

PR #2745, opened 2017-09-01
WIP: Delay after starting keepalived and before reconfiguring

### LINKED ISSUE #2744: [keepalived-vip] kube-keepalived-vip sends SIGHUP to keepalived before keepalived has installed signal handler
The kube-keepalived-vip daemon starts the keepalived daemon and almost immediately sends it a SIGHUP. Because the default action on a SIGHUP is to terminate the process the keepalived process will continually exit and be restarted.

The fix here should be ensure that kube-keepalived-vip waits until keepalived is fully operational with an installed SIGHUP handler.

barclaac:
This fixes #2744 for me.

This isn't a really good fix because almost any time that you fix a race condition by using a sleep there's something wrong.

k.started definitely should not be set to true before the k.cmd.Start() call.

Anyone got a better solution to the sleep? I thought about waiting on the PID file creation by keepalived but unfortunately by code inspection keepalived creates that file before the signal handler is installed so to truly fix the race condition we would need an upstream fix in keepalived.

Open to other suggestions though.

### DIFF
--- a/keepalived-vip/keepalived.go
+++ b/keepalived-vip/keepalived.go
@@ -23,6 +23,7 @@ import (
 	"os/exec"
 	"syscall"
 	"text/template"
+	"time"
 
 	"github.com/golang/glog"
 	k8sexec "k8s.io/kubernetes/pkg/util/exec"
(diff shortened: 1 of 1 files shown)


BlaineEXE:
I would be concerned that 1000ms might work for your testing  but not work for someone else. I would suggest that instead of waiting a fixed amount of time before trying, it might be more appropriate to try to get the results of the command in a back-off loop with a timeout of several seconds before the fatal error is returned.

PR #2753, opened 2017-09-07
Use correct property name for log.dir

golisandeep3:
The correct property name for log.dir  is log.dirs


golisandeep3:
signed the CLA

PR #2754, opened 2017-09-07
Fix kafka data directory property and version 

golisandeep3:
The property name for kafka data directory( log.dir) is wrong.It should be log.dirs


artemyarulin:
Omg, I spent like a day trying to understand why the hell pod was consuming main disk under the load test. Please merge :)

josephglanville:
You will also likely need to create a sub directory to avoid `lost+found` being picked up by Kafka.

PR #2758, opened 2017-09-12
Deprecation notice for service-loadbalancer.

cmluciano:
This project does not have any active maintainers and is supplanted to
enhancements that exist in Ingress.


bowei:
might also want to add some people to the OWNERS just in case

PR #2762, opened 2017-09-20
metadata-proxy: fix build problems, bump dependencies, bump version to 0.1.4

ikehz:
Without `--no-cache`, the `apt-get update` commands were being cached and not updating properly.


tallclair:
/lgtm

@ixdy We should probably use `--no-cache` for all images that install distro packages.

ihmccreery:
Superseded by https://github.com/GoogleCloudPlatform/k8s-metadata-proxy/pull/1.

PR #2764, opened 2017-09-22
[ansible] support installation via system containers

jasonbrooks:
This PR provides the option of installing flannel, etcd, the and the kube master and node components using system containers.

@ingvagabund

### DIFF
--- a/ansible/inventory/group_vars/all.yml
+++ b/ansible/inventory/group_vars/all.yml
@@ -134,3 +134,12 @@ etcd_client_port: '2379'
 # when the scheme vars above are set to "https" you need to set these to true!
 flannel_etcd_use_certs: true
 apiserver_etcd_use_certs: true
+
+kube_use_node_system_container: false
+kube_use_master_system_container: false
+kube_use_etcd_system_container: false
+kube_use_flannel_system_container: false
+
+system_images_registry: "candidate-registry.fedoraproject.org"
+system_images_registry_namespace: "f26"
+system_images_tag: "latest"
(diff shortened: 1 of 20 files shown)


ingvagabund:
@jasonbrooks the verifier does not like header of the `oc_atomic_container.py`. It needs to conform to `https://github.com/kubernetes/contrib/blob/master/hack/boilerplate/boilerplate.py.txt`. Can you update `oc_atomic_container.py`?

 Otherwise, it looks good.

jasonbrooks:
@ashcrow @giuseppe I lifted the atomic container module [directly from](https://github.com/openshift/openshift-ansible/blob/master/roles/lib_openshift/library/oc_atomic_container.py) openshift ansible, is it all right if I change the header to match: https://github.com/kubernetes/contrib/blob/master/hack/boilerplate/boilerplate.py.txt

ashcrow:
Travis error:

```

Boilerplate header is wrong for: /home/travis/gopath/src/k8s.io/contrib/ansible/library/oc_atomic_container.py

!!! 'golint' problems: 

/home/travis/gopath/src/k8s.io/contrib/kubeform/pkg/provider/provider.go:428:2: redundant if ...; err != nil check, just return error instead.
```

ashcrow:
@jasonbrooks I have no issue with that as it's Apache License header ➡️  Apache License header. In this case it will no longer be generated but copied in.

ingvagabund:
The `golint` issue fixed here: https://github.com/kubernetes/contrib/pull/2767

ingvagabund:
@jasonbrooks once you rebase and update the header, the CI should pass. Just fixed the golint stuff.

PR #2766, opened 2017-09-25
Remove api-servers from kubelet environment

strigazi:
In kubernetes 1.8.x the kubelet parameter "--api-servers" has been
removed, remove it from from the systemd environment.


ingvagabund:
Related PR: https://github.com/kubernetes/kubernetes/commit/38aee0464dbcfb2ba22392537ee8f714e991ea1c

PR #2783, opened 2017-10-26
Improve graph visibility and minor cleanups.

porridge:
When adding timeseries to graph, start with higher percentiles, since their
values are usually strictly higher than lower percentiles, which avoids
obscuring graph data.

It also orders data in the onHover labels more naturally.

Swap grey colors for a more visible 90th percentile.

Add .gitignore, missing whitespace and newlines.

### DIFF
--- a/perfdash/.gitignore
+++ b/perfdash/.gitignore
@@ -1,0 +1,2 @@
+/.idea/
+/perfdash
--- a/perfdash/google-gcs-downloader.go
+++ b/perfdash/google-gcs-downloader.go
@@ -55,11 +55,11 @@ func (g *GoogleGCSDownloader) getData() (TestToBuildData, error) {
 				artifacts, err := g.GoogleGCSBucketUtils.ListFilesInBuild(
 					job, buildNumber, fmt.Sprintf("artifacts/%v_%v", filePrefix, strings.ToLower(test)))
 				if err != nil || len(artifacts) == 0 {
-					fmt.Printf("Error while looking for data in build %v: %v", buildNumber, err)
+					fmt.Printf("Error while looking for data in build %v: %v\n", buildNumber, err)
 					continue
 				}
 				if len(artifacts) > 1 {
-					fmt.Printf("WARNING: found multiple files with data, reading only one")
+					fmt.Printf("WARNING: found multiple files with data, reading only one\n")
 				}
 				metricsFilename := artifacts[0][strings.LastIndex(artifacts[0], "/")+1:]
 				testDataResponse, err := g.GoogleGCSBucketUtils.GetFileFromJenkinsGoogleBucket(job, buildNumber, fmt.Sprintf("artifacts/%v", metricsFilename))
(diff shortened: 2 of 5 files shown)


porridge:
Before:

before

After:

after

porridge:
(Note how before the low percentile graphs are obscured by the red 99th percentile graph.)

porridge:
Perhaps @shyamjvs has time to review this?

perfdash/.gitignore
```diff
@@ -0,0 +1,2 @@
+/.idea/
```
shyamjvs: Guess you can skip the second backslash.

shyamjvs:
Lgtm - Could you squash the commits?

porridge:
Why do you think they should be squashed? They represent separate changes to address independent concerns. I shiver at the mere thought of a commit entitled "Change graph colors and .gitignore file"!

shyamjvs:
As long as they're doing logically different (or significantly disjoint) things, it feels ok to have separate commits. For e.g, a commit for adding whitespaces and another for adding newlines seems unnecessary. If one is looking at the commit history of a file or hunting down for a regression, too many commits feels a bit noisy. You could probably add the gitignore in a separate one, but for the rest the PR title seems appropriate also for the commit msg.

porridge:
With all due respect sir, I actually strongly disagree. :-)

Note that the newlines are added in go files, and the whitespace in the js file, in a different directory, and different layer of the app. Lumping them into a single commit does not seem right.

Also, I have the opposite feeling w.r.t. trawling through `git blame` - it's much easier to recognize and skip a harmless commit when its title says "add newline" than when it's "assorted changes in this general area", at which point you need to scan through it all, and waste time scrolling through pages of nitpicks that the author decided to pack together with their main change.

Another argument for keeping commits small and focused is that they are much easier to roll back. If someone later decides that one of the commits was doing something wrong (say the colors are bad), it can be rolled back without losing the log message formatting changes.

BTW [developer docs](https://github.com/kubernetes/community/blob/master/contributors/devel/pull-requests.md#2-smaller-is-better-small-commits-small-prs) clearly favor small commits.

Ideally these would be separate PRs, but then again, every PR produces another merge commit, so I tried to strike a balance here.

shyamjvs:
> Note that the newlines are added in go files, and the whitespace in the js file, in a different directory, and different layer of the app. Lumping them into a single commit does not seem right.

When you add whitespaces, fix typos, etc IMO it is ok to change multiple files in the same commit, even if they're at slightly different code paths.

> Also, I have the opposite feeling w.r.t. trawling through git blame - it's much easier to recognize and skip a harmless commit when its title says "add newline" than when it's "assorted changes in this general area", at which point you need to scan through it all, and waste time scrolling through pages of nitpicks that the author decided to pack together with their main change.

Changes to whitespaces, addition of gitignore, etc are relatively low risk changes and I think can be put into one commit (named 'Add gitignore and missing whitespaces in files' or sth) without affecting readability for someone who's skimming through commits. It wouldn't be needed to open the changes if it was a single commit or a bunch of them to know if it's safe to skip. And you probably would never need to rollback such a commit. Besides, having such extra commits can even hurt while mechanical bisection - where you may need some extra rounds.

> Another argument for keeping commits small and focused is that they are much easier to roll back. If someone later decides that one of the commits was doing something wrong (say the colors are bad), it can be rolled back without losing the log message formatting changes.

I agree - but there is a balance on how granular the changes should be. In principle, it might be possible to even change each line in a separate commit (given that it meets some conditions like compilability, correctness, etc). Something like a color change is small enough imo to not need a rollback - it can be changed in another commit.

That said, I generally agree that splitting changes into smaller independent parts is useful for reasons you stated. The line might be slightly blurry :)

shyamjvs:
/lgtm

This might be an interesting discussion for another day.

PR #2790, opened 2017-10-30
Show resource usage data.

porridge:


porridge:
Bouncing the PR to kick Travis.

perfdash/parser.go
```diff
@@ -43,6 +43,52 @@ func parseTestOutput(data []byte, buildNumber int, job string, testName string,
...
+		result[testName] = BuildData{Job: job, Version: "v1", Builds: map[string][]perftype.DataItem{}}
+	}
+	usage := make(podNameToUsage)
+	for percentile, items := range obj {
+		for _, item := range items {
+			name := RemoveDisambiguationInfixes(item.Name)
+			if _, ok := usage[name]; !ok {
+				usage[name] = make(usageAtPercentiles)
+			} // TODO(porridge): keep max usage?
+			usage[name][percentile] = resourceUsage{float64(item.Cpu), float64(item.Memory)}
```
gmarek: You keep random one, right?

porridge: Yes, effectively the last one spit out by the test to the `.json` file.

gmarek: Max is probably more interesting as it should be more stable than a random one.

porridge: Right, but max over which dimension? :-/

porridge: Or should we keep a franken-record of `{max(cpu), max(memory)}` rather than `max({cpu,memory})`?

gmarek: Max(cpu), max(memory) - this one should be +- stable.

porridge: Done, PTAL.

PR #2796, opened 2017-11-10
Add more data sources and paralellize fetching.

porridge:


shyamjvs:
@porridge Could you fix the gofmt and golint errors related to presubmit?

perfdash/config.go
```diff
@@ -42,36 +45,85 @@ type Jobs map[string]Tests
...
-				},
-				"LoadResources": TestDescription{
-					Name:             "load",
-					OutputFilePrefix: "ResourceUsageSummary",
-					Parser:           parseResourceUsageData,
-				},
+				Prefix:       "gce-100Nodes-master",
+				Descriptions: performanceDescriptions,
+			},
+			"ci-kubernetes-e2e-gci-gce-scalability-stable1": Tests{
```
shyamjvs: Having names of our CI jobs hard-coded here doesn't seem too clean. We're changing it from time and time and every time we do that, we'll have to change here also. Could you we do something better? Maybe just get the names of the jobs from [config.json](https://github.com/kubernetes/test-infra/blob/master/jobs/config.json) and select ones owned by sig-scale?

porridge: But how to retrieve the contents of this file? It's in a different repo, with its own change cadence.

porridge:
Done, PTAL.

perfdash/config.go
shyamjvs: Maybe just `wget https://github.com/kubernetes/test-infra/blob/master/jobs/config.json`?

porridge:
Hm, I'm not sure github.com is recommended for serving files to automated
systems this way, but I can give it a try. How about doing it in a separate
PR though? Reloading configuration in a safe way may require some thought
and effort, and I'd like to start collecting mileage on the bunch of
changes done so far.

shyamjvs:
Changing that in a separate PR sgtm. However, maybe add a TODO there for now?

shyamjvs:
oops.. too late. It merged it right away.

perfdash/config.go
porridge: FTR, opened https://github.com/kubernetes/contrib/issues/2812 about this.

PR #2797, opened 2017-11-13
Add test phases graph and a TODO.

porridge:


shyamjvs:
Holding.. Because there seems to be no options in the 'phases' dropdown next to the one where we're selecting the job_summary.

PR #2805, opened 2017-11-28
Add a demo for init containers

thockin:
@grodrigues3


thockin:
I filed THREE bugs based on this demo.

PR #2806, opened 2017-11-28
Tolerance data populating error

xinxiaogang:
If there is error during data populating, we need to skip the wrong build and continue on the next.


lavalamp:
`contrib` isn't really open for new changes; I'll only approve changes which are adding someone to an OWNERS file who promises to find the code they own a new home.

PR #2821, opened 2018-01-09
scratch-debugger: fix mounted path for socket

ahmetb:
The manifest file for the debug container makes the host's docker socket
available on /mnt/rootfs/var/run/docker.sock, but the script still tries to
use non-existing /run/docker.sock on the container's fs.

/cc @tallclair


ahmetb:
Superseded by #2824.

PR #2822, opened 2018-01-09
scratch-container: allow customizing docker path

ahmetb:
scratch-container/debug.sh assumes the docker CLI on the guest OS will always
be at /usr/bin/docker. This is not the case in many distributions like Minikube
or Kubernetes on Docker (Linuxkit).

Making it customizable through HOST_DOCKERPATH variable.

This will conflict with PR #2821 as it makes edits to the same line. But I'll resolve them depending on which one gets merged first.

/cc @tallclair


ahmetb:
Superseded by #2824.

Issue #2322, opened 2017-01-18
[nginx-ingress-controller] Why is header "connection: close" is added when request is passed to upstream in Nginx ingress?


tjliupeng:
Hi,

With the Nginx ingress controller, the header "connection: close" is added when the request is passed to upstream. I read the upstream server log and find that. Why? What I hope is the original header relating to connection status should be passed to upstream server without any change.

tjliupeng:
Connection is hop-by-hop header. It won't be forwarded by proxy. Adding configuration below

proxy_set_header Connection $http_connection

to nginx.conf to solve it.

rcarmo:
I'm trying to track down why the ingress controller is closing keepalive connections to my backends, and... this appears to be part of the reason. Is there a specific repo where I can discuss this?

PR #2824, opened 2018-01-09
staging-container: stop relying on host docker CLI

ahmetb:
This change supersedes #2320, #2321 and #2322:

- Stop relying on host docker: (1) it might be dynamically compiled and
  therefore might not work on busybox (2) it might not be at /usr/bin/docker.
- Removed mounting of entire host root filesystem to /mnt/rootfs as it's not
  needed anymore. Mounting only the /var/run/docker.sock now.
- Fixed a stderr redirection bug that caused log line to be saved as a file.

/assign @tallclair

### DIFF
--- a/scratch-debugger/debug.sh
+++ b/scratch-debugger/debug.sh
@@ -22,6 +22,7 @@ set -o pipefail
 TMP_SUBDIR="${TMP_SUBDIR:-debug-tools}"
 CONTEXT="${KUBECONTEXT:-}"
 ARCH="${ARCH:-amd64}"
+DOCKER_DOWNLOAD_URL="${DOCKER_DOWNLOAD_URL-https://download.docker.com/linux/static/stable/x86_64/docker-17.12.0-ce.tgz}"
 
 # Parse arguments & flags
 while [[ $# -gt 0 ]]; do
@@ -113,12 +114,16 @@ case ${ARCH} in
     exit 1
 esac
 
-DOCKERCMD="/mnt/rootfs/usr/bin/docker -H unix:///run/docker.sock"
+DOCKERCMD="/tmp/docker/docker -H unix:///mnt/docker.sock"
 
 # Command for installing busybox image from the debugger container into the target container.
 INSTALLCMD="set -x;" # Print commands, for debugging.
+# Download docker client
+INSTALLCMD="${INSTALLCMD} wget -qO/tmp/docker.tgz ${DOCKER_DOWNLOAD_URL}"
+# Extract docker client
+INSTALLCMD="${INSTALLCMD} && tar zxvf /tmp/docker.tgz -C /tmp"
 # Create the directory structure for the install.
-INSTALLCMD="${INSTALLCMD} mkdir -p ${INSTALL_DIR}"
+INSTALLCMD="${INSTALLCMD} && mkdir -p ${INSTALL_DIR}"
 # Copy the directory structure into the target container.
 INSTALLCMD="${INSTALLCMD} && ${DOCKERCMD} cp /tmp ${CONTAINER_ID}:/"
 # Copy the busybox binary into the install location.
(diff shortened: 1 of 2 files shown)


scratch-debugger/debug.sh
```diff
@@ -22,6 +22,7 @@ set -o pipefail
 TMP_SUBDIR="${TMP_SUBDIR:-debug-tools}"
 CONTEXT="${KUBECONTEXT:-}"
 ARCH="${ARCH:-amd64}"
+DOCKER_DOWNLOAD_URL="${DOCKER_DOWNLOAD_URL-https://download.docker.com/linux/static/stable/x86_64/docker-17.12.0-ce.tgz}"
```
tallclair: nit: use `${VAR:-default}` (`:-` vs `-`) syntax, for consistency

scratch-debugger/debug.sh
```diff
@@ -113,12 +114,16 @@ case ${ARCH} in
...
 INSTALLCMD="set -x;" # Print commands, for debugging.
+# Download docker client
+INSTALLCMD="${INSTALLCMD} wget -qO/tmp/docker.tgz ${DOCKER_DOWNLOAD_URL}"
+# Extract docker client
+INSTALLCMD="${INSTALLCMD} && tar zxvf /tmp/docker.tgz -C /tmp"
 # Create the directory structure for the install.
-INSTALLCMD="${INSTALLCMD} mkdir -p ${INSTALL_DIR}"
+INSTALLCMD="${INSTALLCMD} && mkdir -p ${INSTALL_DIR}"
 # Copy the directory structure into the target container.
 INSTALLCMD="${INSTALLCMD} && ${DOCKERCMD} cp /tmp ${CONTAINER_ID}:/"
```
tallclair: This line needs to be changed, because we don't want to copy the downloaded docker into the target container. It was wrong before - it assumed `INSTALL_DIR` was under `/tmp`.

scratch-debugger/debug.sh
```diff
@@ -146,19 +151,13 @@ spec:
...
       # Mount the node FS for direct access to docker.
       volumeMounts:
-        - name: rootfs
-          mountPath: /mnt/rootfs
-          readOnly: true
-        - name: rootfs-run
-          mountPath: /mnt/rootfs/var/run
+        - name: docker-sock
+          mountPath: /mnt/docker.sock
           readOnly: true
```
tallclair: The container can talk to the socket even if readOnly is set to true?

tallclair (commented):
Thanks for fixing this up!

scratch-debugger/debug.sh
ahmetb: yeah this is a typo, good catch!

scratch-debugger/debug.sh
ahmetb: oh I didn't realize the entire `/tmp` got copied there. I will wget docker into `/docker-cli`.

scratch-debugger/debug.sh
ahmetb: yes.

ahmetb:
@tallclair thanks for reviewing. I moved docker CLI out of `/tmp`. It looks kinda uglier now (`/docker-cli/docker/docker`) but if you have any ideas to make it better I'm open to it.

foxish:
If this is maintained, please consider adding an OWNERS file to the directory. @timstclair @ahmetb 
Also, can we find a long term home for this besides contrib?

ahmetb:
I don't think it deserves it's own repo under kubernetes or kubernetes-incubator, either. It's just a hack script, the way I see it. I'm open to being an OWNERS of this.

We have the same problem in k/examples repo, and for that, we encourage people to host these stuff on their personal accounts. Do you have a better suggestion @foxish?

ahmetb:
@tallclair ping for approval

scratch-debugger/debug.sh
tallclair: Yeah, it's a hack since docker doesn't have a `mkdir` equivalent, nor can you copy into a directory that doesn't exist in the target.

tallclair (approved):
/lgtm

I'm happy to add an OWNERS file, but I agree this doesn't deserve it's own repository. This is a temporary solution until [debug containers](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/node/troubleshoot-running-pods.md) are merged.

foxish:
/approve

Based on Tim's lgtm.

PR #2833, opened 2018-01-25
Modify the sample error description

Davidchinacloud:
virtual_ipaddress  should same as virtual_server  in /etc/keepalived/keepalived.conf

Signed-off-by: LinWengang <linwengang@chinacloud.com.cn>


wokaoo1234:
signed a CLA  ok

wokaoo1234:
signed a CLA ok

PR #2842, opened 2018-01-30
isLeader endpoint for leader election

Kyrremann:
We are using the leader election containers in my project, but we want to make it as easy as possible for our developers to find out if a pod is the leader. Hence the `/isLeader` endpoint.

This is also based on problems where we have had a new leader for a set of pods, but the other (non-leader) pods reported that the deleted leader was still the leader, and not the newly created pod.

Flow:
- starting 3 pods (A, B, C)
- B is leader
- deleting B
- starting D
- D is selected as new leader
- A and C both report B to be the leader via `localhost:4040` endpoint, but logs (`kubectl logs`)
 the correct leader (D)
- D reports itself to be the leader, both via endpoint and logs

Deleting one or both A and C, creating a new E and/or F, make them report the correct leader from the endpoint.

### DIFF
--- a/election/example/main.go
+++ b/election/example/main.go
@@ -30,7 +30,7 @@ import (
 	"k8s.io/kubernetes/pkg/api"
 	"k8s.io/kubernetes/pkg/client/restclient"
 	client "k8s.io/kubernetes/pkg/client/unversioned"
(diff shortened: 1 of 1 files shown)


Kyrremann:
I have signed the CLA.

PR #2851, opened 2018-02-13
[keepalived-vip] Propagate error when network interface name couldn't be detected

twz123:
Be fail-fast in case when the network interface name auto detection fails. Otherwise, an empty interface name would be used, which in turn crashes keepalived with a segfault.


twz123:
Ping... Someone here to review this?

PR #2853, opened 2018-02-13
Remove nginx-slim image path

### LINKED ISSUE #2852: The link of images/nginx-slim/README.md is NotFound
https://github.com/kubernetes/contrib/blob/master/images/nginx-slim/README.md contains the link to nginx-slim image under kubernetes/ingress repository.
However, today the link is not found.

oomichi:
Since the commit of kubernetes/ingress-nginx[1], the path has been
changed from images/nginx-slim to images/nginx. This commit updates
the path.

[1]: https://github.com/kubernetes/ingress-nginx/commit/811829de60f8e0c2c69e67a0e6044eab4a5b1627

fixes #2852

### DIFF
--- a/images/nginx-slim/README.md
+++ b/images/nginx-slim/README.md
@@ -1,3 +1,0 @@
-The `nginx-slim` image has moved to the
-[kubernetes/ingress](https://github.com/kubernetes/ingress/tree/master/images/nginx-slim)
-repository.


images/nginx-slim/README.md
```diff
@@ -1,3 +1,3 @@
 The `nginx-slim` image has moved to the
```
foxish: That looks like the `nginx` image. We should probably remove the nginx-slim directory entirely?

lavalamp:
FYI I consider contrib closed for new development; I will approve PRs adding other people into individual subdirectory OWNERS files, though, as long as they agree to eventually move the contents elsewhere.

images/nginx-slim/README.md
oomichi: @foxish Thanks for your review, that makes sense.

PR #2855, opened 2018-02-15
Load benchmark metrics.

porridge:
Also minor improvement to error handling.

This is for https://github.com/kubernetes/kubernetes/issues/58154#issuecomment-362551793

/hold
Needs to wait for https://github.com/kubernetes/test-infra/pull/6843 to go in.
/cc @shyamjvs

### DIFF
--- a/perfdash/config.go
+++ b/perfdash/config.go
@@ -91,6 +91,20 @@ var (
 		},
 	}
 
+	// benchmarkDescriptions contains metrics exported by test/integration/scheduler_perf
+	benchmarkDescriptions = map[string]TestDescription{
+		"BenchmarkResults": {
+			Name:             "benchmark",
+			OutputFilePrefix: "BenchmarkResults",
+			Parser:           parseResponsivenessData,
+		},
+	}
+
+	jobTypeToDescriptions = map[string]map[string]TestDescription{
+		"performance": performanceDescriptions,
+		"benchmark":   benchmarkDescriptions,
+	}
+
 	// TestConfig contains all the test PerfDash supports now. Downloader will download and
 	// analyze build log from all these Jobs, and parse the data from all these Test.
 	// Notice that all the tests should have different name for now.
@@ -131,16 +145,30 @@ func getProwConfig() (Jobs, error) {
 	}
 	jobs := Jobs{}
 	for _, periodic := range conf.Periodics {
+		var thisPeriodicConfig Tests
 		for _, tag := range periodic.Tags {
 			if strings.HasPrefix(tag, "perfDashPrefix:") {
 				split := strings.SplitN(tag, ":", 2)
-				jobs[periodic.Name] = Tests{
-					Prefix:       strings.TrimSpace(split[1]),
-					Descriptions: performanceDescriptions,
-				}
-				break
+				thisPeriodicConfig.Prefix = strings.TrimSpace(split[1])
+				continue
+			}
+			if strings.HasPrefix(tag, "perfDashJobType:") {
+				split := strings.SplitN(tag, ":", 2)
+				jobType := strings.TrimSpace(split[1])
+				thisPeriodicConfig.Descriptions = jobTypeToDescriptions[jobType]
+				continue
+			}
+			if strings.HasPrefix(tag, "perfDash") {
+				fmt.Fprintf(os.Stderr, "warning: unknown perfdash tag name: %q\n", tag)
 			}
 		}
+		if thisPeriodicConfig.Prefix == "" && thisPeriodicConfig.Descriptions == nil {
+			continue
+		}
+		if thisPeriodicConfig.Prefix == "" || thisPeriodicConfig.Descriptions == nil {
+			return nil, fmt.Errorf("invalid perfdash config of periodic %q: none or both of prefix and job type must be specified", periodic.Name)
+		}
+		jobs[periodic.Name] = thisPeriodicConfig
 	}
 	fmt.Printf("Read config with %d jobs\n", len(jobs))
 	return jobs, nil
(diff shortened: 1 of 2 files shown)


perfdash/config.go
```diff
@@ -131,16 +145,30 @@ func getProwConfig() (Jobs, error) {
...
+		var thisPeriodicConfig Tests
 		for _, tag := range periodic.Tags {
 			if strings.HasPrefix(tag, "perfDashPrefix:") {
 				split := strings.SplitN(tag, ":", 2)
-				jobs[periodic.Name] = Tests{
-					Prefix:       strings.TrimSpace(split[1]),
-					Descriptions: performanceDescriptions,
-				}
-				break
+				thisPeriodicConfig.Prefix = strings.TrimSpace(split[1])
```
shyamjvs: Check before this that split actually is of length at least 2?

perfdash/config.go
```diff
@@ -131,16 +145,30 @@ func getProwConfig() (Jobs, error) {
...
+				continue
 			}
+			if strings.HasPrefix(tag, "perfDashJobType:") {
+				split := strings.SplitN(tag, ":", 2)
+				jobType := strings.TrimSpace(split[1])
```
shyamjvs: Same as above.

perfdash/config.go
porridge: I do, 2nd arg to HasPrefix contains a colon.

perfdash/config.go
porridge: Ditto.

porridge:
Reopening to trigger travis now that kubernetes/test-infra#6843 is in.
/hold cancel

perfdash/config.go
```diff
@@ -131,16 +145,30 @@ func getProwConfig() (Jobs, error) {
...
+				continue
+			}
+			if strings.HasPrefix(tag, "perfDash") {
+				fmt.Fprintf(os.Stderr, "warning: unknown perfdash tag name: %q\n", tag)
+			}
+		}
+		if thisPeriodicConfig.Prefix == "" && thisPeriodicConfig.Descriptions == nil {
+			continue
+		}
+		if thisPeriodicConfig.Prefix == "" || thisPeriodicConfig.Descriptions == nil {
```
shyamjvs: I believe we shouldn't be crashing perfdash if one job config wasn't defined properly. Let's print to stderr and proceed by neglecting it?

perfdash/config.go
```diff
@@ -91,6 +91,20 @@ var (
 		},
 	}
 
+	// benchmarkDescriptions contains metrics exported by test/integration/scheduler_perf
+	benchmarkDescriptions = map[string]TestDescription{
+		"BenchmarkResults": {
+			Name:             "benchmark",
+			OutputFilePrefix: "BenchmarkResults",
```
shyamjvs: Shouldn't this be BenchmarkResults_benchmark? (To not conflict with the other human-readable file)

perfdash/config.go
porridge: This is what will happen once perfdash is up and running.
Config errors are only fatal at startup.

perfdash/config.go
porridge: Nope. I tested, it works fine.
Also, the human-readable file does not have a date suffix and has a .txt extension.

perfdash/config.go
shyamjvs: Ah that's true.

perfdash/config.go
shyamjvs: Sg.

perfdash/config.go
shyamjvs: But i still don't understand why do we need to return nil from this function because of 1 bad job (instead of returning the valid jobs we already find)?

porridge: Because with the amount of garbage perfdash prints at startup, it will be impossible to notice just from a warning that there is a broken config, leading to confusion and time spent on debugging.
I prefer to fail fast rater that trying to DWIM.

shyamjvs: I'm still not sure if that's a good enough reason to sacrifice all our results because someone made a small spelling mistake in one of the jobs. The cost of losing all our metrics because of one such mistake (unless someone debugs it) is much higher than making it difficult to debug. Moreover, why do you say that debugging would be hard? Wouldn't it be visible that one of the jobs we were expecting the results for is missing on the dashboard, and we go back looking for issues with it. And if noone notices it, it wasn't probably worth to have the job on the dashboard in the first place :)

shyamjvs: Basically, if someone cares about a job, he will notice missing metrics and look for debugging it. For others, we don't want to block information of all remaining jobs.

porridge: Three problems:
- the person who notices the lack of a graph might be someone who is not aware that this is a result of a misconfiguration, they could simply assume that such data is not available, and not escalate to perfdash owners,
- due to the amount of noise in perfdash logs, looking for the cause of a missing graph will be somewhat daunting,
- in general I really feel strongly that a program should refuse to take garbage as input, otherwise madness ensues (have you tried programming in Visual Basic? ;-) See for example how kubectl exits 1 if one of the concatenated configs on its input is invalid, but some are OK.

shyamjvs: > the person who notices the lack of a graph might be someone who is not aware that this is a result of a misconfiguration

I would suppose that any person with a serious purpose would know the job he wants the results for (i.e scheduler-benchmark, gce at 100-node scale, etc). So if he's trying to find it, he'll know it's missing (either because the job is missing, or its perfdash is missing - both of which should ideally lead him to escalate). This may not be true for more general purpose audience though.

Overall, your reasons do look quite reasonable to me - so I'm fine with keeping it this way.

PR #2860, opened 2018-02-22
Update links

tonglil:
Link gce examples to the gce ingress repo.

What are people's thoughts about removing the examples here and only keep the link?
I am not a fan of having duplicate docs floating around.


tonglil:
/remove-lifecycle rotten

Anyone able to review?

tonglil:
@mirake PTAL
https://github.com/kubernetes/contrib/blob/master/ingress/OWNERS is pretty out of date as well

PR #2866, opened 2018-03-05
Directly write on-start and on-change script output to stdout

lenalebt:
We had a problem with a hanging on-start script, and peer-finder waited for the script to terminate before writing it's output to stdout. This change directly writes the output to stdout as it happens, making it way easier to debug hanging startup scripts.

### DIFF
--- a/peer-finder/peer-finder.go
+++ b/peer-finder/peer-finder.go
@@ -29,6 +29,7 @@ import (
 	"sort"
 	"strings"
 	"time"
+	"bufio"
 
 	"k8s.io/apimachinery/pkg/util/sets"
 )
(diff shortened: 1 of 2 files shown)


peer-finder/peer-finder.go
```diff
@@ -62,11 +63,22 @@ func lookup(svcName string) (sets.String, error) {
 func shellOut(sendStdin, script string) {
 	log.Printf("execing: %v with stdin: %v", script, sendStdin)
 	// TODO: Switch to sending stdin from go
-	out, err := exec.Command("bash", "-c", fmt.Sprintf("echo -e '%v' | %v", sendStdin, script)).CombinedOutput()
+	cmd := exec.Command("bash", "-c", fmt.Sprintf("echo -e '%v' | %v", sendStdin, script))
+
+	stdout, err := cmd.StdoutPipe()
```
eparis: would `cmd.Output()` do this a lot more clearerly?

lenalebt: No. Afaik, `cmd.Output()` will pause until the program terminates and then return it's output. This approach is streaming and will write the data to stdout as it happens (line-by-line). It allows to find problems with hanging on-start and on-change scripts that will never terminate (as it was the case for me).

lenalebt:
I just realized that there have been formatting issues, which I resolved now. Anything I can do to get this merged :)?

lenalebt:
We realized internally that this approach has some drawbacks; it sometimes does not recognize that the connected command terminates and then hangs. We have a solution (by my coworker) that solves these issues. We will create a new pull request for that.

andrenarchy:
@lenalebt did your alternative solution work out in the end? I just created #2944 since we ran into the same issue. If you found out more it'd be great to share it in the issue. Thanks! :)

lenalebt:
Hi @andrenarchy - indeed, we have a more-or-less working solution. I am not 100% sure what exactly we changed, but I have added the source we used to compile our own version as a gist here: https://gist.github.com/lenalebt/9b2ed4abe659fe78a8f68b796d24fad7

Maybe you can use it to solve your issue, or post a new PR for this?

PR #2868, opened 2018-03-11
[startup-script] Don't cache apk update results

yuvipanda:
This makes the image smaller.

### DIFF
--- a/startup-script/Dockerfile
+++ b/startup-script/Dockerfile
@@ -1,5 +1,4 @@
 FROM alpine
-RUN apk update && \
-   apk add bash util-linux
+RUN apk add --no-cache bash util-linux
 ADD manage-startup-script.sh /
 CMD /manage-startup-script.sh


yuvipanda:
/cc @eparis @foxish @lavalamp (from OWNERS file)

This snippet is great, am very glad I found it :)

I'd also love to pin the FROM to specific version of alpine so this is fully reproducible.

yuvipanda:
ALthough I just realized that nsenter is in base alpine, so am unsure why these extra packages are here at all.

yuvipanda:
For those wondering, bash seems to be required when passing the STARTUP_SCRIPT (which is multiline) to `bash -c`. With just sh that seems to not work very well.

yuvipanda:
After a day of being very frustrated, it turns out util-linux is needed because even though alpine ships with an `nsenter`, it is from busybox and really old. I wish this had been documented, so will do so soon.

yuvipanda:
Anyone? :)

yuvipanda:
Anyone? :)

tonglil:
I would like to see this approved as well

foxish (approved):

foxish:
I don't think this code is maintained at this point. If it is useful, it should find a new home. 

/lgtm
/approve

PR #2869, opened 2018-03-12
Upgrade to keepalived 1.4.2

lorellalou:
This Pull Request only upgrade the version ok keepalived to version 1.4.2
It was tested in our environment on baremetal and AWS and work fine.


lrolaz:
CLA Signed

jonaz:
Why is not the Docker image on k8s.gcr.io/kube-keepalived-vip updated with this change?

PR #2871, opened 2018-03-16
fix some mistake

chenhonggc:
Fix some minor bugs.


chenhonggc:
@lavalamp  Are there any problems with these commit？

PR #2878, opened 2018-04-05
Adding metrics for apiserver request count.

krzysied:


perfdash/config.go
```diff
@@ -79,6 +79,11 @@ var (
 			OutputFilePrefix: "APIResponsiveness",
 			Parser:           parseRequestCountData,
 		},
+		"DensityApiserverRequestCount": {
```
shyamjvs: `DensityRequestCount` should be enough. Same below.

shyamjvs: Ok, seems like that's already taken. How about `DensityRequestCountByClient`?

perfdash/parser.go
```diff
@@ -137,3 +137,58 @@ func parseRequestCountData(data []byte, buildNumber int, testResult *BuildData)
 		}
 	}
 }
+
+type metricValue struct {
```
shyamjvs: You probably shouldn't need to redefine a struct here. It should already be present somewhere in kubernetes/kubernetes code. Just try to import that package.

perfdash/parser.go
```diff
@@ -137,3 +137,58 @@ func parseRequestCountData(data []byte, buildNumber int, testResult *BuildData)
...
+	Metric	map[string]string	`json: metric`
+	Value	[]interface{}		`json: value`
+}
+
+type e2eMeteric []metricValue
+
+type e2eMetrics struct {
+	ApiServerMetrics		map[string]e2eMeteric	`ApiServerMetrics`
+	ClusterAutoscalerMetrics	map[string]e2eMeteric	`ClusterAutoscalerMetrics`
```
shyamjvs: Why do you need this?

perfdash/config.go
krzysied: Done

perfdash/parser.go
krzysied: I will add import, however I want to do this in a separate pull request. It requires changes in godep dependencies.

shyamjvs: sgtm

perfdash/parser.go
```diff
@@ -137,3 +137,58 @@ func parseRequestCountData(data []byte, buildNumber int, testResult *BuildData)
...
+type metricValue struct {
+	Metric map[string]string `json: metric`
+	Value  []interface{}     `json: value`
+}
+
+type e2eMeteric []metricValue
+
+type e2eMetrics struct {
+	ApiServerMetrics         map[string]e2eMeteric `ApiServerMetrics`
+	ClusterAutoscalerMetrics map[string]e2eMeteric `ClusterAutoscalerMetrics`
```
shyamjvs: Can we somehow get rid of this? It looks strange that we're specifically also getting ClusterAutoscalerMetrics.

perfdash/parser.go
```diff
@@ -137,3 +137,58 @@ func parseRequestCountData(data []byte, buildNumber int, testResult *BuildData)
...
+}
+
+func createApiserverRequestCountData(data *perftype.DataItem, metric *metricValue) error {
+	data.Unit = ""
+	data.Data = make(map[string]float64)
+	data.Labels = metric.Metric
+	delete(data.Labels, "__name__")
```
shyamjvs: I think you probably can also get rid of `contentType` label - it's not too useful IMO.

perfdash/parser.go
shyamjvs: Can you please add a TODO above the struct saying that we should eventually replace it with the dependency?

shyamjvs:
Have few minor comments - looks great overall.
Thanks!

perfdash/parser.go
krzysied: done

perfdash/parser.go
krzysied: ClusterAutoscalerMetrics removed

perfdash/parser.go
krzysied: done

perfdash/parser.go
```diff
@@ -137,3 +137,59 @@ func parseRequestCountData(data []byte, buildNumber int, testResult *BuildData)
 		}
 	}
 }
+
+//TODO: replace this strucure with proper structure from metric dependecy.
```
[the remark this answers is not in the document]
krzysied: done

shyamjvs:
/lgtm
Thanks.

PR #2887, opened 2018-04-17
small typo fix

AdamDang:
small typo fix


bowei:
I would like to see this if there was an actual code change rather than single character comment fix.
/close

PR #2900, opened 2018-05-24
Fix for ingress-nginx auth example

venezia:
[ingress-nginx v0.9.0](https://github.com/kubernetes/ingress-nginx/releases/tag/nginx-0.9.0) changed default annotation prefixes.  This PR updates the documentation so it is correct for default installations of ingress-nginx versions release after 12/4/2017

### DIFF
--- a/ingress/controllers/nginx/OWNERS
+++ b/ingress/controllers/nginx/OWNERS
@@ -1,0 +1,4 @@
+approvers:
+- venezia
+reviewers:
+- venezia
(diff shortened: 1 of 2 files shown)


venezia:
/assign @bprashanth
Small change, but it is incorrect for anyone trying to use this example with recent versions of ingress-nginx

lavalamp:
I usually make people promise to work on moving things to a better home than contrib before adding them to approvers ;) So, ah, keep your eyes peeled, I guess.

/lgtm
/approve

Issue #2903, opened 2018-05-25
scratch-debugger: debugger fails to start if the /docker-cli directory does not exist


jeromefroe:
The `scratch-debugger/debug.sh` script fails to work if the `/docker-cli` directory does not exist because `wget` is unable to download docker into it. For example:

```
$ ./debug.sh httpbin-7cb89bfd94-tk4j2
Debug Target Container:
  Pod:          httpbin-7cb89bfd94-tk4j2
  Namespace:    default
  Node:         ip-172-20-62-198.ec2.internal
  Container:    httpbin
  Container ID: 64e9e36eee378541839bbfd66726b622b7234abe2a99af7d9d8d6d034529b859
  Runtime:      docker

  "Installing busybox to /tmp/debug-tools ..."

...

+ wget -qO/docker-cli/docker.tgz https://download.docker.com/linux/static/stable/x86_64/docker-17.12.0-ce.tgz
wget: can't open '/docker-cli/docker.tgz': No such file or directory
pod "debugger-8gq6n" deleted
```

By adding a command to ensure `docker-cli` exists prior to the `wget` command I was able to get the container to start.

pavolloffay:
Is there any solution?

PR #2904, opened 2018-05-25
scratch-debugger: Ensure /docker-cli directory exists

jeromefroe:
This diff addresses #2903 by ensuring that the `/docker-cli` exists. Presently, if the directory doesn't exists, the `wget` command will fail because it can't download docker into the directory.


pavolloffay:
Can somebody verify/merge? I have run into the same issue

PR #2905, opened 2018-05-30
 Add a readme and deprecation note for rescheduler

ravisantoshgudimetla:
Since we are moving to priorty and preemption, we don't need rescheduler to move pods around to make room for new pods coming in.


rescheduler/README.md
```diff
@@ -0,0 +1,2 @@
+### Important Note
+Rescheduler has been deprecated and will be removed in subsequent releases. Please make sure that rescheduler is not enabled with priority and preemption in kube-scheduler.
```
bsalamat: To be more clear, please change the second sentence to:
 Please make sure that rescheduler is not run when priority and preemption is enabled in kube-scheduler. Priority and preemption is enabled by default in Kubernetes 1.11 and newer.

ravisantoshgudimetla: Thanks @bsalamat for the review. Done. PTAL.

ravisantoshgudimetla:
closing this PR in order to avoid confusion. For more info, please see https://github.com/kubernetes/kubernetes/pull/64592#issue-191919104

PR #2931, opened 2018-07-19
consider all addresses on an interface

guoshimin:
An interface can have multiple addresses. We should consider all addresses on an interface when we are looking for an interface with a given address.


guoshimin:
Ping @bowei @bprashanth

PR #2947, opened 2018-09-06
Use -domain and -ns flags when performing lookup

oliverisaac:
In previous version of peer-finder if you specify -domain or -ns it doesn't actually look in those domains and namespaces. This PR switches the lookup from using just `$svc` to using `$svc.$ns.svc.$domain`. 

Note that due to [go 1.11 not supporting compressed hostnames in SRV records](https://groups.google.com/forum/#!topic/golang-nuts/vAbjprJNPV0) building this on go 1.11 will cause it to no work when using kube-dns.
I have successfully tested it with go 1.10.4 connecting to a kube-dns instance.

### DIFF
--- a/peer-finder/peer-finder.go
+++ b/peer-finder/peer-finder.go
@@ -127,7 +127,7 @@ func main() {
 	}
 
 	myName := strings.Join([]string{hostname, *svc, domainName}, ".")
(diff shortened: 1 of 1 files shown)


oliverisaac:
@foxish @lavalamp : Is there some way I can help get this merged?

PR #2963, opened 2018-10-17
Delete duplicate 'the'

xichengliudui:
Delete duplicate 'the'


mikedanese:
These need to be fixed in the upstream library then revendored.

PR #2965, opened 2018-10-24
Update Zookeeper Docker Container to 3.4.13 and Ubuntu 18.04

cablespaghetti:
I noticed that the Zookeeper container in this repository was a few versions out of date and while I was there I upgraded it to Ubuntu 18.04 from 16.04. I am running this in my environment and it works properly.

Built image here for convenient testing: https://hub.docker.com/r/cablespaghetti/k8szk

### DIFF
--- a/statefulsets/zookeeper/Dockerfile
+++ b/statefulsets/zookeeper/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:16.04
+FROM ubuntu:18.04
 
 ENV ZK_USER=zookeeper \
     ZK_DATA_DIR=/var/lib/zookeeper/data \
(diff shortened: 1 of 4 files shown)


cablespaghetti:
/assign @kow3ns 

I think the tests have failed because of a glitch in Travis. I am not sure now to re-run them. :(

rlees85:
Worth updating openjdk to version 9 too as the latest version of Zookeeper supports this?

cablespaghetti:
For some reason on my follower nodes I'm getting intermittent health/liveness check failures still (although less than with previous netcat version I was running). I'll investigate this.

rlees85:
Also (PLEASE correct me if i'm wrong) I don't think applications in kubernetes/contrib are maintained. There is another Zookeeper implementation here https://github.com/kow3ns/kubernetes-zookeeper but I have no idea if that is maintained either

cablespaghetti:
You may be right. Although the current incubator helm chart for Zookeeper uses this image as a base. The other repo seems to be quite dead as well. :(

PR #2990, opened 2019-01-08
restart policy should be 'always'

llamahunter:
The 'on-failure' restart policy means that if the kublet exits 'normally', it will not be restarted.  Problem is that systemd considers processes that are killed due to unhandled SIGPIPE as 'clean exits'.  See the 'Restart=' section of https://www.freedesktop.org/software/systemd/man/systemd.service.html, in particular:
"clean exit means an exit code of 0, or one of the signals SIGHUP, SIGINT, SIGTERM or SIGPIPE"
"If set to on-failure, the service will be restarted when the process .. is terminated by a signal (including on core dump, but *excluding the aforementioned four signals*)"
I saw this happen on my own cluster.  Kubelet got a SIGPIPE somehow, and exited, but because the systemd restart policy was 'on-failure', it didn't restart.

### DIFF
--- a/init/systemd/kubelet.service
+++ b/init/systemd/kubelet.service
@@ -17,7 +17,7 @@ ExecStart=/usr/bin/kubelet \
 	    $KUBELET_HOSTNAME \
 	    $KUBE_ALLOW_PRIV \
 	    $KUBELET_ARGS
-Restart=on-failure
+Restart=always
 KillMode=process
 
 [Install]


rootfs:
what caused SIGPIPE in your setup? I suggest we fix that problem instead of changing restart mode (which may cause premature restarts if that error persists)

llamahunter:
The SIGPIPE is caused by journald restarting.  Note that a similar fix was made recently to the AWS EKS AMI for running kubernetes in EC2.  They chose to treat SIGPIPE as a failure, but still leave the restart policy as on-failure.  I guess it's less of a delta, but I can't figure out *why* you would want kubelet to not be running if you've told systemd that you want it running.  Seems to me that the restart policy should be 'always'.

llamahunter:
fwiw, here's the AWS AMI change.  https://github.com/awslabs/amazon-eks-ami/commit/e13d401e531e10f1a66a62d7b54ce109f2375859

llamahunter:
See also https://github.com/awslabs/amazon-eks-ami/pull/110

llamahunter:
and https://github.com/awslabs/amazon-eks-ami/issues/122

travcunn:
@llamahunter I think the restart policy should look like this:

```
Restart=on-failure
RestartForceExitStatus=SIGPIPE
```

In my opinion, this is better because of a user runs SIGKILL against the process, the user intends to kill the process and they won't want systemd to restart it. Here is the final change that made its way into the amazon-eks-ami: https://github.com/awslabs/amazon-eks-ami/pull/137

llamahunter:
Regardless, the current systemd unit file needs to change.

PR #2995, opened 2019-02-07
Fixing non-success response handling

krzysied:
Returning error instead of `nil` when there is non-success response.


krzysied:
/assign @gmarek 
@gmarek PTAL

PR #3010, opened 2019-04-11
Rebase reshceduler to distroless

yuwenma:
This is part of the effort described in kep kubernetes/enhancements#900
Comparing to busybox, Distroless can mitigate the influence due to CVE issues.

Since the rescheduler doesn't have heavy dependencies (only calls rescheduler), this should be a safe move.


bsalamat (commented):
@yuwenma Rescheduler is deprecated and decommissioned since K8s 1.12. K8s priority and preemption is the new mechanism that removed the need to have Rescheduler. So, we no longer deploy Rescheduler. Given that, do we still need this change?

yuwenma:
@bsalamat I see. If so, I'll close the PR (the rebasing is targeting the timeline for the v1.15 release). Thanks for the explanation.

PR #3011, opened 2019-04-16
fix absend response by adding small delay in zookeeper check scripts

langemar:
Running the pipes will close nc after sending commands immediately, which might not let zk time to respond to the request. Given a small delay fix false positives.

Not waiting for the response will (in case of zkOk) result in a lot of failing liveness probes and will restart the pod unnecessarily.


langemar:
CLA

